Issue #30, opened 2022-01-06
Make contract calls with structs as inputs work


digorithm:
As of now, contract calls with structs as inputs don't work because of the way we're crafting the `script_data` and `script` being passed to the transaction. 

We need to develop a way to read custom types in contract calls.

adlerjohn:
Is this an issue with the SDK, or the script you're constructing?

digorithm:
> Is this an issue with the SDK, or the script you're constructing?

The script, I believe. Or how accessing custom types in the VM works. I'm in talks with @vlopes11 to figure this out.

PR #36, opened 2022-01-12
Add support for contract calls with struct as inputs

digorithm:
Before the changes introduced here, using the SDK to call contracts
that takes structs as input would cause weird bugs -- it didn't really
cause a panic on the VM, but it made the access to fields within
a struct retrieve garbage data.

That was happening because in the `script_data` we sent in the tx
we weren't including a `call_data_offset` that's computed as:

```rust
let call_data_offset = script_data_offset as usize
                        + ContractId::LEN + 2 * WORD_SIZE;
```

Once that was included, struct field access worked perfectly.

Curiously, this `call_data_offset` isn't needed if the input is a
primitive type (u8, bool, ...). So a conditional check that adds
the `call_data_offset` to the `script_data` iff the ABI method called
takes a struct is now in place.

This enables users to successfully perform the following:

```Rust
// ...
let contract_instance = MyContract::new(compiled, client);

let counter_config = CounterConfig {
    dummy: true,
    initial_value: 42,
};

let result = contract_instance
    .initialize_counter(counter_config)
    .call()
    .await
    .unwrap();

assert_eq!(42, result);
```

Closes https://github.com/FuelLabs/fuels-rs/issues/30

### DIFF
--- a/fuels-abigen-macro/tests/test_projects/complex_types_contract/Forc.toml
+++ b/fuels-abigen-macro/tests/test_projects/complex_types_contract/Forc.toml
@@ -1,0 +1,9 @@
+[project]
+author  = "Rodrigo Araujo"
+license = "MIT"
+name = "contract_test"
+entry = "main.sw"
+
+[dependencies]
+std = { path = "../lib-std" }
+core = { path = "../lib-core" }
--- a/fuels-rs/src/contract.rs
+++ b/fuels-rs/src/contract.rs
@@ -12,7 +12,7 @@ use fuel_types::{Bytes32, Immediate12, Salt, Word};
 use fuel_vm::consts::{REG_CGAS, REG_RET, REG_ZERO, VM_TX_MEMORY};
 use fuel_vm::prelude::Contract as FuelContract;
 use fuels_core::ParamType;
-use fuels_core::{Detokenize, Selector, Token};
+use fuels_core::{Detokenize, Selector, Token, WORD_SIZE};
 use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
 use std::marker::PhantomData;
@@ -59,6 +59,7 @@ impl Contract {
         gas_price: Word,
         gas_limit: Word,
         maturity: Word,
+        custom_inputs: bool,
     ) -> Result<Vec<Receipt>, String> {
         // Based on the defined script length,
         // we set the appropriate data offset.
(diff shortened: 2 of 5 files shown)


fuels-abigen-macro/tests/test_projects/complex_types_contract/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+author  = "Rodrigo Araujo"
+license = "MIT"
```
adlerjohn, line 3: In a separate PR, can you make this change to current tests?
```suggestion
author  = "Fuel Labs <contact@fuel.sh>"
license = "Apache-2.0"
```

fuels-abigen-macro/tests/test_projects/complex_types_contract/Forc.toml
```diff
@@ -0,0 +1,9 @@
...
+name = "contract_test"
+entry = "main.sw"
+
+[dependencies]
+std = { path = "../lib-std" }
+core = { path = "../lib-core" }
```
adlerjohn, line 9: In a separate PR, can you remove the local version of these and use the versions on GitHub instead?

fuels-rs/src/contract.rs
```diff
@@ -87,16 +88,37 @@ impl Contract {
...
-        // 2. The function selector
-        // 3. The encoded arguments, in order
+        // Primitive types as input:
+        // ┌─────────────┬─────────────┬──────────────┐
+        // │ contract_id │ fn selector │ encoded args │
+        // └─────────────┴─────────────┴──────────────┘
+        // Structs as input:
+        // ┌─────────────┬─────────────┬─────────────────┬──────────────┐
+        // │ contract_id │ fn selector │call_data_offset │ encoded args │
+        // └─────────────┴─────────────┴─────────────────┴──────────────┘
```
adlerjohn, line 98: Can you do something like (only did first line here because GitHub editor sucks)
```suggestion
        // ┌─────────────┬─────────────┬─────────────────┬──────────────┐
        // │ contract_id │ fn selector │call_data_offset │ encoded args │
        // └─────────────┴─────────────┴─────────────────┴──────────────┘
        //  ^^^^^^^^^^^^^ ContractId::LEN
```^

adlerjohn (changes requested):
Please open issues for the two things to do in future PRs.

vnepveu (approved):
Still understanding the repo but OK from my POV

fuels-rs/src/contract.rs
digorithm: Not sure what you mean by you doing only the first line, did you mean adding the `ContractId::LEN` to both diagrams? Like 

image

adlerjohn: No, I meant do it for all the fields, not just contract_id. And ideally on both boxes. Updated the comment to show 2 lines.

digorithm: Uh... it was getting really unreadable like that, and I agree that we should mention the sizes. So I've changed to a simple list. Let me know what you think.

adlerjohn (approved):

Issue #39, opened 2022-01-13
Create initial abstraction for signers


digorithm:
The SDK should have an entrypoint abstraction for signers, in order to enable transaction signing. To begin with, we could keep things simple and generate it all from within the SDK, and eventually add support for Ledger, MetaMask, and so on.

PR #40, opened 2022-01-15
Add initial abstraction for signers and wallets

digorithm:
Two new main abstractions are introduced here: signers and wallets.

Wallets can be created from a secret key and used to sign messages and,
eventually, transactions. The signer trait is generic enough to enable
different sources to be signers, such as local wallets, trezor, ledger,
metamask, etc. These changes enable only the creation of local wallets.

All encryption is done through `secp256k1`.

Here's an example of the new capability:

```Rust
// Generate your secret key
let mut rng = StdRng::seed_from_u64(2322u64);
let mut secret_seed = [0u8; 32];
rng.fill_bytes(&mut secret_seed);

let secret =
    SecretKey::from_slice(&secret_seed)
                        .expect("Failed to generate random secret!");

// Create a new local wallet with the newly generated key
let wallet = LocalWallet::new_from_private_key(secret)?;

let message = "my message";
let signature = wallet.sign_message(message.as_bytes()).await?;

// Recover address that signed the message
let recovered_address = signature.recover(message).unwrap();

assert_eq!(wallet.address(), recovered_address);

// Verify signature
signature.verify(message, recovered_address).unwrap();
```

Closes https://github.com/FuelLabs/fuels-rs/issues/39.

### DIFF
--- a/fuels-core/src/signature.rs
+++ b/fuels-core/src/signature.rs
@@ -74,7 +74,7 @@ impl Signature {
         Ok(())
     }
 
-    /// Recovers the Ethereum address which was used
+    /// Recovers the Fuel address which was used
     /// to sign the given message. Note that this message
     /// can be either the original message or its digest (hashed message).
     /// Both can be used to recover the address of the signature.
--- a/fuels-core/src/signature.rs
+++ b/fuels-core/src/signature.rs
@@ -1,0 +1,195 @@
+use crate::Bits256;
+use fuel_tx::{crypto::Hasher, Bytes64};
+use fuel_types::Address;
+use fuel_vm::crypto::secp256k1_sign_compact_recover;
+use std::{convert::TryFrom, fmt, str::FromStr};
+use thiserror::Error;
+
+/// An error involving a signature.
+#[derive(Debug, Error)]
+pub enum SignatureError {
+    /// Invalid length
+    #[error("invalid signature length, got {0}, expected 64")]
+    InvalidLength(usize),
+    /// When parsing a signature from string to hex
+    #[error(transparent)]
+    DecodingError(#[from] hex::FromHexError),
+    /// Thrown when signature verification failed (i.e. when the address that
+    /// produced the signature did not match the expected address)
+    #[error("Signature verification failed. Expected {0}, got {1}")]
+    VerificationError(Address, Address),
+    /// Error in recovering public key from signature
+    #[error("Public key recovery error")]
+    RecoveryError,
+}
+
+/// Recovery message data.
+#[derive(Clone, Debug, PartialEq)]
+pub enum RecoveryMessage {
+    /// Message bytes
+    Data(Vec<u8>),
+    /// Message hash
+    Hash(Bits256),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
+/// An ECDSA signature. Encoded as:
+///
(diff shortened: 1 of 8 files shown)


QuinnLee (dismissed):

fuels-core/Cargo.toml
```diff
@@ -9,10 +9,16 @@ repository = "https://github.com/FuelLabs/fuels-rs"
 description = "Fuel Rust SDK core."
 
 [dependencies]
+fuel-tx = "0.1"
 fuel-types = { version = "0.1", default-features = false }
+fuel-vm = "0.1"
```
adlerjohn, line 14: Noting that the verification of signatures doesn't really depend on `fuel-vm`, and should be abstracted out into a lighter crate. Can you file and issue for this in `fuel-vm`?

fuels-core/src/signature.rs
```diff
@@ -0,0 +1,195 @@
...
+
+        if recovered != address {
+            return Err(SignatureError::VerificationError(address, recovered));
+        }
+
+        Ok(())
+    }
+
+    /// Recovers the Ethereum address which was used to sign the given message.
+    pub fn recover<M>(&self, message: M) -> Result<Address, SignatureError>
```
adlerjohn, line 78: Sometimes users only have the message hash, not the original message. There should be a method to recover from a message directly (you can enforce the message is 32 bytes in that case, I'd say).

adlerjohn (changes requested):

adlerjohn:
In general, handling key material requires lots of care to harden it. Supporting external hardware wallets (Ledger for a start) is a requirement to completing the signing features (can be done in a future PR).

fuels-core/Cargo.toml
digorithm: Yeah, I noticed that too. It *shouldn't* depend on the VM, but it does as of now because the signing/verification functions live there and reimplementing it here isn't a good idea. 

In the future, yes, this should 100% be extracted out, I'll file an issue as soon as I can!

fuels-core/src/signature.rs
digorithm: I know. This method works both for original message and hashed message, check below. There's a match checking whether it's "Data" or "Hash".

digorithm:
> In general, handling key material requires lots of care to harden it. Supporting external hardware wallets (Ledger for a start) is a requirement to completing the signing features (can be done in a future PR).

True. That's why this PR has "initial abstraction" in the title. One thing at a time 🙂

fuels-core/src/signature.rs
adlerjohn: Ohhh now I get it. Can you document this feature in this section and the above?

fuels-core/src/signature.rs
```diff
@@ -0,0 +1,198 @@
...
+        let recovered = self.recover(message)?;
+
+        if recovered != address {
+            return Err(SignatureError::VerificationError(address, recovered));
+        }
+
+        Ok(())
+    }
+
+    /// Recovers the Ethereum address which was used
```
adlerjohn, line 77:
```suggestion
    /// Recovers the Fuel address which was used
```

adlerjohn (changes requested):

adlerjohn (approved):

Issue #41, opened 2022-01-17
Unwrap on `get_receipt_value()` error


adlerjohn:
Repro: https://github.com/ControlCplusControlV/Sway-Example-Repo/tree/3beeb4c5e0415e1109a730f50ad40dcd22487154

```console
$ forc test
   Compiling core_test v0.1.0 (/home/devel/local/ControlCplusControlV/Sway-Example-Repo)
warning: unused variable: `contract_id`
  --> tests/harness.rs:21:18
   |
21 |     let (client, contract_id) = Contract::launch_and_deploy(&compiled).await.unwrap();
   |                  ^^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_contract_id`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: unused variable: `result`
  --> tests/harness.rs:29:9
   |
29 |     let result = contract_instance.sqrtu32(100).call().await;
   |         ^^^^^^ help: if this is intentional, prefix it with an underscore: `_result`

warning: `core_test` (test "integration_tests") generated 2 warnings
    Finished test [unoptimized + debuginfo] target(s) in 10.04s
     Running tests/harness.rs (target/debug/deps/integration_tests-07366f577abc0bae)

running 1 test
  Compiled library "core".
  Compiled library "core".
  Compiled library "lib-std" with 2 warnings.
  Compiled script "core_test" with 3 warnings.
  Bytecode size is 12724 bytes.
thread 'harness' panicked at 'called `Option::unwrap()` on a `None` value', /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/fuels-rs-0.1.2/src/contract.rs:329:65
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test harness ... FAILED

failures:

failures:
    harness

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 84.04s

error: test failed, to rerun pass '--test integration_tests'
```

Looks like this line is trying to `unwrap()` even though the result being unwrapped is not guaranteed to be ok.

https://github.com/FuelLabs/fuels-rs/blob/v0.1.2/fuels-rs/src/contract.rs#L329

PR #42, opened 2022-01-17
Handle calls with empty returns

digorithm:
Calls that don't return any values weren't being properly handled before. This fixes https://github.com/FuelLabs/fuels-rs/issues/41.


adlerjohn (approved):
utACK

PR #45, opened 2022-01-18
Make fuel-core && fuel-gql-client optional deps

Voxelot:
The abi-gen-macro crate transitively includes fuel-core and fuel-gql-client even though those deps aren't used at all in the codegen process. This was causing circular dependencies when testing abi gen in the fuel-core indexer service. 

Ideally, codegen is refactored to live in fuels-core without requiring fuel-core or fuel-gql-client. I attempted this, but it was going to be too much of a major change, and resorted to feature flagging for now (basically the only thing left in fuels-rs after my refactor was the script module and some contract deployment and call functions).

### DIFF
--- a/fuels-abigen-macro/Cargo.toml
+++ b/fuels-abigen-macro/Cargo.toml
@@ -18,12 +18,10 @@ bytes = { version = "1.0.1", features = ["serde"] }
 forc = { version = "0.2", features = ["test", "util"], default-features = false }
 fuels-core = { version = "0.1.3", path = "../fuels-core" }
 fuel-asm = { version = "0.1", features = ["serde-types"] }
-fuel-gql-client = { version = "0.1", default-features = false }
-fuel-core = { version = "0.1", default-features = false }
 fuel-tx = "0.1"
 fuel-types = "0.1"
 fuel-vm = "0.1"
-fuels-rs = { version = "0.1.3", path = "../fuels-rs" }
+fuels-rs = { version = "0.1.3", path = "../fuels-rs", default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
 itertools = "0.10.1"
 proc-macro2 = "1.0"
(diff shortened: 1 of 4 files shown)


rakita (dismissed):
lgtm.
It seems okay that `fuels-rs` uses `fuel-core`,`fuel-gql-client`.
"Problem" (it is not even that problematic, it just add's one more step) is in example lib inside `fuel-core`. The solution is maybe to just remove it: https://github.com/FuelLabs/fuel-core/pull/114#discussion_r786328493

fuels-abigen-macro/Cargo.toml
```diff
@@ -38,3 +36,8 @@ strum_macros = "0.21"
 syn = "1.0.12"
 thiserror = { version = "1.0.26", default-features = false }
 tokio = "1.15.0"
+
+[dev-dependencies]
+fuel-gql-client = { version = "0.1", default-features = false }
+fuel-core = { version = "0.1", default-features = false }
```
adlerjohn, line 42:
```suggestion
fuel-core = { version = "0.1", default-features = false }
fuel-gql-client = { version = "0.1", default-features = false }
```

adlerjohn (commented):
CI is failing due to warnings being errors I think.

Voxelot:
We can talk about where to move the example, but that will affect the infra team's progress so that should be coordinated with @tjsharp1. 

The abi gen macros are in a standalone crate used only for codegen of abi types, and these types have no integration or need to depend on fuel-core and fuel-gql-client. It's a lot of overhead to make everyone who wants to be able to codegen their abi types compile a lot of unrelated dependencies like graphql libraries etc.

adlerjohn (commented):
lgtm, but will defer to @digorithm. Very ugly indeed 😂

tjsharp1 (approved):

digorithm (approved):

Voxelot:
If it's too ugly I can close the PR and investigate a cleaner refactor. Moving the code-gen to fuels-core might be a pretty significant gutting of fuels-rs though.

Voxelot:
Marking as a draft, for now, to allow @digorithm time to investigate other options and prevent someone from accidentally slapping the merge button.

digorithm:
@Voxelot Do you mind closing this one since https://github.com/FuelLabs/fuels-rs/pull/66 was merged?

Voxelot:
Gladly, thank you!

Issue #34, opened 2022-01-10
Update UtxoId to be (TxID,OutputIndex)


digorithm:
Once https://github.com/FuelLabs/fuel-tx/pull/53 is merged, the SDK needs to update references to the UtoxId to match the new format, which is the tuple (TxID,OutputIndex).

PR #46, opened 2022-01-19
Fix breaking changes introduced by new UtxoId

digorithm:
Closes https://github.com/FuelLabs/fuels-rs/issues/34

### DIFF
--- a/fuels-abigen-macro/Cargo.toml
+++ b/fuels-abigen-macro/Cargo.toml
@@ -15,14 +15,14 @@ proc-macro = true
 Inflector = "0.11"
 anyhow = "1"
 bytes = { version = "1.0.1", features = ["serde"] }
-forc = { version = "0.2", features = ["test", "util"], default-features = false }
+forc = { version = "0.3", features = ["test", "util"], default-features = false }
 fuels-core = { version = "0.1.3", path = "../fuels-core" }
 fuel-asm = { version = "0.1", features = ["serde-types"] }
-fuel-gql-client = { version = "0.1", default-features = false }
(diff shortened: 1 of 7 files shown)


fuels-rs/src/script.rs
```diff
@@ -43,6 +43,8 @@ impl Script {
             binary_outfile: None,
             offline_mode: false,
             silent_mode: true,
+            print_ir: false,
+            use_ir: false,
```
adlerjohn, line 47: To be clean, can you put these in the same order as the fields are defined in?

fuels-abigen-macro/Cargo.toml
```diff
@@ -15,14 +15,14 @@ proc-macro = true
...
 bytes = { version = "1.0.1", features = ["serde"] }
-forc = { version = "0.2", features = ["test", "util"], default-features = false }
+forc = { version = "0.3", features = ["test", "util"], default-features = false }
 fuels-core = { version = "0.1.3", path = "../fuels-core" }
 fuel-asm = { version = "0.1", features = ["serde-types"] }
-fuel-gql-client = { version = "0.1", default-features = false }
-fuel-core = { version = "0.1", default-features = false }
-fuel-tx = "0.1"
+fuel-gql-client = { version = "0.2", default-features = false }
+fuel-core = "0.2"
```
adlerjohn, line 22: Shouldn't remove `default-features = false` here and in the other manifest file.

adlerjohn (changes requested):

adlerjohn (approved):

PR #47, opened 2022-01-19
Bump versions to 0.2.0

digorithm:


adlerjohn (changes requested):
While technically semver works, can you use the exact version for workspace dependencies? We have to think harder about the ramifications of using only up to minor version. Also, it makes it harder to do a find and replace.

adlerjohn (approved):

Issue #35, opened 2022-01-12
Improve the data returned by contract calls


digorithm:
As of now, we return the parsed return value equivalent to the contract call's return type, e.g. if `my_method()` returns a `bool`, the call to this method using the SDK will return only the `bool`. It might be useful to return more context about the call, such as the other receipts, containing all the rest of the context of the call.

In more detail, this is fairly simple. In `fuels-rs/src/contract.rs` we have:

```Rust
pub async fn call(self) -> Result<D, Error> {
    let receipts = Contract::call(
        self.contract_id,
        Some(self.encoded_selector),
        Some(self.encoded_args),
        &self.fuel_client,
        self.utxo_id,
        self.balance_root,
        self.state_root,
        self.input_index,
        self.gas_price,
        self.gas_limit,
        self.maturity,
        self.custom_inputs,
    )
    .await
    .unwrap();

    let returned_value = Self::get_receipt_value(&receipts).unwrap().to_be_bytes();

    let mut decoder = ABIDecoder::new();

    let decoded = decoder.decode(&self.output_params, &returned_value)?;

    Ok(D::from_tokens(decoded)?)
}
```

We need to find an ergonomic way to return, along with the `D::from_tokens(decoded)` (the decoded Tokens), the other receipts that are already in there.

PR #48, opened 2022-01-19
feat: return receipts in data returned by contract calls

vnepveu:
- Introduce the CallResult struct, that holds the call return in its
  `value` field, and the receipts in the `receipts` field.
- This closes issue #35

### DIFF
--- a/fuels-rs/src/contract.rs
+++ b/fuels-rs/src/contract.rs
@@ -362,7 +362,7 @@ where
         .await
         .unwrap();
 
-        let encoded_value = match receipts.iter().find(|&receipts| receipt.val().is_some()) {
+        let encoded_value = match receipts.iter().find(|&receipt| receipt.val().is_some()) {
             Some(r) => r.val().unwrap().to_be_bytes(),
             None => [0u8; 8],
         };
--- a/fuels-rs/src/contract.rs
+++ b/fuels-rs/src/contract.rs
@@ -297,12 +297,12 @@ impl<D> ContractCall<D>
 where
     D: Detokenize,
 {
-    /// Call a contract's method. Note that it will return
-    /// the method's value as an actual typed value `D`.
-    /// For instance, if your method returns a `bool`, this will be a
-    /// `Result<bool, Error>`. Also works for structs! If your method
-    /// returns `MyStruct`, `MyStruct` will be generated through the `abigen!()`
-    /// and this will return `Result<MyStruct, Error>`.
+    /// Call a contract's method. Return a Result<CallResponse, Error>.
+    /// The CallResponse structs contains the method's value in its `value`
+    /// field as an actual typed value `D` (if your method returns `bool`, it will
+    /// be a bool, works also for structs thanks to the `abigen!()`).
+    /// The other field of CallResponse, `receipts`, contains the receipts of the
+    /// transaction
     pub async fn call(self) -> Result<CallResponse<D>, Error> {
         let mut receipts = Contract::call(
             self.contract_id,
--- a/fuels-rs/src/contract.rs
+++ b/fuels-rs/src/contract.rs
@@ -33,6 +33,14 @@ pub struct Contract {
     pub compiled_contract: CompiledContract,
 }
 
+/// CallResponse is a struct that is returned by a call to the contract. Its first
+/// value is the value of interest, and the other one are receipts, are per FuelVM specs
+#[derive(Debug)]
+pub struct CallResponse<D> {
+    pub value: D,
+    pub receipts: Vec<Receipt>,
+}
+
 impl Contract {
     pub fn new(compiled_contract: CompiledContract) -> Self {
         Self { compiled_contract }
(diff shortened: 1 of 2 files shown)


vnepveu:
Note: failing tests are passing when I run them individually, for some weird reason (probably config issue?)

digorithm:
> Note: Sometimes some tests are failing but passing when I run them individually, for some weird reason (probably config issue?).

The tests were designed to be self-contained enough so that state doesn't leak between tests. E.g. we spawn a short-lived `fuel-core` per test case in `harness.rs`. 

Please provide repro steps if you found this not to be the case. Also, check it against `master` to see if this was introduced on this branch. 

So far I haven't experienced any flakiness in `fuels-rs` tests, and if that's the case now we have to do the best we can to avoid flakiness in tests.

vnepveu:
Tell me if you think I should add more tests!

vnepveu:
It's true that it doesn't happen on master… But I have little to no idea how to investigate, since the behavior seems erratic at best, and seems really linked to a performance issue (some tests don't complete in less than 60 secs and are killed).

fuels-rs/src/contract.rs
```diff
@@ -353,24 +361,26 @@ where
...
+                } else {
+                    encoded_value = receipt.val();
+                }
             }
         }
-        None
+        let encoded_value = match encoded_value {
+            Some(val) => val.to_be_bytes(),
+            None => [0u8; 8],
+        };
```
digorithm, line 378: This gets the job done, but we can make this much simpler and efficient: 

```Rust
    let receipts = Contract::call(
        // ...
    )
    .await
    .unwrap();

  // No need to create a whole new `contract_receipts` (we already have the receipts returned from the call!)
  // No need for a `for-loop` with two nested `if`s, better for readability.
  // So here we're basically doing the `receipt.val()` look-up and parsing in one go. 
  let encoded_value = match receipts.iter().find(|&receipt| receipt.val().is_some()) {
      Some(r) => r.val().unwrap().to_be_bytes(),
      None => [0u8; 8],
  };
  
  let mut decoder = ABIDecoder::new();
  
  let decoded_value = decoder.decode(&self.output_params, &encoded_value)?;
  
  Ok(CallResponse {
      value: D::from_tokens(decoded_value)?,
      receipts, // And here we pass the `receipts` that returned from the call itself
  })
}
```
Much shorter, eh? Let me know what you think! Also all the comments I placed here are just for explanation, they wouldn't go in the actual code! 😅

digorithm (changes requested):
It's looking good! Just a few nits and comments.

fuels-rs/src/contract.rs
vnepveu: But shouldn't we call `.val()` on the `receipts` other than just the *one* value? And also we are returning the `value` in receipts (so twice?).

digorithm: Not sure I understand what you mean.

> But shouldn't we call .val() on the receipts other than just the one value?

Should we? What we want: the value in the receipt type that contains the value returned by the method we call, so only one: the `Return` receipt that, which is the one that contains `.val() == Some` (https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#return-receipt)

> And also we are returning the value in receipts (so twice?)

Not sure what you mean by this.

vnepveu: Ok go it!
> Not sure what you mean by this.
I just mean that the `encoded_value` (the one from which we derive `value`) is also inside the `receipts` vector so it's one thing we could remove from the receipts vector. We are sending the same information twice (once decoded as `value`, the other one still encoded, the first non-None item in the `receipts` vector). I'm not sur I make sense :sweat_smile:

digorithm: Oh, I see what you mean! Yeah, we could remove that receipt out of the `receipts`, since `value` is already being sent to the user. I guess we can declare `receipts` as `mut` and remove from it once we decode the `Return` receipt.

vnepveu:
CI passing as well! :tada:

fuels-rs/src/contract.rs
```diff
@@ -294,8 +303,8 @@ where
     /// `Result<bool, Error>`. Also works for structs! If your method
     /// returns `MyStruct`, `MyStruct` will be generated through the `abigen!()`
     /// and this will return `Result<MyStruct, Error>`.
-    pub async fn call(self) -> Result<D, Error> {
-        let receipts = Contract::call(
+    pub async fn call(self) -> Result<CallResponse<D>, Error> {
```
digorithm, line 306: Do you mind updating this function's doc-level comments?

digorithm (changes requested):

fuels-rs/src/contract.rs
vnepveu: sure, forgot

digorithm:
> Changing docs strings broke CI? wth

Haha, that's actually what I'm currently investigating. It's failing to compile Sway code when we call forc::build during a test. So weird.

digorithm (approved):
Good stuff! 😄

PR #51, opened 2022-01-22
ci: build all

adlerjohn:


vnepveu (approved):

vnepveu:
this will reduce compilation time?

adlerjohn:
> this will reduce compilation time?

I don't think so? Presumably it was being built on the first test only. I guess we'll find out.

adlerjohn:
cc @digorithm

digorithm (approved):

PR #56, opened 2022-01-23
Remove use of `[[tx-input]]` from manifest.

adlerjohn:
`[[tx-input]]` should never have been used outside of the Sway repo anyways. This PR removes the use of it in the SDK.


vnepveu (approved):
The idea is that once compiled, a contract/script doesn't need inputs/outputs, but only calls to that contracts' methods do? Checking that I understand

digorithm (approved):
I love the smell of freshly deleted code 🎉

PR #58, opened 2022-01-23
Remove malleable fields from being set by the SDK.

adlerjohn:
The SDK was setting values for UTXO ID, state root, and balances root. These are not meant to be set by the SDK, but rather malleated by the block producer (i.e. a node).

Also removed the logic to handle input index, since it wasn't actually used. Added a TODO to allow passing in multiple contracts that will each have an input-output pair, at which point the input index will be needed. Filed #59 to track this feature.


digorithm (approved):
LGTM!

Issue #52, opened 2022-01-22
Refactor `no-std` feature to not break build


adlerjohn:
> Having a “no-std” feature isn't really the idiomatic way. Typically you make it so that disabling default features is all you need to do to be no-std friendly.
@Voxelot

Currently, #51 fails because turning _on_ features breaks the build. Features should be refactored so that turning _off_ features makes the crate no-std friendly.

PR #60, opened 2022-01-24
Temporarily remove `no_std` support

digorithm:
Currently, we created a `no_std` feature to handle wasm builds. However, this causes `cargo build --all-features` to break, which shouldn't happen. Instead, we should have a `std` feature-set set as default, as per the official Rust docs:

> For example, if you want to optionally support no_std environments, do not use a no_std feature. Instead, use a std feature that enables std. For example: (https://doc.rust-lang.org/cargo/reference/features.html#feature-unification)

This PR temporarily removes the `no_std` so the CI won't break and this issue (https://github.com/FuelLabs/fuels-rs/issues/62) captures the future work that needs to be done in order to properly enable `no_std` compatibility.

Closes https://github.com/FuelLabs/fuels-rs/issues/52.


fuels-core/src/lib.rs
```diff
@@ -1,10 +1,13 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
 use core::fmt;
 use fuel_types::bytes::padded_len;
 use fuel_types::Word;
 use strum_macros::EnumString;
 
 pub mod abi_decoder;
-#[cfg(not(feature = "no-std"))]
+#[cfg(any(feature = "std"))]
```
adlerjohn, line 10: Is there a reason not to use

```suggestion
#[cfg(feature = "std")]
```

digorithm: No special reason.

adlerjohn (commented):
CI failing for some reason

fuels-core/Cargo.toml
```diff
@@ -24,5 +24,5 @@ thiserror = { version = "1.0.30", default-features = false }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
 
 [features]
-default = ["fuel-types/default", "sha2"]
-no-std = ["fuel-types/serde-types-minimal"]
+default = ["fuel-types/default", "sha2", "std"]
+std = []
```
Voxelot, line 28: You'll likely need to make more of the deps optional, and only enable their "std" features when enabled in this crate.

Haven't tested myself but you'll need to change it to something like this:

```suggestion
std = ["hex/std", "fuel-tx/std", "fuel-types/std", "thiserror"]
```

Voxelot:
You'll also need to update CI to run a build/test in no-std mode. Something similar to this: https://github.com/FuelLabs/fuel-types/blob/master/.github/workflows/ci.yml#L49

fuels-core/Cargo.toml
digorithm: Oh, interesting. TIL. Thanks!

digorithm:
Uuuugh, this is very weird. The following test https://github.com/FuelLabs/fuels-rs/blob/master/fuels-rs/tests/calls.rs#L17 is failing because `forc` can't build this very simple Sway script (https://github.com/FuelLabs/fuels-rs/blob/master/fuels-abigen-macro/tests/test_projects/simple_script/src/main.sw). 

However, this is only happening in this CI server (!), and locally I can build it just fine:

```plaintext
simple_script git:(rodrigo/no-std-config) forc build
  Compiled library "core".
  Compiled library "core".
  Compiled library "lib-std" with 2 warnings.
  Compiled script "Fuel example project".
  Bytecode size is 396 bytes.
  ```
  
  Crazy, uh? I then decided to put an `#[ignore]` on this failing test, because it honestly isn't testing anything special that the main harness file isn't testing... and the CI passes.  
  
  Note that this is happening everywhere else, not just this branch. Even on `master` that was passing a couple of days ago. That leads me to believe it isn't on `fuels-rs`, a small likelihood that's on `forc`, and a pretty big chance that Github Actions is doing something fishy. 
  
  cc @adlerjohn @Voxelot @vnepveu

**Comment by digorithm** [comment]
**Update**

After a discussion that happened in the infra sync call with @Voxelot and @tjsharp1, we've decided to unblock the `fuels-rs` build by simplifying things (no `no_std` compatibility for now), and then, in another issue, I'll do some refactoring to turn the necessary bits of `fuels-rs` `no_std` so that it won't impact the indexer. In the meantime, the indexer should continue using the current `fuels-rs` pinned version.

File: .github/workflows/ci.yml [hunk 3]
```diff
@@ -65,24 +65,24 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Checkout repository
-      uses: actions/checkout@v2
+      - name: Checkout repository
```
adlerjohn, line 68: Can you revert these formatting changes? Orthogonal to this PR.

adlerjohn:
Can you also update the description of the PR? I'm not actually sure what this PR does.

digorithm:
> Can you also update the description of the PR? I'm not actually sure what this PR does.

Nice catch, that wasn't making any sense anymore. Title and description updated.

.github/workflows/ci.yml
digorithm: Done

fuels-rs/tests/calls.rs
```diff
@@ -10,6 +10,7 @@ use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
 
 #[tokio::test]
+#[ignore]
```
adlerjohn, line 13: Can you file an issue for this _and then link that issue here in a code comment_?

adlerjohn (approved):
LGTM, but would prefer weigh-in from @Voxelot

Voxelot (approved):

PR #66, opened 2022-01-27
Reorganize fuels-core and fuels-contract

digorithm:
This PR doesn't introduce any new feature / breaking changes.

It moves things around in order to:

1. Make `fuel-core` and `fuel-gql-client` dev dependencies on
`fuels-abigen-macro`, which solves
(https://github.com/FuelLabs/fuels-rs/pull/45/files);
2. Move files that should be `core`, such as codegen, to `fuels-core`;
3. Rename `fuels-rs` to `fuels-contract`, as it now only deals with
contract related code, instead of trying to be a "catch-all" crate;
`fuels-rs` before as a real mixed bag of code and it never really had a
specific purpose, such as `fuels-signers`. Now it being `fuels-contract`
it can be concerned with a single purpose.

Now we have the abigen macro depending only on abigen related code
and the whole `fuels-rs` org makes more sense:

1. `fuels-abi-cli`: CLI with some utility tools
2. `fuels-abigen-macro`: macro crate
3. `fuels-contract`: contract calling related code
4. `fuels-core`: code that all other crates depend on
5. `fuels-signers`: signature/wallet related code

This will also help us, in the future, to make `fuels-core`
`no_std` compatible.


digorithm:
> > This PR doesn't introduce any new feature / breaking changes.
> 
> I think it does introduce a breaking change if it removes `fuels-rs` as a thing, no?

That's true, I meant no "internal" breaking changes 😅. But yes, it's definitely an external breaking change that will require a minor bump on the semver.

vnepveu (dismissed):
This way the repo is much cleaner, love it!

adlerjohn (commented):
No real complaints, but would like an okay from @Voxelot too

Voxelot:
Looks like the abigen macros are still causing a circular dep on the fuel-core repo due to dependencies on forc and fuels-contract. 

```
pushd fuels-abigen-macro/ && cargo tree -e features,normal | grep "fuel-gql" && popd
..
│       ├── fuel-gql-client v0.2.1
│       ├── fuel-gql-client v0.2.1 (*)
```

However, it seems these deps and many others (like tokio) could be safely moved to `dev-dependencies` within `fuels-abigen-macro`.

digorithm:
> Looks like the abigen macros are still causing a circular dep on the fuel-core repo due to dependencies on forc and fuels-contract.
> 
> ```
> pushd fuels-abigen-macro/ && cargo tree -e features,normal | grep "fuel-gql" && popd
> ..
> │       ├── fuel-gql-client v0.2.1
> │       ├── fuel-gql-client v0.2.1 (*)
> ```
> 
> However, it seems these deps and many others (like tokio) could be safely moved to `dev-dependencies` within `fuels-abigen-macro`.
> 
> Apart from that, overall the re-org makes sense to me.

True! Checking this led me to see lots of unused deps in `fuels-abigen-macro`, thanks! It looks like now it doesn't depend on `fuel-core`/`fuel-gql-client` anymore.

Voxelot (approved):
lgtm!

adlerjohn (approved):

Issue #61, opened 2022-01-25
ABIgen custom types do not check that the declared type matches the concrete type


vnepveu:
See `expand_input_param`

PR #67, opened 2022-01-27
test: add unit tests for ABI function generation

vnepveu:
This PR implements unit tests for the function expansion portion of the ABIgen.
It closes issue #6.

Issue #63, opened 2022-01-25
`expand_fn_outputs` is not consistent with type checking


vnepveu:
Having a `&[Property{type_field: "enum xxx"}]` requires providing valid `components` but not with type_field: "struct xx"

PR #73, opened 2022-01-28
refactor: check for types in custom name extraction

vnepveu:
This PR refactors code to check that the type we convert to is the one declared in ABI custom type generation.
This closes issue #61.
EDIT: built on top of #67 which should be merged first.


fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -96,7 +98,7 @@ pub fn expand_internal_struct(prop: &Property) -> Result<TokenStream, Error> {
         }
     }
 
-    let name = ident(&extract_struct_name_from_abi_property(prop).to_class_case());
+    let name = ident(&extract_custom_type_name_from_abi_property(prop, "struct")?.to_class_case());
```
digorithm, line 101: How about making this second argument an enum instead of a string? e.g.:

```rust
extract_custom_type_name_from_abi_property(prop, CustomType::Struct)
```

Then, in the function, we can use [strum](https://docs.rs/strum/latest/strum/) (we use it all over the repo already) to match the expected type, which is an enum, that we convert to string and match against `type_field[0]` like you already do.

This prevents hard-to-track bugs like calling this function passing `(prop, "strut")` or something hah.

fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -227,32 +247,53 @@ mod tests {
     use std::str::FromStr;
 
     #[test]
-    #[should_panic(expected = "index out of bounds")]
-    fn test_extract_struct_name_from_abi_property_default() {
+    fn test_extract_custom_type_name_from_abi_property_bad_data() {
         let p: Property = Default::default();
-        let _ = extract_struct_name_from_abi_property(&p);
+        let result = extract_custom_type_name_from_abi_property(&p, "");
```
digorithm, line 252: Here, for instance, having the second arg be an enum will eliminate this type of bug completely!

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -99,6 +95,7 @@ fn expand_fn_outputs(outputs: &[Property]) -> Result<TokenStream, Error> {
             }
         }
         _ => {
+            // TODO: inconsistent behavior especially between structs and enum
```
digorithm, line 98: Can you expand more on this `TODO`?

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -209,17 +206,544 @@ fn expand_input_param(
...
+        [0, 0, 0, 0, 151, 212, 222, 69],
+        &[],
+        &[bimbam.into_token(), ]
+    )
+    .expect("method not found (this should never happen)")
+}
+        "#,
+        );
+        let expected = expected.unwrap().to_string();
+        // println!("{}", result.unwrap().to_string());
```
digorithm, line 266:
```suggestion
```

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -209,17 +206,544 @@ fn expand_input_param(
...
+        let abi_parser = ABIParser::new();
+        let result = expand_function(&the_function, &abi_parser, &custom_enums, &custom_structs);
+        // Some more editing was required because it is not rustfmt-compatible (adding/removing parentheses or commas)
+        let expected = TokenStream::from_str(
+            r#"
+#[doc = "Calls the contract's `HelloWorld` (0x00000000b6425163) function"]
+pub fn HelloWorld(
+    &self,
+    bim_bap_struct: BikeHelmet,
+    burgundy_beef_enum: SomeFrenchCuisine
```
digorithm, line 386: You name your test stuff like a drunk GPT-3 ML model, I love it 😂.

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -209,17 +206,544 @@ fn expand_input_param(
...
+        assert_eq!(result.unwrap().to_string(), "()");
+        let result = expand_fn_outputs(&[Property {
+            name: "unused".to_string(),
+            type_field: "bool".to_string(),
+            components: None,
+        }]);
+        assert_eq!(result.unwrap().to_string(), "bool");
+        let result = expand_fn_outputs(&[Property {
+            name: "unused".to_string(),
+            type_field: String::from("struct streaming_services"),
```
digorithm, line 427: IIRC we're strictly following Rust naming convention in Sway, so a struct shouldn't come as `streaming_services`, it should be `StreamingServices`. Regardless, I believe `fuels-rs` should support both and turn them (through the codegen) into `StreamingServices`. Just in case.

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -209,17 +206,544 @@ fn expand_input_param(
...
+                Property {
+                    name: String::from("unused"),
+                    type_field: String::from("thistypedoesntexist"),
+                    components: None,
+                },
+            ]),
+        }]);
+
+        // the function has inconsistent  behavior for enum compared to struct:
+        // here we have to provide actual types in the components, not with the struct
```
digorithm, line 443: Yeah, I remember that. Do you mind creating an issue to eliminate this inconsistency? That's mostly because `expand_fn_outputs` doesn't explicitly handle enums, only structs:

```rust
if outputs[0].type_field.contains("struct ") {
    let tok: proc_macro2::TokenStream =
        extract_struct_name_from_abi_property(&outputs[0])
            .parse()
            .unwrap();
    Ok(tok)
} else {
    expand_type(&parse_param(&outputs[0])?)
}
```

digorithm (changes requested):
It's looking really good, really appreciate all those tests! Left a couple of comments.

fuels-core/src/code_gen/custom_types_gen.rs
vnepveu: Yes that's what I wanted to do but didn't know how you'd feel about me introducing a new type. Happy to make it cleaner!

fuels-core/src/code_gen/custom_types_gen.rs
vnepveu: Yes!

fuels-core/src/code_gen/functions_gen.rs
vnepveu: I had created issue #63

fuels-core/src/code_gen/functions_gen.rs
vnepveu: My idea was to feed it `snake_case` and put into light the fact that it turns them into `CamlCase`

vnepveu:
Normally done w/ all changes!

fuels-core/src/code_gen/functions_gen.rs
digorithm: Oh, I see, even better then! Good stuff 😄

fuels-core/src/code_gen/functions_gen.rs
digorithm: Oh!! I didn't see it on the board because you forgot to add the Github Project to it 😅. No biggie though, just added to this one, just remember to do it on the next issues as well.

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -99,6 +96,7 @@ fn expand_fn_outputs(outputs: &[Property]) -> Result<TokenStream, Error> {
             }
         }
         _ => {
+            // There, everything will be expanded, which goes against the `1` case
```
digorithm, line 99: Do you mind removing this comment? IMO it would just make it more confusing for someone reading it without full context. We've already captured the tech debt as an issue, I think that's enough. Thoughts?

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -209,17 +207,549 @@ fn expand_input_param(
...
+                },
+                Property {
+                    name: String::from("unused"),
+                    type_field: String::from("bool"),
+                    components: None,
+                },
+            ]),
+        };
+        let result = expand_fn_outputs(&[some_struct.clone(), some_struct]);
+        // TODO: this should panic after the function is refactored
```
digorithm, line 571: We need more context on this `TODO`. Maybe a link to the respective Github issue?

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -209,17 +207,549 @@ fn expand_input_param(
...
+                    type_field: String::from("enum nameunused"),
+                    // Not parsed, so can be empty but not None
+                    components: Some(vec![]),
+                },
+            ],
+            name: "PipPopFunction".to_string(),
+            outputs: vec![],
+        };
+        let mut custom_structs = HashMap::new();
+        // TODO: inconsistent behavior
```
digorithm, line 659: Same for these `TODO: inconsistent behavior`. Let's link them to the external Github issue so we give more context to other people, just in case.

digorithm (changes requested):

fuels-core/src/code_gen/functions_gen.rs
vnepveu: Not even sure what I meant, removing because it will probably resurface during the refactor.

fuels-core/src/code_gen/functions_gen.rs
vnepveu: same: it will go down with the refactoring, removing the comment.

digorithm (approved):
Great job, looking much cleaner, and love the tests!

PR #79, opened 2022-02-02
Add --no-verify flag to publish crate action

digorithm:
Our crate publishing step is still blocked by this PR: https://github.com/katyo/publish-crates/pull/394. 

However, as shown by `katyo` (author of `publish-crates`), one can work around the cyclic deps issue by using the `--no-verify` flag, which is equivalent to the `--no-verify` flag in here: https://doc.rust-lang.org/cargo/commands/cargo-publish.html. Here's the concrete example linked by the author: https://github.com/DelSkayn/rquickjs/blob/master/.github/workflows/ci.yml#L349. 

This PR enables that in order to try and unblock us for the release of the `v0.3.0`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,4 +85,5 @@ jobs:
       uses: katyo/publish-crates@v1
       with:
           publish-delay: 30000
+          args: --no-verify
           registry-token: ${{ secrets.CARGO_REGISTRY_TOKEN }}


Voxelot (approved):

adlerjohn (approved):
Hmm. This doesn't make me the most comfortable, but if it works then I guess it works. Can you open an issue to track undoing this when the fix lands upstream?

digorithm:
> Hmm. This doesn't make me the most comfortable, but if it works then I guess it works. Can you open an issue to track undoing this when the fix lands upstream?

Same, not a fan of this at all, what comforts me is that this is only triggered by our CI, which in turn is controlled by the contributors who have access to it. 

Still, not ideal and temporary while we wait for @Voxelot's PR on the publish-crates repo. Ah yes, I'll create an issue to undo this asap.

Voxelot:
It might help to add `--dry-run` to catch any issues that might come from packaging without relying on crates.io.

adlerjohn:
`--dry-run` didn't work for me, because it will fail on missing local dep. Maybe it works when combined with `--no-verify`, but should be tested in the dummy repo first. https://github.com/FuelLabs/fuel-dummy-test/pull/21

Issue #13, opened 2021-11-30
Remove workaround placed to handle temporary mandatory arguments


digorithm:
As discussed in https://github.com/FuelLabs/fuels-rs/pull/11#discussion_r758243462, once the Sway compiler removes the need to have gas, color, and coin amount as mandatory arguments to every ABI method, we should go back to this place in the `fuels-rs` code and remove the workaround.

adlerjohn:
We can tackle this after https://github.com/FuelLabs/sway/pull/859 lands and a new release of the compiler is cut

adlerjohn:
https://github.com/FuelLabs/sway/pull/859 has been merged, so we only need a new release of the compiler.

vnepveu:
Compiler was updated to 0.6 with https://github.com/FuelLabs/sway/commit/35e5a2f12e137953b82470e9637e4969ca064e4a

PR #83, opened 2022-02-04
refactor: add an argument to enable strict argument checking in abigen macro

### LINKED ISSUE #75: abigen: exclude first 3 fields instead of name matching
https://github.com/FuelLabs/fuels-rs/blob/eb0f17c5907cb17b54c302f60283ea4f8f483605/fuels-core/src/code_gen/functions_gen.rs#L139

Currently, only fields with specific names are ignored. This isn't really documented anywhere and is the source of endless confusion. Moreover, currently the compiler requires exactly 4 fields from ABI methods: 3 implicit and one for user parameters. Therefore, there's no reason to prune on matching names when the first 3 can simply be pruned.

vnepveu:
This PR refactors the function `expand_function_arguments` to avoid checking for template arguments (`gas_, amount_, color_`) every time. It propagates the argument up to the abigen macro so user can chose if the expansion should check for types or not. 
It closes #75

### DIFF
--- a/fuels-core/src/code_gen/abigen.rs
+++ b/fuels-core/src/code_gen/abigen.rs
@@ -35,6 +35,10 @@ pub struct Abigen {
 
     /// Generate no-std safe code
     no_std: bool,
+
+    /// Check that the arguments are exactly as follows:
+    /// [gas_(u64), amount_(u64), color_(b256), custom(T)]
+    strict_checking: bool,
 }
 
 impl Abigen {
@@ -65,11 +69,17 @@ impl Abigen {
             abi_parser: ABIParser::new(),
             rustfmt: true,
             no_std: false,
+            strict_checking: false,
         })
     }
 
     pub fn no_std(mut self) -> Self {
         self.no_std = true;
+        self
+    }
+
+    pub fn strict_checking(mut self) -> Self {
+        self.strict_checking = true;
         self
     }
 
@@ -160,6 +170,7 @@ impl Abigen {
                 &self.abi_parser,
                 &self.custom_enums,
                 &self.custom_structs,
+                &self.strict_checking,
             )?;
             tokenized_functions.push(tokenized_fn);
         }
--- a/fuels-core/src/code_gen/abigen.rs
+++ b/fuels-core/src/code_gen/abigen.rs
@@ -35,19 +35,11 @@ pub struct Abigen {
 
     /// Generate no-std safe code
     no_std: bool,
-
-    /// Check that the arguments are exactly as follows:
-    /// [gas_(u64), amount_(u64), color_(b256), custom(T)]
-    strict_checking: bool,
 }
 
 impl Abigen {
     /// Creates a new contract with the given ABI JSON source.
-    pub fn new<S: AsRef<str>>(
-        contract_name: &str,
-        abi_source: S,
-        strict_checking: bool,
-    ) -> Result<Self, Error> {
+    pub fn new<S: AsRef<str>>(contract_name: &str, abi_source: S) -> Result<Self, Error> {
         let source = Source::parse(abi_source).unwrap();
         let mut parsed_abi: JsonABI = serde_json::from_str(&source.get().unwrap())?;
 
@@ -73,17 +65,11 @@ impl Abigen {
             abi_parser: ABIParser::new(),
             rustfmt: true,
             no_std: false,
-            strict_checking,
         })
     }
 
     pub fn no_std(mut self) -> Self {
         self.no_std = true;
-        self
-    }
-
-    pub fn strict_checking(mut self) -> Self {
-        self.strict_checking = true;
         self
     }
 
@@ -174,7 +160,6 @@ impl Abigen {
                 &self.abi_parser,
                 &self.custom_enums,
                 &self.custom_structs,
-                self.strict_checking,
             )?;
             tokenized_functions.push(tokenized_fn);
         }
@@ -288,6 +273,21 @@ mod tests {
                 "type":"contract",
                 "inputs":[
                     {
+                        "components": null,
+                        "name": "gas_",
+                        "type": "u64"
+                    },
+                    {
+                        "components": null,
+                        "name": "amount_",
+                        "type": "u64"
+                    },
+                    {
+                        "components": null,
+                        "name": "color_",
+                        "type": "b256"
+                    },
+                    {
                         "name":"arg",
                         "type":"u32"
                     }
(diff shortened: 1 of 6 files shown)


vnepveu:
Honestly had trouble coming up with this PR, so really taking *any* criticisms to improve the design I chose :smile_cat:

digorithm (changes requested):
You see, there seems to be a (very valid) misunderstanding that will lead to worse ergonomics of the `abigen!` macro -- which is already relatively hard to understand for newcomers.

Adding a `bool` to the `abigen!` will make things much more confusing, and here's why this ended up being a valid (mis)direction: this happened because in _some_ of the tests in `harness.rs`, the JSON ABI being used was _handcrafted_ and didn't follow what the compile generates, that's because it was handcrafted _before_ the compiler was able to generate the JSON ABI. 

In practice, this should never happen. In other words, I don't believe there's a reason to complicate the devEx around `abigen!` because users will use the JSON ABI generated by the compiler, which includes the mandatory 3 params. Meaning it should always follow the "`strict_checking`" logic introduced in this PR. 

The original issue (https://github.com/FuelLabs/fuels-rs/issues/75) was created because, in Sway, if you give a different name to one of the mandatory params, it will show up in the JSON ABI generated by the compiler that way. E.g. if your sway fn is `do_something(gass: u64, amount: u64, color_: b256, MyInput: u64)` (notice the inconsistency), your JSON ABI will be generated the same way and the SDK shouldn't care, it should only care that's `u64, u64, b256`. This is a common mistake by Fuel users as we've seen in discussions on Discord. 

I propose that we remove this extra `bool` param in the `abigen!` and use only the (correct) logic introduced in this PR that does the strict checking: 

```rust
if strict_checking {
    if n_inputs != 4 {
        return Err(Error::MissingData(format!(
            "Expected 4 inputs because `strict_checking` is true, found {}",
            n_inputs
        )));
    }
    let given_types: Vec<String> = fun.inputs[..3]
        .iter()
        .map(|x| x.type_field.clone())
        .collect();
    let expected = ["u64", "u64", "b256"];
    if given_types != expected {
        return Err(Error::InvalidType(format!(
            "Expected the 3 first types to be {:?}, found {:?}",
            expected, given_types
        )));
    };
    // Ignore the first three arguments so we don't have to provide them when calling
    // the contracts' methods
    first_index = 3
}
```

Except that this will be the only path (no `if strict_checking {}`). 

Then we update all tests to include the 3 mandatory params in the JSON ABI, _as it should've been_. So that it matches the output of the compiler. 

This will effectively solve the friction that users have while incorrectly naming the 3 mandatory params _in the Sway code_ and not complicate the `abigen!` macro (and not break it for current users!).

In the future, and I already have this documented somewhere, we should have the option of passing, instead of a compiler-generated JSON ABI, a human-readable list of function signatures to `abigen!` that could be easily handcrafted, just like ethers provide (https://docs.ethers.io/v5/api/utils/abi/formats/#abi-formats--human-readable-abi). This should be the place for users to handcraft the ABI, not in the JSON ABI that's generated by the compiler.

cc @adlerjohn

vnepveu:
Ok I understand, it's just that John asked in #75 that we have "two" versions of the `abigen`, I thought about doing one `abigen_strict` and one `abigen` but figured factoring it would make more sense. Working on the requested changes: updating the logic and the tests.

digorithm:
> Ok I understand, it's just that John asked in #75 that we have "two" versions of the `abigen`, I thought about doing one `abigen_strict` and one `abigen` but figured factoring it would make more sense. Working on the requested changes: updating the logic and the tests.

Yup, I believe John asked that because you mentioned the tests that didn't include the 3 params. These tests are "invalid" because they should be using the compiler output. This distinction shouldn't be necessary under the (now corrected) assumption that the tests themselves should've included the 3 params.

adlerjohn:
Ah I guess I wasn't clear. Yeah a `abigen!` and `abigen_strict!` would have been better than a Boolean parameters into a single `abigen!`. Either way, if @digorithm says that abigen should only work on things output from the compiler, then he's the boss.

digorithm:
> Ah I guess I wasn't clear. Yeah a `abigen!` and `abigen_strict!` would have been better than a Boolean parameter into a single `abigen!`. Either way, if @digorithm says that abigen should only work on things output from the compiler, then he's the boss.

IMO, there's a place for handcrafted input and another place for auto-generated inputs. That's one of the things that's nice about `ethers-rs` (and `ethers.js`); you have the JSON ABI that's auto-generated and more strict and the handcrafted human-readable version (https://github.com/FuelLabs/fuels-rs/pull/83#issuecomment-1030321543). 

Trying to do both, in JSON, and on top of that adding more complexity to the macro... what a pain 😢.

vnepveu:
Updated according the remarks, which changed a lot of tests along the way. I'm always trying to factorize as much as possible, but I guess for macros it's not necessarily a good idea?

fuels-core/src/code_gen/abigen.rs
```diff
@@ -464,8 +503,8 @@ mod tests {
             assert!(contract.custom_structs.contains_key(name));
         }
 
-        let bindings = contract.generate().unwrap();
-        bindings.write(std::io::stdout()).unwrap();
+        let bindings = contract.generate();
+        assert!(matches!(bindings, Err(Error::MissingData(_))));
```
digorithm, line 507: Do you mind adding a comment here explaining why we're expecting to get this error? The name of the test itself doesn't mention that, the reader would be confused. Would be nice to explain that we're just testing multiple inputs generations at once and that would exceed the limit (4).

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -115,35 +113,31 @@ fn expand_function_arguments(
...
+        return Err(Error::InvalidType(format!(
+            "Expected the 3 first types to be {:?}, found {:?}",
+            expected, given_types
+        )));
+    };
+    // Ignore the first three arguments so we don't have to provide them when calling
+    // the contracts' methods
+    let first_index = 3;
+    let mut args = Vec::with_capacity(1);
+    let mut call_args = Vec::with_capacity(1);
```
digorithm, line 140: We should add one more comment here to give some clarity around this logic. The for loop below: 

```rust
for (i, param) in fun.inputs[first_index..].iter().enumerate() {}
```

Only runs once under the assumption that we have a fixed input size of 4. A newcomer to the codebase would then ask, "why not just expand on the last element of `fun.inputs` then? Why the for loop?".

The only reason we're keeping this for loop is that we might, in the future, lift this restriction on the max number of inputs (4). So the for loop will make more sense then.

digorithm (changes requested):
It's looking great! I played with it a bit and now inconsistent/wrong names for the first 3 params don't break anything 🎉. Great work!

I just left some requests for additional comments.

vnepveu:
It also closes #13 right?

adlerjohn:
> It also closes https://github.com/FuelLabs/fuels-rs/issues/13 right?

Not really, because #13 is about removing workaround altogether. I should probably be re-worded given the changes that have been made since then, but we still need an issue to track removing workarounds for the mandatory arguments altogether.

digorithm (approved):
Ship it!

Issue #90, opened 2022-02-09
Add random change to script transaction


digorithm:
Full context and discussion here: https://github.com/FuelLabs/fuels-rs/issues/89. 

The overall idea is to enable contracts calls that may be view-only to be able to be called multiple times without causing tx id collision. 

This will be a temporary workaround for the lack of explicit view-only ABI methods and different ways of calling different types of functions (`.call()`, `.send()`, etc, as you see in Solidity).

Later, we should explore other potential solutions.

Issue #89, opened 2022-02-08
Response errors; unexpected block execution error OutputAlreadyExists When Running Multiple Tests


ControlCplusControlV:
So was finally getting around to debugging my [Sushi Sway Repo](https://github.com/ControlCplusControlV/Sushi-Sway/tree/45e3e0509bfc347bf0818cd45a55394b22d3242b) my withdraw test was failing with the following error

```
thread 'withdraw' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "Response errors; unexpected block execution error OutputAlreadyExists" }', /hom/me/.cargo/registry/src/github.com-1ecc6299db9ec823/fuels-rs-0.2.1/src/script.rs:28:56
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test withdraw ... FAILED

failures:

failures:
    withdraw
```
It seems there seems to be some concurrency failure when running multiple tests at once. If this an error in my code though let me know and I can close the issue

digorithm:
Tracking this here https://github.com/FuelLabs/fuels-rs/issues/90

PR #92, opened 2022-02-10
Create random input coin in tx

digorithm:
Fixes https://github.com/FuelLabs/fuels-rs/issues/90.


digorithm:
@adlerjohn Need your input on this one. All tests and the new test that reproduces the bug discussed in https://github.com/FuelLabs/fuels-rs/issues/89 are passing and seem to be working fine. 

I'm just curious if I'm missing something that could cascade from using a random coin input. So far all is good, but I'm wondering about cases not covered in the tests.

vnepveu (approved):
LGTM, but maybe we should also create an issue about potential concurrency issues with tests as John and CCCV mentioned (or just create tests that "show" there aren't any as a sanity check).

adlerjohn (approved):

digorithm:
Fair enough, I'll update this comment on a next upcoming PR.

PR #95, opened 2022-02-10
fix: enable having an empty fourth argument for non-user methods

### LINKED ISSUE #91: Support contract methods with only 3 params (no user input param)
Right now we're strictly looking for a 4th param:

```rust
if n_inputs != 4 {
    return Err(Error::MissingData(format!(
        "Expected exactly 4 inputs, found {}",
        n_inputs
    )));
}
```

But there are cases like this one:

```
fn get_counter(gas_: u64, amount_: u64, color_: b256, input: ()) -> u64 {
  get(COUNTER_KEY)
}
```

Where the user doesn't send the 4th input. 

This causes:

```
help: message: called `Result::unwrap()` on an `Err` value: MissingData("cannot parse custom type with no components")
```

Because since `()` isn't a primitive type, it's trying to create a custom type without components. 

We should allow the 4th param to be `()` and not try expanding it when going through the `abigen!`.

vnepveu:
This PR fixes #91 and implements test to demonstrate the efficacy of the fix.
* before this fix, methods that had a forth argument automatically set
  to `()` failed (because it was perceived as a custom type). We now
  filter out these cases.


digorithm (changes requested):
Hah, such a simple change, nice!

Could you add a full (with an actual test project with a single ABI method with a `()` input) test case under `harness.rs`?

vnepveu:
Ok I'm definitely sure there is "flakiness" in the way the contract calls are run. One moment I get the expected values, the other it's all zeroes. Tried multiple things, didn't diagnose properly. I'm going to try to repro this. @sezna could would you mind looking at my harness/sway code to detect what could be causing this, if it's human error on my side? Thanks :smile_cat:

digorithm:
> Ok I'm definitely sure there is "flakiness" in the way the contract calls are run

You mean in this new test you just added right? Other tests have never shown any sort of flakiness.

vnepveu:
> > Ok I'm definitely sure there is "flakiness" in the way the contract calls are run
> 
> You mean in this new test you just added right? Other tests have never shown any sort of flakiness.

Yes of course! I'm just really disturbed because it's really simple contracts and the outputs are really variable. Other tests obviously run fine.

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -147,6 +147,10 @@ fn expand_function_arguments(
         // Note that _any_ significant change in the way the JSON ABI is generated
         // could affect this function expansion.
         // TokenStream representing the name of the argument
+        if param.type_field == "()" || param.type_field == "" {
```
adlerjohn, line 150: Is there a reason to have both these cases? Presumably the compile will only use one, so only one should be used here.

fuels-abigen-macro/tests/test_projects/empty-arguments/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+name = "method_four_arguments"
+version = "0.1.0"
+authors = ["Fuel Labs <contact@fuel.sh>"]
+edition = "2021"
+license = "Apache-2.0"
+
+[dependencies]
+fuels-abigen-macro = "0.2"
+fuels-core = "0.2"
+fuels-rs = "0.2"
```
adlerjohn, line 11: Shouldn't this use a local dep instead of a pinned dep? You want this test to test the current SDK.

adlerjohn (changes requested):

adlerjohn (commented):
Merge conflicts

adlerjohn:
I was able to repro the failing test, _but only sometimes_. Other times, it passes. I think the SDK isn't handling an async error properly, and instead returning a success with a value of `0`.

fuels-core/src/code_gen/functions_gen.rs
vnepveu: I will remove the `""` if you think that's confusing

fuels-abigen-macro/tests/test_projects/empty-arguments/Cargo.toml
vnepveu: True! Using `forc init` is a trap!

vnepveu:
Amusing: doing the changes in [0cb759c](https://github.com/FuelLabs/fuels-rs/pull/95/commits/0cb759cf3c707daf9d8b141444b490255285bf3d) make the test run fine for me now. Reverting "by hand" the changes made in the commit do *not* make it fail again.

digorithm (approved):

sezna:
Just checking in here -- was the nondeterminism resolved?

PR #100, opened 2022-02-15
remove the Url crate from fuels-core

### LINKED ISSUE #93: abigen! macro fails to read .json file on Windows
I found a bug where the `abigen!` macro works for me when compiling on my mac, but fails when building on windows.

The error i'm getting is:

```
message: called `Result::unwrap()` on an `Err` value: unable to canonicalize file from working dir C:\Users\conta\Desktop\counter_project with path /C:/Users/conta/Desktop/counter_project/my_contract_abi.json
```

I'll take a look into this.

JoshuaBatty:
This removes the use of the `Url` crate for loading ABI .json files. This approach was incompatible with Windows paths so has been removed in order to support that OS. [See here](https://github.com/FuelLabs/fuels-rs/issues/93) for more discussion on this issue.

closes #93

### DIFF
--- a/fuels-core/Cargo.toml
+++ b/fuels-core/Cargo.toml
@@ -28,4 +28,3 @@ sway-types = { version = "0.1" }
 sway-utils = { version = "0.1" }
 syn = "1.0.12"
 thiserror = "1.0.30"
-url = "2.1"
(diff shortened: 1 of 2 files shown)


digorithm (approved):
Short and sweet, good stuff!

PR #103, opened 2022-02-16
feat: capture error from reverting transactions in the SDK

### LINKED ISSUE #101: Create `examples/` directory centralizing all SDK examples
Currently, all our examples are scattered across many tests, both in the "central" `harness.rs` under `fuels-abigen-macro` and in smaller tests across the repo.

We should have a root-level `examples/` directory containing many different examples with the most common use-cases.

### LINKED ISSUE #102: Investigate why some errors at the fuel-core / fuel-vm level aren’t being captured by the SDK
Sometimes, some errors do not get captured at all by the SDK. For an example of that, see branch https://github.com/FuelLabs/fuels-rs/tree/repro-storage-slots where the 2nd transaction is actually a reverting transaction with `MemoryOverflow` error, and nothing appears in the receipts.

vnepveu:
* Previously, a reverting transaction would not be captured by the SDK.
* This closes #102 .

### DIFF
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -3,6 +3,7 @@ use fuel_gql_client::client::FuelClient;
 use fuel_tx::Salt;
 use fuels_abigen_macro::abigen;
 use fuels_contract::contract::Contract;
+use fuels_contract::errors::Error;
 use fuels_core::Token;
 use rand::rngs::StdRng;
 use rand::{Rng, SeedableRng};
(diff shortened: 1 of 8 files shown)


fuels-contract/src/script.rs
```diff
@@ -25,11 +25,14 @@ impl Script {
...
 
     pub async fn call(self, fuel_client: &FuelClient) -> Result<Vec<Receipt>, Error> {
-        let tx_id = fuel_client.submit(&self.tx).await.unwrap();
+        let tx_id = fuel_client.submit(&self.tx).await.unwrap().0.to_string();
 
-        let receipts = fuel_client.receipts(&tx_id.0.to_string()).await.unwrap();
-
-        Ok(receipts)
+        let receipts = fuel_client.receipts(&tx_id).await.unwrap();
+        let status = fuel_client.transaction_status(&tx_id).await.unwrap();
```
digorithm, line 31: Since we're already making some improvements to the error handling in this section, can we get rid of these `unwrap()`s and do some neat error conversions so that we return an `Error` here instead of panicking?

digorithm (changes requested):
This is awesome; major debugging QoL improvement. Just left one small request.

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1335,6 +1336,30 @@ async fn abigen_different_structs_same_arg_name() {
...
+    // Build the contract
+    let salt: [u8; 32] = rng.gen();
+    let salt = Salt::from(salt);
+
+    let compiled =
+        Contract::compile_sway_contract("tests/test_projects/revert_transaction_error", salt)
+            .unwrap();
+
+    let (client, contract_id) = Contract::launch_and_deploy(&compiled).await.unwrap();
+    println!("Contract deployed @ {:x}", contract_id);
```
vlopes11, line 1357: Nit: shouldn't include `println` statements

vlopes11 (commented):
Small nit

adlerjohn:
cc @digorithm for re-review

digorithm (approved):
Nice 👍

Issue #68, opened 2022-01-28
Create an interface in `Wallet` to send transactions


digorithm:
`Contract`s already can send transactions but we need a way to send "standalone" transactions. For instance, how it's done with `ethers.js`:

```js
walletSigner.sendTransaction(tx).then((transaction) => {
// ...
})
```

We should extend `fuels-signers`'s Wallet abstraction to include this functionality.

Here's a more complete reference for how it could look like: https://ethereum.org/en/developers/tutorials/send-token-etherjs/

Issue #31, opened 2022-01-06
Coin selection API


digorithm:
Once https://github.com/FuelLabs/fuel-core/issues/121 is done, the SDK needs an API to select which UTXOs to use as inputs.

Voxelot:
https://github.com/FuelLabs/fuel-core/issues/121 is done and will be ready to start integrating into the SDK.

PR #110, opened 2022-02-23
Send transactions and transfer funds between wallets

digorithm:
Closes https://github.com/FuelLabs/fuels-rs/issues/68.
Closes https://github.com/FuelLabs/fuels-rs/issues/31. 

Initially, this was targeting only transaction sending, but the use case I was solving for was... sending coins, so one thing led to another, and I needed to have a proper coin selecting API in place first. So I bundled it all together in this PR.

This enables not only sending signed transactions but also adds an API in the wallet/signer for sending coins between wallets: 

```rust
// Setup two sets of coins, one for each wallet, each containing 1 coin with 1 amount.
let (pk_1, mut coins_1) = setup_address_and_coins(1, 1);
let (pk_2, coins_2) = setup_address_and_coins(1, 1);

coins_1.extend(coins_2);

// Setup a provider and node with both set of coins
let provider = setup_test_provider(coins_1).await;

let wallet_1 = LocalWallet::new_from_private_key(pk_1, provider.clone()).unwrap();
let wallet_2 = LocalWallet::new_from_private_key(pk_2, provider).unwrap();

let wallet_1_initial_coins = wallet_1.get_coins().await.unwrap();
let wallet_2_initial_coins = wallet_2.get_coins().await.unwrap();

// Check initial wallet state
assert_eq!(wallet_1_initial_coins.len(), 1);
assert_eq!(wallet_2_initial_coins.len(), 1);

// Transfer 1 from wallet 1 to wallet 2
let _receipts = wallet_1
    .transfer(&wallet_2.address(), 1, Default::default())
    .await
    .unwrap();

// Currently ignoring the effect on wallet 1, as coins aren't being marked as spent for now
let _wallet_1_final_coins = wallet_1.get_coins().await.unwrap();
let wallet_2_final_coins = wallet_2.get_coins().await.unwrap();

// Check that wallet two now has two coins
assert_eq!(wallet_2_final_coins.len(), 2);

// Transferring more than balance should fail
let result = wallet_1
    .transfer(&wallet_2.address(), 2, Default::default())
    .await;

assert!(result.is_err());
let wallet_2_coins = wallet_2.get_coins().await.unwrap();
assert_eq!(wallet_2_coins.len(), 2); // Not changed
```

It also handles `Output::Change`. Note that the sent UTXOs aren't marked as spent because that depends on a work in progress on `fuel-core`. 

**Note for when this is merged: add examples in the Sway book showing this functionality.**


digorithm:
@Voxelot adding you as a reviewer for this one because it uses some new stuff introduced in `fuel-core` and it implements some stuff we discussed recently (UTXOs, etc). Would appreciate your review of it.

fuels-signers/src/provider.rs
```diff
@@ -0,0 +1,108 @@
...
+                None,
+            )
+            .await?;
+
+        Ok(res)
+    }
+
+    /// Craft a transaction used to transfer funds between two addresses.
+    pub fn build_transfer_tx(&self, inputs: &[Input], outputs: &[Output]) -> Transaction {
+        let script = Opcode::RET(0x10).to_bytes().to_vec();
```
vnepveu, line 89: `// the script does nothing so just return and add a receipt (RET) immediately `

vnepveu: Also why `RET(0x10)` and not `RET(0x0)`?

fuels-signers/src/provider.rs
```diff
@@ -0,0 +1,108 @@
...
+            outputs: outputs.to_vec(),
+            witnesses: vec![],
+            metadata: None,
+        }
+    }
+
+    // @todo
+    // - Get transaction(s)
+    // - Get block(s)
+}
```
vnepveu, line 108: We should probably have unit tests for provider

fuels-signers/src/util.rs
```diff
@@ -0,0 +1,80 @@
...
+        let public = PublicKey::from_secret_key(&secp, &secret).serialize_uncompressed();
+        let public = Bytes64::try_from(&public[1..]).unwrap();
+        let hashed = Hasher::hash(public);
+
+        let coins: Vec<(UtxoId, Coin)> = (1..=num_of_coins)
+            .map(|_i| {
+                let coin = Coin {
+                    owner: Address::from(*hashed),
+                    amount,
+                    color: Default::default(),
```
vnepveu, line 51: Since `color` is ~ to `asset_id`, shouldn't we have 1 color per `num_of_coins` ie `num_of_coins` is the number of different coins in the wallet? If `num_of_coins` is the number of different UTXOs containting the same `asset_id`, then the name should be `num_of_utxo`

fuels-signers/src/wallet/mod.rs
```diff
@@ -83,8 +100,144 @@ impl Wallet {
...
+    /// `get_spendable_coins`.
+    pub async fn get_spendable_coins(&self, color: &Color, amount: u64) -> io::Result<Vec<Coin>> {
+        Ok(self
+            .provider
+            .get_spendable_coins(&self.address(), *color, amount)
+            .await?)
+    }
 }
 
 #[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
```
vnepveu, line 243: `transfer` should probably have unit tests too, since it is a non-trivial process ( checking the amount, change …)

vnepveu (changes requested):
I feel like we could use some more unit tests. Also some remarks regarding UTXOs that might be completely wrong. Otherwise LGTM!

fuels-signers/src/provider.rs
digorithm: > `// the script does nothing so just return and add a receipt (RET) immediately `

Doesn't the naming `build_transfer_tx` and the docstring already imply that?

> Also why RET(0x10) and not RET(0x0)?

https://github.com/FuelLabs/fuel-core/blob/master/fuel-client/tests/tx.rs#L428

fuels-signers/src/provider.rs
digorithm: The tests on the wallet are already testing the provider, as they're just thin wrappers on the provider's methods.

fuels-signers/src/wallet/mod.rs
digorithm: https://github.com/FuelLabs/fuels-rs/pull/110/files#diff-f3c02edc378fe2fe43610a513ab6b6d5cd57233e86d1794f2f564030ce385d50R178-R216

fuels-signers/src/provider.rs
digorithm: Fair enough, I'll add this comment 😄

vnepveu (dismissed):

fuels-signers/Cargo.toml
```diff
@@ -11,8 +11,10 @@ description = "Fuel Rust SDK signers."
 [dependencies]
 async-trait = { version = "0.1.50", default-features = false }
 bytes = { version = "1.1.0", features = ["serde"] }
+fuel-core = { version = "0.3.2", default-features = false, features = ["test-helpers"] }
+fuel-gql-client = { version = "0.3.2", default-features = false }
 fuel-tx = "0.5"
-fuel-types = { version = "0.1", default-features = false }
+fuel-types = { version = "0.1", default-features = false, features = ["random"] }
```
Voxelot, line 17: Typically we only enable the "random" features in the dev-deps section, since they're only used in testing. This makes use of v2 feature resolver to avoid unifying random impls into the production code.

digorithm: Yeah, nice catch. Just updated it.

fuels-signers/src/util.rs
```diff
@@ -0,0 +1,80 @@
+#[allow(missing_docs)]
+/// Testing utilities
+pub mod test_helpers {
```
Voxelot, line 3: Since these are publically exported, I'd recommend putting these under a feature flag.

```suggestion
#[cfg(feature = "test-helpers")]
pub mod test_helpers {
```

fuels-signers/Cargo.toml
```diff
@@ -11,6 +11,8 @@ description = "Fuel Rust SDK signers."
 [dependencies]
 async-trait = { version = "0.1.50", default-features = false }
 bytes = { version = "1.1.0", features = ["serde"] }
+fuel-core = { version = "0.3.2", default-features = false, features = ["test-helpers"] }
```
Voxelot, line 14:
```suggestion
fuel-core = { version = "0.3.2", default-features = false }
```

fuels-signers/Cargo.toml
```diff
@@ -25,3 +27,4 @@ tokio = { version = "1.10.1", features = ["full"] }
```
Voxelot, line 27: If you add your own test-helpers flag, you can conditionally enable fuel-core's test-helpers based on your flag:

```suggestion
[features]
test-helpers = ["fuel-core/test-helpers"]
```

fuels-signers/Cargo.toml
Voxelot: fuel-core is only needed when using the test-helpers

fuels-signers/src/util.rs
digorithm: That's what I did initially, but we have doc tests (https://github.com/FuelLabs/fuels-rs/pull/110/files#diff-dabbabe6828f1fb7638ca0ace7dd311af93e4f83fc42d234246fdef19d086cf0R140) that's importing these `test_helpers`, and in doc level examples you can't import anything behind a feature flag without doing very hacky things: https://users.rust-lang.org/t/doctests-that-require-a-non-default-feature-is-it-possible/29529/4

fuels-signers/src/provider.rs
Voxelot: Using `REG_ONE` might be less cryptic and also valid vs `0x10`.

fuels-signers/src/wallet/mod.rs
```diff
@@ -83,8 +100,144 @@ impl Wallet {
...
+        let mut outputs: Vec<Output> = vec![];
+
+        let mut amount_to_send = 0;
+
+        let output_coin = Output::coin(*to, amount, color);
+        outputs.push(output_coin);
+
+        for coin in spendable {
+            if amount_to_send >= amount {
+                break;
```
Voxelot, line 191: This is counter productive to the random-improve algorithm. The intention is that the use should spend **_all_** the coins returned by the API, as the complete combination of coins should reduce dust in the future.

Voxelot: For more details, check out the algorithm docs here: https://input-output-hk.github.io/cardano-coin-selection/haddock/cardano-coin-selection-1.0.1/Cardano-CoinSelection-Algorithm-RandomImprove.html

Basically we want the input to be on avg 2*amount, that way the change output will be approximately the right size for the next transaction.

digorithm: Right, so the API is returning the exact amount of coins to be spent (or error). But do we still have to create one `Input::Coin` per spendable coin? So we would still keep the loop to create the inputs but remove the `break;`? i.e.

```rust
for coin in spendable {
    let input_coin = Input::coin(
        UtxoId::from(coin.utxo_id),
        coin.owner.into(),
        coin.amount.0,
        color,
        0,
        0,
        vec![],
        vec![],
    );

    inputs.push(input_coin);

    amount_to_send += coin.amount.0;
}

// Calculate change, if applicable
if amount_to_send > amount {
    let change = Output::change(self.address(), amount_to_send - amount, color);
    outputs.push(change);
}
```

fuels-signers/src/util.rs
Voxelot: I think you should be able to just simply enable all the deps and features you need in the dev-deps to get the rust docs working..

e.g. 
```
[dev-dependencies]
fuels-rs = { path = ".", version = "0.x", features = ["test-helpers"] }
```

digorithm: Oh, I see, TIL. That works!

fuels-signers/src/wallet/mod.rs
Voxelot: Yep that's correct, although you don't need to calculate the change here since the block producer will do that automatically based on the gasPrice vs gasUsed etc...

Voxelot: You still need to set who the owner of the change will be, but the amount is automatically calculated for every change output based on the color and freebalances.

digorithm: So that logic above can become just this?

```rust
for coin in spendable {
    let input_coin = Input::coin(
        UtxoId::from(coin.utxo_id),
        coin.owner.into(),
        coin.amount.0,
        color,
        0,
        0,
        vec![],
        vec![],
    );

    inputs.push(input_coin);
}

outputs.push(Output::change(self.address(), 0, color));
```

fuels-signers/src/wallet/mod.rs
```diff
@@ -1,15 +1,17 @@
+use crate::provider::{Provider, ProviderError};
```
Voxelot, line 1: `src/wallet/mod.rs` is the outdated Rust 2015 style. We should just have a `src/wallet.rs` file instead.

fuels-signers/src/wallet/mod.rs
```diff
@@ -22,6 +24,8 @@ use thiserror::Error;
 /// use fuels_signers::{LocalWallet, Signer};
 /// use secp256k1::SecretKey;
 /// use rand::{rngs::StdRng, RngCore, SeedableRng};
+/// use fuels_signers::provider::Provider;
+/// use fuels_signers::util::test_helpers::setup_local_node;
 ///
 /// # async fn foo() -> Result<(), Box<dyn std::error::Error>> {
```
Voxelot, line 30:
```suggestion
/// async fn foo() -> Result<(), Box<dyn std::error::Error>> {
```

fuels-signers/src/wallet/mod.rs
Voxelot: yep, exactly!

digorithm: Neat!

Voxelot: I think this release of the node your using should have correct change amounts already, but the unreleased version on master is more complete.

digorithm: Yes!! I just noticed (last commit here) that the test for sending + change is still working even without calculating the change here. 😄

fuels-signers/src/wallet.rs
```diff
@@ -0,0 +1,264 @@
...
+                vec![],
+            );
+
+            inputs.push(input_coin);
+        }
+
+        // Build transaction and sign it
+        let mut tx = self.provider.build_transfer_tx(&inputs, &outputs);
+        let sig = self.sign_transaction(&tx).await.unwrap();
+        self.append_signatures(&mut tx, &[sig]);
```
Voxelot, line 205: Would it make sense to combine these into one API?

digorithm: You mean once you `sign_transaction(&tx)` it would mutate `&tx` by appending the sign to it?

Voxelot: Yep exactly. It would simplify the most common usecases. We could expose the underlying keystore signature api for arbitrary data if people need that later.

digorithm: That's a good idea!

Voxelot: I don't think this is a blocker though.

Voxelot (approved):

vnepveu (dismissed):

fuels-signers/src/wallet.rs
```diff
@@ -0,0 +1,264 @@
...
+        to: &Address,
+        amount: u64,
+        color: Color,
+    ) -> io::Result<Vec<Receipt>> {
+        let spendable = self.get_spendable_coins(&color, amount).await?;
+
+        let mut inputs: Vec<Input> = vec![];
+        let outputs: Vec<Output> = vec![
+            Output::coin(*to, amount, color),
+            Output::change(self.address(), 0, color),
```
vnepveu, line 184: The fact that change is 0 is not trivial: you have to know that the node itself will "fill in" the right value

fuels-signers/src/wallet.rs
digorithm: I did it now anyway, it was a pretty simple change.

fuels-signers/src/wallet/mod.rs
Voxelot: Nice!

fuels-signers/src/wallet.rs
```diff
@@ -0,0 +1,277 @@
...
+    pub fn set_provider(&mut self, provider: Provider) {
+        self.provider = provider
+    }
+
+    /// Append the witnesses (signatures) for the transaction.
+    /// Note that it mutates the `Transaction` being passed into it
+    /// effectively updating its `Witness` field with the `Signature`s.
+    /// Previous witnesses in this `Transaction` are preserved, allowing
+    /// the caller to implement multisig (transactions with many signatures).
+    pub fn append_signatures(&self, tx: &mut Transaction, sigs: &[Signature]) {
```
Voxelot, line 116: This api is no longer used, should we just remove it to avoid confusion?

Voxelot:
Eventually, we'll need a way to track which coins are on a pending tx and exclude them from the coin selection API..

digorithm:
> Eventually, we'll need a way to track which coins are on a pending tx and exclude them from the coin selection API..

Would that be happening in the client itself? i.e. when executing a method from the coin selection API it would check for the pending TX and not include that coin when returning the result to the caller?

Voxelot (approved):

Voxelot:
Yeah I'm assuming we could query the txpool from the graphql API and filter any coins that are being tracked in that dependency graph.

digorithm:
Yeah, sounds like a good plan. Do we have an issue for that already?

Voxelot:
Just added: https://github.com/FuelLabs/fuel-core/issues/185

vnepveu (approved):

Voxelot:
The sdk might also want to track recently used utxos, just in case a series of requests to `coins_to_spend` get round-robined between multiple nodes before their txpools sync up.

digorithm:
> The sdk might also want to track recently used utxos, just in case a series of requests to `coins_to_spend` gets round-robined between multiple nodes before their txpools sync up. In which case, maybe the `coins_to_spend` could take a list of utxos to exclude from the search provided by the sdk.

That makes sense, let's track that on a separate issue as well.

PR #112, opened 2022-02-24
Handle return data larger than a word

digorithm:
When the returned data from a contract call is larger than a word,
it isn't stored in `Return.val`, but in `ReturnData.data`.

Before the changes in here, we were pulling the data only from
`Return.val`, which fortunately handled most cases. Until now,
with some methods returning, for instance, a `b256`, which would cause
the data to _not_ be in `Return.val` as we expected, leading to errors.

This PR introduces some changes to fix this by looking whether the
expected return type is larger than a word, if it is, we fetch the
returned data from `ReturnData.data` instead of `Return.val`.


digorithm:
@nfurfaro tagging you in this PR because it fixes the issue you found and includes the test cases you used to find the problem. These test cases are now part of the test suite for the SDK.

nfurfaro (approved):
LGTM!

fuels-core/src/lib.rs
```diff
@@ -44,6 +44,25 @@ impl Default for ParamType {
     }
 }
 
+impl ParamType {
+    // Checks whether the `ParamType` is bigger than a `WORD`
+    pub fn bigger_than_word(&self) -> bool {
+        match *self {
+            Self::B256 => true,
+            Self::String(size) => {
```
vnepveu, line 52: `return (size > 8)`

fuels-abigen-macro/tests/test_projects/large-return-data/abi.json
```diff
@@ -0,0 +1 @@
+[{"inputs":[{"components":null,"name":"gas","type":"u64"},{"components":null,"name":"coin","type":"u64"},{"components":null,"name":"color","type":"b256"},{"components":null,"name":"value","type":"()"}],"name":"get_id","outputs":[{"components":null,"name":"","type":"b256"}],"type":"function"},{"inputs":[{"components":null,"name":"gas","type":"u64"},{"components":null,"name":"coin","type":"u64"},{"components":null,"name":"color","type":"b256"},{"components":null,"name":"value","type":"()"}],"name":"get_small_string","outputs":[{"components":null,"name":"","type":"str[8]"}],"type":"function"},{"inputs":[{"components":null,"name":"gas","type":"u64"},{"components":null,"name":"coin","type":"u64"},{"components":null,"name":"color","type":"b256"},{"components":null,"name":"value","type":"()"}],"name":"get_large_string","outputs":[{"components":null,"name":"","type":"str[9]"}],"type":"function"}]
```
vnepveu, line 1: I think we usually format on multiple lines

vnepveu (changes requested):
Nitpicks

vnepveu (approved):

PR #113, opened 2022-02-25
feat: connect to already deployed contracts

### LINKED ISSUE #88: Connect contract instance to existing contract
Currently, we're only able to create a contract instance by compiling and deploying it: 

```rust
let contract_instance = MyContract::new(compiled, client);
```

We should be able to create a contract instance from an already deployed contract by referencing to its deployed address, similar to `ethers.js` / `ethers-rs`. Something similar to this:

```rust
 let contract_instance = MyContract::connect("0x...");
```

vnepveu:
Closes #88.
This PR changes the way to instantiate a contract. Since the compiled code is actually not needed
to interact with the contract, we just provide the `contract_id`. 
This PR also creates a "default instantiation" for contracts, mainly for test purposes.

Comment: not sure about some design choices around string or `new_default`.


README.md
```diff
@@ -163,7 +163,7 @@ impl MyContract {
     pub fn initialize_counter(&self, arg: u64) -> ContractCall<u64> {
         Contract::method_hash(
             &self.fuel_client,
-            &self.compiled,
+            &self.contrat_id,
```
nfurfaro, line 166: typo : contract_id

README.md
```diff
@@ -173,7 +173,7 @@ impl MyContract {
     pub fn increment_counter(&self, arg: u64) -> ContractCall<u64> {
         Contract::method_hash(
             &self.fuel_client,
-            &self.compiled,
+            &self.contrat_id,
```
nfurfaro, line 176: same typo as above.

adlerjohn (commented):
Looks reasonable. Will defer to @digorithm for final review.

fuels-core/src/code_gen/abigen.rs
```diff
@@ -111,18 +111,25 @@ impl Abigen {
...
                     impl #name {
-                        pub fn new(compiled: CompiledContract, fuel_client: FuelClient) -> Self {
-                            Self{ compiled, fuel_client }
+                        pub fn new(contract_id: String, fuel_client: FuelClient) -> Self {
+                            let contract_id = ContractId::from_str(&contract_id).unwrap();
+                            Self{ contract_id, fuel_client }
+                        }
+                        pub fn new_default(fuel_client: FuelClient) -> Self {
+                            let contract_id = ContractId::new([0xFFu8;32]);
+                            Self{ contract_id, fuel_client }
```
digorithm, line 132: Can we document this? It's not immediately clear, from the name alone, what the differences are between instantiating a contract with `new` and `new_default`.

digorithm (changes requested):
Looks good, great job removing the need for passing the compiled to contract around, much better with just the contract id -- just left a tiny request for comment.

fuels-core/src/code_gen/abigen.rs
```diff
@@ -111,18 +111,25 @@ impl Abigen {
...
 
                     impl #name {
-                        pub fn new(compiled: CompiledContract, fuel_client: FuelClient) -> Self {
-                            Self{ compiled, fuel_client }
+                        pub fn new(contract_id: String, fuel_client: FuelClient) -> Self {
+                            let contract_id = ContractId::from_str(&contract_id).unwrap();
+                            Self{ contract_id, fuel_client }
+                        }
+                        pub fn new_default(fuel_client: FuelClient) -> Self {
+                            let contract_id = ContractId::new([0xFFu8;32]);
```
adlerjohn, line 131: What is the reason for this function?

digorithm: I'm asking myself if there's a real good reason for it to exist. The only reason it exists is that in our test suite (`harness.rs`) we have some tests that instantiate a contract but don't make actual calls -- they're only testing stuff like the encoded data for a generated method call. So we're not deploying it. I'm guessing this `new_default` covers this need _but_ I guess we could get away by just having `new()` that takes the contract id, and in the tests that are not deploying, we just pass a zeroed contract id, like `new_default()` is doing; simplifying the whole API.

adlerjohn: Could `new_default` be non-`pub` if it's only to be used for testing? Or can the casts `impl` a `new_default` only in the test suite for this struct?

vnepveu: The idea of this `new_default` is indeed to allow "cleaner" testing. Upon reading your comments, I realize it's not good to have this function exist in *every* contract. I pushed the alternative, which is to feed a *null* `contract_id` to the tests.

Cargo.lock
```diff
@@ -1,4444 +0,0 @@
-# This file is automatically @generated by Cargo.
```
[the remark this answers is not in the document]
vnepveu: Uh I messed up I think

fuels-abigen-macro/Cargo.toml
```diff
@@ -26,4 +26,4 @@ fuels-contract = { version = "0.4.3", path = "../fuels-contract" }
 fuels-signers = { version = "0.4.3", path = "../fuels-signers", features=["test-helpers"] }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
 sha2 = "0.9.5"
-tokio = "1.15.0"
+tokio = "1.15.0"
```
adlerjohn, line 29: newline

If you're using vscode, turn on adding newline at end of file

vnepveu: It's the rustfmt formatter it's weird

digorithm (approved):
LGTM. Let's keep in mind that this is a breaking change -- we're moving from creating a contract instance with the compiled contract to taking only the contract id. All tests out there will break with this change. It will require minimal changes on the user side, though, similar to the changes in `harness.rs` in this PR.

adlerjohn (approved):

PR #114, opened 2022-02-25
refactor: remove the `launch_and_deploy` function

### LINKED ISSUE #98: Remove `launch_and_deploy` function, refactor launching node
The `launch_and_deploy()` function is generally messy and confusing because it conflates two things: launching a node and deploying a contract.

https://github.com/FuelLabs/fuels-rs/blob/f02c231bde373466fc94237cd77134e00114726d/fuels-contract/src/contract.rs#L223-L236

It also makes writing tests more brittle and non-uniform, e.g. if you deploy two contracts for a single test, the second contract has to use `deploy`. Then if you rearrange things to deploy the second contract first, all of a sudden your test will break.

## Action

- [ ] Remove `launch_and_deploy()`.
- [ ] Add a new function `launch()` that launches a new node with the provided configuration (e.g. port, and genesis state), so long as there's not a node already running at that port. When done, the node should be shut down.
- [ ] Maybe add a new function `connect()` that connects to an already-running node. When done, the node should not be shut down. Can be left to a future issue.

vnepveu:
This PR closes #98.
This PR replace the `launch_and_deploy` function with a simpler `launch`
function whose only responsibility is to launch a node. 
Implementation of a `connect` function is tracked in issue #112 and will be done later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,9 @@ let compiled =
 let contract_id = Contract::deploy(compiled_contract, fuel_client).await.unwrap();
 ```
 
-Alternatively, if you want to launch a local node for every deployment, which is usually useful for smaller tests where you don't want to keep state between each test, you can use `Contract::launch_and_deploy()`:
+Alternatively, if you want to launch a local node for every deployment, which is usually 
+useful for smaller tests where you don't want to keep state between each test, you can use 
+`Contract::launch(Config::local_node())`:
 
 ```Rust
 // Build the contract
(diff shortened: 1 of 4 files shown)


digorithm:
This is much better, but one thing sounded a bit weird to me... `Contract::launch(...)` doesn't that sound like we're "launching" a contract and not a node? I was thinking if it wouldn't make more sense to move the `launch()` method to `Provider` (`provider.rs`) so that users could

```rust
let client = Provider::launch(...); // Launch a test provider (node)
let contract_id = Contract::deploy(&compiled, &client).await.unwrap(); // Deploy the contract
```

vnepveu:
Good remark, changing.

digorithm:
Tiny conflict in one of the `Cargo.toml`s. Ping me when it's resolved, PR is good to go!

digorithm (dismissed):

adlerjohn:
Is this ready for re-review? cc @digorithm

digorithm:
> Is this ready for re-review? cc @digorithm

I believe it is.

vnepveu:
It is! I had to cancel some changes I did trying to fix CI/tests

digorithm (approved):
Good stuff!

PR #116, opened 2022-02-27
Bump to v0.4.1.

adlerjohn:
There shouldn't have been any breaking changes since v0.4.0 I think.


digorithm (approved):
Yeah, we added new functionalities and some fixes, but no breaking changes.

PR #118, opened 2022-02-27
Bring back no-verify temporarily

digorithm:
Currently the crate publishing action is complaining about this reasonable use-case here: https://github.com/FuelLabs/fuels-rs/blob/master/fuels-signers/Cargo.toml#L31

@Voxelot will patch the action itself to accept this use-case, but for now we'll bring back the `no-verify` option so we can publish a new release.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --verbose
+          args: --all-features --verbose
 
   publish:
     # Only do this job if publishing a release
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --all-features --verbose
+          args: --verbose
```
adlerjohn, line 59: Why remove --all-features?

digorithm: Ah, that was an editing mistake!

adlerjohn (approved):

PR #120, opened 2022-02-28
Remove dev dependency

digorithm:
Removes a `fuels-signers` dev dependency to itself, which might be causing crate publishing issues.


adlerjohn (approved):
Looks fine to me, but not sure if this will actually solve the issue.

digorithm:
> Looks fine to me, but not sure if this will actually solve the issue.

Me neither. I just tried removing this dependency and everything worked just fine -- which I found surprising. This might have been the cause of that cyclic dep issue that led us to add `--no-verify` again.

PR #124, opened 2022-03-01
Add caching for rust dependencies

digorithm:
Using https://github.com/marketplace/actions/rust-cache for caching. This should speed up our CI pipeline.

Note that, for it to be effective, `Cargo.lock` must be committed.


adlerjohn:
Do we want to commit the lockfile? There's technically a binary in `fuels-abi-cli`, so maybe the lockfile should have been committed the whole time?

digorithm:
> Do we want to commit the lockfile? There's technically a binary in fuels-abi-cli, so maybe the lockfile should have been committed the whole time?

We do want to commit the lockfile: https://github.com/marketplace/actions/rust-cache#cache-effectiveness. It's also how we enable caching on `fuel-core` and probably other repos.

adlerjohn (approved):

PR #142, opened 2022-03-08
docs: fix docs for new contract instantiation

vnepveu:
Some things I forgot during refactoring

### DIFF
--- a/fuels-abigen-macro/README.md
+++ b/fuels-abigen-macro/README.md
@@ -40,7 +40,8 @@ fn compile_bindings_from_inline_contract() {
         "#
     );
 
-    let contract_instance = SimpleContract::new();
+    let dummy_client = FuelClient::new("").unwrap();
+    let contract_instance = SimpleContract::new(contract_id.to_string(),dummy_client);
 
     let contract_call = contract_instance.takes_ints_returns_bool(42 as u32, 10 as u16);
 
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ And, then, you're able to use to call the actual methods on the deployed contrac
 
 ```Rust
 //...
-let contract_instance = MyContract::new(compiled, client);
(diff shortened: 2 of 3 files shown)


fuels-abigen-macro/README.md
```diff
@@ -40,7 +40,8 @@ fn compile_bindings_from_inline_contract() {
         "#
     );
 
-    let contract_instance = SimpleContract::new();
+    let dummy_client = FuelClient::new("").unwrap();
```
adlerjohn, line 43: The string here is supposed to be the URL? Instead of a string with a magic value, could we use an `Option`?

digorithm: Yeah, `""` is just too... ugly. Passing a `None` would make more sense, even though very few times a user will pass a `None` (or `""`) to this method.

adlerjohn: Changing this can be left to a future PR, since it's a functional change.

adlerjohn (approved):

digorithm (approved):

fuels-abigen-macro/README.md
adlerjohn: @vnepveu can you file an issue?

digorithm: Hmm the problem is that this is on the `fuel-core` side https://github.com/FuelLabs/fuel-core/blob/master/fuel-client/src/client.rs#L52.

adlerjohn: Ah, then we can file an issue in `fuel-core`!

vnepveu: Yes filing the issue. Should I update the `""` bit to a random non-empty string?

adlerjohn: Well, no, since it's a URL. If the correct thing _currently_ is `""`, then the docs should say `""`.

PR #143, opened 2022-03-08
Add support for multi-contract contract calls

digorithm:
This PR enables users to call contracts that are calling other external contracts. It introduces a chain method `.set_contracts(&[...])` to do so. Like this:

```Rust
let response = my_contract
    .my_method(...)
    .set_contracts(&[another_contract_id]) // Add this to set the external contract
    .call()
    .await
    .unwrap();
```


README.md
```diff
@@ -270,6 +270,70 @@ abigen!(
...
+    .my_method(...)
+    .set_contracts(&[another_contract_id]) // Add this to set the external contract
+    .call()
+    .await
+    .unwrap();
+```
+
+Here's a more concrete example:
+
+```Rust
```
adlerjohn, line 288: This is why moving all these docs into The Fuels-rs Book is a `good idea`: you could have them as real files then just include them.

fuels-abigen-macro/tests/test_projects/foo-caller-contract/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
+target
```
adlerjohn, line 2: Could you put this in `fuels-abigen-macro/tests/test_projects/.gitignore` instead?

digorithm: This is what's generated by `forc init` (for every test case I use the whole `forc init` workflow), didn't want to change the default output from `forc init` in case people wanted to play with these test projects 🤔.

adlerjohn: Got it

fuels-abigen-macro/tests/test_projects/foo-caller-contract/src/main.sw
```diff
@@ -0,0 +1,16 @@
...
+use foo::FooContract;
+use std::constants::ETH_ID;
+
+abi FooCaller {
+    fn call_foo_contract(gas_: u64, amount_: u64, color_: b256, value: bool) -> bool;
+}
+
+impl FooCaller for Contract {
+    fn call_foo_contract(gas_: u64, amount_: u64, color_: b256, value: bool) -> bool {
+        let foo_contract = abi(FooContract, 0x7b4837e641d659a0662183f0fdfeca3fb4fa1248d62e4721ff28808bf11bd8c7);
```
vnepveu, line 12: maybe document what this ID is/comes from

fuels-contract/src/contract.rs
```diff
@@ -309,12 +335,23 @@ pub struct ContractCall<D> {
...
 
 impl<D> ContractCall<D>
 where
     D: Detokenize,
 {
+    /// Sets external contracts as dependencies to this contract's call.
+    /// Effectively, this will be used to create Input::Contract/Output::Contract
+    /// pairs and set them into the transaction.
+    /// Note that this is a builder method, i.e. use it as a chain:
+    /// `my_contract_instance.my_method(...).set_contracts(&[another_contract_id]).call()`.
```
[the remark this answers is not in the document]
digorithm: That's a good thought that would be very cool from a UX perspective. But unfortunately no, it's not desirable. Setting an external contract as a dependency to a contract method call is something that happens within a single transaction, therefore this happens _per call_. This effectively updates the `Transaction` with `Input::Contract`s and `Output::Contract`s. 

For instance, there are cases where in a call/transaction we use contract ID `A`, and in another call, we use a contract ID `B`[0]. If you had set the contract ID at the instantiation level (when creating the contract instance), this wouldn't be possible to achieve.  

[0]: You could imagine a `transfer` method that would use `n` underlying contracts to perform the transfer. In a first call you would like to `contract.transfer(10).set_contracts(vec![A])` so it would use `A`, and in another call you would like to use the contract `B` for that: `contract.transfer(10).set_contracts(vec![B])`.

TL;DR: these Input/Output contracts shouldn't be bound to the contract instance, it should be bound to a transaction instance.

fuels-abigen-macro/tests/test_projects/foo-caller-contract/src/main.sw
digorithm: It's the `FooContract` address... right there next to itself: 
```
      vv the contract,           vv the address
abi(FooContract, 0x7b4837e641d659a0662183f0fdfeca3fb4fa1248d62e4721ff28808bf11bd8c7)
```

fuels-contract/src/contract.rs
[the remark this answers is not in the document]
vnepveu: Ok got it, thanks for explaining.

fuels-abigen-macro/tests/test_projects/foo-caller-contract/src/main.sw
vnepveu: But I mean did you choose the address yourself? Is it deterministacally always the same contract address?

vnepveu (approved):

PR #144, opened 2022-03-08
fix: support the new Sway array type format

### LINKED ISSUE #111: Fix support for arrays
The SDK had support for static non-string arrays when the format was `T[n]` as specified here https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#array

But now arrays in Sway are defined as `[T; size]` for type `T` and the generated ABI looks like `"type":"[T; size]"`. 

This broke the array support in the SDK and we should fix it by changing the parsing from `T[n]` to `[T; size]`. 

- [x] Change ABI parsing from `T[n]` to `[T; size]`;
- [x] Update all tests in harness to include the right json ABI;
- [x] Update the `fuel-specs`'s ABI section.

vnepveu:
This PR closes #111 by fixing the ABI parsing of the array type from `T[n]` to `[T; n]`.
It also refactors the parsing of `str[n]` parameters and implements tests for both parsing.

### DIFF
--- a/fuels-core/src/json_abi.rs
+++ b/fuels-core/src/json_abi.rs
@@ -592,6 +592,36 @@ pub fn parse_custom_type_param(param: &Property) -> Result<ParamType, Error> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::ParamType;
+
+    #[test]
+    fn parse_string_and_array_param() {
+        let array_prop = Property {
+            name: "some_array".to_string(),
+            type_field: "[bool; 4]".to_string(),
+            components: None,
+        };
+        let expected = "Array(Box::new(ParamType::Bool),4)";
+        let result = parse_array_param(&array_prop).unwrap().to_string();
+        assert_eq!(result, expected);
+
+        let string_prop = Property {
+            name: "some_array".to_string(),
+            type_field: "str[5]".to_string(),
+            components: None,
+        };
+        let expected = "String(5)";
+        let result = parse_string_param(&string_prop).unwrap().to_string();
+        assert_eq!(result, expected);
+
+        let expected = "Invalid type: Expected parameter type `str[n]`, found `[bool; 4]`";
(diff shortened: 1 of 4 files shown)


fuels-core/src/json_abi.rs
```diff
@@ -516,41 +516,52 @@ pub fn parse_param(param: &Property) -> Result<ParamType, Error> {
...
+    let split: Vec<&str> = param.type_field.split("; ").collect();
+    if split.len() != 2 {
+        return Err(Error::InvalidType(format!(
+            "Expected parameter type `[T; n]`, found `{}`",
+            param.type_field
+        )));
     }
+    let (type_field, size) = (split[0], split[1]);
+    let type_field = type_field[1..].to_string();
+    let param_type = ParamType::from_str(&type_field).unwrap();
```
digorithm, line 560: We should avoid using `unwrap()` when it's possible / not too hard to avoid it. In this case here, it's not too hard to change `unwrap()` to `?` to better handle the error and avoid panicking. The error from parsing a bad `type_field` would be very cryptic to the end-user that could've misspelled the type in the JSON ABI.

fuels-core/src/json_abi.rs
```diff
@@ -516,41 +516,52 @@ pub fn parse_param(param: &Property) -> Result<ParamType, Error> {
...
+
+    // Grab size the `n` in "[T; n]"
+    let size: usize = size[..size.len() - 1].parse().unwrap();
```
digorithm, line 563: Same comment here about `unwrap()`.

digorithm (changes requested):
Looking great, good work on this. Just left some comments on better error handling.

fuels-core/src/json_abi.rs
vnepveu: I'm still not used to this you're absolutely right

digorithm (approved):

Issue #140, opened 2022-03-05
Update `fuel-core` to v0.4.0


digorithm:
`fuel-core` v0.4.0 introduces many breaking changes to `fuels-rs`, such as coin color => asset id renaming and more (https://github.com/FuelLabs/fuel-core/pull/197).

PR #147, opened 2022-03-09
Update to `fuel-core` 0.4

digorithm:
`fuel-core` `0.4.x` brought in lots of changes, such as:

- Renaming `color` to `asset id`;
- Some structural changes to the GraphQL schema;
- Some changes to its types;
- And some more.

This PR makes the necessary changes to support these changes. Also, `Hasher` was moved from `fuel-tx` to `fuel-crypto`, so we introduce `fuel-crypto` as a dependency to `fuels-rs`. 

Closes https://github.com/FuelLabs/fuels-rs/issues/140.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -521,6 +521,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "borrown"
+version = "0.1.0"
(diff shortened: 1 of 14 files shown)


vnepveu (approved):
Nice!

adlerjohn (approved):

PR #149, opened 2022-03-10
Rename color to asset ID.

adlerjohn:
Remove stragglers missed in #147.

Remaining instances of the word `color` should only be as ABI parameters, which will be removed with #13, and a single use of the stdlib `ETH_COLOR` constant, which can be changed in the stdlib first.


digorithm (approved):
Oh right, nice catch. What I did on the other PR was only related to the color -> asset ID _type_ changes. Thanks!

vnepveu (approved):
:ok_hand:

Issue #148, opened 2022-03-10
Bump `forc` to `0.6`


adlerjohn:
Probably requires #13 to be done simultaneously.

mitchmindtree:
Also requires https://github.com/FuelLabs/sway/pull/917 (along with a follow-up forc `0.6.1` release) for the tests to pass as witnessed by @vnepveu (see [here](https://github.com/FuelLabs/sway/issues/915)).

vnepveu:
After trying to update `forc` to `0.6`, it seems some tests are broken for no reason: 
```
---- abigen_different_structs_same_arg_name stdout ----
  Bytecode size is 156 bytes.
Contract deployed @ 9906afc4d317e5d16ff3694bd14c0f908f1a5c21a685f6e778e18b2b363d73da
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `43`', fuels-abigen-macro/tests/harness.rs:1337:5
```. If you look a the concerned contract, there is *afaik* no reason its behavior should be changed by the `forc` update.

digorithm:
@vnepveu https://github.com/FuelLabs/fuels-rs/blob/master/fuels-core/src/code_gen/functions_gen.rs#L117-L135 

We're still checking for the 3 mandatory arguments `(u62, u62, b256)` and encoding it in the call. `forc`/`sway` `0.6.x` removed it, so if you compile it with `forc` `0.6.x`, it will be deployed with all methods _not_ including these args in the function signature but the SDK will still be encoding it, leading to all sorts of errors. 

I'm currently working on these changes to support `forc` `0.6` though.

Issue #125, opened 2022-03-02
Use compiled binary output from `forc`to build Rust bindings


digorithm:
Currently, we have a method under `Contract` to compile Sway code:

```rust
pub fn compile_sway_contract(
    project_path: &str,
    salt: Salt,
) -> Result<CompiledContract, Error> {
    let build_command = BuildCommand {
        debug_outfile: None,
        minify_json_abi: false,
        path: Some(project_path.into()),
        print_finalized_asm: false,
        print_intermediate_asm: false,
        binary_outfile: None,
        offline_mode: false,
        output_directory: None,
        silent_mode: true,
        print_ir: false,
        use_ir: false,
    };

    let (raw, _) =
        forc_build::build(build_command).map_err(|message| Error::CompilationError(message))?;

    Ok(CompiledContract { salt, raw })
}
```

Although that works fine, it adds a dependency on `forc` as a library. `fuels-rs` should be language agnostic: it should operate only on the compiled binary and the ABI, both targeting the FuelVM.

Recently, `forc` was updated to output the binary data to an output directory. `fuels-rs` can leverage that in order to remove the dependency on `forc` as a library. Instead of calling `forc::build`, take a path to the compiled binary data, the same way that it takes the path to the json ABI file. 

Then, it would be on the user to use `forc` to generate the compiled binary and the json ABI file, then pass both to `fuels-rs`'s deploy method and contract instantiation method (i.e. `Contract::new()`).

PR #150, opened 2022-03-11
Changes to support forc 0.6

digorithm:
Major bundle of updates due to changes in `forc` `0.6.x`:

1. Now that `forc build` outputs the compiled binary, we can remove `forc` as a dependency in `fuels-rs` and load the pre-compiled binary instead of building it. Closes https://github.com/FuelLabs/fuels-rs/issues/125;
2. Sway `0.6` also removes the 3 mandatory arguments (coin, asset_id, gas), and thus we can remove the many workarounds to support those args, including the encoding of those args in the function selector. Closes https://github.com/FuelLabs/fuels-rs/issues/13;
3. Because of (1), we can also better organize the tests under `harness.rs` by using the structure from `forc build` output. Also removed gas, coin, color/asset_id from all test projects;
4. Deleted some tests that were either redundant or outdated due to the changes introduced here.

Closes https://github.com/FuelLabs/fuels-rs/issues/148.


fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -114,43 +114,18 @@ fn expand_function_arguments(
...
-        .map(|x| x.type_field.clone())
-        .collect();
-    let expected = ["u64", "u64", "b256"];
-    if given_types != expected {
-        return Err(Error::InvalidType(format!(
-            "Expected the 3 first types to be {:?}, found {:?}",
-            expected, given_types
-        )));
-    };
     // Ignore the first three arguments so we don't have to provide them when calling
```
vnepveu, line 117: Remove this comment

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -114,43 +114,18 @@ fn expand_function_arguments(
...
     // the contracts' methods
-    let first_index = 3;
     let mut args = Vec::with_capacity(1);
```
vnepveu, line 119: keep the line 117 and use `n_inputs` as vec capacity (instead of 1)

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -114,43 +114,18 @@ fn expand_function_arguments(
...
-    for (i, param) in fun.inputs[first_index..].iter().enumerate() {
+
+    for (i, param) in fun.inputs.iter().enumerate() {
         // For each [`Property`] in a function input we expand:
         // 1. The name of the argument;
         // 2. The type of the argument;
         // Note that _any_ significant change in the way the JSON ABI is generated
         // could affect this function expansion.
         // TokenStream representing the name of the argument
-        if param.type_field == "()" {
```
vnepveu, line 150: why is this not needed anymore?

vnepveu: It's not related to #13 but to #95 and I'm not sure we can just remove it

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -260,31 +235,11 @@ fn expand_input_param(
 mod tests {
     use super::*;
     use std::str::FromStr;
-    fn generate_base_inputs() -> Vec<Property> {
```
vnepveu, line 263: good ridance haha

fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -690,27 +630,6 @@ pub fn hello_world(
         assert_eq!(result, expected);
     }
 
-    #[test]
-    fn test_expand_function_argument_empty_fourth() {
```
vnepveu, line 694: why is this not needed anymore?

vnepveu (changes requested):
Just some questions about the "empty" types arguments, otherwise gg, lgtm!

fuels-core/src/code_gen/functions_gen.rs
digorithm: Uh... We don't need `with_capacity`at all. Was trying to preserve as much as I could, but this isn't really necessary anymore.

fuels-core/src/code_gen/functions_gen.rs
digorithm: Sway now supports calls with actually 0 arguments (`fn foo()` instead of `fn foo(input: ()))`. So we can remove this check above. See tests with proper empty arguments down below.

fuels-core/src/code_gen/functions_gen.rs
digorithm: https://github.com/FuelLabs/fuels-rs/pull/150/files#r825151275

.gitignore
```diff
@@ -5,5 +5,5 @@
 # These are backup files generated by rustfmt
 **/*.rs.bk
 
-# Forc output
-**/out/
\ No newline at end of file
+# Forc lock
+Forc.lock
```
mitchmindtree, line 9: Maybe it's worth including the `Forc.lock` files, purely to keep track of which commits were used to produce the ABI and binary?

fuels-abigen-macro/tests/test_projects/foo-caller-contract/Forc.toml
```diff
@@ -7,4 +7,4 @@ name = "foo-caller-contract"
 [dependencies]
 core = { git = "http://github.com/FuelLabs/sway-lib-core" }
 std = { git = "http://github.com/FuelLabs/sway-lib-std" }
-foo = { path = "../foo-contract-abi/src/" }
+foo = { path = "../foo-contract-abi" }
```
mitchmindtree, line 10: :+1:

fuels-contract/src/contract.rs
```diff
@@ -255,28 +255,9 @@ impl Contract {
...
-            print_ir: false,
-            use_ir: false,
-        };
-
-        let (raw, _) = forc_build::build(build_command).map_err(Error::CompilationError)?;
-
-        Ok(CompiledContract { salt, raw })
+    pub fn load_sway_contract(binary_filepath: &str, salt: Salt) -> Result<CompiledContract> {
+        let bin = std::fs::read(binary_filepath)?;
+        Ok(CompiledContract { raw: bin, salt })
```
mitchmindtree, line 260: :clinking_glasses:

.gitignore
digorithm: Yeah, that's a good idea!

fuels-core/src/code_gen/functions_gen.rs
vnepveu: ok!

fuels-core/src/code_gen/functions_gen.rs
vnepveu: ok!

vnepveu (approved):
LGTM!

fuels-abigen-macro/tests/test_projects/call_empty_return/Forc.lock
```diff
@@ -0,0 +1,21 @@
...
+
+[[package]]
+name = 'core'
+source = 'git+https://github.com/FuelLabs/sway-lib-core?reference=master#c331ed20ebc9d646acec6b8ee8f408627ce3b573'
+dependencies = []
+
+[[package]]
+name = 'std'
+source = 'git+https://github.com/FuelLabs/sway-lib-std?reference=master#4eb73be7be90e40099ba9fe34cd2d5f1faa01c6f'
+dependencies = ['core git+http://github.com/FuelLabs/sway-lib-core?reference=master#c331ed20ebc9d646acec6b8ee8f408627ce3b573']
```
mitchmindtree, line 21: I just noticed that `core` is getting pinned multiple times due to the difference in `http` (via `std` lib's dependency) and `https` URLs (via `contract_test`'s dependency). I think `https` as in the `contract_test` manifest is the better convention and we should consider updating `std`'s manifest to do the same, however this is something that `forc` should resolve either way. I've opened https://github.com/FuelLabs/sway/issues/931 to track this, but it shouldn't cause any issues in the meantime and I think it's fine to land this as is.

mitchmindtree (approved):

adlerjohn (approved):
LGTM, but can we get an issue in `sway` to have `forc test` first run `forc build` if that's not the case already? Since testing with the SDK will now no longer involve compiling the contract.

mitchmindtree:
> can we get an issue in sway to have forc test first run forc build if that's not the case already?

Yep this is [the current behaviour](https://github.com/FuelLabs/sway/blob/7c34f24b68544a4b1fbb07765f470d43be97e594/forc/src/cli/commands/test.rs#L24) as of https://github.com/FuelLabs/sway/pull/796.

Issue #138, opened 2022-03-04
Add params to a `Contract`'s `.call()` method


digorithm:
Extend `.call()` to take in a struct containing optional params (for `gas_limit`, `gas_price`, `byte_price`) and add those to the crafted transaction.

In `ethers.js`, this is the equivalent to the overrides:

```js
var options = { gasPrice: 1000000000, gasLimit: 85000, nonce: 45, value: 0 };
var fooPromise = contract.foo(address, options);
```

However, only `gas_limit`, `gas_price`, `byte_price` should be initially addressed, as `value` would require more complex changes.

PR #158, opened 2022-03-17
Support tx parameter overrides

digorithm:
This PR adds support for setting parameters to contract calls and deployments. These parameters are byte price, gas limit, and gas price for now.

It's done through a new method `tx_params()` that's chained to the generated ABI method:

```Rust
let result = contract_instance
        .tx_params(Parameters::new(None, Some(1_000_000), None)) // <--- this here.
        .await
        .unwrap();
```

This doesn't break the way contract calls were made previously. i.e. `initialize_counter(42).call();` (without `.tx_params()`) works just like before and, also just like before, it will use default values for these parameters. If the user wants to override the default values, then they can chain `tx_params()` to the method call.

However, this does break previous contract deployments, as the deployment method now requires these params to be set:

```Rust
 let contract_id =
        Contract::deploy(&compiled, &provider, &wallet, Parameters::default())
            .await
            .unwrap();
```

Closes https://github.com/FuelLabs/fuels-rs/issues/138.


fuels-contract/src/contract.rs
```diff
@@ -91,6 +90,11 @@ impl Contract {
         #[allow(clippy::iter_cloned_collect)]
         let script = vec![
             Opcode::ADDI(0x10, REG_ZERO, script_data_offset),
+            // @todo currently there's no way to forward an amount.
+            // This would be done by programmatically changing
+            // $rB (`REG_ZERO` right now) to actually point to an
+            // amount. This would forward the amount in $rB using
+            // $rC as the asset_id.
```
digorithm, line 97: @adlerjohn Need your thoughts on this here. The issue being tackled here is this one https://github.com/FuelLabs/fuels-rs/issues/138. However, it only addresses the byte price and gas price/limit params. The original issue, on top of these params, also wants to be able to set amount/asset_id to be forwarded in the call (and be consumed with `msg_amount()` in Sway). 

But these aren't as simple as updating a value in the `Transaction` (like gas price, byte price, gas limit). I believe setting the amount/asset to forward means updating this assembly block, more specifically  `Opcode::CALL(0x10, REG_ZERO, 0x10, REG_CGAS)`, to actually be updated programmatically. Does that make sense? Is that the best/right approach?

adlerjohn: Your changes don't actually resolve #138 since that's about what gets forwarded in the call to the contract, which as you say here would require changes to the how it's called. You should unlink #138 from this PR since it's completely orthogonal.

adlerjohn (commented):
I'm not the biggest fan of the name `with_params` and `Parameters`. My initial reaction to `.call().with_params()` is that those are parameters _for the call_, not the transaction. Maybe `.tx_params()` and `TxParameters`?

digorithm:
> I'm not the biggest fan of the name `with_params` and `Parameters`. My initial reaction to `.call().with_params()` is that those are parameters _for the call_, not the transaction. Maybe `.tx_params()` and `TxParameters`?

Yeah, I wasn't attached to the `with_params`/`params` naming. Will update it soon.

fuels-contract/src/contract.rs
digorithm: It's not completely orthogonal. The example I mentioned in https://github.com/FuelLabs/fuels-rs/issues/138 is orthogonal because it's covering only `value` (which is the value to be forwarded). But in the original discussion the motivation for that issue was to also programmatically set `gas_price`, `gas_limit`, and `byte_price`, which has been a blocker for @nfurfaro. 

At the beginning of this work, I tried to address it all (including `value`) but realized only `gas_price`, `gas_limit`, and `byte_price` could be done with this approach, whereas `value` (and `asset_id`) would require a totally different approach. Hence why I didn't address it here and am leaving it for another PR. I was also planning on creating another issue, extending https://github.com/FuelLabs/fuels-rs/issues/138.

digorithm: Updated https://github.com/FuelLabs/fuels-rs/issues/138 to limit its scope and created this one https://github.com/FuelLabs/fuels-rs/issues/159.

fuels-abigen-macro/tests/harness.rs
```diff
@@ -643,19 +644,27 @@ async fn example_workflow() {
...
     let contract_instance = MyContract::new(contract_id.to_string(), provider, wallet);
 
     let result = contract_instance
         .initialize_counter(42) // Build the ABI call
+        .tx_params(TxParameters::new(None, Some(1_000_000), None))
         .call() // Perform the network call
         .await
         .unwrap();
 
+    println!("result: {:?}\n", result);
```
adlerjohn, line 666: This and the one below look like debugging prints? Should they be removed?

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1195,3 +1212,105 @@ async fn test_contract_calling_contract() {
...
+    )
+    .unwrap();
+
+    let (provider, wallet) = setup_test_provider_and_wallet().await;
+
+    let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
+        .await
+        .unwrap();
+
+    println!("id: {:?}\n", id);
```
adlerjohn, line 1291: Same here

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+license = "Apache-2.0"
+name = "token-ops"
+version = "0.1.0"
+
+[dependencies]
+fuel-gql-client = { version = "0.2", default-features = false }
+fuel-tx = "0.3"
+fuels-abigen-macro = "0.3"
+fuels-contract = "0.3"
+fuels-core = "0.3"
```
adlerjohn, line 13: Is there a reason not use a local path for these? Now you're testing against a pinned version, which isn't testing upgrades.

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+authors = ["Fuel Labs"]
+edition = "2021"
+license = "Apache-2.0"
+name = "token-ops"
+version = "0.1.0"
+
+[dependencies]
+fuel-gql-client = { version = "0.2", default-features = false }
+fuel-tx = "0.3"
```
adlerjohn, line 10: Should use the latest version for these.

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+authors = ["Fuel Labs"]
+edition = "2021"
+license = "Apache-2.0"
+name = "token-ops"
+version = "0.1.0"
```
adlerjohn, line 6: Use 0.0.0 for things that aren't meant to be published.

fuels-abigen-macro/tests/test_projects/token-ops/Forc.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+authors = ["Fuel Labs"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "token-ops"
+
+[dependencies]
+core = { git = "http://github.com/FuelLabs/sway-lib-core" }
+std = { git = "http://github.com/FuelLabs/sway-lib-std" }
```
adlerjohn, line 9: These should include a `branch`

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
digorithm: Nope, no reason. It's just the output from `forc init` @ `0.6`.

digorithm: And we don't use any of these `fuels-*` because we're not creating tests under that forc project, just using the compiled contract + ABI.

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
digorithm: We actually don't need any of these `Cargo.toml`s because we're just using the compiled contracts and ABI for testing.

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
adlerjohn: Ah I see. Then maybe to avoid headaches they should just be removed if they're not used?

fuels-abigen-macro/tests/test_projects/token-ops/Cargo.toml
adlerjohn: Ah, then maybe remove altogether?

digorithm: Yup!

adlerjohn (commented):
LGTM, but will defer to @vnepveu

fuels-contract/src/constants.rs
```diff
@@ -0,0 +1,3 @@
+pub const GAS_LIMIT: u64 = 1_000_000;
```
vnepveu, line 1: We should probably put `NATIVE_ASSET_ID` in this file since we're here

fuels-contract/src/constants.rs
```diff
@@ -0,0 +1,3 @@
...
+pub const GAS_PRICE: u64 = 0;
```
vnepveu, line 2: Wouldn't `DEFAULT_*` be a better name?

fuels-contract/src/contract.rs
```diff
@@ -69,9 +70,7 @@ impl Contract {
         encoded_selector: Option<Selector>,
         encoded_args: Option<Vec<u8>>,
         fuel_client: &FuelClient,
-        gas_price: Word,
-        gas_limit: Word,
-        byte_price: Word,
+        call_parameters: TxParameters,
```
vnepveu, line 73: argument name should be `tx_parameters`

fuels-contract/src/contract.rs
```diff
@@ -236,19 +244,15 @@ impl Contract {
...
         let custom_inputs = args.iter().any(|t| matches!(t, Token::Struct(_)));
 
+        let maturity = 0;
         Ok(ContractCall {
             contract_id,
             encoded_args,
-            gas_price,
-            gas_limit,
-            byte_price,
+            call_parameters: params,
```
vnepveu, line 255: this should be named `tx_parameters` as well

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1195,3 +1208,104 @@ async fn test_contract_calling_contract() {
...
+
+    let (provider, wallet) = setup_test_provider_and_wallet().await;
+
+    let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
+        .await
+        .unwrap();
+
+    let instance = TestFuelCoinContract::new(id.to_string(), provider.clone(), wallet.clone());
+
+    let target = testfuelcoincontract_mod::ContractId { value: id.into() };
```
vnepveu, line 1289: what's `testfuelcoincontract_mod`

vnepveu (changes requested):
Some naming issues, 1/2 nits and 1/2 questions to make sure I understand. Otherwise great!!!

fuels-contract/src/constants.rs
vnepveu: We could also put this inside: 
`NULL_CONTRACT_ID: &str = "0x00…"` so as to avoid the `null_contract_id` function inside the harness.

digorithm: Yup, might as well move all relevant constants here.

fuels-abigen-macro/tests/harness.rs
digorithm: The generated module/namespace for potentially conflicting names, such as ContractID (which is created by the user in this case here). The name is based on the contract/project name. See sway-lib-std's token-ops tests for more context/rationale.

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1195,3 +1208,104 @@ async fn test_contract_calling_contract() {
...
+
+    // Test for insufficient gas.
+    let result = contract_instance
+        .initialize_counter(42) // Build the ABI call
+        .tx_params(TxParameters::new(Some(1_000), Some(100), None))
+        .call() // Perform the network call
+        .await
+        .expect_err("should error");
+
+    assert_eq!("Contract call error: Response errors; unexpected block execution error InsufficientGas { provided: 1, required: 100000 }", result.to_string());
```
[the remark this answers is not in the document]
digorithm: Yup!

vnepveu (approved):

PR #160, opened 2022-03-18
feat: sign transactions before submitting them to the client

### LINKED ISSUE #130: Integrate wallet/signer abstraction with the contract abstraction
Now that we have a basic wallet and signer interface to sign transactions, it needs to be integrated with the contract abstraction in `fuels-rs`. Currently, all transactions sent by a contract instance are unsigned -- they work because the client isn't checking for signatures, but eventually, it will.

Contract instantiation, then, needs to take in a signer, and have it accessible through the contract interface (and also use the signer to sign the transaction before sending it).

The implementation and API design and details are still up for discussion, though.

vnepveu:
* This PR closes #130 by signing transactions before submitting them to the client.
* The "sign transactions before submission" portion isn't tested because I struggled 
to find a way to meaningfully test the whole workflow. Feel free to chime in with ideas
if you think this is important, but to me the unit tests are enough.

### DIFF
--- a/fuels-contract/src/contract.rs
+++ b/fuels-contract/src/contract.rs
@@ -105,7 +105,7 @@ impl Contract {
         .copied()
         .collect::<Vec<u8>>();
 
-        assert!(script.len() == script_len, "Script length *must* be 16");
+        assert_eq!(script.len(), script_len, "Script length *must* be 16");
 
         // `script_data` consists of:
         // 1. Contract ID (ContractID::LEN);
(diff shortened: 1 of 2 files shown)


digorithm (approved):
> The "sign transactions before submission" portion isn't tested because I struggled
to find a way to meaningfully test the whole workflow. Feel free to chime in with ideas
if you think this is important, but to me the unit tests are enough.

IMO that's good enough _for now_.

I think that for this to be properly tested, the Sway contract that we deploy with a given signature, would have to check, in its ABI method, say, `foo()`, if the caller of `foo()` (who signed the call) is the one who signed the deployment, something like Solidity's `owner()` modifier. This way we could verify that they match. But that would be basically testing Sway + core (including `tx`, VM) abilities to verify signatures -- not sure this would be valuable at the SDK level. That could be a longer discussion and we can prioritize that for later.

adlerjohn (approved):

Issue #139, opened 2022-03-04
Extend data handling for types larger than a `WORD`


digorithm:
This is a follow-up to https://github.com/FuelLabs/fuels-rs/pull/112. Currently, we're checking for returned data size only for `String`s and `b256`: 

```rust
impl ParamType {
    // Checks whether the `ParamType` is bigger than a `WORD`
    pub fn bigger_than_word(&self) -> bool {
        match *self {
            Self::B256 => true,
            Self::String(size) => size > 8,
            _ => false,
            // More types will be handled later.
            // Currently, the support for arrays in the SDK is broken
            // due to a change to the array definition in Sway.
        }
    }
}
```

We need to extend this for the other types, more specifically custom types; `Struct([u8])` should be smaller than a `WORD`, but `Struct([u8, B256])` isn't.

PR #161, opened 2022-03-18
Handle `ReturnData` for other types

digorithm:
Closes https://github.com/FuelLabs/fuels-rs/issues/139. 

More context in https://github.com/FuelLabs/fuels-rs/issues/139. Summary: depending on the size of the returned data, the returned data will live either under the `Return` receipt or `ReturnData` receipt. This handles it for the types that weren't being properly handled before.


fuels-abigen-macro/tests/harness.rs
```diff
@@ -1075,6 +1082,27 @@ async fn test_large_return_data() {
...
+
+    // First `value` is from `CallResponse`.
+    // Second `value` is from Sway `ContractId` type.
+    assert_eq!(
+        res.value.value,
+        [
+            255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255,
+            255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255
+        ]
+    );
```
digorithm, line 1105: @nfurfaro I believe this is the relevant case for you: `get_contract_id()` returns the Sway type `ContractId`, which wraps the `b256` in a `Struct`, which wasn't being handled before. Now it seems to be working fine!

fuels-core/src/lib.rs
```diff
@@ -44,15 +44,29 @@ impl Default for ParamType {
...
+                1 => params[0].bigger_than_word(),
+                _ => true,
+            },
+            // Enums are always in `ReturnData`.
+            Self::Enum(_params) => true,
+            // Arrays seem to always be inside `ReturnData`.
+            Self::Array(_params, _l) => true,
+            // The other primitive types are inside `Return`,
+            // thus smaller than one `WORD`.
             _ => false,
```
digorithm, line 69: @vnepveu I believe this will be related to your current work on supporting multiple return values, make sure to rebase it on top of this once it's merged!

nfurfaro (commented):
LGTM, but I'll defer to @vnepveu

fuels-abigen-macro/tests/test_projects/large-return-data/src/main.sw
```diff
@@ -3,25 +3,65 @@ contract;
...
-  fn get_id() -> b256;
-  fn get_small_string() -> str[8];
-  fn get_large_string() -> str[9];
+    fn get_id() -> b256;
+    fn get_small_string() -> str[8];
+    fn get_large_string() -> str[9];
+    fn get_large_struct() -> LargeStruct;
+    fn get_small_struct() -> SmallStruct;
+    fn get_large_array() -> [u32;
+    2];
```
adlerjohn, line 24: https://github.com/FuelLabs/sway/issues/726 strikes again. cc @eureka-cpu

fuels-core/src/lib.rs
```diff
@@ -44,15 +44,29 @@ impl Default for ParamType {
 }
 
 impl ParamType {
-    // Checks whether the `ParamType` is bigger than a `WORD`
+    // Checks if the `ParamType` is bigger than a `WORD`.
+    // This is important because, depending on whether its
```
adlerjohn, line 48:
```suggestion
    // This is important because, depending on whether it's
```

fuels-core/src/lib.rs
```diff
@@ -44,15 +44,29 @@ impl Default for ParamType {
...
+            Self::String(size) => size > &8,
+            Self::Struct(params) => match params.len() {
+                // If only one component in this struct
+                // check if this element itself is bigger than a `WORD`.
+                1 => params[0].bigger_than_word(),
+                _ => true,
+            },
+            // Enums are always in `ReturnData`.
+            Self::Enum(_params) => true,
+            // Arrays seem to always be inside `ReturnData`.
```
adlerjohn, line 65: Is this expected or a bug? cc @otrho

otrho: I'm guessing at this stage it's always true since even an enum with a single variant will still have a tag and a value, even if it's just `unit`.  So it can't fit in a register.  After we add some gnarly optimisations this may no longer be true, but only if we want to go there.

adlerjohn: Oh I was pointing at the ``Arrays seem to always be inside `ReturnData`.``, not the enums 😂

digorithm: Unfortunately I had to do a lot of guesswork to get this working and figure out where each returned type would be located at. 

As for arrays, in the tests I did, they were always in ReturnData (there are few new array-as-output tests added in this PR). But then again, I didn't fuzz or anything, just tested a few cases and focused on solving some existing blockers. 

But, intuitively (which could be wrong of course), if any primitive type takes a whole word. Even if it doesn't fill it all we pad it. So an array should be larger than a word? Maybe unless is an array of size zero or one? 😅

fuels-core/src/lib.rs
vnepveu: Yes will do!

vnepveu (approved):
LGTM

fuels-core/src/lib.rs
adlerjohn: An array of size 0 or 1 should only take up a single word. Could you add a test for that @digorithm?

fuels-abigen-macro/tests/test_projects/large-return-data/src/main.sw
eureka-cpu: Likely what's happening is the formatter is adding a newline character when it finds a `;` and this edge case wasn't taken into account. Currently working on the `format stucts and enums` draft that I had yet to finish, so will take a look at this too.

adlerjohn (approved):
Can handle arrays of size 0 or 1 in a future PR. Make sure there's an issue for it.

digorithm:
Tracking it here: https://github.com/FuelLabs/fuels-rs/issues/162.

PR #165, opened 2022-03-21
Use signing/recovery from fuel_crypto in fuels_signers

### LINKED ISSUE #65: Use `fuel-crypto` instead of `fuel-vm` for signatures
Currently `fuels-rs` signers and wallet abstractions require cryptography-related functions from `fuel-vm`, e.g. in `fuels-core` `lib.rs`:

```rust
use fuel_vm::crypto::secp256k1_sign_compact_recover;
```

Recently we've created `fuel-crypto` (https://github.com/FuelLabs/fuel-crypto/) that will host all cryptography-related code that will be used downstream and eventually `fuel-vm` won't host this type of code anymore.

`fuels-rs` needs to be modified to use the signing/recovery functions in `fuel-crypto` instead of `fuel-vm`.

Br1ght0ne:
Close #65.

Things left:
- [x] `impl FromStr for fuel_crypto::Signature`
- [x] port remaining tests that require `Signature::from_str`
- [x] fix tests failing after rebase


digorithm:
Since there's still `impl FromStr for fuel_crypto::Signature so I can port tests straight away` left, shouldn't we keep this PR as a draft? Still, I'll be doing an initial review soon 😄

Br1ght0ne:
> Since there's still `impl FromStr for fuel_crypto::Signature so I can port tests straight away` left, shouldn't we keep this PR as a draft? Still, I'll be doing an initial review soon 😄

Whoops, pressed too early. Will add this impl soon :)

fuels-signers/src/lib.rs
```diff
@@ -112,9 +111,8 @@ mod tests {
         let signature = wallet.sign_transaction(&mut tx).await.unwrap();
         let message = Message::new(tx.id());
 
-        // TODO(oleksii): impl FromStr for fuel_crypto::Signature
         // Check if signature is what we expect it to be
-        // assert_eq!(signature.compact, Signature::from_str("0xa1287a24af13fc102cb9e60988b558d5575d7870032f64bafcc2deda2c99125fb25eca55a29a169de156cb30700965e2b26278fcc7ad375bc720440ea50ba3cb").unwrap().compact);
+        assert_eq!(signature, Signature::from_str("0x87493a4e6e02e70e6a380f9c9ad2167cc01ccd6cf058c371316911c80e2415c8b7a406268b09b985581ebbe70de81ff59bba99d24dae5295d0a137b7b9a8507c").unwrap());
```
Br1ght0ne, line 115: Please note that the signature here has changed. Not sure if it's normal. `sign_and_verify` didn't change, while `sign_tx_and_verify` did.

Br1ght0ne:
Weird that GH wants to merge/rebase this with `master`. I did the rebase on top of fresh `master`, no other changes. 🤔

adlerjohn (changes requested):
Remove lockfile and update from `master` using the button if needed

fuels-signers/src/lib.rs
```diff
@@ -48,26 +46,26 @@ mod tests {
         let mut secret_seed = [0u8; 32];
         rng.fill_bytes(&mut secret_seed);
 
-        let secret =
-            SecretKey::from_slice(&secret_seed).expect("Failed to generate random secret!");
+        let secret = unsafe { SecretKey::from_bytes_unchecked(secret_seed) };
```
vnepveu, line 49: This should be both commented to explain why we are doing an `unsafe` block, and IMO we should figure out the proper way to make it non-unsafe. We can't really compromise memory safety (even in this trivial case) for something like signing.

fuels-signers/src/lib.rs
```diff
@@ -48,26 +46,26 @@ mod tests {
...
-        assert_eq!(signature.compact, Signature::from_str("0x8eeb238db1adea4152644f1cd827b552dfa9ab3f4939718bb45ca476d167c6512a656f4d4c7356bfb9561b14448c230c6e7e4bd781df5ee9e5999faa6495163d").unwrap().compact);
+        assert_eq!(signature, Signature::from_str("0x8eeb238db1adea4152644f1cd827b552dfa9ab3f4939718bb45ca476d167c6512a656f4d4c7356bfb9561b14448c230c6e7e4bd781df5ee9e5999faa6495163d").unwrap());
 
         // Recover address that signed the message
-        let recovered_address = signature.recover(message).unwrap();
+        let message = Message::new(message);
+        let recovered_address = signature.recover(&message).unwrap();
 
-        assert_eq!(wallet.address, recovered_address);
+        assert_eq!(wallet.address.as_ref(), recovered_address.hash().as_ref());
```
vnepveu, line 65: why is it passed `as_ref`?

fuels-signers/src/lib.rs
vnepveu: But we have to "prove" to ourselves that it is because the tx format changed

Br1ght0ne: Well, the sign/recover/verify workflow is intact…

>  the transaction format/content changed

Then the old  `secp256k1`-based tests should have caught it, I think?

fuels-signers/src/util.rs
```diff
@@ -37,14 +36,10 @@ pub mod test_helpers {
 
         let secret_seed = rng.gen::<[u8; 32]>();
 
-        let secret =
-            SecretKey::from_slice(&secret_seed).expect("Failed to generate random secret!");
+        let secret = unsafe { SecretKey::from_bytes_unchecked(secret_seed) };
```
vnepveu, line 39: Same remark as before

fuels-signers/src/wallet.rs
```diff
@@ -32,8 +29,7 @@ use thiserror::Error;
 ///   let mut secret_seed = [0u8; 32];
 ///   rng.fill_bytes(&mut secret_seed);
 ///
-///   let secret =
-///       SecretKey::from_slice(&secret_seed).expect("Failed to generate random secret!");
+///   let secret = unsafe { SecretKey::from_bytes_unchecked(secret_seed) };
```
vnepveu, line 32: see above

fuels-signers/src/wallet.rs
```diff
@@ -223,20 +216,16 @@ impl Signer for Wallet {
...
         message: S,
     ) -> Result<Signature, Self::Error> {
-        let message = message.as_ref();
-        let message_hash = Hasher::hash(message);
-
-        let sig =
-            secp256k1_sign_compact_recoverable(self.private_key.as_ref(), &*message_hash).unwrap();
-        Ok(Signature { compact: sig })
+        let message = Message::new(message);
+        let sig = Signature::sign(&self.private_key, &message);
```
vnepveu, line 220: if the Signature::sign cannot fail, then we shouldn't return a `Result`. Otherwise, I think you could use `?`

fuels-signers/src/wallet.rs
```diff
@@ -91,11 +88,8 @@ impl Wallet {
         private_key: SecretKey,
         provider: Provider,
     ) -> Result<Self, WalletError> {
-        let secp = Secp256k1::new();
-
-        let public = PublicKey::from_secret_key(&secp, &private_key).serialize_uncompressed();
-        let public = Bytes64::try_from(&public[1..])?;
-        let hashed = Hasher::hash(public);
+        let public = PublicKey::from(&private_key);
+        let hashed = public.hash();
```
vnepveu, line 92: same remark as below regarding using `Result` -> it's not useful if you can't get "an error" out

vnepveu: (I think at least lol)

vnepveu (changes requested):
Some remarks about the `unsafe`, and also I'm not sure about the error handling but this might come from a misunderstanding on my end :) Otherwise looks good!

fuels-signers/src/lib.rs
Br1ght0ne:
```
mismatched types
expected struct `fuel_tx::Address`, found struct `fuel_tx::Bytes32`
```

I think they're both FuelVM's atomic types, with underlying arrays. At type-level, we can't know they're the same length, so we convert to `&[u8]` (an unbounded slice) with `as_ref` to compare.

fuels-signers/src/lib.rs
Voxelot: It's interesting that this is the only place the signature changed, while the signature on the hardcoded message "my message" above wasn't affected. Definitely seems like a change in the tx id would've affected this signature earlier though.

adlerjohn:
> merge commit or rebase?

Merge commit, so that reviewers know what's changed. On the topic of the lockfile, you need to remove because it's not supposed to be there: https://github.com/FuelLabs/fuels-rs/pull/179

Updating from `master` will also update your gitignore so it's not added again by accident.

fuels-signers/src/lib.rs
vnepveu: So what should we do? IMO the 4 options are:
- Previous implementation was faulty, and this is "the correct one" for some obscure elliptic-curve level reason
- The new implementation is faulty for the the same kind of reason
- The tx_id format changed, and for some reason it was not caught by the previous tests
- There is a dependency/version issue that for instance would prevent the tests to detect the change

vnepveu: Maybe let's merge master back into it just to make sure it's not deps/version issue (I'm shooting blindfolded haha)

Voxelot: Figured it out. Message::new() from fuel-crypto always hashes the input. Before this PR, we skipped hashing the tx id since it was already guaranteed to be 32 bytes. The reason why the signature in the other test didn't change is that we had a step in fuels-rs [to hash the message](https://github.com/FuelLabs/fuels-rs/blob/master/fuels-signers/src/wallet.rs#L227), which matches the new behavior of Message::new().

If we want to restore the original behavior, we'd have to use the "unsafe" method `Message::from_bytes_unchecked`. This may be appropriate since we already know a transaction id comes from the output of a hash function.

Br1ght0ne: @Voxelot Thank you!

fuels-signers/src/wallet.rs
Br1ght0ne: `Wallet::sign_message` is provided by the [`Signer`](https://github.com/FuelLabs/fuels-rs/blob/master/fuels-signers/src/lib.rs#L20) trait implementation. As the `Signer` doc mentions, there could be other signing backends in the future, which might want to return a `Result`. And yes, current `Signature::sign` method can't fail.

fuels-signers/src/lib.rs
vlopes11: Alternatively, you can create a message from a [Hasher](https://docs.rs/fuel-crypto/0.4.2/fuel_crypto/struct.Message.html#impl-From%3C%26%27_%20Hasher%3E)

fuels-signers/src/lib.rs
vlopes11: The proper way would be via [TryFrom](https://docs.rs/fuel-crypto/0.4.2/fuel_crypto/struct.SecretKey.html#impl-TryFrom%3CBytes32%3E).

However, from the lines above, looks like you are creating a new random key. Hence, in this case, the proper way would be to call [Secret::random](https://docs.rs/fuel-crypto/0.4.2/src/fuel_crypto/secret.rs.html#122)

Br1ght0ne:
Needs UTXO validation in `send_transaction` and `transfer_coins_with_change`.

packages/fuels-signers/src/lib.rs
```diff
@@ -1,11 +1,9 @@
 pub mod provider;
-pub mod signature;
 pub mod util;
 pub mod wallet;
```
Voxelot, line 4: I think we should re-export fuel-crypto here so that users of the SDK don't need to take a separate dependency on fuel-crypto in order to work with Signature and key types.

```suggestion
pub use fuel_crypto;
```

digorithm:
@Br1ght0ne Please take a look at the changes I introduced at https://github.com/FuelLabs/fuels-rs/pull/165/commits/a5026c68c911add58fd4ddf8687256fcc7968cff. 

Here's a quick explanation: 

1. `tx.id()` [hashes the transaction ID](https://github.com/FuelLabs/fuel-tx/blob/master/src/transaction/id.rs#L40);
2. `fuel-crypto`'s `Message::new()` [hashes the bytes passed onto it](https://github.com/FuelLabs/fuel-crypto/blob/master/src/message.rs#L26).

Before, this is what you were doing:

```Rust
let id = tx.id();
let sig = Signature::sign(&self.private_key, &Message::new(&id));
```

Effectively hashing the transaction bytes twice. This caused the recovered public key to be wrong because the `Message` was wrong (double hashed), causing a mismatch between the input coin's owner (correct) and the recovered public key (incorrect), [which is the error you had been seeing all this time](https://github.com/FuelLabs/fuel-tx/blob/master/src/transaction/validation.rs#L64-L66).

packages/fuels-signers/Cargo.toml
```diff
@@ -30,6 +29,7 @@ tokio = { version = "1.10.1", features = ["full"] }
 hex = { version = "0.4.3", default-features = false, features = ["std"] }
 fuel-types = { version = "0.3", default-features = false, features = ["random"] }
 fuels-test-helpers = { path = "../fuels-test-helpers", default-features = false }
+fuels = { path = "../fuels" }
```
adlerjohn, line 32: Should this be before `fuels-test-helpers`?

adlerjohn (approved):

vnepveu (approved):
All my questions have been nicely addressed afai can see! lgtm!

Issue #159, opened 2022-03-17
Programmatically set amount and asset to be forwarded in contract calls


digorithm:
This extends https://github.com/FuelLabs/fuels-rs/issues/138. In `ethers.js` one can:

```js
var options = { gasPrice: 1000000000, gasLimit: 85000, nonce: 45, value: 0 };
var fooPromise = contract.foo(address, options);
```

We've addresses gas price, limit and byte price in https://github.com/FuelLabs/fuels-rs/issues/138. But `value` requires a more complex change that's currently described in a comment under `fuels-contract`'s `contract.rs`:

```rust
let script = vec![
    Opcode::ADDI(0x10, REG_ZERO, script_data_offset),
    // @todo currently there's no way to forward an amount.
    // This would be done by programmatically changing
    // $rB (`REG_ZERO` right now) to actually point to an
    // amount. This would forward the amount in $rB using
    // $rC as the asset_id.
    Opcode::CALL(0x10, REG_ZERO, 0x10, REG_CGAS),
    Opcode::RET(REG_RET),
    Opcode::NOOP,
]
.iter()
.copied()
.collect::<Vec<u8>>();
```

So, in order for us to introduce a new API to set the amount/asset to be forwarded, the `script` needs to be updated not to hardcode the values in `Opcode::CALL(0x10, REG_ZERO, 0x10, REG_CGAS)` and be able to update these values (`$rB` and `$rC`) before sending the transaction. 

This will make it possible to interact with cases like this, in Sway:

```rust
fn deposit() {
        assert(msg_asset_id() == ETH_ID || msg_asset_id() == TOKEN_ID);
        
        let sender = ~Address::from(0x0000000000000000000000000000000000000000000000000000000000000000);
        let key = key_deposits(sender, msg_asset_id());

        let total_amount = get::<u64>(key) + msg_amount();
        store(key, total_amount);
    }
``` 

Where `msg_amount()` and `msg_asset_id()` will be the values we're passing using the SDK's new API for value/asset forwarding.

PR #167, opened 2022-03-24
Enable amount/asset forwarding in tx params

digorithm:
Adds a `CallParameters` to support forwarding `amount` and `asset_id` through a new chain method: `call_params()`.

It also extends the generated `script` and `script_data` to enable this support.

Example: 

```Rust
let tx_params = TxParameters::new(None, Some(1_000_000), None);

// Forwarding 1 (Some(1)) coin of type `NATIVE_ASSET_ID` (`0x00...`).
let call_params = CallParameters::new(Some(1), None);

let response = instance
    .get_msg_amount()
    .tx_params(tx_params) // Set the transaction params
    .call_params(call_params) // Set the call params
    .call()
    .await
    .unwrap();
```

```
CallResponse { value: 1, receipts: [Call { id: eb41245813e928b57b2c57ee14b0dfa4073687d3c69792f8bc14beed6825ffae, to: 
eb41245813e928b57b2c57ee14b0dfa4073687d3c69792f8bc14beed6825ffae, amount: 1, asset_id: 
0000000000000000000000000000000000000000000000000000000000000000, gas: 999967, a: 2830212025, b: 1, pc: 1712, is: 1712 }] }
```

In Sway, this amount is accessible through the standard library method `msg_amount()`, e.g.:

```Rust
abi TestFuelCoin {
    fn get_msg_amount() -> u64;
}

impl TestFuelCoin for Contract {
    fn get_msg_amount() -> u64 {
        msg_amount()
    }
}
```

Closes https://github.com/FuelLabs/fuels-rs/issues/159.


fuels-contract/src/contract.rs
```diff
@@ -78,44 +78,55 @@ impl Contract {
...
+        // 2. Contract ID (ContractID::LEN);
+        // 3. Function selector (1 * WORD_SIZE);
+        // 4. Calldata offset, if it has structs as input,
         // computed as `script_data_offset` + ContractId::LEN
         //                                  + 2 * WORD_SIZE;
-        // 4. Encoded arguments.
+        // 5. Encoded arguments.
         let mut script_data: Vec<u8> = vec![];
 
+        script_data.extend(tx_parameters.asset_id.to_vec());
```
digorithm, line 128: @luizstacio @AlicanC This whole section here is what's needed on `fuels-ts` to implement the same thing. Let me know if it's not clear enough and/or if you have more questions!

Cargo.lock
```diff
@@ -86,9 +86,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.55"
+version = "1.0.56"
```
adlerjohn, line 89: Revert changes to the lockfile

fuels-contract/src/parameters.rs
```diff
@@ -1,10 +1,15 @@
...
+    DEFAULT_BYTE_PRICE, DEFAULT_GAS_LIMIT, DEFAULT_GAS_PRICE, NATIVE_ASSET_ID,
+};
 
 #[derive(Debug)]
 pub struct TxParameters {
     pub gas_price: u64,
     pub gas_limit: u64,
     pub byte_price: u64,
+    pub amount: u64,
+    pub asset_id: AssetId,
```
adlerjohn, line 12: I wouldn't put these as tx parameters since they're more parameters for a call rather than parameters for a tx. This will come into play later for sure, if we do something like allow a builder that can push multiple calls into a single tx.

ethers-rs uses dedicated functions: https://docs.rs/ethers/latest/ethers/contract/builders/struct.ContractCall.html#method.value

Instead of a dedicated function per param, maybe something like `CallParameters` could suffice. Should also include gas to forward as an optional, but that maybe could be left for a future PR since it involves more logic.

adlerjohn (changes requested):

fuels-contract/src/parameters.rs
digorithm:
```Rust
 let response = instance
        .get_msg_amount()
        .tx_params(tx_params)
        .call_params(call_params)
        .call()
        .await
        .unwrap();
```

Or

```Rust
 let response = instance
        .get_msg_amount()
        .tx_params(tx_params)
        .call(Some(call_params))
        .await
        .unwrap();
```
?

adlerjohn: John votes for the former so as to avoid `None` all the time when using default call params.

adlerjohn (commented):
The core changeset is good, but just to get into a good habit, could you break apart updating fuel-vm, updating tests, fixing unrelated tests, and fixing some constants all into different PRs?

digorithm:
> The core changeset is good, but just to get into a good habit, could you break apart updating fuel-vm, updating tests, fixing unrelated tests, and fixing some constants all into different PRs?

Hmm... All the changes done in this PR are related to each other, I think?

1. Updating `fuel-vm` is needed to actually make `Enable amount/asset forwarding in tx params` work;
2. Updating tests... also needed for `Enable amount/asset forwarding in tx params`;
3. Fixing unrelated tests, they were not unrelated, I think? They were affected by the core changes done here;
4. Fixing some constants... I added a new one, related to this PR, needed to get `Enable amount/asset forwarding in tx params` to work.

adlerjohn (approved):

Issue #164, opened 2022-03-20
Make single fuels-rs crates that exports everything needed for tests


adlerjohn:
Rather than needing to set

```toml
fuel-gql-client = {{ version = "0.4", default-features = false }}
fuel-tx = "0.6"
fuels-abigen-macro = "0.6"
fuels-contract = "0.6"
fuels-core = "0.6"
rand = "0.8"
tokio = {{ version = "1.12", features = ["rt", "macros"] }}
```

in `Cargo.toml` for Sway tests, there should be a single `fuels-rs` crate that exports all needed dependencies, such that users only need a single one.

PR #168, opened 2022-03-24
Re-export fuels-rs as single crate

digorithm:
_Important: This is a very critical breaking change, review this carefully._

Closes https://github.com/FuelLabs/fuels-rs/issues/164. 

This PR makes necessary changes to turn `fuels-rs` into a single crate, instead of having to import `fuels-contract`, `fuels-signers`, `fuels-core`, etc. 

The only exception is `fuels-abigen-macro`, since it's a `proc_macro` package, it has be separated in its own package, for a lot of very tricky reasons. Still, this feels much more ergonomic, turning:

```rust
use fuels_contract::contract::Contract;
use fuels_contract::errors::Error;
use fuels_contract::parameters::{CallParameters, TxParameters};
use fuels_core::constants::NATIVE_ASSET_ID;
use fuels_core::Token;
use fuels_signers::util::test_helpers::{
    setup_address_and_coins, setup_test_provider, setup_test_provider_and_wallet,
};
use fuels_signers::{provider::Provider, LocalWallet};
```

Into:

```rust
use fuels_rs::prelude::*;
```

Where the `prelude` exports all necessary types/functions/names to bootstrap a `fuels-rs` project. Users can also import things separately through `fuels_rs::{...}`. The point is that now everything is under `fuels_rs::` and that will be the crate that will be published on crates.io. And `fuels-abigen-macro` for reasons mentioned above.


src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    pub use super::core::Token;
+
+    pub use super::signers::provider::*;
+
+    pub use super::signers::LocalWallet;
+
+    pub use super::contract::errors::Error;
+
+    pub use super::signers::util::test_helpers::*;
```
adlerjohn, line 59: Alphabetize and maybe remove newlines?

adlerjohn:
Reasoning for proc macro needing to be its own crate: https://stackoverflow.com/questions/56713877/why-do-proc-macros-have-to-be-defined-in-proc-macro-crate/56714256#56714256

adlerjohn (commented):
Looks fine to be, but want @vnepveu at least to verify everything works

fuels-abi-cli/Cargo.toml
```diff
@@ -15,4 +15,4 @@ fuels-core = { version = "0.7.0", path = "../fuels-core" }
 hex = "0.4"
```
vnepveu, line 15: I feel like `fuels-contract/core` should be replaced by `fuels-rs`? Similarly to other manifests

fuels-abigen-macro/Cargo.toml
```diff
@@ -22,8 +22,7 @@ syn = "1.0.12"
 [dev-dependencies]
 fuel-core = { version = "0.4", default-features = false }
 fuel-gql-client = { version = "0.4", default-features = false }
-fuels-contract = { version = "0.7.0", path = "../fuels-contract" }
```
vnepveu, line 25: same remark about `fuels-core` in the `dependencies` section

fuels-signers/Cargo.toml
```diff
@@ -28,7 +28,7 @@ tokio = { version = "1.10.1", features = ["full"] }
```
vnepveu, line 28: same question about `fuels-core`

vnepveu (changes requested):
I think some imports were left out, otherwise lgtm!

fuels-abi-cli/Cargo.toml
digorithm: No, that doesn't seem to be the way it's done; when referring to local dependencies, you can point to them directly. [See this in ethers-rs for a good example](https://github.com/gakonst/ethers-rs/blob/master/ethers-signers/Cargo.toml#L17).

vnepveu: Oh nvm, just realized the only instance where you did this was in the special `fuels-abigen-macro` repo, my bad

vnepveu (approved):
Misunderstood so it's actually ok!

adlerjohn:
Converting to draft so it doesn't get merged by accident.

adlerjohn (changes requested):
Need to add `fuels-rs` to the CI version checker

Cargo.toml
```diff
@@ -7,3 +17,9 @@ members = [
     "fuels-core",
     "fuels-signers",
 ]
+
+[dependencies]
+fuels-abigen-macro = { version = "0.7.0", path = "./fuels-abigen-macro"}
+fuels-contract = { version = "0.7.0", path = "./fuels-contract"}
+fuels-core = { version = "0.7.0", path = "./fuels-core"}
+fuels-signers = { version = "0.7.0", path = "./fuels-signers"}
```
adlerjohn, line 25:
```suggestion
fuels-abigen-macro = { version = "0.7.1", path = "./fuels-abigen-macro"}
fuels-contract = { version = "0.7.1", path = "./fuels-contract"}
fuels-core = { version = "0.7.1", path = "./fuels-core"}
fuels-signers = { version = "0.7.1", path = "./fuels-signers"}
```

Br1ght0ne:
32342e6 resolves conflicts with `master`. Additional changes to make it build:
- update `fuels-rs` version to 0.8.1 to match the dependencies version on `master`
- add `core::Tokenizable` and `signers::Signer` to re-exports

`cargo t --workspace --all-features` passes locally.

Cargo.toml
```diff
@@ -7,3 +17,9 @@ members = [
     "fuels-core",
     "fuels-signers",
 ]
+
+[dependencies]
+fuels-abigen-macro = { version = "0.8.1", path = "./fuels-abigen-macro"}
+fuels-contract = { version = "0.8.1", path = "./fuels-contract"}
+fuels-core = { version = "0.8.1", path = "./fuels-core"}
+fuels-signers = { version = "0.8.1", path = "./fuels-signers"}
```
adlerjohn, line 25:
```suggestion
fuels-abigen-macro = { version = "0.8.1", path = "./fuels-abigen-macro" }
fuels-contract = { version = "0.8.1", path = "./fuels-contract" }
fuels-core = { version = "0.8.1", path = "./fuels-core" }
fuels-signers = { version = "0.8.1", path = "./fuels-signers" }
```

adlerjohn (changes requested):
1. Shouldn't `fuels-rs` just be a workspace member?
2. Is there a circular dependency, since `fuels-rs` depends on `fuels-abigen-macro`, and `fuels-abigen-macro` depends on `fuels-rs`?

digorithm:
> 1. Shouldn't `fuels-rs` just be a workspace member?
> 
>     2. Is there a circular dependency, since `fuels-rs` depends on `fuels-abigen-macro`, and `fuels-abigen-macro` depends on `fuels-rs`?

1. Hmm, not sure. Two reasons: I'm using [ethers-rs organization as an inspiration](https://github.com/gakonst/ethers-rs/blob/master/Cargo.toml) and the way it is right now worked w.r.t shrinking the number and complexity of imports (main goal);

2. Yeah, that's a valid concern, but again... everything seems to be working fine? It might be because `fuels-abigen-macro` only depends on `fuels-rs` for testing (`dev-dependencies` instead of `dependencies`)?

adlerjohn:
Is it okay for something to be both a package and a workspace?

digorithm:
> Is it okay for something to be both a package and a workspace? 

I'm not sure about this. Summoning @Voxelot. Thoughts on it?

> Ethers-rs shouldn't necessarily be copied blindly.

Very true. However, I don't have tons of experience with package management in Rust, especially because every couple of years it changes; I was used to the pre-2018 way of structuring/packaging, and now different projects seem to use different approaches. 

Findings working references is the way I'm approaching this, I'm also completely open to suggestions. I followed (not blindly copied 😉) `ethers-rs` because it's both relatively easy to understand and... it works!

digorithm:
Alright, made another big change here https://github.com/FuelLabs/fuels-rs/pull/168/commits/be2a845474a398ad34ea7a41c37a1786397f7381. This time inspired by another great project's structure: [Yew](https://github.com/yewstack/yew/blob/master/Cargo.toml). It turns `fuels-rs` into a workspace-only.

Let me know whether that's better or worse (it's the same from a DevEx perspective), we can revert to the previous commit if this isn't preferable.

Now it looks like this:

```
├── Cargo.lock
├── Cargo.toml
├── LICENSE
├── README.md
├── packages
│   ├── fuels-abigen-macro
│   │   ├── Cargo.toml
│   │   ├── src
│   │   └── tests
│   ├── fuels-contract
│   │   ├── Cargo.toml
│   │   └── src
│   ├── fuels-core
│   │   ├── Cargo.toml
│   │   └── src
│   ├── fuels-rs
│   │   ├── Cargo.toml
│   │   └── src
│   └── fuels-signers
│       ├── Cargo.toml
│       └── src
└── tools
    └── fuels-abi-cli
        ├── Cargo.toml
        ├── examples
        └── src
```

Br1ght0ne:
I think that with the latter system, we will still need to use `/packages/fuels-rs/examples/foo` instead of `/examples/foo` in #101. We're back to using a [virtual manifest](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-manifest), which does not support top-level examples natively :(

digorithm:
> I think that with the latter system, we will still need to use `/packages/fuels-rs/examples/foo` instead of `/examples/foo` in #101. We're back to using a [virtual manifest](https://doc.rust-lang.org/cargo/reference/workspaces.html#virtual-manifest), which does not support top-level examples natively :(

I think it does support top-level examples? Look here https://github.com/yewstack/yew/blob/master/Cargo.toml#L10. 

We would have a top-level (sibling to `packages/` and `tools/`) called `examples/` and the examples would go in there. Seems to be a working strategy for Yew.

adlerjohn (dismissed):
I'm okay with this, but would like @Voxelot to chime in as well.

Br1ght0ne:
Yes, you're right—though that would require a separate crate for each example. I was leaning towards [example targets](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples) (`cargo run --example foo`), but workspace members are fine too, makes sense :)

adlerjohn (commented):
Merge conflicts due to version bump 😬

adlerjohn (approved):

.github/workflows/ci.yml
```diff
@@ -84,6 +84,7 @@ jobs:
         ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuels-core/Cargo.toml
         ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuels-abigen-macro/Cargo.toml
         ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuels-abi-cli/Cargo.toml
+        ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} fuels-rs/Cargo.toml
```
adlerjohn, line 87: Ah wait, fix these paths

adlerjohn (changes requested):

Voxelot:
I agree that example targets are preferable where that makes sense, but may not be a great fit for every usecase.

Overall I think this new packaging structure looks fine, although I did notice there's a lot of unused `harness.rs` files under `fuels-abigen-macro/tests/test_projects/*`. At first I was wondering if those should be included in the workspace testing, but it appears they are just copied and pasted artifacts from the std lib?

Voxelot:
Should they just be .gitignored then?

adlerjohn:
Maybe not ignored, but could be manually removed if they're not relevant to the particular tests?

Voxelot:
If they're just autogenerated files that never get run, then they should be cleaned up. Otherwise we should figure out how to include them in the workspace restructure if the plan is to include them.

adlerjohn (dismissed):

vnepveu (dismissed):
Nice!

adlerjohn (commented):
bump to v0.9.1

adlerjohn (approved):

adlerjohn (approved):

adlerjohn:
Merging this just to avoid further conflicts

PR #169, opened 2022-03-25
Fix multi-argument contract calls

digorithm:
Calls with multiple arguments, for instance

```rust
fn get(x: u64, y: u64) -> u64 {
    x
}
```

Were broken due to how we compute the call data offset. Before, we added this extra offset only if the argument was a custom type. Turns out, if you have more than one argument, you also need that. The only moment you don't need this extra offset is when the method takes only one arg and it's not a custom type.

This PR makes some simple, non-breaking changes to check whether we should or not compute this offset.


digorithm:
cc @AlicanC @luizstacio this probably should also be implemented on `fuels-ts`. Do you mind confirming that?

luizstacio:
> Were broken due to how we compute the call data offset. Before, we added this extra offset only if the argument was a custom type. Turns out, if you have more than one argument, you also need that. The only moment you don't need this extra offset is when the method takes only one arg and it's not a custom type.

I had implemented it on this PR: https://github.com/FuelLabs/fuels-ts/pull/170 just do a double-check, please.

fuels-contract/src/contract.rs
```diff
@@ -275,12 +276,20 @@ impl Contract {
...
-            custom_inputs,
+            compute_calldata_offset,
             external_contracts: None,
             wallet: wallet.clone(),
         })
     }
 
+    // Returns true if the method call takes custom inputs or has more than one argument. This is used to determine whether we need to compute the `call_data_offset`.
+    fn should_compute_call_data_offset(args: &[Token]) -> bool {
+        match args.iter().any(|t| matches!(t, Token::Struct(_))) {
```
vnepveu, line 287: Shouldn't we also check for `Enum` and in the future `Tuples`?

vnepveu (changes requested):
One question, otherwise LGTM!

fuels-contract/src/contract.rs
digorithm: Yes for enums. Tuples might or might not be a thing, we'll see once the time comes.

vnepveu: yes!

vnepveu (approved):
nice!

PR #172, opened 2022-03-25
Support non-native asset forwarding

digorithm:
Before, forwarding a non-native asset ID wasn't having an effect -- the asset ID always ended up as being the native one. This PR fixes this. Now the `Call` registers the asset ID passed by the user:

```
CallResponse { value: 0, receipts: [Call { id: 0xeb41245813e928b57b2c57ee14b0dfa4073687d3c69792f8bc14beed6825ffae, 
to: 0xeb41245813e928b57b2c57ee14b0dfa4073687d3c69792f8bc14beed6825ffae, amount: 0, 
asset_id: 0xeb41245813e928b57b2c57ee14b0dfa4073687d3c69792f8bc14beed6825ffae, 
gas: 999967, a: 2830212025, b: 1, pc: 1712, is: 1712 }] }
```

However, as noted in one of the tests, the VM is complaining about the asset ID being forwarded not being found. This only happens when `amount > 0`. Not sure where the issue is located at, but doesn't seem to be SDK-related.

### DIFF
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -1312,8 +1312,21 @@ async fn test_amount_and_asset_forwarding() {
     let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
         .await
         .unwrap();
-
+    let target = testfuelcoincontract_mod::ContractId { value: id.into() };
+    let asset_id = testfuelcoincontract_mod::ContractId { value: id.into() };
     let instance = TestFuelCoinContract::new(id.to_string(), provider.clone(), wallet.clone());
+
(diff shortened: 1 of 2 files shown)


adlerjohn:
> VM is complaining about the asset ID being forwarded not being found. This only happens when amount > 0

Yes, that's expected because the initial genesis coins are all of the native asset! This can either be an additional feature in the SDK, to allow the user to choose additional assets to own at genesis, or it can be accomplished by the user themselves by simply deploying a fungible token contract and minting themselves coins. I think the SDK should support the former feature.

nfurfaro (approved):
Looks good!

adlerjohn (approved):
utACK

Issue #171, opened 2022-03-25
The sdk needs to support variable outputs.


nfurfaro:
Ref: slack thread: https://fuellabs.slack.com/archives/C02CX4JVCSJ/p1647526324734319?thread_ts=1647464981.548979&cid=C02CX4JVCSJ

adlerjohn:
This can be accomplished by something like `.append_variable_outputs(num: u64)` in the tx builder which appends `num` variable outputs to the tx.

PR #173, opened 2022-03-25
Support setting variable outputs in contract calls

digorithm:
Introduces a new non-breaking API `append_variable_outputs(num: u64)` that appends `num` `Output::Variable`s to the crafted transaction. 

This is a builder method, like `call_params()` and `tx_params()`:

```rust
let response = instance
        .get_msg_amount()
        .tx_params(tx_params)
        .call_params(call_params)
        .append_variable_outputs(10)
        .call()
        .await
        .unwrap();
```

Closes https://github.com/FuelLabs/fuels-rs/issues/171.

### DIFF
--- a/fuels-contract/src/contract.rs
+++ b/fuels-contract/src/contract.rs
@@ -7,7 +7,7 @@ use anyhow::Result;
 use fuel_asm::Opcode;
 use fuel_gql_client::client::FuelClient;
 use fuel_tx::{
-    AssetId, ContractId, Input, Output, Receipt, StorageSlot, Transaction, UtxoId, Witness,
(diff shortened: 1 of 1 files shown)


nfurfaro (approved):
I think this is good, but would like another set of eyes on it if possible.

fuels-contract/src/contract.rs
```diff
@@ -396,6 +404,24 @@ where
         self
     }
 
+    /// Sets the variable outputs transaction.
```
adlerjohn, line 407: I would just make this comment a bit better

digorithm: Fun fact: this was accidentally auto-generated by GitHub Copilot. Safe to say ML won't be taking over.

adlerjohn (approved):
utACK

PR #175, opened 2022-03-25
feat: Support forwarding amounts of size up to `u64` in contract calls

luizstacio:
### Summary

Enable forwarding amount of size up to `u64` in contract calls.

### Motivation

Before, we were forcing the amount to be an `Immediate12`, which doesn't work for many use cases where we would need a `u64`. Passing a bigger number to the amount being forwarded would cause an overflow. 

closes: #174

### DIFF
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -1332,8 +1332,10 @@ async fn test_amount_and_asset_forwarding() {
     balance_result = instance.get_balance(target, asset_id).call().await.unwrap();
     assert_eq!(balance_result.value, 5_000_000);
 
-    // Forward 1 coin of native asset_id
+    // Forward 1_000_000 gas_limit of native asset_id
     let tx_params = TxParameters::new(None, Some(1_000_000), None);
+    // Forward 1_000_000 coin amount of native asset_id
+    // this is a big number for checking that amount can a u64
     let call_params = CallParameters::new(Some(1_000_000), None);
 
     let response = instance
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -3,9 +3,7 @@ use fuels_abigen_macro::abigen;
 use fuels_contract::contract::Contract;
 use fuels_contract::errors::Error;
 use fuels_contract::parameters::{CallParameters, TxParameters};
-use fuels_core::constants::{
-    NATIVE_ASSET_ID, DEFAULT_INITIAL_BALANCE
-};
+use fuels_core::constants::{DEFAULT_INITIAL_BALANCE, NATIVE_ASSET_ID};
 use fuels_core::Token;
 use fuels_signers::util::test_helpers::{
     setup_address_and_coins, setup_test_provider, setup_test_provider_and_wallet,
@@ -1276,7 +1274,11 @@ async fn test_gas_errors() {
     // Test for insufficient gas.
     let result = contract_instance
         .initialize_counter(42) // Build the ABI call
-        .tx_params(TxParameters::new(Some(DEFAULT_INITIAL_BALANCE), Some(100), None))
+        .tx_params(TxParameters::new(
+            Some(DEFAULT_INITIAL_BALANCE),
+            Some(100),
+            None,
+        ))
         .call() // Perform the network call
         .await
         .expect_err("should error");
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -1335,7 +1335,7 @@ async fn test_amount_and_asset_forwarding() {
     // Forward 1_000_000 gas_limit of native asset_id
     let tx_params = TxParameters::new(None, Some(1_000_000), None);
     // Forward 1_000_000 coin amount of native asset_id
-    // this is a big number for checking that amount can a u64
+    // this is a big number for checking that amount can be a u64
     let call_params = CallParameters::new(Some(1_000_000), None);
 
     let response = instance
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -3,7 +3,9 @@ use fuels_abigen_macro::abigen;
 use fuels_contract::contract::Contract;
 use fuels_contract::errors::Error;
 use fuels_contract::parameters::{CallParameters, TxParameters};
-use fuels_core::constants::NATIVE_ASSET_ID;
+use fuels_core::constants::{
+    NATIVE_ASSET_ID, DEFAULT_INITIAL_BALANCE
+};
 use fuels_core::Token;
 use fuels_signers::util::test_helpers::{
     setup_address_and_coins, setup_test_provider, setup_test_provider_and_wallet,
@@ -1274,12 +1276,12 @@ async fn test_gas_errors() {
     // Test for insufficient gas.
     let result = contract_instance
         .initialize_counter(42) // Build the ABI call
-        .tx_params(TxParameters::new(Some(1_000), Some(100), None))
+        .tx_params(TxParameters::new(Some(DEFAULT_INITIAL_BALANCE), Some(100), None))
         .call() // Perform the network call
         .await
         .expect_err("should error");
 
-    assert_eq!("Contract call error: Response errors; unexpected block execution error InsufficientGas { provided: 100, required: 100000 }", result.to_string());
+    assert_eq!("Contract call error: Response errors; unexpected block execution error InsufficientGas { provided: 1000000000, required: 100000000000 }", result.to_string());
 
     // Test for running out of gas. Gas price as `None` will be 0.
     // Gas limit will be 100, this call will use more than 100 gas.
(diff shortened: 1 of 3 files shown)


adlerjohn:
Can you provide some more context of what this changes is supposed to be for?

digorithm:
> Can you provide some more context of what this changes is supposed to be for?

@adlerjohn  Just updated the description to do that!

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1274,12 +1274,16 @@ async fn test_gas_errors() {
...
+            Some(DEFAULT_INITIAL_BALANCE),
+            Some(100),
+            None,
+        ))
         .call() // Perform the network call
         .await
         .expect_err("should error");
 
-    assert_eq!("Contract call error: Response errors; unexpected block execution error InsufficientGas { provided: 100, required: 100000 }", result.to_string());
+    assert_eq!("Contract call error: Response errors; unexpected block execution error InsufficientGas { provided: 1000000000, required: 100000000000 }", result.to_string());
```
adlerjohn, line 1286: Why is this error about insufficient gas and not balance???

fuels-core/src/constants.rs
```diff
@@ -8,8 +8,8 @@ pub const WORD_SIZE: usize = core::mem::size_of::<Word>();
 
 // This constant is used to determine the amount in the 1 UTXO
 // when initializing wallets for now.
-pub const DEFAULT_COIN_AMOUNT: u64 = 1;
-pub const DEFAULT_INITIAL_BALANCE: u64 = 100;
+pub const DEFAULT_COIN_AMOUNT: u64 = 1_000_000;
+pub const DEFAULT_INITIAL_BALANCE: u64 = 1_000_000_000;
```
adlerjohn, line 12: Technically orthogonal to this issue, but what's the different between coin amount and initial balance?

fuels-abigen-macro/tests/harness.rs
digorithm: IIRC, because that's what the node returns given the configuration parameters (`DEFAULT_INITIAL_BALANCE x 100`). It's just testing `tx_params` is properly setting things in the transaction.

fuels-core/src/constants.rs
digorithm: One is for testing, the other is a temporary thing to setup some arbitrary coins to a transaction: https://github.com/FuelLabs/fuels-rs/blob/master/fuels-contract/src/contract.rs#L152-L155

Naming could be improved though, very confusing indeed!

adlerjohn: If it's for testing then it shouldn't be pub

digorithm: Yeah. @luizstacio do you mind removing `pub` from `DEFAULT_INITIAL_BALANCE`?

fuels-contract/src/contract.rs
```diff
@@ -88,9 +88,12 @@ impl Contract {
...
             vec![
                 // Load call data to 0x10.
-                Opcode::ADDI(0x10, REG_ZERO, data_offset + 32),
+                Opcode::ADDI(0x10, REG_ZERO, data_offset + 40),
                 // Load gas forward to 0x11.
-                Opcode::ADDI(0x12, REG_ZERO, call_parameters.amount as Immediate12),
+                // Load the amount into 0x12
+                Opcode::ADDI(0x12, REG_ZERO, data_offset + 32),
+                // Load word into 0x12
+                Opcode::LW(0x12, 0x12, 0),
```
adlerjohn, line 96:
```suggestion
                // Load the amount into 0x12
                Opcode::ADDI(0x12, REG_ZERO, data_offset + 32),
                Opcode::LW(0x12, 0x12, 0),
```

fuels-contract/src/contract.rs
```diff
@@ -103,16 +106,22 @@ impl Contract {
 
         // `script_data` consists of:
         // 1. Asset ID to be forwarded
-        // 2. Contract ID (ContractID::LEN);
-        // 3. Function selector (1 * WORD_SIZE);
-        // 4. Calldata offset, if it has structs as input,
+        // 2. Amount to be forwarded
+        // 3. Contract ID (ContractID::LEN);
+        // 4. Function selector (1 * WORD_SIZE);
```
adlerjohn, line 111:
```suggestion
        // 2. Amount to be forwarded (ContractID::LEN).
        // 3. Contract ID (ContractID::LEN).
        // 4. Function selector (WORD_SIZE).
```

fuels-contract/src/contract.rs
```diff
@@ -103,16 +106,22 @@ impl Contract {
...
+        // 5. Calldata offset, if it has structs as input,
         // computed as `script_data_offset` + ContractId::LEN
         //                                  + 2 * WORD_SIZE;
-        // 5. Encoded arguments.
+        // 6. Encoded arguments.
         let mut script_data: Vec<u8> = vec![];
 
+        // Insert assent_id to be forwarded
```
adlerjohn, line 118:
```suggestion
        // Insert asset_id to be forwarded
```

fuels-core/src/constants.rs
luizstacio: I add some comments on it but would require more work to remove pub https://github.com/FuelLabs/fuels-rs/issues/180

adlerjohn (commented):
I would change the magic `40` and `32` to use some combination of constants `ContractId::LEN` and `WORD_SIZE`. Other than that, will defer to @digorithm

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1323,14 +1327,14 @@ async fn test_amount_and_asset_forwarding() {
...
+    instance.mint_coins(5_000_000).call().await.unwrap();
 
     balance_result = instance.get_balance(target, asset_id).call().await.unwrap();
-    assert_eq!(balance_result.value, 11);
+    assert_eq!(balance_result.value, 5_000_000);
 
     // Forward 1 coin of native asset_id
     let tx_params = TxParameters::new(None, Some(1_000_000), None);
-    let call_params = CallParameters::new(Some(1), None);
+    let call_params = CallParameters::new(Some(1_000_000), None);
```
vnepveu, line 1337: We should probably document the fact that using "big" numbers is what is also being tested here

fuels-contract/src/contract.rs
```diff
@@ -82,16 +82,23 @@ impl Contract {
         wallet: LocalWallet,
     ) -> Result<Vec<Receipt>, Error> {
         // Script to call the contract. The offset that points to the `script_data` is loaded at the
-        // register `0x12`. Note that we're picking `0x12` simply because it could be any
+        // register `0x13`. Note that we're picking `0x13` simply because it could be any
```
vnepveu, line 85: why did 0x12 changed to 0x13 if "any non-reserved" register can be picked?

fuels-contract/src/contract.rs
```diff
@@ -82,16 +82,23 @@ impl Contract {
...
         // non-reserved register. Then, we use the Opcode to call a contract: `CALL` pointing at the
         // register that we loaded the `script_data` at.
+        let forward_data_offset = ContractId::LEN + WORD_SIZE;
         let (script, offset) = script_with_data_offset!(
             data_offset,
             vec![
                 // Load call data to 0x10.
-                Opcode::MOVI(0x10, data_offset + 32),
+                Opcode::MOVI(0x10, data_offset + forward_data_offset as Immediate18),
```
vnepveu, line 93: Why is it cast as an Immediate18 if we want to support `u64` (which would be a word-size, not a u32 like Immediate18). I'm not familiar with this.

vnepveu (changes requested):
Just some questions because there are things I don't know, otherwise LGTM!!!

fuels-contract/src/contract.rs
adlerjohn: That's an address, and it's the address of the call data. The subsequent two instructions move the memory address of the 8-byte amount in script data into a register, than load from that memory address. That's what does the `u64` amount handling.

vnepveu: Oh ok!!! ty for the explaination

fuels-contract/src/contract.rs
luizstacio: As the address are not reserved. We have chosen to keep an order for adding forwarding values first and then the data offset.

For now, looks like the order will be;
- 0x10 call data offset
- 0x11 gas_price
- 0x12 amount
- 0x13 asset_id

These are soft rules as addresses can be changed

fuels-abigen-macro/tests/harness.rs
luizstacio: What is the best place for documenting it?

adlerjohn: In a comment in the test

fuels-contract/src/contract.rs
vnepveu: These soft rules make sense! But it shoud probably be written in the docstring imo

fuels-abigen-macro/tests/harness.rs
```diff
@@ -1323,14 +1327,16 @@ async fn test_amount_and_asset_forwarding() {
...
 
-    instance.mint_coins(11).call().await.unwrap();
+    instance.mint_coins(5_000_000).call().await.unwrap();
 
     balance_result = instance.get_balance(target, asset_id).call().await.unwrap();
-    assert_eq!(balance_result.value, 11);
+    assert_eq!(balance_result.value, 5_000_000);
 
-    // Forward 1 coin of native asset_id
+    // Forward 1_000_000 gas_limit of native asset_id
```
adlerjohn, line 1335:
```suggestion
```

fuels-contract/src/contract.rs
```diff
@@ -81,17 +81,31 @@ impl Contract {
...
-        // Script to call the contract. The offset that points to the `script_data` is loaded at the
-        // register `0x12`. Note that we're picking `0x12` simply because it could be any
-        // non-reserved register. Then, we use the Opcode to call a contract: `CALL` pointing at the
-        // register that we loaded the `script_data` at.
+        // Script to call the contract.
+        // We use the Opcode to call a contract: `CALL` pointing at the
+        // following registers;
+        //
+        // 0x10 Script data offset
+        // TODO: 0x11 Gas price
```
adlerjohn, line 89:
```suggestion
        // 0x11 Gas price  TODO: https://github.com/FuelLabs/fuels-rs/issues/184
```

vnepveu (approved):
Great job, lgtm!

PR #189, opened 2022-03-28
feat: differentiate calling and simulating a contract call

### LINKED ISSUE #153: Call and send to contract
There should be a distinction between calling a contract, which dry runs essentially, and sending to a contract, which actually sends a valid tx to be included on-chain. The former can be used for e.g. view or pure functions,and does not require the tx to be unique since it's never included on-chain. The latter can be used for stateful writes, and requires the tx to be unique in some way.

Names subject to change, not sure if "call" and "send" are the best names, but IIRC this is what ethers uses.

If a notion of state mutability is added to the ABI, then the SDK could actually warn if you attempt to call a method that mutates the state.

vnepveu:
* This PR closes #153 by adding a `simulate` methods for running transactions without submitting them.

- The existing `call` method is here to signify that there is an actual transaction
  being sent on chain, and so there is a state modification of the blockchain.
- The new `simulate` method means the function being called is essentially `pure`, and 
does not need an actual transaction, but a `dry-run` is enough, so it is simulated by the client. Keep in mind that the transactions are the same, it's just that the simulated one is ran on a fork of the current state.

### DIFF
--- a/fuels-contract/src/contract.rs
+++ b/fuels-contract/src/contract.rs
@@ -68,10 +68,10 @@ impl Contract {
     /// data is for instance the function selector. (script, script_data) is returned as a tuple
     /// of hex-encoded value vectors
     pub fn build_script(
-        contract_id: ContractId,
-        encoded_selector: Option<Selector>,
-        encoded_args: Option<Vec<u8>>,
-        call_parameters: CallParameters,
+        contract_id: &ContractId,
+        encoded_selector: &Option<Selector>,
+        encoded_args: &Option<Vec<u8>>,
+        call_parameters: &CallParameters,
         compute_calldata_offset: bool,
     ) -> Result<(Vec<u8>, Vec<u8>), Error> {
         // Script to call the contract.
@@ -174,10 +174,10 @@ impl Contract {
         simulate: bool,
     ) -> Result<Vec<Receipt>, Error> {
         let (script, script_data) = Self::build_script(
-            contract_id,
-            encoded_selector,
-            encoded_args,
-            call_parameters,
+            &contract_id,
+            &encoded_selector,
+            &encoded_args,
+            &call_parameters,
             compute_calldata_offset,
         )?;
         let mut inputs: Vec<Input> = vec![];
--- a/fuels-contract/src/contract.rs
+++ b/fuels-contract/src/contract.rs
@@ -168,6 +168,7 @@ impl Contract {
         compute_calldata_offset: bool,
         external_contracts: Option<Vec<ContractId>>,
         wallet: LocalWallet,
+        simulate: bool,
     ) -> Result<Vec<Receipt>, Error> {
         let (script, script_data) = Self::build_script(
             contract_id,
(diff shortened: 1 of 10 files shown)


vnepveu:
EDIT: after Slack discussion, we are doing 
```call->call and adding a simulate``` instead of doing `call -> send and adding a call`

**Comment by vnepveu** [comment]
EDIT: further discussion on Slack showed that I overlooked essential parts of the requirements, re-working on it.

File: fuels-contract/src/contract.rs [hunk 1]
```diff
@@ -159,7 +148,35 @@ impl Contract {
...
+        Ok((script, script_data))
+    }
 
+    /// Calls an already-deployed contract code.
+    /// Note that this is a "generic" call to a contract
+    /// and it doesn't, yet, call a specific ABI function in that contract.
+    /// We need a wallet to pay for the transaction fees (even though they are 0 right now)
+    #[allow(clippy::too_many_arguments)] // We need that many arguments for now
+    pub async fn call(
+        contract_id: ContractId,
```
Voxelot, line 160: Might be worth factoring these args into a params struct with nice defaults and builder pattern.

vnepveu: Definitely agree,tracking this in #190

fuels-contract/src/contract.rs
```diff
@@ -445,14 +466,14 @@ where
...
-    /// transaction
-    pub async fn call(self) -> Result<CallResponse<D>, Error> {
-        let mut receipts = Contract::call(
+    /// Call a contract's method on the node. If `simulate==true`, then the call is done in a
+    /// read-only manner, using a `dry-run`. Return a Result<CallResponse, Error>. The CallResponse
+    /// structs contains the method's value in its `value` field as an actual typed value `D` (if
+    /// your method returns `bool`, it will be a bool, works also for structs thanks to the
+    /// `abigen!()`). The other field of CallResponse, `receipts`, contains the receipts of the
+    /// transaction.
+    pub async fn call_or_simulate(self, simulate: bool) -> Result<CallResponse<D>, Error> {
```
adlerjohn, line 475:
```suggestion
    async fn call_or_simulate(self, simulate: bool) -> Result<CallResponse<D>, Error> {
```

fuels-contract/src/contract.rs
```diff
@@ -159,7 +148,35 @@ impl Contract {
         if let Some(e) = encoded_args {
             script_data.extend(e)
         }
+        Ok((script, script_data))
+    }
 
+    /// Calls an already-deployed contract code.
```
digorithm, line 154:
```suggestion
    /// Calls a contract method.
```

Just a nit; I don't see when one would call a non-deployed contract method.

fuels-contract/src/contract.rs
```diff
@@ -159,7 +148,35 @@ impl Contract {
...
+    /// Note that this is a "generic" call to a contract
+    /// and it doesn't, yet, call a specific ABI function in that contract.
+    /// We need a wallet to pay for the transaction fees (even though they are 0 right now)
```
digorithm, line 157: I think this is a very confusing comment. `"generic"` here is very vague.

> it doesn't, yet, call a specific ABI function in that contract

This isn't correct, is it? It does call a specific ABI function in the contract. The function is `encoded_selector` with `encoded_args` as arguments. And it is called at the end, through `script.call(fuel_client).await` (or `.simulate`).

> We need a wallet to pay for the transaction fees

And we do have a wallet in here: `wallet: LocalWallet`.

fuels-contract/src/contract.rs
```diff
@@ -445,14 +466,14 @@ where
...
-    /// The CallResponse structs contains the method's value in its `value`
-    /// field as an actual typed value `D` (if your method returns `bool`, it will
-    /// be a bool, works also for structs thanks to the `abigen!()`).
-    /// The other field of CallResponse, `receipts`, contains the receipts of the
-    /// transaction
-    pub async fn call(self) -> Result<CallResponse<D>, Error> {
-        let mut receipts = Contract::call(
+    /// Call a contract's method on the node. If `simulate==true`, then the call is done in a
+    /// read-only manner, using a `dry-run`. Return a Result<CallResponse, Error>. The CallResponse
+    /// structs contains the method's value in its `value` field as an actual typed value `D` (if
```
digorithm, line 471:
```suggestion
    /// struct contains the method's value in its `value` field as an actual typed value `D` (if
```

fuels-contract/src/contract.rs
adlerjohn: I think that's just copy-paste of the original comment, which as you say is outdated.

fuels-contract/src/contract.rs
```diff
@@ -242,7 +259,11 @@ impl Contract {
 
         let script = Script::new(tx);
 
-        script.call(fuel_client).await
+        if simulate {
+            script.simulate(fuel_client).await
+        } else {
+            script.call(fuel_client).await
+        }
```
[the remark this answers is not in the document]
adlerjohn: Would a `match` be better?

fuels-contract/src/contract.rs
vnepveu: The wallet bit was just to justify that it's there even though transactions are free

Voxelot:
@vnepveu looks like there's some build errors to address

vnepveu:
fixed it, ready for review now

adlerjohn (approved):

digorithm (approved):

PR #191, opened 2022-03-29
Allow assets to be forwarded to contracts on calls

Voxelot:
Previously, contract calls only allowed the default asset to be sent in the call data. This change allows non-default asset ids to be spent when calling a contract.


fuels-abigen-macro/tests/harness.rs
```diff
@@ -1363,8 +1363,22 @@ async fn test_amount_and_asset_forwarding() {
...
+    let address = wallet.address();
+    let address = testfuelcoincontract_mod::Address {
+        value: address.into(),
+    };
+    instance
+        .transfer_coins_to_output(1_000_000, c_id, address)
+        .append_variable_outputs(1)
+        .call()
+        .await
+        .unwrap();
```
vnepveu, line 1379: we should probably add a test showing that the balance of the contract is the one we expecte

vnepveu (changes requested):
I think an additional test could be added, otherwise gg for the fix

fuels-abigen-macro/tests/harness.rs
digorithm: Not that necessary to test for that. 

If there's an issue with the amount, the contract id, the address being passed, or the internal tx building process, `unwrap()` will catch it and fail the test. If the script/script_data encoding breaks, the VM will return an error. If all works, then checking if the balance of the contract is as expected would be testing the client and VM internals, which shouldn't be `fuels-rs` responsibility.

vnepveu: Oh ok!!! smart

Issue #99, opened 2022-02-15
Better handle generated custom types that are equivalent between Sway and `fuel_types`


digorithm:
# Motivation 

In cases where a Sway contract is using the `stdlib`'s native structs such as:

```rust
use std::{address::Address, contract_id::ContractId};

pub struct ParamsGetBalance {
    target: b256,
    asset_id: ContractId,
}
```

the SDK's `abigen!` will generate these as custom structs:

```json
{
    "components": [
        {
            "components": null,
            "name": "value",
            "type": "b256"
        }
    ],
    "name": "asset_id",
    "type": "struct ContractId"
}
```

```rust
 pub struct ContractId {
    pub value: [u8; 32],
}
```

And, then, when using it in the SDK:

```rust
#[tokio::test]
async fn mint() {
    let salt = get_new_salt();

    abigen!(TestFuelCoinContract, "test_projects/token_ops/src/abi.json",);

    let compiled = Contract::compile_sway_contract("test_projects/token_ops", salt).unwrap();

    let (client, fuelcoin_id) = Contract::launch_and_deploy(&compiled).await.unwrap();

    let fuel_coin_instance = TestFuelCoinContract::new(compiled, client);

    let c = testfuelcoincontract_mod::ContractId {
        value: fuelcoin_id.into(),
    };

    let default_balance_params = ParamsGetBalance {
        target: fuelcoin_id.into(),
        asset_id: c,
    };

    let balance_result = fuel_coin_instance.get_balance(default_balance_params);

    let result = fuel_coin_instance.mint_coins(11).call().await.unwrap();
}
```

Note that you have to reference the generated `ContractId` through its namespace: `testfuelcoincontract_mod::ContractId`. But, underneath, it's equivalent to `fuel_types::ContractId`: they're `[u8; 32]`;

This works fine, but it would be more ergonomic if we treated Sway's `ContractId` (and `Address`) as `fuel_types::ContractId`. So that we could eliminate this conversion step:

```rust
 let c = testfuelcoincontract_mod::ContractId {
    value: fuelcoin_id.into(),
};
```

And simply use `fuelcoin_id: fuel_types::ContractId` to a contract method that takes the Sway's `ContractId`.

# Potential solution 

The `abigen!` process would have to detect that the Sway type being generated is named as `ContractId` (and others such as `Address`) and allow for `fuel_types::ContractId` to be used instead. This require significant changes to the `abigen!` macro.

[...]

PR #194, opened 2022-03-30
Accept Sway-native types directly and without wrappers

digorithm:
Closes https://github.com/FuelLabs/fuels-rs/issues/99 (you can also find more context in that issue).

This PR enables support for the Sway-native types `ContractId` and `Address` in the ABI methods. This means no need to create unergonomic wrappers. 

Before:

```Rust
// This is the contract id we want to pass around.
let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
  .await
  .unwrap();
  
// We have to create these ugly wrappers to make it work.
let target = testfuelcoincontract_mod::ContractId { value: id.into() };
let asset_id = testfuelcoincontract_mod::ContractId { value: id.into() };

// ...

let mut balance_result = instance
    .get_balance(target.clone(), asset_id.clone())
    .call()
    .await
    .unwrap();
```

Now:

```Rust
// This is the contract id we want to pass around.
let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
    .await
    .unwrap();

// ... 

// Can use `id` (which is a `fuel_tx::ContractId`) directly in the generated ABI method.
let mut balance_result = instance.get_balance(id, id).call().await.unwrap();
```

### DIFF
--- a/fuels-core/src/code_gen/abigen.rs
+++ b/fuels-core/src/code_gen/abigen.rs
@@ -111,16 +111,14 @@ impl Abigen {
         } else {
             (
                 quote! {
-                    // aliasing to `FuelContractId` to prevent conflicts between
-                    // `fuel_tx::ContractId` and user-defined structs named `ContractId`.
-                    use fuel_tx::ContractId as FuelContractId;
+                    use fuel_tx::{ContractId, Address};
                     use fuels_contract::contract::{Contract, ContractCall};
                     use fuels_signers::{provider::Provider, LocalWallet};
                     use std::str::FromStr;
                 },
                 quote! {
                     pub struct #name {
-                        contract_id: FuelContractId,
+                        contract_id: ContractId,
                         provider: Provider,
                         wallet: LocalWallet
                     }
@@ -128,7 +126,7 @@ impl Abigen {
                     impl #name {
                         pub fn new(contract_id: String, provider: Provider, wallet: LocalWallet)
                         -> Self {
-                            let contract_id = FuelContractId::from_str(&contract_id).unwrap();
+                            let contract_id = ContractId::from_str(&contract_id).unwrap();
                             Self{ contract_id, provider, wallet }
                         }
                         #contract_functions
@@ -180,6 +178,13 @@ impl Abigen {
         let mut seen_struct: Vec<&str> = vec![];
 
         for prop in self.custom_structs.values() {
+            // Skip custom type generation if the custom type is a Sway-native type.
+            // This means ABI methods receiving or returning a Sway-native type
+            // can receive or return that native type directly.
+            if prop.type_field.contains("ContractId") || prop.type_field.contains("Address") {
+                continue;
+            }
+
             if !seen_struct.contains(&prop.type_field.as_str()) {
                 structs.extend(expand_internal_struct(prop)?);
                 seen_struct.push(&prop.type_field);
--- a/fuels-abigen-macro/tests/harness.rs
+++ b/fuels-abigen-macro/tests/harness.rs
@@ -1,4 +1,4 @@
-use fuel_tx::{AssetId, Receipt, Salt};
+use fuel_tx::{AssetId, ContractId, Receipt, Salt};
 use fuels_abigen_macro::abigen;
 use fuels_contract::contract::Contract;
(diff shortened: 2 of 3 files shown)


fuels-core/src/code_gen/abigen.rs
```diff
@@ -180,6 +178,13 @@ impl Abigen {
         let mut seen_struct: Vec<&str> = vec![];
 
         for prop in self.custom_structs.values() {
+            // Skip custom type generation if the custom type is a Sway-native type.
+            // This means ABI methods receiving or returning a Sway-native type
+            // can receive or return that native type directly.
+            if prop.type_field.contains("ContractId") || prop.type_field.contains("Address") {
```
Voxelot, line 184: Should we also support AssetId? Since a call param is actually using an asset id which is only happens to be a contract id in the case of native tokens.

digorithm: That has me confused a bit. Every Sway case I see in the "wild" is using Sway's `ContractId` (https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/contract_id.sw#L5) to represent an `asset_id` -- which works with the current solution here. I don't even see a proper `AssetId` in the Sway stdlib? But maybe I didn't look deep enough.

But yeah, I can throw in a `prop.type_field.contains("AssetId")` here if `AssetId` is an actual thing in Sway 🤔

adlerjohn: There is no `AssetId` in Sway, but there is an `AssetId` in Rust! https://github.com/FuelLabs/fuel-types/blob/a852e072c9654b1d50f9495e8ba1de074fb71725/src/types.rs#L254

This should automatically get converted to a `ContractId` as @Voxelot said.

adlerjohn: Note that in the not-too-distant future Sway will have an `AssetId` as well: https://github.com/FuelLabs/sway/issues/1064

nfurfaro: what @adlerjohn said.

digorithm: Yeah, I know that we have `AssetId` in Rust. But this line we're commenting on is checking the custom types that were defined in Sway code, coming from the JSON ABI. Actual example:

```
{
  "name": "asset_id",
  "type": "struct ContractId",
  "components": [
    {
      "name": "value",
      "type": "b256",
      "components": null
    }
  ]
},
```

So, here, in this section of the code, we're checking for that `struct ContractId`, which won't be `struct AssetId` until Sway has `AssetId` natively. 

The generated code, then, ends up with `asset_id: ContractId`, e.g.: 

```Rust
#[doc = "Calls the contract\'s `get_balance` (0x000000006fc66a82) function"]
pub fn get_balance(&self,target:ContractId,asset_id:ContractId) -> ContractCall<u64>{
  Contract::method_hash(&self.provider,self.contract_id, &self.wallet,[0,0,0,0,111,198,106,130], &[ParamType::U64], &[target.into_token(),asset_id.into_token(),]).expect("method not found (this should never happen)")
}
```

So I'm guessing that by

> This should automatically get converted to a ContractId as @Voxelot said.

You mean the generated `get_balance`, when receiving `(ContractId, AssetId)` should convert that `AssetId` to `ContractId`? As in:

```Rust
 let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
    .await
    .unwrap();

// ...

// Turning it into an actually AssetId
let asset = AssetId::from(*id);

// `id` is `ContractId`, asset is `AssetId`
balance_result = instance.get_balance(id, asset).call().await.unwrap();
```

Even though that method was generated as `get_balance(&self,target:ContractId,asset_id:ContractId)`? 

If that's it... It feels a bit hacky to me. Kinda forcing and making up for something that's missing in Sway itself (`AssetId`). 

Though I don't know, am open to anything.

adlerjohn: I'd be okay with opening an issue for this. If no one complains then it doesn't need to be worked on.

adlerjohn (approved):

vnepveu (approved):
Much cleaner this way :partying_face:

Issue #201, opened 2022-04-04
Update SDK for Sway's change to use IR by default.


otrho:
Sway is switching to use IR by default here: https://github.com/FuelLabs/sway/pull/1130

Contract calls are slightly different compared to current Sway -- user args are always wrapped in a struct, even if there is only one.  This is a bit more consistent and disregards the need for special casing.

Also, values are returned from a contract call via `RET` or `RETD` based now on the value type, not size.  `unit`, `bool`, `byte` and `u64` are 'copy' types returned via `RET`, otherwise `RETD` is used.

digorithm:
Note: this is momentarily blocked by https://github.com/FuelLabs/sway/issues/1146.

mohammadfawaz:
`forc v0.10.0` is now [available](https://github.com/FuelLabs/sway/releases/tag/v0.10.0).

otrho:
I think the design we want is to actually keep the special casing for small struct args but use the newer 'by-type` conventions for `RET` vs `RETD` as mentioned in the description.

PR #203, opened 2022-04-05
Compute calldata offset on `B256` argument

digorithm:
Due to some changes somewhere (VM, compiler? Hard to say because as of now we don't have proper E2E tests and too many fast-moving pieces to pinpoint the actual root cause), contract calls with a single `B256` argument stopped working because we're not including the calldata offset in the `script_data`.  

This workaround (the check for calldata offset) will be removed *entirely* once [this issue is done](https://github.com/FuelLabs/fuels-rs/issues/201) which is currently blocked by [this one in the compiler](https://github.com/FuelLabs/sway/issues/1146).

However, this current problem (calls with `B256` arg) is blocking @nfurfaro on his auth testing work. This PR unblocks him (see the test case I added to this PR, which is now passing).

This is an extremely simple solution: we're now computing calldata offset if the arg is a `B256`: 

```Rust
fn should_compute_call_data_offset(args: &[Token]) -> bool {
    match args
        .iter()
        .any(|t| matches!(t, Token::Struct(_) | Token::Enum(_) | Token::B256(_)))
    {
        true => true,
        false => args.len() > 1,
    }
}
```

That's the only significant change introduced by this PR. The rest of this PR is just adding the projects to test that specific case.


packages/fuels-abigen-macro/tests/test_projects/auth_testing_contract/src/main.sw
```diff
@@ -0,0 +1,50 @@
+contract;
```
mohammadfawaz, line 1: The `fuels-rs` repo probably needs this: https://github.com/FuelLabs/sway/blob/master/.gitattributes for syntax highlighting.

packages/fuels-contract/src/contract.rs
```diff
@@ -341,7 +341,7 @@ impl Contract {
     fn should_compute_call_data_offset(args: &[Token]) -> bool {
         match args
             .iter()
-            .any(|t| matches!(t, Token::Struct(_) | Token::Enum(_)))
+            .any(|t| matches!(t, Token::Struct(_) | Token::Enum(_) | Token::B256(_)))
```
mohammadfawaz, line 344: Tuples should probably be on this list as well? Then there's probably the question about strings and arrays.

Maybe it's better if we invert the condition to check for copy types since they are a limited well defined set that will likely never change?

digorithm: We don't support tuples: https://github.com/FuelLabs/fuels-rs/blob/0c5b9f2c8b89745b1f29e7d8fc15e834e066967d/packages/fuels-core/src/lib.rs#L103-L115

> Maybe it's better if we invert the condition to check for copy types since they are a limited well defined set that will likely never change?

That's a good idea, but this whole thing will go away very soon, as soon as [this is done](https://github.com/FuelLabs/fuels-rs/issues/201). So I didn't wanna shake things too much, but I'm not married to this idea. What do you think?

mohammadfawaz: Ah of course, you're right... let's keep it as is :)

packages/fuels-abigen-macro/tests/test_projects/auth_testing_contract/src/main.sw
```diff
@@ -0,0 +1,50 @@
...
+                ret = true;
+            } else {
+                ret = false;
+            };
+        };
+
+        ret
+    }
+
+    fn returns_msg_sender_address(expected_id: Address) -> bool {
```
vnepveu, line 33: same remark

packages/fuels-contract/src/contract.rs
vnepveu: I agree with Rodrigo as well

packages/fuels-abigen-macro/tests/test_projects/auth_testing_contract/src/main.sw
```diff
@@ -0,0 +1,50 @@
...
+use auth_testing_abi::*;
+use std::result::*;
+use std::assert::assert;
+
+impl AuthTesting for Contract {
+    fn is_caller_external() -> bool {
+        caller_is_external()
+    }
+
+    fn returns_msg_sender(expected_id: ContractId) -> bool {
```
vnepveu, line 15: Also this function is not used, so IMO it should either be tested or removed.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1424,3 +1424,37 @@ async fn test_multiple_args() {
...
+    .unwrap();
+
+    let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
+        .await
+        .unwrap();
+
+    let auth_instance = AuthContract::new(id.to_string(), provider.clone(), wallet.clone());
+
+    let result = auth_instance
+        .returns_msg_sender_address(wallet.address())
```
vnepveu, line 1454: `// Contract returns true if the msg_sender matches the provided argument`

vnepveu (changes requested):
Some nits, otherwise LGTM!

vnepveu (approved):
image

Issue #108, opened 2022-02-22
Abigen support for tuple return values


simonr0204:
`abigen!` panics with :

`error: proc macro panicked
...
  = help: message: called Result::unwrap() on an Err value: MissingData("cannot parse custom type with no components")`
 
 when parsing an ABI containing functions which return tuples, e.g. 
 
 ` fn my_tuple_function(gas_: u64, amount_: u64, asset_id_: b256, params: () ) -> (u64, u64);`

vnepveu:
So after trying a few things (experiences can be found at https://github.com/FuelLabs/fuels-rs/tree/vnepveu/feat-support-tuple-type):
- The "major" blocker for this PR is the "no-orphans" rule of Rust.

Basically, by definition, a tuple is a _foreign type_. And the thing with _foreign types_ is that you can implement only traits that are defined **in the current crate**. Issue being that we need to implement the `Tokenizable` trait, which is defined in a particular crate (`fuels-core` but that's not the issue). 
Passing it over to @digorithm.

A solution I've observed, but is very complicated, in `ethers-rs`: https://github.com/gakonst/ethers-rs/blob/78c87b11269fe94da6d09a04035afd8367fd3ae5/ethers-core/src/abi/tokens.rs#L79.

PR #205, opened 2022-04-05
Add basic support for tuple inputs/outputs

digorithm:
This PR introduces the support for contract calls that take and return tuples. 

Closes https://github.com/FuelLabs/fuels-rs/issues/108.

Here's an example:

```Rust
contract;

abi MyContract {
    fn returns_tuple(input: (u64, u64)) -> (u64, u64);
}

    fn returns_tuple(input: (u64, u64)) -> (u64, u64) {
        input
    }
}
```

And then, in Rust:

```Rust
// ...Setup

let response = instance.returns_tuple((1, 2)).call().await.unwrap();
assert_eq!(response.value, (1, 2));
```


packages/fuels-abigen-macro/tests/test_projects/tuples/out/debug/tuples-abi.json
```diff
@@ -0,0 +1,20 @@
+[
+  {
+    "type": "function",
+    "inputs": [
+      {
+        "name": "input",
+        "type": "(u64, u64)",
```
vnepveu, line 7: Maybe test with 4 different types eg `(u64, bool)` and `(u32, u8)` just to be extra safe?

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -181,6 +181,23 @@ impl ABIDecoder {
                     new_offset: res.new_offset,
                 };
 
+                Ok(result)
+            }
+            ParamType::Tuple(types) => {
```
vnepveu, line 186: I think this should be unit-tested

vnepveu: Although the implementation is straigthforward so idk… :sweat_smile:

packages/fuels-core/src/json_abi.rs
```diff
@@ -237,6 +237,7 @@ impl ABIParser {
 
                 Ok(Token::Enum(Box::new((discriminant as u8, token))))
             }
+            _ => Err(Error::InvalidType(param.to_string())),
```
vnepveu, line 240: I think there should be `Token::Tuple` somewhere?

packages/fuels-core/src/json_abi.rs
```diff
@@ -808,7 +837,7 @@ mod tests {
             .unwrap();
         println!("encoded: {:?}\n", encoded);
 
-        let expected_encode = "00000000058734b9000000000000000100000000000000020000000000000003";
+        let expected_encode = "000000005898d3a4000000000000000100000000000000020000000000000003";
```
vnepveu, line 840: the encoding changed because whitespaces are now removed, right?

packages/fuels-core/src/lib.rs
```diff
@@ -66,6 +67,7 @@ impl ParamType {
             Self::Array(_params, _l) => true,
             // The other primitive types are inside `Return`,
             // thus smaller than one `WORD`.
+            Self::Tuple(_) => true,
```
vnepveu, line 70: I think we should check if `len(Tuple) > 1`

packages/fuels-core/src/json_abi.rs
```diff
@@ -526,12 +532,35 @@ pub fn parse_param(param: &Property) -> Result<ParamType, Error> {
...
+                // Try to parse tuple (T, T, ..., T)
+                return parse_tuple_param(param);
+            }
             // Try to parse enum or struct
             parse_custom_type_param(param)
         }
     }
 }
 
+pub fn parse_tuple_param(param: &Property) -> Result<ParamType, Error> {
```
vnepveu, line 545: This should be unit-tested

packages/fuels-core/src/lib.rs
```diff
@@ -273,6 +283,60 @@ impl Tokenizable for u64 {
...
+            }
+
+            fn into_token(self) -> Token {
+                Token::Tuple(vec![
+                    $( self.$no.into_token(), )+
+                ])
+            }
+        }
+    }
+}
```
vnepveu, line 319: I think for clarity, we should maybe include an example of the macro's expansion as doc somewhere (eg for tuples of len 2), because reading this and checking its validity is pretty hard imo.

vnepveu (changes requested):
I think some tests should be added, otherwise I'm impressed! Very concise :tada:

packages/fuels-abigen-macro/tests/test_projects/tuples/out/debug/tuples-abi.json
digorithm: I love correct software as much as the next person, but if you look closely at the tuple handling code:

```Rust
pub fn parse_tuple_param(param: &Property) -> Result<ParamType, Error> {
  let mut params: Vec<ParamType> = Vec::new();

  let mut chars = param.type_field.chars();
  chars.next(); // Remove "("
  chars.next_back(); // Remove ")"

  for ele in chars.as_str().split(',') {
      let tuple_type = Property {
          name: "".to_owned(),
          type_field: ele.trim().to_owned(),
          components: None,
      };
      params.push(parse_param(&tuple_type)?);
  }

  Ok(ParamType::Tuple(params))
}
```

After we strip `'('` and `')'`, it goes through regular/primitive `ParamType` parsing: `units`, `bool`, etc., that's been tested extensively. 

The same applies to your comment [here](https://github.com/FuelLabs/fuels-rs/pull/205#discussion_r843551352). Where now, instead of encoding, we're talking about decoding:

```Rust
ParamType::Tuple(types) => {
    let mut tokens = vec![];
    let mut new_offset = offset;

    for t in types {
        let res = self.decode_param(t, data, new_offset)?;
        new_offset = res.new_offset;
        tokens.push(res.token);
    }

    let result = DecodeResult {
        token: Token::Tuple(tokens),
        new_offset,
    };

    Ok(result)
}
```

Which, as you said in that comment, the implementation is fairly straightforward, just going through each inner type and doing ` let res = self.decode_param(t, data, new_offset)?;`, which has been fairly tested already.

As much as I _love_ tests, I personally refrain from writing unnecessary and/or redundant code (code == liability).

packages/fuels-core/src/json_abi.rs
digorithm: This (`tokenize`) is actually only used for the JSON ABI CLI tool, which is out of the scope for this issue/PR. The SDK works without any of this `tokenize()`. I've updated it to include a `todo!()` in here instead of `Error`, and the issue for future work is here https://github.com/FuelLabs/fuels-rs/issues/206.

packages/fuels-core/src/lib.rs
digorithm: Sure, by the way, this whole block is going away [very soon](https://github.com/FuelLabs/fuels-rs/issues/201).

packages/fuels-core/src/json_abi.rs
digorithm: https://github.com/FuelLabs/fuels-rs/pull/205#discussion_r844102421.

packages/fuels-core/src/lib.rs
digorithm: Here's the generated impl for a tuple of len 2:

```Rust
impl<A,B, >Tokenizable for(A,B,)where A:Tokenizable,B:Tokenizable,{
  fn from_token(token:Token) -> Result<Self,InvalidOutputType>{
    match token {
      Token::Tuple(mut tokens) => {
        let mut it = tokens.drain(..);
        Ok((A::from_token(it.next().expect("All elements are in vector."))?,B::from_token(it.next().expect("All elements are in vector."))?,))
      },
      other => Err(InvalidOutputType({
        let res = $crate::fmt::format(unsafe {
          std::fmt::Arguments::new_v1(&[], &[std::fmt::ArgumentV1::new(&(other),std::fmt::Display::fmt),])
        });
        res
      })),
    
      }
  }
  fn into_token(self) -> Token {
    Token::Tuple((<[_]>::into_vec(box[(self.0.into_token()),(self.1.into_token())])))
  }
}
```

If you tell me that this is a lot different/more readable from/than the macro definition:

```Rust
macro_rules! impl_tuples {
    ($num: expr, $( $ty: ident : $no: tt, )+) => {
        impl<$($ty, )+> Tokenizable for ($($ty,)+) where
            $(
                $ty: Tokenizable,
            )+
        {
            fn from_token(token: Token) -> Result<Self, InvalidOutputType> {
                match token {
                    Token::Tuple(mut tokens) => {
                        let mut it = tokens.drain(..);
                        Ok(($(
                          $ty::from_token(it.next().expect("All elements are in vector."))?,
                        )+))
                    },
                    other => Err(InvalidOutputType(format!(
                        "Expected `Tuple`, got {:?}",
                        other,
                    ))),
                }
            }

            fn into_token(self) -> Token {
                Token::Tuple(vec![
                    $( self.$no.into_token(), )+
                ])
            }
        }
    }
}
```

Or the `impl Tokenizable for Vec<T>` right above https://github.com/FuelLabs/fuels-rs/blob/fe6727e74d52cf7fd2118a23642a28638133eacd/packages/fuels-core/src/lib.rs#L211-L232 

Which is very similar to the generated one for tuples...

Then yes, I'll 20 extra lines of documentation code to clarify it 😛.

packages/fuels-abigen-macro/tests/test_projects/tuples/out/debug/tuples-abi.json
vnepveu: Makes sense, I often forget that code (even tests) is a liability!

packages/fuels-core/src/json_abi.rs
vnepveu: Oh ok! Just realized the role of json_abi wasn't clear for me then haha

packages/fuels-core/src/lib.rs
vnepveu: Fair enough, it's not more readable! :stuck_out_tongue_closed_eyes:

vnepveu (approved):
Thank you for your answers, they make perfect sense. GG

adlerjohn (approved):
utACK

PR #207, opened 2022-04-06
refactor: add an argument builder for `Contract::call` method

### LINKED ISSUE #190: Factor `Contract::call` params into a arg builder with nice defaults

vnepveu:
This PR closes #190 by implementing a `ContractCallArguments` struct that holds
the arguments for for the `Contract::call` method. Fields which are often
repetitive are abstracted away, but can be set using a `set_*` method on the struct.

### DIFF
--- a/packages/fuels-contract/src/contract.rs
+++ b/packages/fuels-contract/src/contract.rs
@@ -12,11 +12,9 @@ use fuel_types::{Bytes32, Salt, Word};
 use fuel_vm::consts::{REG_CGAS, REG_ONE};
 use fuel_vm::prelude::Contract as FuelContract;
 use fuel_vm::script_with_data_offset;
+use fuels_core::constants::{DEFAULT_COIN_AMOUNT, DEFAULT_MATURITY, NATIVE_ASSET_ID, WORD_SIZE};
 use fuels_core::errors::Error;
-use fuels_core::{
-    constants::DEFAULT_COIN_AMOUNT, constants::WORD_SIZE, Detokenize, Selector, Token,
-};
-use fuels_core::{constants::NATIVE_ASSET_ID, ParamType};
+use fuels_core::{Detokenize, ParamType, Selector, Token};
 use fuels_signers::provider::Provider;
 use fuels_signers::{LocalWallet, Signer};
 use std::marker::PhantomData;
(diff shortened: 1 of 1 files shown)


packages/fuels-contract/src/contract.rs
```diff
@@ -45,6 +43,56 @@ pub struct CallResponse<D> {
...
+            contract_id,
+            encoded_selector,
+            encoded_args,
+            tx_parameters: TxParameters::default(),
+            call_parameters: CallParameters::default(),
+            variable_outputs,
+            maturity: DEFAULT_MATURITY as Word,
+            compute_calldata_offset,
+            external_contracts,
+            // TODO: the fact that transactions are not simulated by default should be a flag
```
adlerjohn, line 77: Issue

packages/fuels-contract/src/contract.rs
```diff
@@ -45,6 +43,56 @@ pub struct CallResponse<D> {
...
+    encoded_args: Option<Vec<u8>>,
+    tx_parameters: TxParameters,
+    call_parameters: CallParameters,
+    variable_outputs: Option<Vec<Output>>,
+    maturity: Word,
+    compute_calldata_offset: bool,
+    external_contracts: Option<Vec<ContractId>>,
+    simulate: bool,
+}
+impl ContractCallArguments {
```
adlerjohn, line 58:
```suggestion
}

impl ContractCallArguments {
```

packages/fuels-contract/src/contract.rs
```diff
@@ -45,6 +43,56 @@ pub struct CallResponse<D> {
...
+    }
+    pub fn set_call_parameters(&mut self, call_params: CallParameters) {
+        self.call_parameters = call_params
+    }
+    pub fn set_maturity(&mut self, maturity: Word) {
+        self.maturity = maturity
+    }
+    pub fn set_simulate(&mut self, simulate: bool) {
+        self.simulate = simulate
+    }
```
adlerjohn, line 93: Hmmm. Do we really want this to be set like the other parameters? It's neither a call not a tx parameter. It's a separate thing, about what the SDK will actually do.

adlerjohn (commented):
Are these changes tested?

vnepveu:
Changes are not tested because I don't see how I could test them, it's just a "repackaging". I mean I could probably unit-test the setter methods though

packages/fuels-contract/src/contract.rs
vnepveu: My idea is that "by default", you usually don't simulate, so having to do `set_simulate` isn't a big deal. I can change it if you think that's not good.

packages/fuels-contract/src/contract.rs
```diff
@@ -45,6 +43,57 @@ pub struct CallResponse<D> {
     pub receipts: Vec<Receipt>,
 }
 
+pub struct ContractCallArguments {
```
adlerjohn, line 46: Ah this explains my confusion. Your change is to the internals rather than the user-facing interface. Builder pattern is much more useful for user-facing things. Internally I think it could even be counter-productive since you _want_ compiler errors when you change things, rather an unexpected behavior at runtime.

```suggestion
struct ContractCallArguments {
```

adlerjohn (changes requested):

packages/fuels-contract/src/contract.rs
vnepveu: Ok so that's why I was confused by this issue. Actually we already  have builder methods for the _user-facing_ call method, see an example of that here:https://github.com/FuelLabs/fuels-rs/blob/5edde493227459246eeec937d1a4855b7f8a40b2/packages/fuels-abigen-macro/tests/harness.rs#L619

adlerjohn: Can we close this PR and issue then? I think the original suggestion to use a builder pattern was also assuming this was user-facing due to `pub`.

vnepveu: I think so yes.

vnepveu:
Closing because it's actually irrelevant.

PR #209, opened 2022-04-06
Use remote stdlib and update tests

adlerjohn:
1. Use the latest stdlib from the `sway` repo. They will be locked, so without further intervention won't change.
2. Rebuild test contracts with `forc` `v0.9.2`.
3. Update test suite.

Note that two tests had to be commented out, related to #201


vnepveu (changes requested):
LGTM, but since you're there renaming and restructuring test projects, would you mind switching all test project name to using `_` (instead of `-`)? This way we all agree on it and are coherent with the`sway` repo.

digorithm (approved):
Many thanks for cleaning this up! 
lgtm

PR #213, opened 2022-04-14
feat: support `Tuple` in `Enum` and `Enum` in `Tuple`

### LINKED ISSUE #53: Support `enums` inside `structs` and `structs` inside `enums` in the ABIgen
For now, those features are not implemented ie we cannot expand nested types defined in an ABI spec

### LINKED ISSUE #63: `expand_fn_outputs` is not consistent with type checking

vnepveu:
This PR closes #53 by supporting nested `struct` and `enums` as return types.
It also closes #63 because of refactoring. 
Some caveats: 
- Currently does not work as `inputs`, but the infrastructure is theoritically here. I'm waiting for the `RESERV00` bug to be solved to see if it works
- Arbitrary depth of nesting has *not* been tested yet!


digorithm:
Moving this summary of our conversation here so it's explicit what needs to be done:

1. We need to support enum output by updating `expand_fn_outputs` to include the enum name instead of its components, which looks like a tuple, leading to the wrong `Tokenizable`;
2. We need to `abigen` `Tokenizable` for enum types.

packages/fuels-abigen-macro/tests/test_projects/enum_inside_struct/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["lulwat"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "enum_inside_struct"
+
+[dependencies]
+std = { git = "https://github.com/FuelLabs/sway-lib-std" }
```
adlerjohn, line 8: Rather than default branch, just leave this implicit. Then it'll use the stdlib version that matches the `forc` version.

packages/fuels-abigen-macro/tests/test_projects/enum_inside_struct/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["lulwat"]
```
adlerjohn, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

packages/fuels-core/src/json_abi.rs
```diff
@@ -522,7 +525,9 @@ pub fn parse_param(param: &Property) -> Result<ParamType, Error> {
         // Simple case (primitive types, no arrays, including string)
         Ok(param_type) => Ok(param_type),
         Err(_) => {
-            if param.type_field.contains("struct") || param.type_field.contains("enum") {
+            if param.type_field.starts_with(STRUCT_KEYWORD)
```
vnepveu, line 528: use `is_custom_type`

vnepveu (commented):
Just re-reading what I did, it's way too big lol

packages/fuels-core/src/json_abi.rs
vnepveu: The `is_custom_type` function should be move alongside the `Property` type definition, and some util methods on `Property` could be added to make the code more readable.

digorithm:
@vnepveu Is this still a draft? Or officially ready for review? Let me know so I can take another deep pass.

vnepveu:
It's definitely ready for review just thought I'd review myself as well

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -223,47 +224,57 @@ impl Abigen {
...
 
-        for inner_struct in inner_structs {
-            let inner_custom_type_name =
-                extract_custom_type_name_from_abi_property(&inner_struct, ty)
-                    .expect("failed to extract custom type name");
-            structs
-                .entry(inner_custom_type_name)
-                .or_insert(inner_struct);
+        for nested_custom_type in nested_custom_types {
+            if is_custom_type(&nested_custom_type) {
```
digorithm, line 255: Isn't this check redundant? We're adding the custom types to `nested_custom_types` by using `get_nested_custom_properties()` [0] which already checks for `is_custom_type()` [1].

[0]:

```rust
// Find inner {structs, enums} in case of nested custom types
for inner_component in prop.components.as_ref().unwrap() {
    nested_custom_types
        .extend(Abigen::get_nested_custom_properties(inner_component));
}
```

[1]:

```rust
 fn get_nested_custom_properties(prop: &Property) -> Vec<Property> {
    let mut props = Vec::new();

    if is_custom_type(prop) {
        props.push(prop.clone());

        for inner_prop in prop.components.as_ref().unwrap() {
            let inner = Abigen::get_nested_custom_properties(inner_prop);
            props.extend(inner);
        }
    }

    props
}
```

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -35,6 +38,8 @@ pub fn expand_internal_struct(prop: &Property) -> Result<TokenStream, Error> {
     // `MyStruct::new_from_tokens()`.
     let mut args = Vec::new();
 
+    let mut contains_enums = false;
```
digorithm, line 41: How about we eliminate the need for this `contains_enums`? 

I don't think we need the `Default` derive at all (personally couldn't find any use case). So this https://github.com/FuelLabs/fuels-rs/blob/087425f3d69afe5df790f0d441078b198ade40be/packages/fuels-core/src/code_gen/custom_types_gen.rs#L134-L138 could be simplified and we could remove the `contains_enums` variable.

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -161,42 +187,85 @@ pub fn expand_internal_struct(prop: &Property) -> Result<TokenStream, Error> {
...
+    let mut param_types = Vec::new();
 
     for (discriminant, component) in components.iter().enumerate() {
-        let field_name = ident(&component.name.to_class_case());
+        let variant_name = ident(&component.name.to_class_case());
+        let dis = discriminant as u8;
 
         let param_type = parse_param(component)?;
-        match param_type {
+        match param_type.clone() {
```
digorithm, line 210:
```suggestion
        match param_type {
```

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -220,15 +295,56 @@ pub fn expand_internal_enum(name: &str, prop: &Property) -> Result<TokenStream,
...
+
+            pub fn new_from_tokens(tokens: &[Token]) -> Self {
+                if tokens.is_empty() {
+                    panic!("Empty tokens array received in `{}::new_from_tokens`",
+                        #enum_name);
+                }
+                // For some reason sometimes we receive arrays that have multiple elements, with the
+                // first token being a `Token::Enum`. We only consider that `Enum` token in that
+                // case
+                // TODO: figure out what is actually happening and if this is normal
```
digorithm, line 307: Can you show me an example that this is reproduced? We should document this if that's the case.

packages/fuels-core/src/constants.rs
```diff
@@ -5,6 +5,8 @@ pub const DEFAULT_GAS_LIMIT: u64 = 1_000_000;
 pub const DEFAULT_GAS_PRICE: u64 = 0;
 pub const DEFAULT_BYTE_PRICE: u64 = 0;
 pub const DEFAULT_MATURITY: u32 = 0;
+pub const STRUCT_KEYWORD: &str = "struct ";
```
digorithm, line 8: Can we add a comment explaining this const? It's very unclear to unfamiliar readers why we have that whitespace at the end.

digorithm (changes requested):
It's looking really good. Great job on this -- this is no simple feat 😃. Just left some comments and improvement suggestions.

packages/fuels-core/src/code_gen/custom_types_gen.rs
vnepveu: It's very hard to reproduce in an example because it can only be "displayed" through a `panic!` print. I will create an issue for this and investigate once it is merged though.

packages/fuels-core/src/code_gen/abigen.rs
vnepveu: Yup, good catch

digorithm (approved):

PR #219, opened 2022-04-20
Add `fuels-types` minimal crate for declaring serializable ABI types. Resolves the cycle between `fuels-rs` and `sway` repos.

mitchmindtree:
This moves the `JsonABI` type along with the associated `Function` and `Property` types from `sway-types` into a new minimal `fuels-types` crate.

By moving these from `sway-types` upstream to the `fuels-rs` repo, we can resolve the remainder of the cyclic dependency between the `sway` and `fuels-rs` repositories.

I think also semantically it makes a little more sense for the Rust representation of these ABI-specific types to be declared under the Rust SDK, and for Sway to depend on these rather than vice versa.

The `fuels-core` crate now depends on these types via the new `fuels-types` crate. The unused `sway-utils` dependency has been removed.

As an immediate follow-up to `fuels-types` landing and getting published to crates.io, we should remove these types from the `sway-types` crate and instead depend on `fuels-types` as necessary throughout the `sway` repo.

This is one of the blockers for https://github.com/FuelLabs/sway/pull/1321.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,7 @@ jobs:
         ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-core/Cargo.toml
         ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-abigen-macro/Cargo.toml
         ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-rs/Cargo.toml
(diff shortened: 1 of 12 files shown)


JoshuaBatty (dismissed):
utACK. Nice one

adlerjohn (commented):
Nice, looks like this finally breaks the circular dependency! I'll defer to @digorithm

vnepveu (approved):
There will be a corresponding PR for Sway if I understand correctly? To make it import from this new crate?

digorithm (approved):
Great stuff, thanks for doing this @mitchmindtree!

PR #228, opened 2022-04-25
Refactor: make a `fuels-test-helpers` crate

### LINKED ISSUE #180: Remove `pub` from constant `DEFAULT_INITIAL_BALANCE`
This was discussed here: https://github.com/FuelLabs/fuels-rs/pull/175#discussion_r835701407

Trying to make it `pub(crate)` would also require changes on `fuels-signers/src/util.rs`.

### LINKED ISSUE #221: Fix version of fuels dependencies in fuels-rs
Should match version

https://github.com/FuelLabs/fuels-rs/blob/7fda121547ee28de91145f8e8034db7e02b79b48/packages/fuels-rs/Cargo.toml#L12-L15

tedbyron:
Ref #180 - `DEFAULT_INITIAL_BALANCE` moved to new crate, but still `pub`
Fixes #221 - commit [5c10cca](https://github.com/FuelLabs/fuels-rs/pull/228/commits/5c10ccaadd0423aa3da336faa20eac4dd088689d)

# What changed
Extract the `utils` mod from `fuels-signer` into its own crate along with `DEFAULT_INITIAL_BALANCE` const from `fuels-core`. 

### Notes
`fuels-test-helpers` is from `utils` mod in `fuels-signer`, but I had to change the functions a bit to avoid having cyclic dependencies between `fuels-test-helpers` and `fuels-signer`. Tests in `fuels-signer` can't use `Provider` or `Wallet` types returned from `fuels-test-helper` or else you get this error:

```
error[E0308]: mismatched types
  --> packages/fuels-signers/src/lib.rs:54:64
   |
54 |         let wallet = LocalWallet::new_from_private_key(secret, provider).unwrap();
   |                                                                ^^^^^^^^ expected struct `Provider`, found struct `fuels_signers::provider::Provider`
```
(https://github.com/rust-lang/rust/issues/79381)

# Reasoning
I think making a separate crate for test helpers will help in the future for all the SDK crates so they can more easily add examples and tests. Having a separate crate is also useful for downstream SDK consumers to add to their dev-dependencies if it not wanted in normal SDK. I could further refactor the crates, but don't want to make breaking changes to the main sdk api.


packages/fuels-rs/src/lib.rs
```diff
@@ -28,6 +28,10 @@ pub mod signers {
     pub use fuels_signers::*;
 }
 
+pub mod test_helpers {
```
tedbyron, line 31: I didn't think about this before, this will be breaking change unless `DEFAULT_INITIAL_BALANCE` is also expotred from `mod core`

vnepveu (changes requested):
nits

packages/fuels-core/src/constants.rs
```diff
@@ -12,10 +12,6 @@ pub const WORD_SIZE: usize = core::mem::size_of::<Word>();
 // when initializing wallets for now.
 pub const DEFAULT_COIN_AMOUNT: u64 = 1_000_000;
```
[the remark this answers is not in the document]
tedbyron: done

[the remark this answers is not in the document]
tedbyron: had to add `fuels-test-helpers` as dep of `fuels-contract` just for that though. Thoughts?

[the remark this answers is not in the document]
vnepveu: I _think_ it's fine, but defer to @digorithm to be sure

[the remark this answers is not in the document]
vnepveu: In the long run, this would be a default set up via configuration, so putting it in `test-helpers` wouldn't make sense.

packages/fuels-rs/Cargo.toml
```diff
@@ -13,3 +13,5 @@ fuels-abigen-macro = { version = "0.9.0", path = "../fuels-abigen-macro" }
 fuels-contract = { version = "0.9.0", path = "../fuels-contract" }
 fuels-core = { version = "0.9.0", path = "../fuels-core" }
 fuels-signers = { version = "0.9.0", path = "../fuels-signers" }
+# This probably shouldn't be here, SDK users should put it in their dev-dependencies.
+fuels-test-helpers = { version = "0.9.0", path = "../fuels-test-helpers" }
```
adlerjohn, line 17: This version `0.9.0` doesn't actually exist on crates.io, and never will if the first version of this crate is `0.9.1`.

adlerjohn (changes requested):

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -0,0 +1,26 @@
...
+use fuel_core::model::coin::Coin;
+use fuel_tx::UtxoId;
+use fuels_signers::provider::Provider;
+use fuels_signers::LocalWallet;
+
+#[cfg(feature = "fuels-signers")]
+pub async fn setup_test_provider_and_wallet() -> (Provider, LocalWallet) {
+    //  We build only 1 coin with amount TEST_COIN_AMOUNT, empirically determined to be
+    //  sufficient right now
+    let (pk, coins) = crate::setup_address_and_coins(1, crate::DEFAULT_INITIAL_BALANCE);
```
vnepveu, line 12: `use crate::DEFAULT_INITAL_BALANCE` and then `DEFAULT_INITIAL_BALANCE`

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -0,0 +1,26 @@
+use std::net::SocketAddr;
+
+use fuel_core::model::coin::Coin;
+use fuel_tx::UtxoId;
+use fuels_signers::provider::Provider;
+use fuels_signers::LocalWallet;
+
+#[cfg(feature = "fuels-signers")]
+pub async fn setup_test_provider_and_wallet() -> (Provider, LocalWallet) {
+    //  We build only 1 coin with amount TEST_COIN_AMOUNT, empirically determined to be
```
vnepveu, line 10: where is `TEST_COIN_AMOUNT`

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -0,0 +1,98 @@
+//! Testing helpers/utilities for Fuel SDK.
+
+use std::net::SocketAddr;
+
+use rand::{Fill, Rng};
+use secp256k1::{PublicKey, Secp256k1, SecretKey};
+
```
vnepveu, line 7: don't add newline

vnepveu (changes requested):
some additional nits

packages/fuels-test-helpers/src/signers.rs
tedbyron: yeah I just ctrl+c ctrl+v the `fuels-signers/utils.rs` mostly

tedbyron: I switched comment to `DEFAULT_INITIAL_BALANCE`

packages/fuels-test-helpers/src/lib.rs
tedbyron: [5c10cca](https://github.com/FuelLabs/fuels-rs/pull/228/commits/5c10ccaadd0423aa3da336faa20eac4dd088689d)

packages/fuels-rs/Cargo.toml
tedbyron: fixed in [5c10cca](https://github.com/FuelLabs/fuels-rs/pull/228/commits/5c10ccaadd0423aa3da336faa20eac4dd088689d), also closes #221

adlerjohn (approved):
Will defer to @vnepveu and @digorithm

vnepveu (approved):
LGTM, thanks a lot :partying_face:

Issue #210, opened 2022-04-07
Compile Sway contracts using latest (or at least version-pinned) `forc` in each CI run


digorithm:
# Problem

Currently, we commit the JSON ABI and compiled binary for each Sway test project, used in our test suite, all compiled manually. This leads to a big problem: these compiled binaries aren't frequently updated. This means updates in the compiler might break the SDK but not get caught in the CI pipeline, because the CI is running our test suite that's running test against currently existing Sway binaries, which was compiled at an earlier point in time, potentially using an earlier version of `forc`. 

# Solution

We must add the Sway compilation process in the CI pipeline, somehow. That means we would stop committing the compiled binaries and JSON ABI files, and these would be generated by, sway, a simple bash script that would install the latest `forc` (or a pinned version), run `forc build` for each test project, and then the test suite would run against these freshly compiled binaries.

adlerjohn:
This may help (though I would prefer a Rust version): https://github.com/FuelLabs/sway/issues/1082

digorithm:
I have a working version of this, but currently blocked by https://github.com/FuelLabs/sway/issues/1335 because building one of the test projects fails with the error mentioned in that issue.

PR #234, opened 2022-04-27
Build Sway projects at every CI run

digorithm:
This PR: 

1. Removes all lock files, ABI files, and binary files for test projects used in the `harness.rs` test suite;
2. Adds respective `gitignore`s to not include lock files, ABI files, and binary files;
3. Adds a Rust binary to build all test projects (`cargo run --bin build-test-projects`). This needs to be run locally when running tests;
4. Updates the CI workflow file to run the above binary in order to build all test projects, from scratch, using the latest `forc`, at every new CI run, guaranteeing fresh binaries when running the test suite.

Closes https://github.com/FuelLabs/fuels-rs/issues/210.
Closes https://github.com/FuelLabs/fuels-rs/issues/201.


.github/workflows/ci.yml
```diff
@@ -68,26 +80,26 @@ jobs:
...
+
+      - name: Verify tag version
+        run: |
+          cargo install toml-cli
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-contract/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-core/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-abigen-macro/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-rs/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} packages/fuels-types/Cargo.toml
+          ./.github/workflows/scripts/verify_tag.sh ${{ github.ref_name }} tools/fuels-abi-cli/Cargo.toml
```
vnepveu, line 100: we would need the new `fuels-helpers` crate here now, no?

packages/fuels-abigen-macro/tests/test_projects/contract_output_test/Forc.lock
```diff
@@ -2,7 +2,7 @@
 name = 'contract_output_test'
```
vnepveu, line 2: This file should be removed? unless it's special

packages/fuels-abigen-macro/tests/test_projects/swayswap/swayswap_abi/Forc.toml
```diff
@@ -0,0 +1,5 @@
+[project]
+authors = ["Mohammad"]
```
vnepveu, line 2: I don't think the `swayswap` project should be in there

packages/fuels-abigen-macro/tests/test_projects/swayswap/swayswap_abi/src/main.sw
```diff
@@ -0,0 +1,23 @@
+library swayswap_abi;
+
```
vnepveu, line 2: same

scripts/build-test-projects/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "build-test-projects"
+description = "Runs `forc build` for all projects under the main test suite"
+version = "0.0.0"
+edition = "2021"
+publish = false
```
vnepveu, line 6: add newline

vnepveu (changes requested):
Just remove the swayswap project and add some newlines, otherwise great work!!! This is going to be awesome to have!

packages/fuels-abigen-macro/tests/test_projects/contract_output_test/Forc.lock
digorithm: Oh no, it should!

packages/fuels-abigen-macro/tests/test_projects/swayswap/swayswap_abi/Forc.toml
digorithm: I believe it should. It's a massive test case for the SDK, it covers a _lot_ of use cases.

Voxelot:
You could consider using the somewhat idiomatic `xtask` approach for these build scripts.

https://github.com/matklad/cargo-xtask

We use it in fuel-core and it works great.

digorithm:
> You could consider using the somewhat idiomatic `xtask` approach for these build scripts.
> 
> https://github.com/matklad/cargo-xtask
> 
> We use it in fuel-core and it works great.

That'd be a nice idea to explore in future PRs -- I have no idea how it works and it would require some exploration time on my end, plus: 
> cargo-xtask is neither an officially recommended workflow, nor a de-facto standard (yet?). It might or might not work for your use case.

For now, I'd rather push this out and unblock _many_ things that are waiting for the next release.

vnepveu (dismissed):
LGTM :tada: :tada:

.github/workflows/ci.yml
```diff
@@ -41,6 +41,18 @@ jobs:
         run: |
           git config --global core.bigfilethreshold 100m
 
+      - name: Install Forc
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: forc
```
adlerjohn, line 48: You should fix a `forc` version here instead of using latest

adlerjohn (changes requested):
above

vnepveu (approved):

adlerjohn (approved):

.github/workflows/ci.yml
vnepveu: Wait isn't it going to be problematic? I feel like the whole point was to use the last compiler version (ie last `forc` version) so we shouldn't pin it?

digorithm: Yeah, ideally we want to test it against the latest. But, for instance, if for some reason we don't wanna use the latest (e.g. latest breaks something on the SDK, we're aware of it, but we're still working on a fix), we might not want to block other PR's CI pipeline because of it. So having this control is nice. 

But, then again, we have to watch ourselves and make sure we're updating this version when there's a new `forc` version.

adlerjohn: If we want the best of both worlds, then we can create a second CI workflow that runs against `latest` `forc` and isn't required to pass to merge PRs. Then if it breaks we know, but it's not blocking anything.

vnepveu: Ok, I feel like we should find a way to ensure this?

digorithm: Not a huge fan of that because I believe it would increase the CI runtime considerably :(. We're currently sitting at ~10min, running another one, in this case, would mean building all Sway projects again.

vnepveu: maybe the solution would be to run the `fuels-rs` CI on every `forc` major release?
But just on releases?

vnepveu: Ah no wait that doesn't work

Issue #215, opened 2022-04-18
Include salt gen process for ergonomics


SilentCicero:
Currently, we force a developer to provide their own Salt and they have to bring their own library to do this. 

But in the vein and philosophy of Ethers.js (as Richard would put it), we want to make this an all-in-one stop shop for a developer to use Fuel. Including rand in the library and just having Salt pre-gen for now is much nicer than having to specify and include this every time. Just include it for ergonomics.

I'd like to see `rand` just included as a dependency for fuels-rs. I don't like specifying rand every-time. I think the general philosophy should be, if I have to do it everytime, then it should just be provided nicely as a default in the repo (even if its extra bulk).

This:
```rust
use fuel_tx::Salt;
use fuels_abigen_macro::abigen;
use fuels_contract::contract::Contract;
use rand::rngs::StdRng;
use rand::{Rng, SeedableRng};

// Generate Rust bindings from our contract JSON ABI
abigen!(MyContract, "./my-contract-abi.json");

#[tokio::test]
async fn harness() {
    let rng = &mut StdRng::seed_from_u64(2322u64);

    // Build the contract
    let salt: [u8; 32] = rng.gen();
    let salt = Salt::from(salt);

    let compiled = Contract::compile_sway_contract("./", salt).unwrap();
    let client = Provider::launch(Config::local_node()).await.unwrap();
    let contract_id = Contract::deploy(&compiled, &client).await.unwrap();
    println!("Contract deployed @ {:x}", contract_id);
```

Should be:
```rust
use fuels_abigen_macro::abigen;
use fuels_contract::contract::Contract;

// Generate Rust bindings from our contract JSON ABI
abigen!(MyContract, "./my-contract-abi.json");

#[tokio::test]
async fn harness() {
    let compiled = Contract::compile_sway_contract("./").unwrap();
    let client = Provider::launch(Config::local_node()).await.unwrap();
    let contract_id = Contract::deploy(&compiled, &client).await.unwrap();
    println!("Contract deployed @ {:x}", contract_id);
```

And if I want to see what the Salt that was generated, there should just be a trait or property of some kind to have visibility into it from the compiler or contract_id property.

My two cents.

digorithm:
Oh, 100%. This is a great idea and the current way we ask the users to do their own rng was supposed to be just something temporary that was introduced at the very early stages of the SDK development. 

[...]

PR #240, opened 2022-04-28
add new load_sway_contract_with_salt method

Salka1988:
WIP.

Closes https://github.com/FuelLabs/fuels-rs/issues/215.


packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -672,24 +664,19 @@ async fn example_workflow() {
     assert_eq!(52, result.value);
 }
 
+// TODO https://github.com/FuelLabs/fuels-rs/issues/201
 #[tokio::test]
+#[ignore]
```
digorithm, line 669: Why are we ignoring this test case? https://github.com/FuelLabs/fuels-rs/issues/201 has been solved and closed.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1549,55 +1430,3 @@ async fn workflow_struct_inside_enum() {
     let expected = Shaker::Cosmopolitan(Recipe { ice: 22, sugar: 99 });
     assert_eq!(result.value, expected);
 }
-
-#[tokio::test]
-async fn test_logd_receipts() {
```
digorithm, line 1554: Why are we deleting this test case?

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1,15 +1,14 @@
...
-use rand::rngs::StdRng;
-use rand::{Rng, SeedableRng};
+use fuels_core::constants::NATIVE_ASSET_ID;
+use fuels_core::Token;
 use sha2::{Digest, Sha256};
 
-/// Note: all the tests and examples below require pre-compiled Sway projects.
-/// To compile these projects, run `cargo run --bin build-test-projects`.
-/// It will build all test projects, creating their respective binaries,
-/// ABI files, and lock files. These are not to be committed to the repository.
```
digorithm, line 11: Why are we deleting this? This is still true and this comment is helpful to newcomers who don't know that they have to run `cargo run --bin build-test-projects`.

Salka1988: I brought back these comments and now they are on the branch.

packages/fuels-abigen-macro/tests/harness.rs
Salka1988: I didn't restore this because I'm not sure why it was deleted.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -672,24 +669,18 @@ async fn example_workflow() {
     assert_eq!(52, result.value);
 }
 
+// TODO https://github.com/FuelLabs/fuels-rs/issues/201
```
digorithm, line 672:
```suggestion
```

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1078,8 +1015,10 @@ async fn test_large_return_data() {
         ]
     );
 
-    let res = contract_instance.get_small_string().call().await.unwrap();
-    assert_eq!(res.value, "gggggggg");
+    // TODO https://github.com/FuelLabs/fuels-rs/issues/201
```
digorithm, line 1018:
```suggestion
```

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1078,8 +1015,10 @@ async fn test_large_return_data() {
...
+    // One word-sized string
+    // let res = contract_instance.get_small_string().call().await.unwrap();
+    // assert_eq!(res.value, "gggggggg");
```
digorithm, line 1021:
```suggestion
    // One word-sized string
    let res = contract_instance.get_small_string().call().await.unwrap();
    assert_eq!(res.value, "gggggggg");
```

This should be working now.

packages/fuels-contract/src/contract.rs
```diff
@@ -384,11 +384,26 @@ impl Contract {
 
         match provider.client.submit(&tx).await {
             Ok(_) => Ok(contract_id),
-            Err(e) => Err(Error::TransactionError(e.to_string())),
+            Err(e) => Err(Error::TransactionError(format!(
+                "{} {}",
+                e.to_string(),
+                "-> maybe use load_sway_contract_with_salt",
```
digorithm, line 390: I think this message will, by and large, be very misleading. 

The overwhelming majority of error cases aren't going to be related to the salt. For instance, it could be due to bad network conditions or poorly formatted transactions. So, in most cases, people will be led to believe that using `load_sway_contract_with_salt()` could make a difference, when in most cases it won't, making it a very frustrating experience. 

I'd say remove the suggestion altogether or parse the error, and, if it has something to do with ID collision, suggest using the load with salt method.

digorithm (changes requested):
It's looking really good, so much better seeing those `rng` + `salt` setups going away 😄.

Just left a couple of comments and suggestions.

leviathanbeak:
@Salka1988 the commit messages have all the same name, they should be squashed into a single descriptive message, we don't want to pollute commit logs

packages/fuels-contract/src/contract.rs
Salka1988: That's why I made issue #249 . In this case, i do not get the error type but the string type that is sent from the FuelClient.query function. I could parse that string and look within it for something that would indicate ID collision, but is that a good solution? Maybe it would be better if the fuel-core in the FuelClient.query function sends enum insted of a string, that would indicate that a certain problem has occurred (from GraphQlError to fuel-core.errors enum) .
Like you said, it would be best to delete suggestion "-> maybe use load_sway_contract_with_salt" until it is decided how we will proceed.

digorithm (changes requested):
`packages/fuels-abigen-macro/tests/test_projects/foo_caller_contract/out/debug/foo_caller_contract.bin`

Shouldn't be committed; we don't commit `.bin`s, lockfiles, etc. See https://github.com/FuelLabs/fuels-rs/blob/master/packages/fuels-abigen-macro/tests/test_projects/foo_caller_contract/.gitignore.

digorithm (approved):
Great job on your first PR!

adlerjohn:
> we don't commit .bins, lockfiles, etc

@digorithm The linked-to .gitignore doesn't ignore `.bin` files. Was `target` renamed to `out` at some point?

packages/fuels-abigen-macro/tests/test_projects/foo_caller_contract/src/main.sw
```diff
@@ -9,7 +9,7 @@ abi FooCaller {
 
 impl FooCaller for Contract {
     fn call_foo_contract(value: bool) -> bool {
-        let foo_contract = abi(FooContract, 0x5b987da578669aa7f733110c3b1e99678fee8f9dd1302e6562c0a6f35bab4b26);
+        let foo_contract = abi(FooContract, 0xfe98f602add19c4b2d0c8be2929e4300f9f154eda43457b8b9ea02ef2c7b2d3c);
```
adlerjohn, line 12: Can you file an issue to refactor this to take the contract ID as a parameter?

PR #243, opened 2022-04-29
Specify minimum supported Rust version in `fuels-rs` manifest

mitchmindtree:
This ensures that users are warned in the case that their installed Rust version predates the version required by `fuels-rs`.

This should be particularly useful for users building `forc`-generated test harnesses, as I believe we've already had a couple of reports of strange compile errors that just required a Rust version update.

You can find documentation on this field [here][1].

[1]: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

---

Originally I had planned on exposing this minimum required Rust version from a function in the crate root [like so](https://github.com/mitchmindtree/fuels-rs/commit/17ad3fd16d48f3c50d688e1333ac8eb8dc59dbac). The idea was that `forc` could use this function to specify the necessary `rust-version` field in the generated test harness manifest files.

However after thinking on this a little more, I realised it should be enough to simply specify the minimum required version here as `fuels` will be a dependency within the generated test harness anyway, meaning the user will be notified either way. As a result, I believe this closes https://github.com/FuelLabs/sway/issues/1239.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
+  DASEL_VERSION: https://github.com/TomWright/dasel/releases/download/v1.24.3/dasel_linux_amd64
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


packages/fuels-rs/Cargo.toml
```diff
@@ -7,6 +7,7 @@ homepage = "https://fuel.network/"
 license = "Apache-2.0"
 repository = "https://github.com/FuelLabs/fuels-rs"
 description = "Fuel Rust SDK."
+rust-version = "1.58.0"
```
adlerjohn, line 10: Should this conform to https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md?

mitchmindtree: Is there a `Cargo.toml` formatting lint we can use in CI?

adlerjohn: https://github.com/FuelLabs/fuel-core/blob/670d16379ed1420061fc9772043b86e7101a3a6e/.github/workflows/ci.yml#L33-L39

adlerjohn: Which is actually https://github.com/FuelLabs/cargo-toml-lint!

mitchmindtree: OK, I think this has been addressed! It was just the ordering of fields right?

mitchmindtree: Ahh, I just need to resolve conflicts

adlerjohn (approved):
LGTM, but will defer to @digorithm

mitchmindtree:
I'm just realising we should probably have another CI check that *uses* the minimum specified version to `cargo check` the workspace and make sure we don't actually require a newer version?

I'm unsure whether or not specifying the `rust-version` like this will cause cargo to automatically error on use of language features that became available after the specified version - I suspect that it won't.

ra0x3 (dismissed):

digorithm (approved):
> I'm unsure whether or not specifying the rust-version like this will cause cargo to automatically error on use of language features that became available after the specified version - I suspect that it won't.

Yeah, that would be very weird. I think it should be fine.

Voxelot:
@ra0x3 fuel-rs is independent of forc. For example in our faucet app, we use the SDK to send genesis coins to users without interacting with contracts or forc at all.

adlerjohn:
Put another way, fuels-rs is independent of the particular high level language used.

mitchmindtree:
Thanks for finishing this off / landing this!

Issue #250, opened 2022-05-04
Make fuel-core an optional dependency


Voxelot:
Fuel Core is a heavy dependency and is only needed in the SDK for local testing. It is fairly burdensome on SDK users in production scenarios to require them to include the entire fuel-core dependency tree in their applications.

Move fuel-core and related SDK apis to a feature flag, allowing users to opt-in to this extra functionality during tests.

PR #251, opened 2022-05-04
make fuel-core optional in fuels-signers

Voxelot:
fixes: #250 

Running a local copy of fuel-core from fuel-signers::Provider is purely a convenience function for local testing. However, when using the SDK in production services, we shouldn't be including the extra dependencies from fuel-core as they're unnecessary when calling a remote fuel-core node.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,6 @@
 [workspace]
+# Use the new resolver to prevent dev-deps and build-deps from enabling debugging or test features in production.
+resolver = "2"
 members = [
     "packages/fuels",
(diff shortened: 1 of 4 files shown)


digorithm (approved):

Voxelot:
found it was also being included by fuels-contracts because it was enabling test-helpers in signers by default, even though it doesn't use the helpers.

digorithm:
> found it was also being included by fuels-contracts because it was enabling test-helpers in signers by default, even though it doesn't use the helpers.

Nice catch!

digorithm (approved):

PR #253, opened 2022-05-05
Ensure minimum supported rust version is verified in CI

Voxelot:
Addresses @mitchmindtree's concern about ensuring backwards compatibility with the rust-version specified in the fuels Cargo.toml.

If the rust-version gets updated in fuels, ci will break unless its version is also updated.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: stable
+          toolchain: 1.58.0
           override: true
 
       - name: Install rustfmt
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -24,16 +24,27 @@ jobs:
...
           override: true
 
+        # selecting a toolchain either by action or manual `rustup` calls should happen
+        # before the cache plugin, as it uses the current rustc version as its cache key
+      - uses: Swatinem/rust-cache@v1
+
+      # Ensure CI is using the same minimum toolchain specified in fuels Cargo.toml
+      - name: Verify Rust Version
+        run: |
+          curl -sSLf "https://github.com/TomWright/dasel/releases/download/v1.24.3/dasel_linux_amd64" -L -o dasel && chmod +x dasel
```
adlerjohn, line 37: Should this URL be abstracted into a constant?

Voxelot: done

Voxelot:
example of ci failure when ci version doesn't match cargo.toml rust version: https://github.com/FuelLabs/fuels-rs/runs/6299722947?check_suite_focus=true#step:5:12

digorithm (approved):
That's awesome -- thanks for addressing this one!

PR #268, opened 2022-05-08
Contract deployment QoL

### LINKED ISSUE #217: Deployment and loading of contract QoL can be improved
Currently, deploying and loading contracts is done in multiple steps, but could be done more ergonomically.

As per my issue:
https://github.com/FuelLabs/fuels-rs/issues/215#issuecomment-1101622119

```rust
use fuels_contract::contract::Contract;

// Generate Rust bindings from our contract JSON ABI
abigen!(MyContract, "./my-contract-abi.json");

let compiled = Contract::load_sway_contract(path,salt).unwrap();
let id = Contract::deploy(&compiled, &provider, &wallet, TxParameters::default())
        .await
        .unwrap();
```

Could just return a contract instance for example and a helper function could do the two activities (as they are often done in tandem anyway):

```rust
use fuels_contract::contract::Contract;

// Generate Rust bindings from our contract JSON ABI
abigen!(MyContract, "./my-contract-abi.json");

let contract_instance = Contract::deploy_compiled(path, MyContract, &wallet, TxParameters::default())
        .await
        .unwrap();
```

Returning the contract instance is far more helpful. Including RNG generation is also nice. Wallet like Ethers.js / RS, could include a provider property to grab.

As well, MyContract could also have the path to the contracts included at that stage, i.e.:

```rust
abigen!(MyContract, "./my-contract-abi.json", "./");

let contract_instance = Contract::deploy_compiled(MyContract, &wallet, TxParameters::default())
        .await
        .unwrap();
```

I think approaches with the mentality above could make everything even more ergonomic allowing the developer to focus on just the logic they need to think about and less config and wiring that should just be inherited.

MujkicA:
Closes #217

### DIFF
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -43,8 +43,8 @@ abigen!(
     "your_project/out/debug/contract_test-abi.json",
 );
 
-// This helper will setup a local node/provider and a test wallet.
-let (provider, wallet) = setup_test_provider_and_wallet().await;
+// This helper will setup a test wallet linked to a local node/provider.
+let wallet = setup_test_provider_and_wallet().await;
 
 // Load the compiled Sway contract (this is the output from `forc build`)
 let compiled = Contract::load_sway_contract(
@@ -56,19 +56,19 @@ let gas_price = 0;
 let gas_limit = 1_000_000;
 let byte_price = 0;
 
-// Deploy the contract
-let contract_id = Contract::deploy(&compiled, provider, wallet,
-TxParameters::new(gas_price, gas_limit, byte_price)).await.unwrap();
+// Deploy the compiled Sway contract (this is the output from `forc build`)
+let contract_id = Contract::deploy("your_project/out/debug/contract_test.bin", wallet,
+    TxParameters::new(gas_price, gas_limit, byte_price)).await.unwrap();
 ```
 
 ## Calling and configuring contract calls
 
 Once you've deployed your contract, as seen in the previous section, you'll likely want to call contract methods and configure some parameters such as gas price, byte price, gas limit, and forward coins in your contract call.
 
-Start by creating an instance of your contract once you have a provider and wallet set up:
+Start by creating an instance of your contract once you have a wallet set up:
 
 ```Rust
-let contract_instance = MyContract::new(contract_id.to_string(), provider, wallet);
+let contract_instance = MyContract::new(contract_id.to_string(), wallet);
 ```
 
 Then we move to configuring contract calls.
@@ -241,7 +241,7 @@ For a more concrete example, see the `test_contract_calling_contract` function i
 
 ## Connecting to existing contracts
 
-If you already have a deployed contract and want to call its methods using the SDK,  but without deploying it again, all you need is the contract ID of your deployed contract. You can skip the whole deployment setup and call `::new(contract_id, provider, wallet)` directly. For example:
+If you already have a deployed contract and want to call its methods using the SDK,  but without deploying it again, all you need is the contract ID of your deployed contract. You can skip the whole deployment setup and call `::new(contract_id, wallet)` directly. For example:
 
 ```Rust
 abigen!(
@@ -249,11 +249,11 @@ abigen!(
     "path/to/abi.json"
 );
 
-let (provider, wallet) = setup_test_provider_and_wallet().await;
+let wallet = setup_test_provider_and_wallet().await;
 
 let contract_id = "0x0123..." // Your contract ID as a string.
 
-let connected_contract_instance = MyContract::new(contract_id, provider, wallet);
+let connected_contract_instance = MyContract::new(contract_id, wallet);
 ```
 
 ## More examples
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -4,7 +4,7 @@ At a high level, the Fuel Rust SDK can be used to build Rust-based applications 
 
 For this interaction to work, the SDK must be able to communicate to a `fuel-core` node; you have two options at your disposal:
 
-1. Use the SDK's native `setup_test_provider_and_wallet()` that runs a short-lived test Fuel node;
+1. Use the SDK's native `launch_provider_and_get_wallet()` that runs a short-lived test Fuel node;
 2. Run a Fuel node outside your SDK code (using `fuel-core`) and point your SDK to that node's IP and port.
 
 The first option is ideal for contract testing, as you can quickly spin up and tear down nodes between specific test cases.
@@ -43,8 +43,8 @@ abigen!(
     "your_project/out/debug/contract_test-abi.json",
 );
 
-// This helper will setup a test wallet linked to a local node/provider.
-let wallet = setup_test_provider_and_wallet().await;
+// This helper will will launch a local node and provide a test wallet linked to it
+let wallet = launch_provider_and_get_wallet().await;
 
 // Load the compiled Sway contract (this is the output from `forc build`)
 let compiled = Contract::load_sway_contract(
@@ -249,7 +249,7 @@ abigen!(
     "path/to/abi.json"
 );
 
-let wallet = setup_test_provider_and_wallet().await;
+let wallet = launch_provider_and_get_wallet().await;
 
 let contract_id = "0x0123..." // Your contract ID as a string.
 
--- a/packages/fuels-abigen-macro/tests/harness.rs
+++ b/packages/fuels-abigen-macro/tests/harness.rs
@@ -28,10 +28,10 @@ async fn compile_bindings_from_contract_file() {
         "packages/fuels-abigen-macro/tests/takes_ints_returns_bool.json",
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     // Calls the function defined in the JSON ABI.
     // Note that this is type-safe, if the function does exist
@@ -83,9 +83,9 @@ async fn compile_bindings_from_inline_contract() {
         "#,
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
     //`SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let contract_call = contract_instance.takes_ints_returns_bool(42_u32);
 
@@ -123,10 +123,10 @@ async fn compile_bindings_array_input() {
         "#,
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let input: Vec<u16> = vec![1, 2, 3, 4];
     let contract_call = contract_instance.takes_array(input);
@@ -168,10 +168,10 @@ async fn compile_bindings_bool_array_input() {
         "#,
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let input: Vec<bool> = vec![true, false, true];
     let contract_call = contract_instance.takes_array(input);
@@ -213,10 +213,10 @@ async fn compile_bindings_byte_input() {
         "#,
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let contract_call = contract_instance.takes_byte(10 as u8);
 
@@ -254,10 +254,10 @@ async fn compile_bindings_string_input() {
         "#,
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let contract_call = contract_instance.takes_string("This is a full sentence".into());
 
@@ -298,10 +298,10 @@ async fn compile_bindings_b256_input() {
         "#,
     );
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let mut hasher = Sha256::new();
     hasher.update("test string".as_bytes());
@@ -361,10 +361,10 @@ async fn compile_bindings_struct_input() {
         bar: "fuel".to_string(),
     };
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let contract_call = contract_instance.takes_struct(input);
 
@@ -426,10 +426,10 @@ async fn compile_bindings_nested_struct_input() {
         foo: inner_struct,
     };
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let contract_call = contract_instance.takes_nested_struct(input);
 
@@ -477,10 +477,10 @@ async fn compile_bindings_enum_input() {
 
     let variant = MyEnum::X(42);
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
 
     // `SimpleContract` is the name of the contract
-    let contract_instance = SimpleContract::new(null_contract_id(), provider, wallet);
+    let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let contract_call = contract_instance.takes_enum(variant);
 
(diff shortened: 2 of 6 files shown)


packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -36,7 +38,11 @@ pub struct Abigen {
 
 impl Abigen {
     /// Creates a new contract with the given ABI JSON source.
-    pub fn new<S: AsRef<str>>(contract_name: &str, abi_source: S) -> Result<Self, Error> {
+    pub fn new<S: AsRef<str>>(
+        contract_name: &str,
+        abi_source: S,
+        binary_filepath: &Option<String>,
```
leviathanbeak, line 44: any special reason this isn't `Option<&str>` instead of `&Option<String>` ?

packages/fuels-abigen-macro/Cargo.toml
```diff
@@ -12,6 +12,7 @@ description = "Fuel Rust SDK marcros to generate types from ABI."
 proc-macro = true
 
 [dependencies]
+async-trait = "0.1.53"
```
leviathanbeak, line 15: `0.1` should suffice

packages/fuels-core/src/code_gen/abigen.rs
MujkicA: adjusted it

packages/fuels-contract/src/contract.rs
```diff
@@ -571,14 +596,14 @@ where
 
     /// Call a contract's method on the node, in a state-modifying manner.
     pub async fn call(self) -> Result<CallResponse<D>, Error> {
-        Ok(Self::call_or_simulate(self, false).await?)
+        Self::call_or_simulate(self, false).await
```
vnepveu, line 599: why did you remove the `?` ? Maybe I don't understand how the `?` works haha

packages/fuels-contract/src/contract.rs
```diff
@@ -22,6 +23,12 @@ use fuels_signers::provider::Provider;
 use fuels_signers::{LocalWallet, Signer};
 use std::marker::PhantomData;
 
+#[async_trait]
+pub trait Deployable {
```
vnepveu, line 27: I don't understand the point of this trait if it is used only once. Traits are used, afaik, to have a sort of generic interface. In our case, only `Contracts` are `Deployable` basically, so it does not make sense IMO to create a whole trait for it, especially if you have to use a new crate for the trait to exist.

packages/fuels-abigen-macro/src/lib.rs
```diff
@@ -21,7 +26,13 @@ pub fn abigen(input: TokenStream) -> TokenStream {
 pub fn wasm_abigen(input: TokenStream) -> TokenStream {
     let args = parse_macro_input!(input as Spanned<ContractArgs>);
 
-    let c = Abigen::new(&args.name, &args.abi).unwrap().no_std();
+    let c = Abigen::new(
+        &args.name,
+        &args.abi,
+        args.binary_filepath.as_ref().map(|v| v.as_ref()),
```
vnepveu, line 32: not a big fan of this `.map()` call, it seems weird to me

packages/fuels-abigen-macro/src/lib.rs
```diff
@@ -90,10 +102,28 @@ impl ParseInner for ContractArgs {
             let literal = input.parse::<LitStr>()?;
             (literal.span(), literal.value())
         };
+
+        if !input.is_empty() {
+            input.parse::<Token![,]>()?;
+        }
+
+        let binary_filepath = if !input.is_empty() {
```
vnepveu, line 110: it should probably be refactored with the statement below IMO

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -608,10 +608,10 @@ async fn create_nested_struct_from_decoded_tokens() {
     assert_eq!(10, nested_struct_from_tokens.x);
     assert!(nested_struct_from_tokens.y.a);
 
-    let (provider, wallet) = setup_test_provider_and_wallet().await;
+    let (_, wallet) = setup_test_provider_and_wallet().await;
```
vnepveu, line 611: the fact that the provider is consistently put as an ignored variable (`_`) tells me that either:
- the function  should be refactored
- a new function should be written

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -335,7 +381,34 @@ mod tests {
...
         // there are 5
-        let _bindings = Abigen::new("test", contract).unwrap().generate().unwrap();
+        let _bindings = Abigen::new("test", contract, None)
+            .unwrap()
+            .generate()
+            .unwrap();
+    }
+
+    #[test]
+    fn generates_deployable() {
```
vnepveu, line 391: This test does not test anything from what I understand, you should probably add an assert statement or expect a panic from the test

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -152,13 +166,42 @@ impl Abigen {
...
                 #code
+                #deployable_trait
 
                 #abi_structs
                 #abi_enums
             }
         })
     }
 
+    fn deployable_trait(&self, name: &Ident) -> TokenStream {
```
vnepveu, line 177: The function should be name `generate_deployable_trait` or similar. I'm not sure of the relevance of it though.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -152,13 +166,42 @@ impl Abigen {
...
+        if let Some(binary_filepath) = self.binary_filepath.as_ref() {
+            return quote! {
+                #[async_trait]
+                impl Deployable for #name {
+                    async fn deploy_compiled(wallet: LocalWallet, params: TxParameters) -> Self {
```
vnepveu, line 182: The function should not be called `deploy_compiled` but rather `compile_and_deploy`. Also this function should IMO not be in the ABIgen, it is a function that should be handled on the SDK side (like for instance the `load_sway` or `compile` function) and should not be generated as part of a macro call.

vnepveu (changes requested):
I'm having trouble understanding your logic here, looking forward to understand it better!

packages/fuels-contract/src/contract.rs
leviathanbeak: `call_or_simulate()` already returns the same type, so the old code is redundant

packages/fuels-core/src/code_gen/abigen.rs
digorithm: > The function should not be called deploy_compiled but rather compile_and_deploy

The SDK isn't responsible for compiling. We are not compiling the Sway code here. We're simply `deploy`ing the `compiled` code, hence `deploy_compiled`.

packages/fuels-core/src/code_gen/abigen.rs
MujkicA: The idea was to just have a simple test case which triggers the generation of the `Deployable` implementation for the binding

packages/fuels-abigen-macro/tests/harness.rs
MujkicA: I don't have a good enough idea of all the use cases users might have. There are some scenarios where direct interaction with the provider is needed, but it feels awkward do it via the wallet.

packages/fuels-contract/src/contract.rs
MujkicA: I've added some more context as to why this is needed in a comment. Please take a look and let me know what you think.

digorithm:
That's looking really nice. Can we make it look like

```Rust
abigen!(
    MyContract,
    "tests/test_projects/contract_test",
);
```

So that `abigen!` could infer both the `.bin` and `.json`? You mentioned something about this potentially leading to some errors on the user side. Could you expand on that? It would be super nice, from a user perspective, to have to include only a single project path.

MujkicA:
> That's looking really nice. Can we make it look like
> 
> ```rust
> abigen!(
>     MyContract,
>     "tests/test_projects/contract_test",
> );
> ```
> 
> So that `abigen!` could infer both the `.bin` and `.json`? You mentioned something about this potentially leading to some errors on the user side. Could you expand on that? It would be super nice, from a user perspective, to have to include only a single project path.

Agreed, it would look a lot better, however we would need some way to make the user know if they are in deployable mode or if loading the binary failed. 
As for the errors, assuming we fetch files with .json/.bin extensions, what happens if the folder structure does not match the default /out/debug or if there are unrelated files with the same extension. I fear such errors might not be transparent and lead to frustration. Or am I being too cautious here?

packages/fuels-contract/src/contract.rs
MujkicA: Actually you were right, we can do without the trait here. I went trough quite a few iterations and the trait is just a remnant. There is still one advantage to it, if the user defines functions `deploy_compiled()`, `new()`, `get_binary_filepath()` on the sway side, abigen will fail. This way they could coexist and be accessed via fully qualified syntax.

packages/fuels-abigen-macro/src/lib.rs
leviathanbeak: @vnepveu what's wrong with the map function? do you have any cleaner way to _map_ `Option<T>` to `Option<U>` 🤔

vnepveu: I  think I'm just not used to seeing `map` being used on something that is not iterable and only has one element. But if that's the only way then let's go!

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: I think having a function that is just `setup_test_wallet` (that the `setup_test_provider_and_wallet` would use) would be good here! This way you only get what you need in the cases where you used `_`, but we can still get the `provider` when it's needed. What do you think?

packages/fuels-abigen-macro/src/lib.rs
leviathanbeak: it's a different type of `map` the one that works on `Option<T>` types:

https://doc.rust-lang.org/std/option/enum.Option.html#method.map

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1049,34 +977,30 @@ async fn test_large_return_data() {
 async fn test_provider_launch_and_connect() {
```
vnepveu, line 977: I think we ought to keep an E2E test for the `load_sway` and `deploy` workflow. Even if it's not the one that we are going to mainly use, we still need to have it tested somewhere. So keep this test and add another one with the "new" workflow (can be in the same test probably)

vnepveu (changes requested):
thank you for the context, much clearer! still some questions

digorithm:
> Agreed, it would look a lot better, however we would need some way to make the user know if they are in deployable mode or if loading the binary failed. 

This is a bit unclear to me. What do you mean by deployable mode?

> As for the errors, assuming we fetch files with .json/.bin extensions, what happens if the folder structure does not match the default /out/debug or if there are unrelated files with the same extension. I fear such errors might not be transparent and lead to frustration. Or am I being too cautious here?

Ah, this is a good point. Then maybe let's not assume `out/debug/` and, instead, assume that the directory we're passing to the `abigen!()` is the directory that contains both the `.bin` and `.json`? So a user would usually pass `/project_path/out/debug/`. That would relax the requirement of having to keep it all in `out/debug/`, but it would still require that both files are in the same directory. I _think_ that's okay?

adlerjohn:
> I assume there are use cases where only the ABI is loaded without the binary.

Yes, since requiring the contract bytecode to interact with a contract defeats the purpose of a separate ABI file. The ABI alone should be sufficient to interact with a contract.

digorithm:
> I assume there are use cases where only the ABI is loaded without the binary. By deployable mode I mean when the binary is loaded and the user wants to deploy it.
> 
> >

Perfect, that makes a lot of sense, thank you!

packages/fuels-contract/src/contract.rs
MujkicA: `new`, other binding methods and any sway methods are generated within the same `impl` block. I ran a quick test with `new` defined in the json ABI to confirm we get a collision. I would therefore move new to the trait impl and keep it for now, maybe with a different name. If you know of a cleaner way to handle this, let me know.

packages/fuels-abigen-macro/src/lib.rs
MujkicA: What would be the better way to do it or what is the problem with the way it is now?

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: Since we are keeping the `load_sway` and `deploy` function, we should keep them tested. The question of removing those two functions can indeed be asked, but I think offering more granularity to the users is always nice, especially if the code is already there.

packages/fuels-core/src/code_gen/abigen.rs
vnepveu: summary of our discussion on Slack:
- The "onchain" workflow (ie loading the contract and deploying it) should not be dependent on the ABIgen workflow
- So we should just compress `load and deploy` into one function, but the contract instantiation `Contract::new()` should be separate. 

@digorithm  this is a decision that creates an extra "line of code" but IMO it's worth it compared to the confusion that could arise from having on-chain deployment being done in the ABIgen, what do you think?

packages/fuels-abigen-macro/src/lib.rs
vnepveu: I would do a full branch with `!input.is_empty`, no? Since you're testing that condition twice?

MujkicA:
Needs to be reopened

packages/fuels-abigen-macro/tests/harness.rs
MujkicA: Following the discussion on slack, I will create a separate issue to confirm if `load` has other use-cases. But I removed `deploy` to enforce a more streamlined approach via `load_and_deploy`.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1085,6 +1058,9 @@ async fn test_provider_launch_and_connect() {
         .unwrap();
     assert_eq!(42, result.value);
 
+    wallet.set_provider(launched_provider);
```
digorithm, line 1061: I'm a bit confused about this test case now that we've made these changes. It feels like there's no difference between what all the other tests are doing; deploying is a different process than connecting to a deployed contract. Every test in here deploys and connects, and I'm not sure about the value of changing the provider here (`wallet.set_provider()`). Is this test case (`test_provider_launch_and_connect`) completely redundant now?

MujkicA: My understanding is that the test checks if behavior is consistent when making calls on a launched vs. connected `provider` instances. The binding now uses the `provider` from the supplied `wallet` to make method calls so I change it to `launched_provider` before making the second call.

digorithm: Fair enough, not entirely redundant then!

digorithm (changes requested):
This is looking great -- much simpler! Also, extra nice because of the simple solution.

Just one more thing, can you update the [SDK book](https://github.com/FuelLabs/fuels-rs/blob/master/docs/src/getting-started/basics.md) to reflect the changes introduced here?

docs/src/getting-started/basics.md
```diff
@@ -43,8 +43,8 @@ abigen!(
     "your_project/out/debug/contract_test-abi.json",
 );
 
-// This helper will setup a local node/provider and a test wallet.
-let (provider, wallet) = setup_test_provider_and_wallet().await;
+// This helper will setup a test wallet linked to a local node/provider.
```
vnepveu, line 46: `will launch a local node and provide a test wallet linked to it`, I think it's important to mention that a local node gets launched by the function.

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: If this is the main test for testing `connected` vs `launched` instances, then the test `connect_to_deployed_contract` can be removed because it does this (actually badly now haha)

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -913,19 +899,17 @@ async fn test_connect_to_deployed_contract() {
         "packages/fuels-abigen-macro/tests/test_projects/contract_test/out/debug/contract_test-abi.json"
```
vnepveu, line 899: remove this test (see comment below on `test_provider_launch_and_connect`)

packages/fuels-contract/src/contract.rs
```diff
@@ -379,20 +379,18 @@ impl Contract {
             })
     }
 
-    /// Deploys a compiled contract to a running node
```
vnepveu, line 382: Don't remove the docstring

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -6,15 +6,14 @@ use fuels_signers::LocalWallet;
 use std::net::SocketAddr;
 
 #[cfg(feature = "fuels-signers")]
-pub async fn setup_test_provider_and_wallet() -> (Provider, LocalWallet) {
+pub async fn setup_test_provider_and_wallet() -> LocalWallet {
```
vnepveu, line 9: Add a docstring

vnepveu (changes requested):
This is looking supeeer clean now, I love it! Some documentation still needed

docs/src/getting-started/basics.md
MujkicA: I found it redundant since it is already stated at the beginning, but I guess it can't hurt to repeat it

packages/fuels-test-helpers/src/signers.rs
digorithm: Was this actually resolved? I'm not seeing the docstring @vnepveu suggested.

packages/fuels-contract/src/contract.rs
digorithm: Was this actually resolved? I'm still seeing the docstring removed.

MujkicA: My bad

adlerjohn (commented):
CI is failing due to some unrelated tests using the `master` branch for the stdlib. This should not be done; tests should use the implicit stdlib that matches the version of `forc` used in CI. Paging @vnepveu and @digorithm to investigate.

vnepveu:
> CI is failing due to some unrelated tests using the `master` branch for the stdlib. This should not be done; tests should use the implicit stdlib that matches the version of `forc` used in CI. Paging @vnepveu and @digorithm to investigate.
Fixed in #280 (normally haha)

docs/src/getting-started/basics.md
```diff
@@ -241,19 +241,19 @@ For a more concrete example, see the `test_contract_calling_contract` function i
 
 ## Connecting to existing contracts
 
-If you already have a deployed contract and want to call its methods using the SDK,  but without deploying it again, all you need is the contract ID of your deployed contract. You can skip the whole deployment setup and call `::new(contract_id, provider, wallet)` directly. For example:
+If you already have a deployed contract and want to call its methods using the SDK,  but without deploying it again, all you need is the contract ID of your deployed contract. You can skip the whole deployment setup and call `::new(contract_id, wallet)` directly. For example:
```
vnepveu, line 244: Just realised that this example isn't accurate. If you have an already deplyed contract, then you already have a node. So you *have* to use a provider that connects to said node to connect the wallet to it. The instruction in this version would confuse users.

packages/fuels-contract/src/contract.rs
```diff
@@ -380,19 +380,18 @@ impl Contract {
     }
 
     /// Deploys a compiled contract to a running node
-    /// To deploy a contract, you need a wallet with enough assets to pay for deployment. This
```
vnepveu, line 383: Re-add docstring

vnepveu (changes requested):
One last detail I hadn't thought of then we merge :tada:

packages/fuels-test-helpers/src/signers.rs
vnepveu: This was not updated actually

docs/src/getting-started/basics.md
MujkicA: This was the case even before my changes. A workflow that involves connecting a provider to a node, instead of starting it via `launch_provider_and_get_wallet` is not introduced. I fear confusion might go the other way as well.

packages/fuels-test-helpers/src/signers.rs
MujkicA: Docstring added

adlerjohn: https://github.com/MujkicA/contract-deployment-qol/commit/55964a5f8eee1b6e7f88c3ac3727d755806a0166#diff-b655abf20ad2575234c19c2517d33a0a4e681de106149f287130a67f1f96155eR8

docs/src/getting-started/basics.md
vnepveu: Can you please open a `docs` issue for this part then. This would be very confusing to a new user I think.

vnepveu (approved):
LGTM!!! :love_you_gesture:

digorithm (approved):
:shipit:

Issue #270, opened 2022-05-09
Add TxParams to transfer


Voxelot:
[`Wallet::transfer`](https://github.com/FuelLabs/fuels-rs/blob/v0.11.0/packages/fuels-signers/src/wallet.rs#L143) doesn't support customization of fee parameters. This is needed to be able to use the SDK against production environments.

QuinnLee:
Created https://github.com/FuelLabs/fuels-ts/issues/255 to track this on the ts side

PR #271, opened 2022-05-09
Add `TxParameters` to transfer functionality

digorithm:
This PR introduces the addition of `TxParameters` to the `wallet.transfer()` API. Enabling the user to configure the `TxParameters` in a transfer call, which was previously hardcoded.

Now, transfers look like this:

```Rust
let receipts = wallet
            .transfer(&target.address(), amount, Default::default(), /* new */ TxParameters::default())
            .await
            .unwrap();
```

Also, I've moved `parameters.rs` from `fuels-contract` to `fuels-core`, to avoid a cyclic dependency problem now that `fuels-signers` need `parameters`, the same for `fuels-contract` needing `parameters`.

Closes https://github.com/FuelLabs/fuels-rs/issues/270.

### DIFF
--- a/packages/fuels-contract/src/contract.rs
+++ b/packages/fuels-contract/src/contract.rs
@@ -1,6 +1,5 @@
 use crate::abi_decoder::ABIDecoder;
 use crate::abi_encoder::ABIEncoder;
-use crate::parameters::{CallParameters, TxParameters};
 use crate::script::Script;
 use anyhow::Result;
 use fuel_asm::Opcode;
@@ -15,7 +14,10 @@ use fuel_vm::script_with_data_offset;
 use fuels_core::errors::Error;
 use fuels_core::ReturnLocation;
 use fuels_core::{
-    constants::DEFAULT_COIN_AMOUNT, constants::WORD_SIZE, Detokenize, Selector, Token,
+    constants::DEFAULT_COIN_AMOUNT,
+    constants::WORD_SIZE,
+    parameters::{CallParameters, TxParameters},
+    Detokenize, Selector, Token,
 };
 use fuels_core::{constants::NATIVE_ASSET_ID, ParamType};
 use fuels_signers::provider::Provider;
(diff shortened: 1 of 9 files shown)


Voxelot (commented):
Should we have any tests that verify the tx params are set correctly on the tx?

digorithm:
> Should we have any tests that verify the tx params are set correctly on the tx?

I think only if we tested `build_transfer_tx` https://github.com/FuelLabs/fuels-rs/blob/31b3111f8ff2e6cfe9094080a23854c1a183a40c/packages/fuels-signers/src/provider.rs#L109-L131

Because we can't see the tx inside transfer https://github.com/FuelLabs/fuels-rs/blob/31b3111f8ff2e6cfe9094080a23854c1a183a40c/packages/fuels-signers/src/wallet.rs#L144

And the call doesn't return the tx info.

But `build_transfer_tx` is so shallow that I don't see it being meaningful enough -- it only creates a `Transaction::Script`, it's super shallow. 

But I'm open to different testing strategies here though!

Voxelot:
We could fetch the transaction back from the fuel-core API, essentially as a round-trip test to ensure fields set by the SDK are received and returned by the fuel-core service.

digorithm:
I like the idea of fetching the tx from fuel-core API! Lemme see if I can get this done easily -- if it gets too tricky I'll leave it for a follow-up PR.

digorithm:
@Voxelot Changed `transfer` a bit to return the tx id. Then I'm using that with the `fuel-core` API to get the tx and then asserting the params. Seems like a decent strategy! Changes here: https://github.com/FuelLabs/fuels-rs/pull/271/commits/33498c1503544ebe41af1dbbe28e5e760e7c3bcb.

digorithm:
Yup, makes sense to me! Here it is: https://github.com/FuelLabs/fuels-rs/pull/271/commits/f264ddffe00411e1b2d263ac74f1f1b07f3860ea.

Voxelot (approved):

PR #273, opened 2022-05-10
Update getting started/setup doc

JoshuaBatty:
Update setup docs to also include `fuel-tx = 0.9` and bump `fuels-abigen-macro` and `fuels` versions from 10 to 12


digorithm (approved):
Thank you!

PR #275, opened 2022-05-12
feat: support `Enum` type as function argument

### LINKED ISSUE #260: feat: support `Enum` as input type
- Currently, trying to use an `Enum` as input type to a contract instance with the SDK fails with the `Revert(0)` error. It is likely that this is due to insufficient support in the SDK encoding rather than Sway.

vnepveu:
This PR closes #260 by implementing support for the `Enum` type in smart contract functions.
The support was lacking because the encoding used was the same for `Struct` and `Enum`. 
The encoding has been specified in https://github.com/FuelLabs/fuel-specs/pull/315.


packages/fuels-abigen-macro/tests/test_projects/use_enum_input/Forc.lock
```diff
@@ -0,0 +1,12 @@
+[[package]]
```
digorithm, line 1: `Forc.lock` files shouldn't be committed! Add `Forc.lock` to this new project's `.gitignore`

vnepveu: A shoot I wasn't sure, almost asked the question haha

digorithm (approved):
LGTM -- great tests!

PR #276, opened 2022-05-12
test: test support for nested custom types in function arguments

### LINKED ISSUE #53: Support `enums` inside `structs` and `structs` inside `enums` in the ABIgen

vnepveu:
This PR definitely closes #53 by testing support for `Struct` inside `Enum` and `Enum`
inside `Struct` types as function arguments in smart contracts. 
- It was only supported as function outputs since #213
- In #275 support was added for the `Enum` type as function argument, and it turns out that nested custom types as function argument worked right-away with this. This PR adds tests to show this.


digorithm:
I'm a bit confused -- what is this PR introducing? The title says it's adding support to nested custom types as inputs, but it's only adding tests (that are passing) to show that it actually works. Which is good and all, I'm just asking because the title itself and the content of the PR doesn't seem to match 😅.

vnepveu:
Hahs you're absolutely right, it turns out that supporting `Enum` made the nested types work immediately. So it *is* just a test PR

digorithm (approved):
Nice!

PR #283, opened 2022-05-16
feat: support unit type variants in `Enum`

### LINKED ISSUE #231: feat: support "empty" enums
Currently, "empty" enums such as 
```
enum SmallEnum {
    v1: (),
    v2: (),
}
```
are not correctly parsed by the SDK, because the `()` type is parsed as a custom type but has no component: 
```
 help: message: called `Result::unwrap()` on an `Err` value: MissingData("cannot parse custom type with no components")
```

vnepveu:
This PR closes #231 by implementing support for unit type variants in `Enums`.

### DIFF
--- a/packages/fuels-core/src/lib.rs
+++ b/packages/fuels-core/src/lib.rs
@@ -29,6 +29,7 @@ pub enum ParamType {
     Bool,
     Byte,
     B256,
+    Unit,
     Array(Box<ParamType>, usize),
     #[strum(serialize = "str")]
     String(usize),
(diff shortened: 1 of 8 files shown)


packages/fuels-core/src/lib.rs
```diff
@@ -107,6 +111,7 @@ impl fmt::Display for ParamType {
 #[derive(Debug, Clone, PartialEq, EnumString)]
 #[strum(ascii_case_insensitive)]
 pub enum Token {
+    Unit,
```
digorithm, line 114: How about adding a comment here explaining `Unit`? If you have no context, `Unit`, both the name and the purpose, can be very confusing. E.g., We use `Unit` when the type being handled is an "empty" `Enum`, so why `Unit` and not `Enum` with an empty `Box<>`? Also maybe hint that the string type coming from the compiler is `"()"`, etc.

digorithm (changes requested):
Short and neat! Just left a small request.

packages/fuels-core/src/lib.rs
vnepveu: Yes you're right!

digorithm (approved):
:shipit:

adlerjohn (approved):
utACK

PR #284, opened 2022-05-16
Fix fuels core dependency

digorithm:
This fixes the issue:

```
Error[E0433]: failed to resolve: use of undeclared crate or module `fuels_core`
 --> tests/harness.rs:7:1
  |
7 | abigen!(MyContract, "out/debug/bimbam-abi.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `fuels_core
```

This was happening because the trait `Detokenize` and one of the errors were being pulled from `fuels_core` without actually pulling `fuels_core` as a dependency. This PR adds both `Detokenize` and the mentioned error in the prelude and solves the issue above.


vnepveu (approved):
Nice catch, gg

PR #290, opened 2022-05-18
Re-export fuel_tx in fuels

### LINKED ISSUE #278: Re-export `fuel-tx` in prelude
The SDK only works with specific versions of `fuel-tx`, so it should be re-exported in the prelude.

Br1ght0ne:
Close #278. Might use some items in `prelude`.

### DIFF
--- a/packages/fuels/src/lib.rs
+++ b/packages/fuels/src/lib.rs
@@ -33,7 +33,8 @@ pub mod test_helpers {
 }
 
 pub mod tx {
-    pub use fuels_tx::*;
+    #[doc(no_inline)]
+    pub use fuel_tx::*;
 }
 
 /// Easy imports of frequently used
--- a/packages/fuels/src/lib.rs
+++ b/packages/fuels/src/lib.rs
@@ -32,6 +32,10 @@ pub mod test_helpers {
     pub use fuels_test_helpers::*;
 }
 
+pub mod tx {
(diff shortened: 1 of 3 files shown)


packages/fuels/src/lib.rs
```diff
@@ -32,6 +32,10 @@ pub mod test_helpers {
     pub use fuels_test_helpers::*;
 }
 
+pub mod tx {
+    pub use fuels_tx::*;
```
Voxelot, line 36:
```suggestion
    #[doc(no_inline)]
    pub use fuel_tx::*;
```

Voxelot:
Maybe we could re-export fuel-tx from fuels-core into the prelude instead of creating a new dep here? Otherwise we'll have to bump the version in several places every upgrade.

Br1ght0ne:
> Maybe we could re-export fuel-tx from fuels-core into the prelude instead of adding a new dep in fuels? Otherwise we'll have to bump the version in several places every upgrade.

Good call! I'm less oriented in the dep tree :)

vnepveu:
I agree with Brandon here

adlerjohn:
Poking @digorithm and @vnepveu for review

Voxelot (approved):

adlerjohn (approved):

digorithm (approved):

Issue #244, opened 2022-04-29
Add logging of all receipts as an option


nfurfaro:
Often, one of the only tools at our disposal for trying to debug is to add println!s in [contract.rs](http://contract.rs/) and use a local version of the SDK to try to log out all receipts before a panic occurs. This is painful and time consuming.

Could we maybe  have this logging always available, but hidden behind a --verbose flag if that seems reasonable?
Open to other  approaches too.

nfurfaro:
>Maybe the sdk could use debug level logging 

Would this work now, or is it another feature we would have to wait for?

PR #292, opened 2022-05-18
Add logging of all receipts as an option

Salka1988:


Br1ght0ne (approved):

packages/fuels-abigen-macro/Cargo.toml
```diff
@@ -18,6 +18,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 rand = "0.8"
 syn = "1.0.12"
+log = "0.4.17"
```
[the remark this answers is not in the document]
leviathanbeak: plus removing the patch number which is unnecessary

packages/fuels-abigen-macro/Cargo.toml
```diff
@@ -14,12 +14,14 @@ proc-macro = true
 [dependencies]
 fuel-tx = "0.9"
 fuels-core = { version = "0.13.0", path = "../fuels-core" }
+log = "0.4"
```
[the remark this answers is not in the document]
leviathanbeak: `fuel-core` is using `tracing` lib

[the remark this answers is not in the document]
Br1ght0ne: `tracing` is not much harder to setup: `tracing-subscriber` is `env-logger`-compatible too. Maybe we should consolidate?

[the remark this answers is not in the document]
vnepveu: agree with using `tracing`

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1221,6 +1229,8 @@ async fn test_multiple_args() {
 
 #[tokio::test]
 async fn test_tuples() {
+    setup_logger();
```
vnepveu, line 1232: why is it used only here?

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -8,12 +8,20 @@ use fuels_abigen_macro::abigen;
...
 use fuels_core::Token;
 use sha2::{Digest, Sha256};
+use std::env;
 
 /// Note: all the tests and examples below require pre-compiled Sway projects.
 /// To compile these projects, run `cargo run --bin build-test-projects`.
 /// It will build all test projects, creating their respective binaries,
 /// ABI files, and lock files. These are not to be committed to the repository.
 
+fn setup_logger() {
```
vnepveu, line 18: I feel like the `setup_logger` should be more general to the crate and accessible, so you can use it in the `contract.rs` for instance, instead of the `dbg!` statement.

packages/fuels-contract/src/contract.rs
```diff
@@ -306,7 +307,9 @@ impl Contract {
         if simulate {
             return script.simulate(fuel_client).await;
         }
-        script.call(fuel_client).await
+        let receipts = script.call(fuel_client).await;
+        debug!("RECEIPTS : {:?}", receipts);
```
vnepveu, line 311: see my comment about the logger

vnepveu (changes requested):
I feel like the logger should maybe be more general and you should be able to enable it outside of the harness

packages/fuels-abigen-macro/tests/harness.rs
Salka1988: https://github.com/FuelLabs/fuels-rs/issues/244 
This issue requires that the results of the receipts be returned. I have used this here just to show how it is used and what results it gives. When I'm done I'll document it in the FuelRs book.

packages/fuels-abigen-macro/tests/harness.rs
Salka1988: I think the same! I could make us a debugger if there is a need with our macro_rules! and types etc.

packages/fuels-abigen-macro/tests/harness.rs
leviathanbeak: if this setup is needed throughout the test suite, we should initialize it once for all the tests, maybe [ctor ](https://docs.rs/ctor/latest/ctor/)crate could help with it.

packages/fuels-abigen-macro/Cargo.toml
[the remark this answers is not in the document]
digorithm: Folks, although `tracing` has simple logging constructions, its main purpose is to do the tracing of spread events using `spans`. This is ideal for highly-concurrent systems, distributed systems, etc. because one can keep track of context and pass around spans representing the same event so that tools like grafana can plot metrics pertaining to each whole execution trace. I recommend reading it more [here](https://docs.rs/tracing/latest/tracing/).

This is very unnecessary for the SDK; it will add unnecessary complexity (the SDK already has enough complexities of its own 😂 ) for something that ought to be simple.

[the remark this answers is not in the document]
Salka1988: I implemented it with both! There is currently “tracing” on the branch but we can revert it to "log" && "env_log".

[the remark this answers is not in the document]
digorithm: Nah, since you already did let's keep it. If it, somehow, becomes a pain later, we can yank it out 😄

packages/fuels-abigen-macro/tests/harness.rs
Salka1988: fn setup_logger() was only to init debugger int test. Now its initialized for every Test.

```rust
#[cfg(test)]
#[ctor::ctor]
fn init() {
    let _ = tracing_subscriber::fmt::try_init();
}
```
You can use for example "event!();" and catch the value that interests you" (event!(Level::DEBUG, "{:?}", receipts);)

Since its TDD, run this on your test like this: "RUST_LOG=fuels_contract::contract=debug cargo test --package fuels-abigen-macro --test harness test_tuples -- --exact --show-output" ----> this is for "test_tuples and my "event" is in "contract" and the flag is "debug"

vnepveu (changes requested):
It would still be nice to have some documentation explaining how to enable the logging. Otherwise it seems good.

packages/fuels-abigen-macro/Cargo.toml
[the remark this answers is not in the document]
vnepveu: I just thought that using the same deps accross repos was nice, didn't think it was such a power tool haha

packages/fuels-contract/src/contract.rs
```diff
@@ -306,7 +308,9 @@ impl Contract {
         if simulate {
             return script.simulate(fuel_client).await;
```
vnepveu, line 309: you should also log in the case of `simulate`

vnepveu (changes requested):
Log receipts in the case of simulating a tx

packages/fuels-contract/src/contract.rs
```diff
@@ -306,7 +308,9 @@ impl Contract {
...
         }
-        script.call(fuel_client).await
+        let receipts = script.call(fuel_client).await;
+        event!(Level::DEBUG, "{:?}", receipts);
```
Voxelot, line 312: isn't this the same thing as `tracing::debug!`?

Salka1988: No it's not the same, it's better than what I used. I didn't know about this approach so I will replace event () with debug ! :)

docs/src/getting-started/basics.md
```diff
@@ -171,6 +171,19 @@ Where `value` will hold the value returned by its respective contract method, re
 
 And lastly, `logs` will hold all logs that happened within that specific contract call.
 
+In order to log out `receipts` values during testing we need to do the following:
+
+1. You have to initialize `tracing_subscriber` in `harness.rs`
```
vnepveu, line 176: This instruction is redundant because you already wrote the code inside `harness.rs`. The formulation might confuse a user, you should probably just keep `2.`. If you want to provide general instructions on how to set up the tests, you can say something along the lines of `
see init_tracing function inside packages/fuels-abigen-macro/harness.rs`



This will help avoid duplication and updates overhead.

docs/src/getting-started/basics.md
```diff
@@ -171,6 +171,19 @@ Where `value` will hold the value returned by its respective contract method, re
...
+```Rust
+#[cfg(test)]
+#[ctor::ctor]
+fn init_tracing() {
+let _ = tracing_subscriber::fmt::try_init();
+}
+```
+2. You have to run `test` as follows: RUST_LOG=receipts cargo test --test harness `name of the test`
```
vnepveu, line 184: Good!

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -14,6 +14,12 @@ use sha2::{Digest, Sha256};
 /// It will build all test projects, creating their respective binaries,
 /// ABI files, and lock files. These are not to be committed to the repository.
 
+#[cfg(test)]
+#[ctor::ctor]
```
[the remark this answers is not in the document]
vnepveu: Please explain what the `ctor` line does

packages/fuels-contract/src/contract.rs
```diff
@@ -306,7 +309,9 @@ impl Contract {
         if simulate {
             return script.simulate(fuel_client).await;
```
vnepveu, line 310:
```rust
if simulate {
        let receipts = script.simulate(fuel_client).await;
        debug!(target: "receipts", "{:?}", receipts);
        return receipts
}
```

packages/fuels-contract/src/script.rs
```diff
@@ -35,6 +36,7 @@ impl Script {
     // modified, it is only simulated using a "dry-run".
     pub async fn simulate(self, fuel_client: &FuelClient) -> Result<Vec<Receipt>, Error> {
         let receipts = fuel_client.dry_run(&self.tx).await?;
+        debug!(target: "receipts", "{:?}", receipts);
```
vnepveu, line 39: The debug statements of receipts should be in the same place, see my suggestion above

vnepveu (changes requested):
Some cleaning left otherwise let's go :rocket:

packages/fuels-contract/src/script.rs
vnepveu: remove this line to avoid logging the receipts twice

vnepveu (changes requested):
Remove the line in `script.rs`

vnepveu (approved):
LGTM! :shipit:

Issue #265, opened 2022-05-07
Return 10 wallets when initializing test provider and wallet


adlerjohn:
Change `setup_test_provider_and_wallet` to instead return a provider and 10 wallets, similar to Ethereum testing infra. Or better yet, a configurable number of wallets.

https://github.com/FuelLabs/fuels-rs/blob/7ff5415357968b83c1c9a94f12bf6a3114c8c62e/packages/fuels-test-helpers/src/signers.rs#L9

PR #293, opened 2022-05-18
Setup multiple test wallets and hide SecretKey

### LINKED ISSUE #264: Return wallet when setting up
The public function `setup_address_and_coins` returns a `SecretKey`, which is useless to an end-user. This should only be done internally. The public function should instead allow initializing a new wallet.

https://github.com/FuelLabs/fuels-rs/blob/7ff5415357968b83c1c9a94f12bf6a3114c8c62e/packages/fuels-test-helpers/src/lib.rs#L23-L26

MujkicA:
Closes #264 #265 

Previously, tests would setup objects in the following order `SecretKey` -> `Coins` -> `Provider` -> `Wallet`, where each object
depends on the previous one for initialization.

To abstract away `SecretKey` from the user, `Wallet` can now be created by supplying `Option<Provider>`
This enables us to create wallets, set addresses on genesis coins and launch a provider in that order.

```Rust
let mut wallet = LocalWallet::new(None);
let coins = setup_coins(wallet.address, 1, 1000000);

let (client, _) = setup_test_client(cons).await;
let provider = Provider::new(client);

wallet.set_provider(provider)
```
When `None` is supplied, calls to `wallet.get_provider()` (e.g. during transactions) will result in an error that suggests using `set_provider()`

`launch_provider_and_get_wallets()` accepts `WalletConfig` to configure the number of wallets returned, the number of coins per wallet, and amount per coin

```Rust
let wallets = 5;
let coins = 3;
let amount = 1_000_000;

let config = WalletsConfig::new(
    Some(wallets),
    Some(coins),
    Some(amount)
);
let wallets: Vec<LocalWallet> = launch_provider_and_get_wallets(config).await;
```
The above example will return a vector of 5 wallets having 5 coins with an amount of 1_000_000 each. 
Passing `WalletConfig::default()` leads to 10 wallets.


docs/src/getting-started/basics.md
```diff
@@ -43,8 +43,8 @@ abigen!(
     "your_project/out/debug/contract_test-abi.json",
 );
 
-// This helper will will launch a local node and provide a test wallet linked to it
-let wallet = launch_provider_and_get_wallet().await;
+// This helper will will launch a local node and provide 10 test wallets linked to it
+let wallet = launch_provider_and_get_wallets(WalletConfig::default()).await;
```
vnepveu, line 47: should be called `wallets` if there are 10

vnepveu, line 47: `wallet` is a `Vec<LocalWallet>` so it's not going to work with the `deploy` call below.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -19,6 +19,13 @@ fn null_contract_id() -> String {
     String::from("0000000000000000000000000000000000000000000000000000000000000000")
 }
 
+async fn get_single_wallet() -> LocalWallet {
```
vnepveu, line 22: I feel that this `single_wallet` should be a method for the `WalletConfig`

docs/src/getting-started/wallets.md
```diff
@@ -9,24 +9,22 @@ Wallets are used for many important things, for instance:
...
-// Generate a random private/secret key.
-let mut rng = rand::thread_rng();
-let secret = SecretKey::random(&mut rng);
-
 // Use the test helper to setup a test provider.
 let (provider, _address) = setup_test_provider(vec![]).await;
 
 // Create the wallet.
-let wallet = LocalWallet::new_from_private_key(secret, provider);
+let wallet = LocalWallet::new_from_private_key(Some(provider));
```
vnepveu, line 21: Reading `new_from_private_key` I would expect to provide my private key, not a `Provider` (which is basically a node)

docs/src/getting-started/wallets.md
```diff
@@ -9,24 +9,22 @@ Wallets are used for many important things, for instance:
...
 The SDK gives you many different ways to create wallets. Let's explore these different ways.
 
-## Creating a wallet from a Private Key
+## Creating a wallet
 
-You can use `fuel_crypto` to generate a random `SecretKey` and use it to create a wallet from it:
+A new wallet can be created from `Option<Provider>`.
 
 ```Rust
-use fuel_crypto::{SecretKey};
```
vnepveu, line 17: I am not convinced by this new workflow. Plus, I don't think it was the point of #264.

docs/src/getting-started/wallets.md
```diff
@@ -9,24 +9,22 @@ Wallets are used for many important things, for instance:
...
-
 // Use the test helper to setup a test provider.
 let (provider, _address) = setup_test_provider(vec![]).await;
 
 // Create the wallet.
-let wallet = LocalWallet::new_from_private_key(secret, provider);
+let wallet = LocalWallet::new_from_private_key(Some(provider));
 ```
 
+> Note: if `None` is supplied instead, any transaction related to the wallet will result
```
vnepveu, line 24: This is kind of confusing. The `set_provider` should be a builder method that overwrites a default provider IMO.

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: More precisely, `single_wallet_config` should already exist in the `WalletConfig` (maybe create an `Enum` around it somewhere

vnepveu, line 22: We should not hide the fact that we are lauching a node behind the scenes.

packages/fuels-contract/src/contract.rs
```diff
@@ -30,6 +29,10 @@ pub struct CompiledContract {
     pub salt: Salt,
 }
 
+// This constant is used to determine the amount in the 1 UTXO
+// when initializing wallets for now.
+pub const DEFAULT_COIN_AMOUNT: u64 = 1_000_000;
```
vnepveu, line 34: I liked it better in `constants`, any reason you need it here?

packages/fuels-core/src/constants.rs
```diff
@@ -8,10 +8,6 @@ pub const DEFAULT_MATURITY: u64 = 0;
 
 pub const WORD_SIZE: usize = core::mem::size_of::<Word>();
 
-// This constant is used to determine the amount in the 1 UTXO
-// when initializing wallets for now.
-pub const DEFAULT_COIN_AMOUNT: u64 = 1_000_000;
```
vnepveu, line 13: any reason for moving the constant out?

packages/fuels-test-helpers/src/wallets_config.rs
```diff
@@ -0,0 +1,34 @@
+pub const DEFAULT_COIN_AMOUNT: u64 = 1_000_000_000;
```
vnepveu, line 1: duplication with the one from `constants.rs` that was moved into `contract.rs`

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -24,3 +37,32 @@ pub async fn setup_test_provider(coins: Vec<(UtxoId, Coin)>) -> (Provider, Socke
...
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::WalletsConfig;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn config() {
```
vnepveu, line 48: improve test naming

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -1,20 +1,33 @@
...
 #[cfg(feature = "fuels-signers")]
-pub async fn launch_provider_and_get_wallet() -> LocalWallet {
-    //  We build only 1 coin with amount DEFAULT_INITIAL_BALANCE, empirically determined to be
-    //  sufficient right now
-    let (pk, coins) = setup_address_and_coins(1, DEFAULT_INITIAL_BALANCE);
-    // Setup a provider and node with the given coins
-    let (provider, _) = setup_test_provider(coins).await;
-
-    LocalWallet::new_from_private_key(pk, provider)
+pub async fn launch_provider_and_get_wallets(config: WalletsConfig) -> Vec<LocalWallet> {
```
vnepveu, line 9: good

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -5,36 +5,26 @@ use fuel_core::{
...
 
-// This constant is used to set a initial balance on wallets
-// mainly used on tests
-pub const DEFAULT_INITIAL_BALANCE: u64 = 1_000_000_000;
-
-pub fn setup_address_and_coins(
-    num_of_coins: usize,
-    amount: u64,
-) -> (SecretKey, Vec<(UtxoId, Coin)>) {
+pub fn setup_coins(owner: Address, num_coins: u64, amount: u64) -> Vec<(UtxoId, Coin)> {
```
vnepveu, line 21: IMO #264 was more about changing the output of the initial `setup_address_and_coin` than creating a function that only creates the coins.

packages/fuels-signers/src/wallet.rs
```diff
@@ -109,8 +100,14 @@ impl From<WalletError> for Error {
 }
 
 impl Wallet {
-    /// Create a new wallet from a private key.
-    pub fn new_from_private_key(private_key: SecretKey, provider: Provider) -> Self {
+    pub fn new(provider: Option<Provider>) -> Self {
```
vnepveu, line 103: the naming of the function should specify that a new wallet is created _at random_, otherwise one might think that the wallet is just a Rust instance.

packages/fuels-signers/src/wallet.rs
```diff
@@ -435,7 +452,7 @@ mod tests {
         // Create first account from mnemonic phrase.
         let wallet = Wallet::new_from_mnemonic_phrase_with_path(
             phrase,
-            provider.clone(),
+            Some(provider.clone()),
             "m/44'/60'/0'/0/0",
```
vnepveu, line 456: use the `DEFAULT_DERIVATION_PREFIX` constant from above

packages/fuels-signers/src/wallet.rs
```diff
@@ -467,7 +484,7 @@ mod tests {
...
```
vnepveu, line 488: use the constant here as well

vnepveu (changes requested):
I feel that there was some confusion between the two issues, but maybe I misunderstood.

packages/fuels-test-helpers/src/wallets_config.rs
MujkicA: IMO these two should be kept separated. This one is the default parameter for the wallet setup config while the other one is   for querying UTXOs.

packages/fuels-test-helpers/src/lib.rs
MujkicA: Maybe I got carried away in the process, but still I don't see the benefit of having those two functionalities bundled in a single call, other than saving a line of code. 
This allows for creating UTXOs of varying sizes without the side effect of creating a wallet each time.

packages/fuels-test-helpers/src/signers.rs
MujkicA: suggestion?

packages/fuels-test-helpers/src/wallets_config.rs
adlerjohn: This should be documented.

packages/fuels-signers/src/wallet.rs
adlerjohn: If the user explicitly want to use the private key, then sure that functionality can be kept. 264 is more about removing dealing with private keys then the user doesn't actually need them, e.g. if they're setting up a wallet then they only need to sign things, not the private key itself.

packages/fuels-test-helpers/src/signers.rs
vnepveu: `test_wallet_config`

packages/fuels-signers/src/wallet.rs
vnepveu: AFAIK the `new_from_private_key` function is still here right?

MujkicA: It's there, but private

packages/fuels-contract/src/contract.rs
leviathanbeak: why is it public if you don't want to leak it 🤔

MujkicA: Renamed it and returned it to constans

docs/src/getting-started/wallets.md
MujkicA: I've explained above why `Option<>` is needed. IMO `set_provider` does what it says it does. Open for ideas.

packages/fuels-signers/src/wallet.rs
```diff
@@ -109,8 +100,14 @@ impl From<WalletError> for Error {
...
-    /// Create a new wallet from a private key.
-    pub fn new_from_private_key(private_key: SecretKey, provider: Provider) -> Self {
+    pub fn new_random(provider: Option<Provider>) -> Self {
+        let mut rng = rand::thread_rng();
+        let private_key = SecretKey::random(&mut rng);
+
+        Self::new_from_private_key(private_key, provider)
+    }
+
+    pub fn new_from_private_key(private_key: SecretKey, provider: Option<Provider>) -> Self {
```
digorithm, line 110: I like this; if you want a quick wallet without thinking too much: `new_random(None)`, otherwise, you have many other ways of constructing a wallet: PKs, mnemonic, encrypted JSON wallet. 👍

packages/fuels-signers/src/wallet.rs
```diff
@@ -121,9 +118,16 @@ impl Wallet {
         }
     }
 
+    pub fn get_provider(&self) -> Result<&Provider, WalletError> {
+        self.provider.as_ref().ok_or(WalletError::NoProvider)
+    }
```
digorithm, line 123: I've been needing this a lot; requiring `Wallet`s to have a `Provider` at creation time was actually a major hassle!

digorithm (approved):
Great stuff! Lots of things here that are sorely needed, both by me (optional `Provider`s 🎉 ) and by the users. 

Good work here -- and good work with the reviews @vnepveu!

vnepveu (approved):
:shipit:

PR #297, opened 2022-05-19
Add support for forc 0.13

digorithm:


vnepveu (approved):
:shipit:

packages/fuels-abigen-macro/tests/test_projects/two_structs/src/main.sw
```diff
@@ -9,8 +9,8 @@ pub struct StructTwo {
 }
 
 abi MyTest {
-    fn something(input: StructTwo) -> u64;
-    fn something_else(input: StructOne) -> u64;
+    fn something(input: StructOne) -> u64;
+    fn something_else(input: StructTwo) -> u64;
```
adlerjohn, line 13: Why were these changed?

digorithm: Because before it was wrong but somehow the compiler never complained 😂 They have the same inner structure (just a `u64`) so I think the compiler was okay using them interchangeably? This was still weird but we never caught it. But now the compiler actually complains, hence why I changed it.

MujkicA (dismissed):

adlerjohn (commented):
CI failing

vnepveu (approved):
I was kind of sure the error message was the one I had changed haha

PR #302, opened 2022-05-22
add customisable node config as parameter

Salka1988:


packages/fuels-test-helpers/src/signers.rs
```diff
@@ -1,19 +1,24 @@
...
+use fuel_core::service::Config;
 use fuel_tx::UtxoId;
 use fuels_signers::provider::Provider;
 use fuels_signers::{LocalWallet, Signer};
 use std::net::SocketAddr;
 
 #[cfg(feature = "fuels-signers")]
-pub async fn launch_provider_and_get_single_wallet() -> LocalWallet {
-    let mut wallets = launch_provider_and_get_wallets(WalletsConfig::new_single(None, None)).await;
+pub async fn launch_provider_and_get_single_wallet(node_config: Config) -> LocalWallet {
```
leviathanbeak, line 10: since `Config::local_node()` is used throughout / most of the tests, why not make `node_config` optional and provide a default value instead of passing `Config::local_node()` throughout the codebase?

Or (maybe) even better why not make a separate function that launches the provider with configurable node?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -69,7 +73,7 @@ pub async fn setup_test_client(coins: Vec<(UtxoId, Coin)>) -> (FuelClient, Socke
         },
         database_type: DbType::InMemory,
         utxo_validation: true,
-        ..Config::local_node()
+        ..node_config // ..Config::local_node()
```
leviathanbeak, line 76: nit: remove the commented line

packages/fuels-test-helpers/src/signers.rs
Salka1988: In new commit I added Launcher in which we can configure everything we need and in this way I tried to avoid modifying existing functions.

MujkicA: Agree with the above. This covers a special case but we get more boilerplate for the common case.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1345,7 +1381,8 @@ async fn workflow_enum_inside_struct() {
         /enum_inside_struct-abi.json"
     );
 
-    let wallet = launch_provider_and_get_single_wallet().await;
+    let launcher = Launcher::default_config();
```
MujkicA, line 1384: Update the docs if we decide to keep `Launcher`

packages/fuels-test-helpers/src/signers.rs
digorithm: Also agree with @leviathanbeak. 

From the first principles, what we want the most is the easiest/simplest dev experience we can come up with. 

The majority of use cases will want to launch a node/provider with default configs. Some users will want to configure that themselves.

So, we're forcing the majority to go from:

```Rust
let wallet = launch_provider_and_get_single_wallet().await;
```

To: 

```Rust
let launcher = Launcher::default_config();
let wallet = launcher.launch_provider_and_get_single_wallet().await;
```

That's not very friendly (compared to the previous one). 

In addition to that, `Launcher` adds one more concept that the user has to hold in their head when working with the SDK, which is not necessarily bad, but it's better if it could be avoided and the user kept needing _fewer_ things to get the work done.

So I think the most user-friendly solution would actually be the simplest solution here: keep `launch_provider_and_get_single_wallet()` as-is and add a new one (e.g. `launch_custom_provider_and_get_single_wallet()`) that takes the config.

MujkicA (dismissed):

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -12,6 +13,24 @@ pub async fn launch_provider_and_get_single_wallet() -> LocalWallet {
...
 
+#[cfg(feature = "fuels-signers")]
+pub async fn launch_custom_provider_and_get_single_wallet(node_config: Config) -> LocalWallet {
+    let mut wallet = LocalWallet::new_random(None);
+
+    let wallet_config = WalletsConfig::new_single(None, None);
+
+    let coins: Vec<(UtxoId, Coin)> = setup_coins(
+        wallet.address(),
+        wallet_config.coins_per_wallet,
```
MujkicA, line 24: There is no need to init `WalletsConfig` here, just pass the constants directly

MujkicA (changes requested):

vnepveu (approved):
Approving but opening issue #312  to discuss if there aren't too many different launching functions (that might be confusing)

MujkicA (approved):

digorithm (approved):

PR #305, opened 2022-05-22
Added `deploy_with_salt` function

### LINKED ISSUE #304: The package `fuels-contract` should add a `deploy_with_salt` function
There should be an easier way to deploy using a salt than to string together multiple methods.
Similar to the [deploy](https://github.com/FuelLabs/fuels-rs/blob/v0.13.0/packages/fuels-contract/src/contract.rs#L385) function there should be a `deploy_with_salt`.

This new function should use this `salt` parameter in conjunction to `load_sway_contract_with_salt` to achieve the same effect.

Braqzen:
Closes #304

### DIFF
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -29,7 +29,11 @@ Alternatively, if you have a Fuel node running separately, you can pass in the `
 
 ## Deploying a Sway contract
 
-Once you have compiled your contract, here's how to deploy it:
+There are two intended ways to deploy a contract
+- `deploy`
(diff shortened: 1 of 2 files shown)


digorithm (changes requested):
Please add the new `deploy_with_salt` to the docs under `Deploying a Sway contract`. Currently, it only mentions the default one (without salt). It would be nice just to hint that if the user wants to define a salt, he can do so with `deploy_with_salt`.

digorithm (approved):
Nice -- thanks for the doc updates!

PR #308, opened 2022-05-23
add get_transactions to Wallet and Provider

Salka1988:
I implemented 2 methods:
- get_transactions() in a Wallet.rs that returns wallet-related transactions 
- get_transactions() in a Provider.rs that returns all transactions

### DIFF
--- a/packages/fuels-signers/src/provider.rs
+++ b/packages/fuels-signers/src/provider.rs
@@ -1,18 +1,19 @@
+use std::io;
+use std::net::SocketAddr;
+
 #[cfg(feature = "fuel-core")]
 use fuel_core::service::{Config, FuelService};
 use fuel_gql_client::client::schema::coin::Coin;
 use fuel_gql_client::client::types::TransactionResponse;
-use fuel_gql_client::client::{FuelClient, PageDirection, PaginationRequest};
+use fuel_gql_client::client::{FuelClient, PageDirection, PaginatedResult, PaginationRequest};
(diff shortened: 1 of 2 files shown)


packages/fuels-signers/src/provider.rs
```diff
@@ -136,7 +137,27 @@ impl Provider {
         Ok(self.client.transaction(tx_id).await.unwrap().unwrap())
     }
 
-    // @todo
     // - Get transaction(s)
+    pub async fn get_transactions(
+        &self,
+        request: PaginationRequest<String>,
+    ) -> io::Result<PaginatedResult<TransactionResponse, String>> {
+        Ok(self.client.transactions(request).await.unwrap())
```
leviathanbeak, line 145:
```suggestion
        self.client.transactions(request).await
```

packages/fuels-signers/src/provider.rs
```diff
@@ -136,7 +137,27 @@ impl Provider {
...
+    }
+
+    // - Get transaction(s) by owner
+    pub async fn get_transactions_by_owner(
+        &self,
+        owner: &str,
+        request: PaginationRequest<String>,
+    ) -> io::Result<PaginatedResult<TransactionResponse, String>> {
+        Ok(self
```
leviathanbeak, line 154: same as above

leviathanbeak:
`unwrap()`s shouldn't be used - the Option / Result is there for a reason

PR #309, opened 2022-05-23
Add deploy for loaded contract

### LINKED ISSUE #282: Confirm
https://github.com/FuelLabs/fuels-rs/blob/3d7623127edf1838068daa1154fe31bb9e6544c9/packages/fuels-contract/src/contract.rs#L402-L412

Now that `deploy` directly accepts the path to a contract binary, there may be no need to expose `load_sway_contract`.
We need to confirm that there is no scenario where the user wants to interact with the raw binary data.

MujkicA:
Closes #282

Since we concluded that `load_sway_contract()` and `load_sway_contract_with_salt()` should still remain exposed to the user, I've added `deploy_loaded()` so that the output of those functions can be deployed. 

I thought about removing `deploy_with_salt()` since its functionality can be replicated with `load_sway_contract_with_salt()` + `deploy_loaded()` but it seems more user friendly to have a similar workflow for deploying with and without salt.

I've also added a small test `same_contract_different_ids` to cover the case when the exact same contract is being deployed twice.

### DIFF
--- a/packages/fuels-abigen-macro/tests/harness.rs
+++ b/packages/fuels-abigen-macro/tests/harness.rs
@@ -1,4 +1,4 @@
-use fuel_tx::{AssetId, ContractId, Receipt};
+use fuel_tx::{AssetId, ContractId, Receipt, Salt};
 use fuels::prelude::{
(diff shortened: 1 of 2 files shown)


vnepveu (approved):
Liking this new workflow, very clean! :shipit:

PR #313, opened 2022-05-24
feat: add a balance API

### LINKED ISSUE #266: Introduce a balance API
`fuel-core` has now introduced a better way to query for balances. The TS SDK already implemented supported for that: https://github.com/FuelLabs/fuels-ts/pull/251. 

We should do the same here and improve balance querying -- currently we do have some mechanism to query for balances, but it should be improved and changed to use the `fuel-core` balance API.

vnepveu:
Closes #266 by implementing a simplified balance API in both the `Provider` and the `Wallet`. 

Question: should this get documented in `The Book`? 
Note: opened an issue for improving tests for multiple assets and put them in there right place (not in the `harness.rs`)


packages/fuels-signers/src/provider.rs
```diff
@@ -132,6 +132,45 @@ impl Provider {
...
+        let pagination = PaginationRequest {
+            cursor: None,
+            results: 9999,
+            direction: PageDirection::Forward,
+        };
+        let balances_vec = self
+            .client
+            .balances(&*address.to_string(), pagination)
+            .await
+            .unwrap()
```
leviathanbeak, line 163: Why not handle the error case or propagate it further? `unwrap()` should be used rarely, if ever - in production code.

packages/fuels-signers/src/provider.rs
```diff
@@ -132,6 +132,45 @@ impl Provider {
...
+    ) -> std::io::Result<u64> {
+        self.client
+            .balance(&*address.to_string(), Some(&*asset_id.to_string()))
+            .await
+    }
+
+    /// Get all the spendable balances of all assets for address `address`. This is different from
+    /// getting
+    /// the coins because we are only returning the sum of UTXOs and not the UTXOs themselves
+    pub async fn get_balances(&self, address: &Address) -> HashMap<String, u64> {
```
leviathanbeak, line 151: `get_asset_balance()` returns `Result<T, Error>` yet `get_balances()` does not, shouldn't there be some standard for client facing API

packages/fuels-signers/src/provider.rs
```diff
@@ -132,6 +141,48 @@ impl Provider {
...
+    /// UTXOs.
+    pub async fn get_asset_balance(
+        &self,
+        address: &Address,
+        asset_id: AssetId,
+    ) -> Result<u64, ProviderError> {
+        Ok(self
+            .client
+            .balance(&*address.to_string(), Some(&*asset_id.to_string()))
+            .await?)
```
leviathanbeak, line 155:
```suggestion
       self.client
            .balance(&*address.to_string(), Some(&*asset_id.to_string()))
            .await
            .map_err(Into::into)
```

packages/fuels-signers/src/provider.rs
```diff
@@ -132,6 +141,48 @@ impl Provider {
...
+            .client
+            .balances(&*address.to_string(), pagination)
+            .await?
+            .results;
+        let balances = balances_vec
+            .iter()
+            .map(|b| (b.asset_id.to_string(), b.amount.clone().try_into().unwrap()))
+            .collect::<Vec<(String, u64)>>()
+            .into_iter()
+            .collect();
```
leviathanbeak, line 182:
```suggestion
       let balances = balances_vec
            .iter()
            .map(|b| (b.asset_id.to_string(), b.amount.clone().try_into().unwrap()))
            .collect();
```

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1545,3 +1545,31 @@ async fn unit_type_enums() {
...
+        .get_provider()
+        .unwrap()
+        .get_balances(&wallet.address())
+        .await
+        .unwrap();
+    let expected_key = "0x".to_owned() + NATIVE_ASSET_ID.to_string().as_str();
+    let expected_key = expected_key.as_str();
+    assert_eq!(balances.len(), 1);
+    assert!(balances.contains_key(expected_key));
+    assert_eq!(*balances.get(expected_key).unwrap(), 231)
```
leviathanbeak, line 1574:
```suggestion
    let expected_key = "0x".to_owned() + NATIVE_ASSET_ID.to_string().as_str();
    assert_eq!(balances.len(), 1);
    assert!(balances.contains_key(&expected_key));
    assert_eq!(*balances.get(&expected_key).unwrap(), 231)
```

packages/fuels-signers/src/provider.rs
```diff
@@ -132,6 +141,48 @@ impl Provider {
...
+    /// Get all the spendable balances of all assets for address `address`. This is different from
+    /// getting
+    /// the coins because we are only returning the sum of UTXOs and not the UTXOs themselves
+    pub async fn get_balances(
+        &self,
+        address: &Address,
+    ) -> Result<HashMap<String, u64>, ProviderError> {
+        // We don't paginate results because there are likely at most ~100 different assets in one
+        // wallet
+        let pagination = PaginationRequest {
```
leviathanbeak, line 167: This is more of a question for my own sanity:

is this a temporary solution? or has this been discussed with @adlerjohn and/or @digorithm and we decided that this should be fine **forever** ?

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1545,3 +1545,31 @@ async fn unit_type_enums() {
         .unwrap();
     assert_eq!(result.value, BimBamBoum::Boum());
 }
+
+#[tokio::test]
+// TODO(vnepveu): test for multiple assets
```
leviathanbeak, line 1550: nit: I'd prefer creating an issue for it, and linking the TODO comment to it

(TODOs tend to get ignored :) )

packages/fuels-signers/src/provider.rs
vnepveu: This has been discussed with @digorithm. We decided that this was fine until someone complains (and we don't think someone will, the pagination can alwas be done later).

vnepveu:
Opened issue #321 to track the testing quality related to this PR.

packages/fuels-signers/src/wallet.rs
```diff
@@ -352,6 +353,17 @@ impl Wallet {
...
             .await
     }
+
+    /// Shallow wrapper around the provider method
+    pub async fn get_asset_balance(&self, asset_id: &AssetId) -> Result<u64, ProviderError> {
+        self.get_provider()?
+            .get_asset_balance(&self.address, *asset_id)
+            .await
+    }
+    /// Shallow wrapper around the provider method
```
MujkicA, line 363: Same as above

packages/fuels-signers/src/provider.rs
```diff
@@ -132,6 +141,46 @@ impl Provider {
...
+        asset_id: AssetId,
+    ) -> Result<u64, ProviderError> {
+        self.client
+            .balance(&*address.to_string(), Some(&*asset_id.to_string()))
+            .await
+            .map_err(Into::into)
+    }
+
+    /// Get all the spendable balances of all assets for address `address`. This is different from
+    /// getting
```
MujkicA, line 159: nit: formatting

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1545,3 +1545,30 @@ async fn unit_type_enums() {
...
+async fn test_provider_balance_api() {
+    let mut wallet = LocalWallet::new_random(None);
+    let coins = setup_coins(wallet.address(), 21, 11);
+    let (provider, _) = setup_test_provider(coins.clone(), Config::local_node()).await;
+    wallet.set_provider(provider);
+    for (_utxo_id, coin) in coins {
+        let balance = wallet
+            .get_provider()
+            .unwrap()
+            .get_asset_balance(&wallet.address(), coin.asset_id)
```
digorithm, line 1560: Very unergonomic to do `wallet.<...>.get_asset_balance(&wallet.address, asset_id)`. Why do I have to pass a reference to my own wallet? I'm using my own `wallet` instance and having to reference itself on its own method. Weird.

Make it `wallet.<...>.get_asset_balance(asset_id)`. If you want to query other wallets' balances, use a provider directly (and make sure `Provider` exposes a `get_asset_balance(address, asset_id)` as well).

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1545,3 +1545,30 @@ async fn unit_type_enums() {
...
+            .await;
```
digorithm, line 1561: We can improve the ergonomics here. 

`wallet.get_provider().unwrap().get_asset_balance(&wallet.address(), asset_id)` is _way_ too much just to fetch the balance.

I see that `get_asset_balance()` is bound to `Provider`, which is good. But add a thin layer bound to `Wallet` so that (1) we can remove the need for `&wallet.address()` when using your own wallet and (2) we can remove the `get_provider()`. 

It would be just a `get_asset_balance(&self, asset_id: AssetId)` where self is `Wallet`. Then inside, it would call the provider's `get_asset_balance()` as-is. Calling `wallet.get_asset_balance(asset_id)` without having a provider set would throw an error.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1545,3 +1545,30 @@ async fn unit_type_enums() {
...
     assert_eq!(result.value, BimBamBoum::Boum());
 }
+
+#[tokio::test]
+// TODO(vnepveu): test for multiple assets
+async fn test_provider_balance_api() {
+    let mut wallet = LocalWallet::new_random(None);
+    let coins = setup_coins(wallet.address(), 21, 11);
+    let (provider, _) = setup_test_provider(coins.clone(), Config::local_node()).await;
+    wallet.set_provider(provider);
```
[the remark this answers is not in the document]
vnepveu: I'm not sure I follow. Line `1552`, where does the `wallet` come from?

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: See comment above. To clarify I can maybe add a different worflow. There is an abstaction for the wallet obviously.

packages/fuels-abigen-macro/tests/harness.rs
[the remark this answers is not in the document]
digorithm: Ah, scratch that, totally forgot `setup_coins` need the wallet's address... ugh, we're stuck with `wallet.set_provider()` in this case 😅

vnepveu:
> RE:
> 
> > Question: should this get documented in The Book?
> 
> 100%! Do you mind updating the book in this PR as well? We're trying to keep doc updates and code updates in sync by bundling them into the same PR as much as we can.

Sure thing, I'm just not sure where to put it? In the `Wallet` section?

digorithm:
Yeah, a new sub-header under the wallet section should do the trick! Maybe call it "Querying for balances" / "Checking balances"?

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: Ok, so I'll write it "canonically", and add the unit tests later then!

docs/src/getting-started/wallets.md
```diff
@@ -118,8 +118,25 @@ let password = "my_master_password";
 let uuid = wallet.encrypt(&dir, password).unwrap();
 ```
 
+## Checking balances and coins
+
+First, one should keep in mind that, with UTXOs, each _coin_ is unique, and corresponds to a unique UTXO. So, when you want to query the balance for a given asset ID, you want to query the sum of the amount in each unspent coin. This is done very easily with a wallet:
```
digorithm, line 123: Since we're talking about the uniqueness of UTXOs, we should also mention, after `each _coin_ is unique, and corresponds to a unique UTXO`, that a coin also holds an amount, maybe use the USD bills analogy: a 10 dollars bill is a single "coin" with the amount `10`, a 50 dollars bill is a single "coin" wit the amount `50`, and, likewise, you can have a single coin here with an arbitrary amount.

The concept of coin and amount is often confusing for newcomers!

digorithm (approved):
Great work! :shipit:

PR #322, opened 2022-05-26
fuel-core 0.8 upgrade (predicates)

Voxelot:
- Updated fuel-core to 0.8, dealing with any breaking API changes related to `Input::coin` becoming `Input::coin_signed`. This will make the SDK compatible with the new UTXO predicate feature in fuel-core.

- Reduced the number of direct deps fuels-rs has to manage by leveraging re-exported types wherever possible.
  - This resolved an issue where some fuels packages depended on different major versions of fuel-asm. Using re-exported types will prevent missed version bumps in the future.
  - It also simplifies the number of dependencies that SDK consumers (sway apps) need to maintain, allowing them to simply import `fuels` for all their needs. 
  - The only place where a direct dependency above fuel-gql-client couldn't be removed was in fuels-core. This is because it's meant to be a lightweight package, and doesn't require the overhead of the client. The fuel-gql-client also isn't WASM friendly, so not including it into fuels-core will make it easier for that package to support no-std usage of the ABI macros.

### DIFF
--- a/packages/fuels-contract/src/contract.rs
+++ b/packages/fuels-contract/src/contract.rs
@@ -1,27 +1,22 @@
-use crate::abi_decoder::ABIDecoder;
-use crate::abi_encoder::ABIEncoder;
-use crate::script::Script;
+use crate::{abi_decoder::ABIDecoder, abi_encoder::ABIEncoder, script::Script};
 use anyhow::Result;
-use fuel_asm::Opcode;
-use fuel_gql_client::client::FuelClient;
-use fuel_tx::{
-    Address, AssetId, ContractId, Input, Output, Receipt, StorageSlot, Transaction, UtxoId,
+use fuel_gql_client::{
+    client::FuelClient,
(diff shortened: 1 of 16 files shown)


packages/fuels-contract/src/contract.rs
```diff
@@ -98,6 +94,7 @@ impl Contract {
         call_parameters: &CallParameters,
         compute_calldata_offset: bool,
     ) -> Result<(Vec<u8>, Vec<u8>), Error> {
+        use fuel_gql_client::fuel_types;
```
vnepveu, line 97: I'm not sure where we are using `fuel_types` below?

vnepveu (approved):
LGTM, just one question. Great work on reducing the imports footprint!

packages/fuels-contract/src/contract.rs
Voxelot: There is a macro hygiene issue with script-with-data-offset. I have a draft pr in fuel vm up that will make this no longer required.

adlerjohn (approved):
Looks fine but will defer to @digorithm

PR #202, opened 2022-04-04
Create `examples/` directory and include examples in docs

Br1ght0ne:
Close #101, #186.

- `examples/` directory
  - [x] Create a wallet from all possible options
  - [ ] Sign and send transactions crafted by the user
  - [x] Transfer funds from one wallet to another
  - [x] Compile, deploy, and interact with contracts using all possible options (mostly what's in `harness.rs`)
  - [x] ~~(new) SwaySwap tests. These are great tests that's been used in the SwaySwap repo. I ported them to `fuels-rs`'s `harness.rs`, but they should go in this `examples/`, so I moved them to this [gist here](https://gist.github.com/digorithm/ea46c1b874c58491e3f9ec9b0443c379). Feel free to copy it onto the examples.~~ SwaySwap tests will probably stay in their repo.
- [x] use examples in `docs` mdBook
  - blocked by #338

PR #324, opened 2022-05-26
feat: log receipts when contract call panics or reverts

### LINKED ISSUE #295: Change call unwrap to not panic if transaction reverts or panics
Currently, if a transaction reverts or panics, `script.call(&client).await` returns an error variant rather than an ok variant. This makes it impossible to view the receipts and actually debug what's wrong. This should be changed to always return the ok variant if the transaction is executed successfully, regardless of whether execution reverted/panicked or not.

vnepveu:
This PR closes #295 by changing the `ContractCallError` variant to log the
receipts in case of reverting or panicking contract calls.

### DIFF
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -284,8 +284,8 @@ If the contract call succeeded, getting the receipts and logs is done this way:
 
 ```rust,ignore
 let response = contract_instance.my_method(args).call().await?;
-logs : Vec<String> = response.logs // This gives out the decoded hex LOGD logs
-receipts : Vec<Receipt> =  response.receipts // This gives out all the receipts of the transaction
+let logs: Vec<String> = response.logs // This gives out the decoded hex LOGD logs
+let receipts: Vec<Receipt> =  response.receipts // This gives out all the receipts of the transaction
 ```
 
 > Note that for this to work, because of the `?`, it means the call has to have succeded.
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -299,7 +299,7 @@ match response {
 }
 ```
 
-> **Note** Is is generally considered good practice, when you expect the call to succeed, to unwrap the response with `?`, this way:
+> **Note:** It is generally considered good practice when you expect the call to succeed, to unwrap the response with `?`, this way:
 > ```rust, ignore
 > let response = contract_instance.my_method(args).call().await?;
 > ```
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -294,17 +294,6 @@ let receipts: Vec<Receipt> =  response.receipts // This gives out all the receip
 
 - If the contract call creates an error, the reason and the receipts of the transaction will be displayed in the Rust `panic`.
 
-- If you want to use the receipts in the case of a `ContractCallError`, you can do:
-
-```rust,ignore
-let contract_call_error = contract_instance.my_error_method(args).call().unwrap_err();
-
-if let ContractCallError(reason, receipts) = contract_call_error {
-    // Do things with `reason` and `receipts`
-}
-```
-> **Note**: It is generally not considered good practice to use `unwrap_err`.
-
 ## More examples
 
 You can find runnable examples under [`fuels-abigen-macro/tests/harness.rs`](https://github.com/FuelLabs/fuels-rs/blob/master/packages/fuels-abigen-macro/tests/harness.rs).
--- a/docs/src/getting-started/basics.md
+++ b/docs/src/getting-started/basics.md
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
 let connected_contract_instance = MyContract::new(contract_id, wallet);
 ```
 
+## Getting the contract call outputs
+
+### Contract call success
+
+If the contract call succeeded, getting the receipts and logs is done this way:
+
+```rust,ignore
+let result = contract_instance.my_method(args).call().await?;
+logs = result.logs.unwrap() // This gives out the decoded hex LOGD logs
+receipts = result.receipts // This gives out all the receipts of the transaction
+```
+
+> Note that for this to work, because of the `?`, it means the call has to have succeded.
+
+### `ContractCallError`
+
+- If the contract call creates an error, the reason and the receipts of the transaction will be displayed in the Rust `panic`.
+
+- If you want to use the receipts in the case of a `ContractCallError`, you can do:
+
+```rust,ignore
+let contract_call_error = contract_instance.my_error_method(args).call().unwrap_err();
+
+if let ContractCallError(reason, receipts) = contract_call_error {
+    // Do things with `reason` and `receipts`
+}
+```
+> It is generally not considered good practice to use `unwrap_err`.
+
 ## More examples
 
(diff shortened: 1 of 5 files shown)


adlerjohn (commented):
Can you update the docs to show how to extract receipts from both variants?

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
 
+## Getting the contract call outputs
+
+### Contract call success
+
+If the contract call succeeded, getting the receipts and logs is done this way:
+
+```rust,ignore
+let result = contract_instance.my_method(args).call().await?;
+logs = result.logs.unwrap() // This gives out the decoded hex LOGD logs
```
adlerjohn, line 287: Wait, why is this a `Result`? If the transaction completed execution then it will have logs, no matter what. If the transaction was invalid, then it won't even run, and won't have receipts other. The entire `result` should be a `Result`, not just the `logs`.

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
+receipts = result.receipts // This gives out all the receipts of the transaction
```
adlerjohn, line 288: @Br1ght0ne moved examples into the `examples/` directory with #202. In order to keep these docs up to date, all new code examples, even one-liners, should be placed there.

vnepveu: In the docs, some examples are not from the `examples` directory. But maybe everything should be you're right.

docs/src/getting-started/basics.md
vnepveu: `result` here is a `CallResponse`, the `?` is there to unwrap it from a `Result<CallResponse, ContractCallError>`. The `logs.unwrap()` is indeed misguided, because it's an `unwrap` on an `Options<Vec<Receipt>>`.
I'm not sure what the example should be then, the point was to show how to "get the receipts" in both cases.

vnepveu: I can change `result` into `response` to make it more clear, you're right.

docs/src/getting-started/basics.md
adlerjohn: Ah I guess it can be deferred to a future PR with #342

docs/src/getting-started/basics.md
adlerjohn: Shouldn't it be `result.unwrap().logs` then?

adlerjohn: An empty `Vec` makes more sense than an `Option`.

vnepveu: Agreed, hadn't thought about it

adlerjohn: #343 to rename

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
+## Getting the contract call outputs
+
+### Contract call success
+
+If the contract call succeeded, getting the receipts and logs is done this way:
+
+```rust,ignore
+let response = contract_instance.my_method(args).call().await?;
+logs : Vec<String> = response.logs // This gives out the decoded hex LOGD logs
+receipts : Vec<Receipt> =  response.receipts // This gives out all the receipts of the transaction
```
adlerjohn, line 288:
```suggestion
let logs: Vec<String> = response.logs // This gives out the decoded hex LOGD logs
let receipts: Vec<Receipt> =  response.receipts // This gives out all the receipts of the transaction
```

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
+```
+
+> Note that for this to work, because of the `?`, it means the call has to have succeded.
```
adlerjohn, line 291:
```suggestion
> **Note**: For this to work, because of the `?`, it means the call has to have succeded.
```

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
+- If you want to use the receipts in the case of a `ContractCallError`, you can do:
+
+```rust,ignore
+let contract_call_error = contract_instance.my_error_method(args).call().unwrap_err();
+
+if let ContractCallError(reason, receipts) = contract_call_error {
+    // Do things with `reason` and `receipts`
+}
+```
+> It is generally not considered good practice to use `unwrap_err`.
```
adlerjohn, line 306:
```suggestion
> **Note**: It is generally not considered good practice to use `unwrap_err`.
```

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,35 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
+> **Note**: For this to work, because of the `?`, it means the call has to have succeded.
+
+### `ContractCallError`
+
+- If the contract call creates an error, the reason and the receipts of the transaction will be displayed in the Rust `panic`.
+
+- If you want to use the receipts in the case of a `ContractCallError`, you can do:
+
+```rust,ignore
+let contract_call_error = contract_instance.my_error_method(args).call().unwrap_err();
```
adlerjohn, line 300: How do you know it's an error or not before unwrapping? This example isn't complete. All the more reason to have actual examples that are tested.

vnepveu: I'm just trying to conform to the instruction `how to read the receipts in case it's an error`. IMO, in case it's an error, the logging done by the `ContractCallError(reason, receipts)` is enough, but I felt like you wanted a code example.

adlerjohn: Please correct me if I'm wrong, but the case here is if the transaction is valid but reverts, right? If so, then the SDK needs to handle such cases. It can't just panic. A code example is needed. You can keep the example, just expand on it.

vnepveu: But we agreed in the issue that the SDK would error on a valid transaction that reverts.

adlerjohn: Sure, it can be an `Err` variant. But the user still needs a way to extract the receipts and logs from there!

vnepveu: What I suggest is we merge as-is so we at least have logging, and then we open an issue to make it not an `Error`

adlerjohn: Okay. Well at least the current is better than what was available before.

adlerjohn (commented):
Seems fine to me, will defer to SDK team. Please file an issue to follow-up on improving this further.

docs/src/getting-started/basics.md
```diff
@@ -276,6 +276,30 @@ let contract_id = "0x0123..." // Your contract ID as a string.
...
+        // Do things with logs and receipts
+    }
+    ContractCallError(reason, receipts) => {
+        println!("ContractCall failed with reason: {}", reason);
+        println!("Transaction receipts are: {:?}", receipts);
+    }
+}
+```
+
+> **Note** Is is generally considered good practice, when you expect the call to succeed, to unwrap the response with `?`, this way:
```
digorithm, line 298:
```suggestion
> **Note:** It is generally considered good practice when you expect the call to succeed, to unwrap the response with `?`, this way:
```

digorithm (approved):
That's a nice improvement! Let's take a look at https://github.com/FuelLabs/fuels-rs/issues/345 later when we have more time and energy; it seems it's a more nuanced issue that could make the UX weaker if we make the wrong decisions there.

PR #330, opened 2022-05-26
Fix crypto dep random feature.

adlerjohn:
Resolves this failing CI: https://github.com/FuelLabs/fuels-rs/actions/runs/2393556289

Adds the `random` feature for `fuel-crypto` in `fuels-signers` and removes unused `fuel-crypto` in `fuels-test-helpers`.


digorithm (approved):
:shipit:

PR #333, opened 2022-05-27
Add unused deps check

adlerjohn:


digorithm (dismissed):

adlerjohn:
Ahh this doesn't work nicely because it depends on the ABI and bin files existing, which requires the rest of `fuels` to be built. Closing for now.

PR #335, opened 2022-05-27
Disable feature unification in CI builds

Voxelot:
We were having build failures during cargo publish dry-run that weren't caught earlier in CI since the build command was unifying features across all crates (including test-helpers). [`cargo make`](https://github.com/sagiegurari/cargo-make) comes with sane default configurations for building the repo in a way that will prevent this from happening in the future.

edit: while cargo make works, we can speed things up a lot by using cargo check on each package in parallel to skip linking. Updated the PR to that approach which cuts the CI pipeline time in half.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
             !packages/fuels-abigen-macro/tests/test_projects/**/Forc.lock
             !packages/fuels-abigen-macro/tests/test_projects/**/.gitignore
 
-  get_workspace_members:
+  get-workspace-members:
     needs: cancel-previous-runs
     runs-on: ubuntu-latest
     outputs:
@@ -102,7 +102,7 @@ jobs:
     needs:
       - setup-test-projects
       - verify-rust-version
-      - get_workspace_members
+      - get-workspace-members
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


digorithm (dismissed):
Thanks!

Voxelot:
going to give parallelizing a whirl

vnepveu (approved):
Thanks a lot for this!

.github/workflows/ci.yml
```diff
@@ -11,83 +11,158 @@ on:
...
       - name: Install Forc
-        uses: actions-rs/cargo@v1
-        with:
-          command: install
-          args: forc --version 0.13.0
+        run: |
+          curl -sSLf https://github.com/FuelLabs/sway/releases/download/v${{ env.FORC_VERSION }}/forc-binaries-linux_amd64.tar.gz -L -o forc.tar.gz
+          tar -xvf forc.tar.gz
+          chmod +x forc-binaries/forc
+          mv forc-binaries/forc /usr/local/bin/forc
```
adlerjohn, line 51: Maybe move all binaries into `/usr/local/bin/`? That way CI can check for Forc formatting in a future PR easily.

digorithm:
@Voxelot What's the current status on this?

.github/workflows/ci.yml
adlerjohn: Nit, can be done in a future PR so as not to block.

PR #348, opened 2022-05-29
ci: run clippy checks

Br1ght0ne:


Br1ght0ne:
This might conflict with #335. Main difference between these PRs - here I used https://github.com/actions-rs/clippy-check to get annotations inline with the PR diff (see example [here](https://github.com/FuelLabs/fuels-rs/pull/348/checks?check_run_id=6639993158)). `clippy-check` might get a special treatment, or we just use it as a status marker, both work.

adlerjohn (commented):
Merge conflicts. Maybe refactor into the matrix?

Br1ght0ne:
Closing because it doesn't fit well with `matrix` strategy, which is working really well.

PR #349, opened 2022-05-29
docs: document `get_spendable_coins` functions

### LINKED ISSUE #301: Document how `get_coins` and `get_spendable_coins` work

vnepveu:
This PR closes #301


digorithm (approved):
Good stuff, but maybe instead of saying that the client is trying to minimize, let's say it's trying to optimize in order to prevent dust accumulation?

Issue #318, opened 2022-05-25
Add support for WASM targets. Add CI job to maintain support.


mitchmindtree:
## Motivation

Currently at `fuel-indexer` we're prioritising updating to the latest versions of the `fuels-*` and `fuel-*` crates in order to catch up and stay in sync with the Fuel ecosystem https://github.com/FuelLabs/fuel-indexer/issues/20.

Previously, [it was thought that the lack of `#![no_std]` support was a blocker for this](https://github.com/FuelLabs/fuels-rs/issues/62), however it appears that what we really need is support for the WASM target which is not the same but imposes similar limitations.

The motivation here is that `fuel-indexer` supports a WASM plugin system, where users of `fuel-indexer` can write and share custom plugins based on their application's particular indexing needs.

As the WASM target is designed for execution in quite a constrained context, it has some limitations compared to regular native targets. Most of these revolve around I/O (e.g. file access, accessing the time, etc).

## WASM target

@tjsharp1 can you confirm which WASM target we need to support here?

My initial assumption was `wasm32-unknown-unknown`, but I just noticed [this line](https://github.com/FuelLabs/fuel-indexer/pull/32/files#diff-50ded71cda3d4275160023b301fa4304b7bc0af6e17887600db3ef82c0b78ea6R11) in the indexer-samples README in https://github.com/FuelLabs/fuel-indexer/pull/32:

```
cargo r --bin index-runner -- --wasm target/wasm32-wasi/release/simple_wasm.wasm
```

does that mean the target we want is `wasm32-wasi`? We'll need to know exactly what target to test for in order to be able to address this.

## Crates requiring attention

Crates that currently do not support WASM are:

- `fuels-contract`
- `fuels-signers`
- `fuels-test-helpers`

@tjsharp1 you mentioned each of these have dependencies that do not support the WASM target. Can you share what these dependencies are and what the errors look like? This might help to identify alternatives or see if we need to enable certain features in these dependencies, etc.

## CI

We should also add a CI job that at least `cargo build`'s the desired WASM target so that we can maintain support going forward and avoid accidental future breakage.

## Potential Temporary Workaround

[...]

Issue #317, opened 2022-05-24
Allow some deps of fuels to be optional


tjsharp1:
`fuels-contract`, `fuels-signers` and `fuels-test-helpers` all have dependencies that don't compile to WASM.
It would be nice to have these as optional dependencies so they don't have to be compiled for a WASM target.

Really all the indexer needs is the type definitions out of the contract ABI, as well as `Tokenizable`, `ParamType` and the other types necessary to deserialize.

digorithm:
Isn't this related to https://github.com/FuelLabs/fuels-rs/issues/62?

hal3e:
@segfault-magnet recently introduced a `std` flag in `fuels` which makes it wasm compatible.

PR #351, opened 2022-05-31
Wasm support

tjsharp1:
Opening as a draft for now to get some eyes on it. There were some build issues on my machine that looked like they may be config related, so I wanted to see if it passes CI.

It looks like this is about all I need in order to make things work in WASM. As long as the codegen output `use` statements that are from `fuels_core`, I think it will compile fine.

I've added a separate directory for WASM tests, since I have to override the default build target. I added the `.cargo/config.toml` to set it to wasm32-unknown-unkown.

I believe this will address #317 and #318

### DIFF
--- a/packages/wasm-tests/Cargo.toml
+++ b/packages/wasm-tests/Cargo.toml
@@ -11,3 +11,6 @@ crate-type = ['cdylib']
 fuels-abigen-macro = { path = "../fuels-abigen-macro" }
 fuels-core = { path = "../fuels-core" }
 getrandom = { version = "0.2", features = ["js"] }
+
+[dev-dependencies]
+webassembly-test = "0.1"
--- a/packages/wasm-tests/Cargo.toml
+++ b/packages/wasm-tests/Cargo.toml
@@ -10,3 +10,4 @@ crate-type = ['cdylib']
 [dependencies]
 fuels-abigen-macro = { path = "../fuels-abigen-macro" }
 fuels-core = { path = "../fuels-core" }
+getrandom = { version = "0.2", features = ["js"] }
--- a/packages/wasm-tests/Cargo.toml
+++ b/packages/wasm-tests/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "wasm-tests"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[lib]
+crate-type = ['cdylib']
+
+[dependencies]
+fuels-abigen-macro = { path = "../fuels-abigen-macro" }
+fuels-core = { path = "../fuels-core" }
--- a/packages/wasm-tests/Cargo.toml
+++ b/packages/wasm-tests/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasm-tests"
-version = "0.1.0"
+version = "0.0.0"
 edition = "2021"
 publish = false
 
--- a/packages/wasm-tests/src/lib.rs
+++ b/packages/wasm-tests/src/lib.rs
@@ -1,0 +1,58 @@
+extern crate alloc;
+use fuels_abigen_macro::wasm_abigen;
+
+wasm_abigen!(
+    no_name,
+    r#"[
+        {
+            "type":"contract",
+            "inputs":[
+                {
+                    "name":"SomeEvent",
+                    "type":"struct SomeEvent",
+                    "components": [
+                        {
+                            "name": "id",
+                            "type": "u64"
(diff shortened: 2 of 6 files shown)


packages/wasm-tests/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "wasm-tests"
+version = "0.1.0"
```
adlerjohn, line 3: We generally use `0.0.0` for unpublished packages.
```suggestion
version = "0.0.0"
```

packages/wasm-tests/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+);
+
+pub fn the_fn() {
+    let a_struct = SomeEvent {
+        id: 20,
+        account: Default::default(),
+    };
+
+    println!("It works! {}", a_struct.id);
+}
```
Voxelot, line 57: When running `cargo test --target wasm32-unknown-unknown` with the following code:

```suggestion
}


#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test() {
        the_fn()
    }
}

```

I encountered a strange error:
```
/home/voxelot/code/fuel/fuels-rs/target/wasm32-unknown-unknown/debug/deps/wasm_tests-b1e4d09cfabc0966.wasm: 1: Syntax error: word unexpected (expecting ")")
error: test failed, to rerun pass '--lib'

```

Voxelot: Given fuels-core still largely relies on std, I'm skeptical about whether any of the `fs` methods will actually be implemented or just panic at runtime. Guess it depends the runtime we use the wasm in.

Voxelot: Alternately, would using the typescript sdk abigen w/ assemblyscript be a more dev friendly experience for wasm? @digorithm

tommyjk21: That first error was because cargo test was trying to run WASM as if it were a native executable... so not gonna work. I put in a WASM test harness to run that test.
I think for mac, the only thing needed is a `brew install llvm`, the one shipped by default does not have a WASM target support.

For loading the json abi file, that only happens in the proc macro. Proc macros are actually compiled prior to the crate that's being built, and only needs to be built for the host architecture, not to WASM. The thing that is important here is that the token stream that is output by the proc macro needs to be supported by WASM. All I need are rust type definitions and a way to deserialize a binary blob into those types, and that's the only thing that the ABI gen should output for WASM.

Voxelot: Ah that makes sense, thanks for completing the test harness example.

As far as my mac issue, I used brew llvm and had the wasm target working, but it got stuck on secp256k:

```
LLVM ERROR: malformed uleb128, extends past end
```

As far as the proc macro, that is reassuring. It's surprising to me that cargo doesn't load proc-macros from build-deps instead of normal deps, considering they're only run at build time. I wonder if that's because types from the macro deps that are used in the token stream need to be transitively imported?

tommyjk21: I thought the `.cargo/config.toml` would've handled that? I guess not though.  I'll put in a separate command for it.

For mac are you env vars pointed at the right binary? I had to do this:
```export PATH="/opt/homebrew/opt/llvm/bin:$PATH"
export CC=/opt/homebrew/opt/llvm/bin/clang
export AR=/opt/homebrew/opt/llvm/bin/llvm-ar
```

Though that issue looks like it's building secp256k to wasm? I don't think we'll ever have a need for making signatures in wasm, so maybe it should be an optional dep of fuels-core?

Voxelot: Ah nice, these exports fixed the mac issue for me!

Having secp256k be optional in fuels-core would make sense since fuels-rs only uses secp256k for signing. If someone was writing fancy indexer logic involving some cryptography, they could source their own no-std libs for that. Looks like we'd just need to disable default features for `fuel-tx` when importing it within `fuels-core`.

Voxelot: Hmm yeah maybe file an issue to investigate this later.

tjsharp1: Okay, filed it here: #371

Voxelot (approved):
lgtm, thanks for getting the CI and tests setup

digorithm (approved):
This is really cool (especially the CI work to support wasm stuff), thanks for putting this one up!

Issue #204, opened 2022-04-05
Improve and give more context to `Error::InvalidData`-related error handling


digorithm:
Currently, we're making extensive use of `Error::InvalidData` all over the codebase. 

For example: 

```Rust
pub fn tokenize_array<'a>(&self, value: &'a str, param: &ParamType) -> Result<Token, Error> {
      if !value.starts_with('[') || !value.ends_with(']') {
          return Err(Error::InvalidData);
      }
```

Even though the error naming is technically correct -- whenever we hit an `Error::InvalidData`, it is because the data being passed to its parent function is invalid -- it provides absolutely no context, making it extremely hard to debug and find where and why this happened.

A simple solution is to provide more context to this `Error` and go through each case and give the context (a simple reason string and/or contextual data) to this `Error` variant.

PR #357, opened 2022-06-02
feat: add context to InvalidData error

hal3e:
This commit should resolve https://github.com/FuelLabs/fuels-rs/issues/204

The InvalidData error type now provides more context. I have gone through all instances and updated the error messages accordingly.

Additionally, I have found a bug where the functions
 `pub fn tokenize_struct(&self, value: &str, params: &[ParamType]) -> Result<Token, Error> {`
 and 
  `pub fn tokenize_array<'a>(&self, value: &'a str, param: &ParamType) -> Result<Token, Error> {`
would not return an error if there were excess opening brackets. For example: 
`let values: Vec<String> = vec!["((42, true)".to_string()];`
would previously pass through the tokenize_struct function and return an error only at the end assertion in the test.

Now both functions return appropriate error messages.

Could somebody please go through the error messages and confirm that they reflect the errors accordingly ?
Should we have test cases to cover cases where the values string is not ok and we are sure that we return the right errors?

### DIFF
--- a/packages/fuels-contract/src/errors.rs
+++ b/packages/fuels-contract/src/errors.rs
@@ -10,8 +10,8 @@ use std::net;
 pub enum Error {
     #[error("Invalid name: {0}")]
     InvalidName(String),
-    #[error("Invalid data")]
-    InvalidData,
+    #[error("Invalid data: {0}")]
+    InvalidData(String),
     #[error("Missing data: {0}")]
     MissingData(String),
     #[error("Serialization error: {0}")]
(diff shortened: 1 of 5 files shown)


hal3e:
Amended the commit twice to fix rustfmt and clippy errors.

digorithm (approved):
Good stuff -- these `Error`s without any context were killing me when debugging complex bugs! 

As for 

> Should we have test cases to cover cases where the values string is not ok and we are sure that we return the right errors?

Yes -- do you mind opening an issue for this?

Br1ght0ne (approved):

hal3e:
Thanks! I will open a new issue.

Issue #206, opened 2022-04-06
Implement tuple tokenization for the encoding/decoding in the JSON ABI CLI tool


digorithm:
Currently, although we have support for tuples in the SDK, the CLI tool makes use of https://github.com/FuelLabs/fuels-rs/blob/0c848517a0f342cc43ac1eba5e4254bd86cb2fea/packages/fuels-core/src/json_abi.rs#L213 to tokenize command line arguments coming in as strings and then proceed with encoding/decoding, but this process hasn't been implemented for tuples:

https://github.com/FuelLabs/fuels-rs/blob/0c848517a0f342cc43ac1eba5e4254bd86cb2fea/packages/fuels-core/src/json_abi.rs#L240-L242

We should finish this implementation, although it's a low priority, as the CLI tool isn't used very much as of now other than for some very specific and quick debugging purposes.

PR #360, opened 2022-06-03
feat: implement tuple tokenization in json abi cli

hal3e:
This commit should close https://github.com/FuelLabs/fuels-rs/issues/206

The tuple tokenization is implemented and I have added a new test case for it.
`#[test]
 fn tuple_encode_and_decode() {`

The expected encoding `let expected_encode = "000000000e8ece68000000000000002a000000000000002a";` was generated with  `sway` and `forc`.

Currently we do not support nested tuples or tuples with custom types. When https://github.com/FuelLabs/fuels-rs/pull/356 is merged we will have to update the test cases to include them.


hal3e:
Rebased the master branch after https://github.com/FuelLabs/fuels-rs/pull/356 was merged.

I added an additional test with nested tuples and custom types. Also fixed a small bug where the enum tokinization did not use the trimmed value string.

packages/fuels-core/src/json_abi.rs
```diff
@@ -1306,6 +1395,135 @@ mod tests {
...
+        let values: Vec<String> = vec!["((42, 42), (John), (1, 0))".to_string()];
+
+        let mut abi = ABIParser::new();
+
+        let function_name = "takes_nested_tuple";
+
+        let encoded = abi
+            .encode_with_function_selector(json_abi, function_name, &values)
+            .unwrap();
+        println!("encoded: {:?}\n", encoded);
```
adlerjohn, line 1520: We probably don't want indiscriminate `println!`s anywhere

hal3e: Thanks for the input. I will delete them all.

hal3e:
Deleted all `println!`

packages/fuels-core/src/json_abi.rs
```diff
@@ -421,6 +419,97 @@ impl ABIParser {
...
+                        }
+                        _ => {}
+                    }
+                }
+                '"' => {
+                    ignore = !ignore;
+                }
+                ',' if nested == 1 && !ignore => {
+                    let sub = &value[last_item..pos];
+                    // If we've encountered an array within a struct property
```
MujkicA, line 479: Should be 
```suggestion
                    // If we've encountered an array within an enum property
```

MujkicA (approved):

packages/fuels-core/src/json_abi.rs
hal3e: Thanks Ahmed. Good catch!

vnepveu:
#356 has been merged, marking as draft so you can add the nested types tests as mentioned :smile:

packages/fuels-core/src/json_abi.rs
```diff
@@ -421,6 +419,97 @@ impl ABIParser {
...
+            return Ok(Token::Tuple(vec![]));
+        }
+
+        let mut result = vec![];
+        let mut nested = 0isize;
+        let mut ignore = false;
+        let mut last_item = 1;
+        let mut params_iter = params.iter();
+
+        for (pos, ch) in value.chars().enumerate() {
```
vnepveu, line 445: Even if this algo is in several places in the code, I've never fully understood it. Could you please use this PR as an opportunity to document it briefly? The variables and process.

packages/fuels-core/src/json_abi.rs
```diff
@@ -1299,13 +1379,139 @@ mod tests {
...
+                    "type": "(u64, u64)",
+                    "components": [
+                      {
+                        "name": "__tuple_element",
+                        "type": "u64",
+                        "components": null
+                      },
+                      {
+                        "name": "__tuple_element",
+                        "type": "u64",
```
vnepveu, line 1406: try with two different types and values as a sanity check

vnepveu (changes requested):
Great work! Simple and efficient, a few nits

packages/fuels-core/src/json_abi.rs
hal3e: I added the comments inside the `tokenize_struct`, `tokenize_array` and `tokenize_tuple` functions

packages/fuels-core/src/json_abi.rs
hal3e: Done :)

packages/fuels-core/src/json_abi.rs
```diff
@@ -342,6 +351,13 @@ impl ABIParser {
             return Ok(Token::Array(vec![]));
         }
 
+        //To parse the value string we use a two pointer/index approach.
+        //The items are comma separated and if an item is tokenized the last_item
+        //index is moved to the current position.
+        //The variable nested is incremented and decremented if a bracket is encountered,
+        //and appropraite errors are thrown if the nested count is not 0
+        //Characters inside quotes are ignored and they are tokenized as one item.
+        //An error is return if there is an odd number of quotes.
```
vnepveu, line 360: I think you can say something along the lines of `for more details about this algo, refer to the tokenize_struct method` so we can avoid having to update the docstring everytime :smiley:

packages/fuels-core/src/json_abi.rs
```diff
@@ -421,6 +437,108 @@ impl ABIParser {
         Ok(Token::Array(result))
     }
 
+    /// Creates a tuple `Token` from an array of parameter types and a string of values.
```
vnepveu, line 440: Is it a `Token::Tuple` or a "tuple `Token`"

vnepveu (changes requested):
One last nit! otherwise excellent

packages/fuels-core/src/json_abi.rs
hal3e: Done

packages/fuels-core/src/json_abi.rs
hal3e: i put in Token::Tuple now to be exact. And I like consistency so I changed the other docstrings as well. There was an wrong docstring for the `tokenize_array` which was merged before. I changed it to match the function

packages/fuels-core/src/json_abi.rs
vnepveu: I don't see it done, I guess you forgot to push?

vnepveu (changes requested):
Last nit about docs then lgtm!

packages/fuels-core/src/json_abi.rs
hal3e: Now it should be ok.

vnepveu (approved):
lgtm :shipit:

PR #361, opened 2022-06-05
Enum encoding

### LINKED ISSUE #352: Enums with different variant widths are not properly handled on the ABI interface
For example, if my enum looks like:
```rust
enum SomeEnum {
    V1: b256,
    V2: u64
}
```
and then I pass `SomeEnum::V2(42)` as an argument to a contract method from the SDK, the SDK is actually sending 2 words (tag + `u64`) and not 5 (tag + 4 words because `b256` is the largest variant).

The reason why this matters is because the compiler is expecting 5 words as it has to account for the largest possible enum that can be sent. When the contract method has multiple arguments, the compiler can get confused if that size is not respected by the SDK (the compiler assumes 5 words and it starts reading into the next argument etc.) 

I think the SDK should do the following:
* Write the tag which is 1 word
* Write the value
* Write zeros until we reach the max possible number of words which is basically `1 + size_of(largest variant)`

segfault-magnet:
closes #352 
blocked by FuelLabs/sway#1872 

### Context on the problem/solution 
`Token`s are used to encode/decode arguments to/from the VM.

Previously Enums were represented as `Token`s that had only the discriminant and the `Token` that is pointed to by that discriminant. This is inadequate to properly encode enums. More context in the issue.

In order to inform the encoder of the enum's variants, a third argument was given to the `EnumSelector` -- a `Vec<ParamType>` which holds the types of the variants of the enum in question.

The encoder then proceeds to determine the biggest variant and pad the enum appropriately. 

The decoder was changed to ignore the extra padding so as not to read into the next argument.

### About doing it in compile time RFC
Theoretically, the proper encoding size for all custom enums could have been determined at compile-time and baked into our custom enums.

I took the liberty of trying out this approach as well, you can take a look at it [here](https://github.com/segfault-magnet/fuels-rs/tree/segfault_magnet/352-enum-encoding-compile-time).

This same process was done inside the `abigen!` macro and the resulting word size was baked into the `EnumSelector` during its creation in `into_tokens`. For decoding the size was also given to the `ParamType` making it of the signature: `ParamType::Enum(variations, usize)`. 

Although I cannot overstate how much I like doing it during compile time, ultimately I decided against it because it seemed to damage code cohesion. The hardcoded enum size was baked into the generated Enum types -- that rubbed me the wrong way. Although Tokens don't exist for anything else but to facilitate encoding/decoding, it felt out of place to have such a hard coded number beside a bunch of types -- as if I'm taking a bit of what used to be the responsibility of the encoder (to determine VM type sizes) and putting it into our custom types.

I'd like to hear your thoughts on this. Do you find the compile-time approach acceptable? Is the damage to cohesion palpable to you as well? 

Although it goes without saying that the compile-time approach is more performant, I found the differences to be negligible. Especially in the light that I don't see anybody passing overly deeply nested types which would require a bunch of recursions to figure out its type should it ever be an enum variant.

Since coding/decoding types is temporally coupled to making a network request, I cannot see that we would ever get any noticeable speedup compared to the time the network request took.

### Where to put an e2e test?
The code is covered by unit tests. Ideally, I'd have an e2e test showing that the VM receives the enum correctly, doesn't mangle other arguments, and properly returns it.

I don't know where to put it... the rest of the tests of that kind live in fuels-abigen-macro -- but they are testing that the code-generation is working correctly, and that is not what I would test. 

Was thinking of creating a similar integration tests folder in fuels-core if that is alright with everyone?


MujkicA:
Commits can be squashed

segfault-magnet:
> Commits can be squashed

Thanks, didn't consider that. I've squashed them :)

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,104 @@
+use crate::{ParamType, WORD_SIZE};
+
+/// If given a non-empty slice it will return the number of `WORD`s needed to fit the biggest of the provided types.
+///
+///
+/// returns: Option<usize>
```
MujkicA, line 6: IMO this is redundant, return type is obvious from the signature

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,104 @@
...
+///
+pub fn max_by_encoding_width(params: &[ParamType]) -> Option<usize> {
+    params.iter().map(encoding_width).max()
+}
+
+/// Calculates the number of `WORD`s the VM expects this parameter to be encoded in.
+///
+/// # Arguments
```
MujkicA, line 14: Same as above

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,104 @@
...
+///
+/// * `param`: the parameter that you plan on encoding
+///
+/// returns: usize
+///
+/// # Panics
```
MujkicA, line 20: Can we return a Result here and have this as an explicit error

MujkicA (changes requested):
Couple nits. The small improvements to readability here and there are neat.

vnepveu (changes requested):
Woah! Great content. I made a lot of comments about documentation and style, but overall this is an impressive first contribution! It could use more documentation because it's not a simple PR and the things you're doing are non-trivial. Congrats! :tada:

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -69,7 +69,7 @@ pub fn expand_custom_struct(prop: &Property) -> Result<TokenStream, Error> {
                 fields.push(quote! {pub #field_name: #enum_name});
                 args.push(quote! {#field_name: #enum_name::new_from_tokens(&tokens[#idx..])});
                 struct_fields_tokens.push(quote! { tokens.push(self.#field_name.into_token()) });
-                param_types.push(quote! { types.push(ParamType::Enum(#enum_name::param_types())) });
+                param_types.push(quote! { types.push(ParamType::Enum(EnumVariants::new(#enum_name::param_types()).unwrap())) });
```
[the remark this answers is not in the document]
segfault-magnet: Using '?' would require changing the signature of:
`fn param_types() -> Vec<ParamType>` 
to
`fn param_types() -> Result<ParamType, Error>`

since these param_types are expanded into the `param_types` method on the custom type.

But it is safe to call unwrap() here because: 
We're going through all the components that make up some custom type. For every component, we parse it into a `ParamType` and proceed to do various things depending on which ParamType it is.

Here we've matched against `ParamType::Enum(EnumVariants)` so we know for a fact that the component (enum) in question must have variants since `EnumVariants` could not have been constructed empty.

Since we know that we successfully created EnumVariants during compile time for this type, we can safely do so later on in runtime when somebody calls `param_types()` on the generated type.

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -46,14 +49,26 @@ impl ABIEncoder {
...
+
+                    let pre_encode_size = self.encoded_args.len();
+
                     // Encode the discriminant of the enum
-                    self.encoded_args.extend(pad_u8(&arg_enum.0));
+                    self.encoded_args.extend(pad_u8(discriminant));
                     // Encode the Token within the enum
-                    self.encode(&[arg_enum.1.to_owned()])?;
+                    self.encode(slice::from_ref(token))?;
+
```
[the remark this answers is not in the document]
segfault-magnet: I'll add a docstring pointing to the fuel-specs detailing how an enum should be encoded.

If that is too detached then I can write the process here as well.

Only thing is I need to wait until the compiler team formalizes the spec.

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -521,6 +542,95 @@ mod tests {
...
+        let expected: Vec<u8> = [enum_discriminant_enc, u64_enc, enum_padding]
+            .into_iter()
+            .flatten()
+            .collect();
+
+        assert_eq!(hex::encode(expected), hex::encode(encoded));
+    }
+
+    #[test]
+    fn encoding_enums_with_deeply_nested_types() {
```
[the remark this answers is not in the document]
segfault-magnet: This test sure is a sore on the eyes.

I'm not sure I understood you:

to 'group every part in one paragraph' 
what is a 'part' -- one of these enums/structs? 

What is a paragraph?

I thought grouping the custom type docstring along with its token and type would improve to the readability -- but I still don't like how it turned out.

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,105 @@
+use crate::{ParamType, WORD_SIZE};
+
+/// If given a non-empty slice it will return the number of `WORD`s needed to fit the biggest of the provided types.
+pub fn max_by_encoding_width(params: &[ParamType]) -> Option<usize> {
```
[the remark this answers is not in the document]
segfault-magnet: I wanted to call it that, but there is one thing that rubs me wrong about the name:
It might be misunderstood to mean that we're trying to encode the whole slice -- all of the types present in it.

That is why I tried to communicate that we're somehow judging them based on their encoding size, and `max_by_encoding_width` was the best I could come up with.

But I just thought of something as I was writing this. Please see how you like the new version :)

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,105 @@
...
+    params.iter().map(encoding_width).max()
+}
+
+/// Calculates the number of `WORD`s the VM expects this parameter to be encoded in.
+pub fn encoding_width(param: &ParamType) -> usize {
```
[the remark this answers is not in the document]
segfault-magnet: I'm not sure about the `get` part, seems like it can be added anywhere you're returning something.

But the expected part is great, communicates that we're only serving the VM.

packages/fuels-core/src/code_gen/custom_types_gen.rs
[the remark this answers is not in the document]
vnepveu: Perfect justification! can you add this reasoning somewhere in the docstrings? I saw you do that reasoning with Ahmed above and I think it's worth writing it down in code.

packages/fuels-core/src/types.rs
```diff
@@ -32,10 +32,8 @@ pub fn expand_type(kind: &ParamType) -> Result<TokenStream, Error> {
             Ok(quote! { (#(#members,)*) })
         }
         ParamType::Enum(members) => {
-            if members.is_empty() {
```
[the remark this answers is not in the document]
vnepveu: ok

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -1,252 +1,287 @@
...
+        Ok(result)
+    }
+
+    /// Will decode an Enum according to TODO: PUT_FUEL_SPECS_ENUM_ENCODING_LINK_HERE
+    ///
+    /// # Arguments
+    ///
+    /// * `data`: slice of encoded data on whose beginning we're expecting an encoded enum
+    /// * `variants`: all types that this particular enum type could hold
+    fn decode_enum(data: &[u8], variants: &EnumVariants) -> Result<DecodeResult, CodecError> {
```
vnepveu, line 215: vry gud

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -1,252 +1,287 @@
...
         ))
     } else {
-        Ok(&data[offset..(offset + len)])
+        Ok(&data[..len])
     }
 }
 
-fn peek_word(data: &[u8], offset: usize) -> Result<ByteArray, CodecError> {
-    peek(data, offset, WORD_SIZE as usize).map(|x| {
+fn peek_word(data: &[u8]) -> Result<ByteArray, CodecError> {
```
vnepveu, line 266: Even if it seems to work since the tests pass, I don't understand why you can simply not care about any offset anymone?

packages/fuels-core/src/abi_encoder.rs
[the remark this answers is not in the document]
vnepveu: I would say something like:


```rust

deeper_enum_variant = xxx
deeper_enum_token = xxx
deeper_enum_discriminant_encoded = xxx


struct_a_type = xxx
struct_a_token = xxx


```
 would be more clear: ie group operations by variable names! wdyt?

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,110 @@
+use crate::{EnumVariants, ParamType, WORD_SIZE};
+
+// Calculates how many WORDs are needed to encode any given variant in `variants`
+pub fn encoding_width_to_fit_any(variants: &EnumVariants) -> usize {
```
vnepveu, line 4: `compute_max_needed_encoding_width` would make more sense imo.

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,110 @@
...
+    variants
+        .param_types()
+        .iter()
+        .map(expected_encoding_width)
+        .max()
+        .expect("Will never panic because EnumVariants must have at least variant inside it!")
```
vnepveu, line 10: `at least one variant`

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -69,7 +69,24 @@ pub fn expand_custom_struct(prop: &Property) -> Result<TokenStream, Error> {
                 fields.push(quote! {pub #field_name: #enum_name});
                 args.push(quote! {#field_name: #enum_name::new_from_tokens(&tokens[#idx..])});
                 struct_fields_tokens.push(quote! { tokens.push(self.#field_name.into_token()) });
-                param_types.push(quote! { types.push(ParamType::Enum(EnumVariants::new(#enum_name::param_types()).unwrap())) });
+
+                //  Calling unwrap here is ok because of the following:
```
vnepveu, line 73: Very good but maybe a little too verbose haha, see if you can reduce

vnepveu (changes requested):
This is starting to look awesome!

packages/fuels-core/src/encoding_utils.rs
segfault-magnet: I went in another direction, this is now called:
`compute_encoding_width_of_enum`

and will include the discriminant size.

packages/fuels-core/src/abi_encoder.rs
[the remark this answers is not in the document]
segfault-magnet: it looks great. I'll update it.

packages/fuels-core/src/abi_decoder.rs
segfault-magnet: A cool trick.

Slices are nothing more than a ptr and a length

So I removed the '+offset' everywhere and just opted to move the slice further right along the data as we continue to decode.

The only thing that is now important is that each decoding step says how many bytes they've already consumed. 

Using that info the root function from which we started (decode) will construct an appropriate slice for the next decoding iteration.

This way we're always reading from the beginning and the offset is encoded in the type :D

packages/fuels-core/src/encoding_utils.rs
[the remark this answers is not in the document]
segfault-magnet: It's now called compute_encoding_width as per the other comment.

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -64,6 +79,10 @@ impl ABIEncoder {
         Ok(self.encoded_args.clone())
     }
 
+    fn zeropad_until_size(&mut self, final_size: usize) {
```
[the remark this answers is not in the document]
segfault-magnet: I've changed the signature, it now just takes the number of bytes you want to pad to the right.

It's called:

`rightpad_with_zeroes(amount)`

packages/fuels-core/src/encoding_utils.rs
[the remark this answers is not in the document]
segfault-magnet: Changed the name as per the other comment. Now it deals with enums and is called
`compute_encoding_width_of_enum`

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -424,6 +432,48 @@ mod tests {
         );
     }
 
+    #[test]
+    fn enum_padding_is_skipped() {
```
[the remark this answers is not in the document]
segfault-magnet: Changed the name

packages/fuels-core/src/abi_encoder.rs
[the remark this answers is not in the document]
segfault-magnet: Added docstring pointing to the specs.

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -548,8 +581,16 @@ mod tests {
         assert_eq!(abi_encoder.function_selector, expected_function_selector);
     }
 
+    // The encoding follows the ABI specs defined  [here](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md)
+    // An enum doesn't change its encoding width according to which variant it
+    // is currently holding. It is always as wide as its biggest variant. If we
+    // store a smaller variant in it, it will be left padded until it reaches
+    // the allotted enum size. Besides the value itself, an enum also stores a
+    // discriminant to indicate which variant it is currently holding.
```
vnepveu, line 589: I think this is a repetition of what's being said in some other place (or should be said some other place haah). So you can remove the above part, but enrich the comment below a little

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -548,8 +581,16 @@ mod tests {
...
     #[test]
     fn enums_are_sized_to_fit_the_biggest_variant() {
+        // Our enum has two variants: B256, and U64. So the enum will set aside
+        // 256b of space or 4 WORDS.
```
vnepveu, line 593: `So the enum will set aside 256B of space (4 WORDS) because it sets aside the storage necessary to hold the largest variant (here, B256)`  or smth like that

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -1,17 +1,19 @@
...
+        .map(compute_encoding_width)
         .max()
-        .expect("Will never panic because EnumVariants must have at least variant inside it!")
+        .map(|width| width + ENUM_DISCRIMINANT_WORD_WIDTH)
+        .expect("Will never panic because EnumVariants must have at least one variant inside it!")
 }
 
 /// Calculates the number of `WORD`s the VM expects this parameter to be encoded in.
-pub fn expected_encoding_width(param: &ParamType) -> usize {
+pub fn compute_encoding_width(param: &ParamType) -> usize {
```
vnepveu, line 16: `compute_parameter_type_encoding_width` would be a better name IMO, even if it's the argument, from the outside it would be clearer, and less likely to be confused with the one for enums

packages/fuels-abigen-macro/tests/test_projects/enums_with_unit_types/src/main.sw
```diff
@@ -0,0 +1,24 @@
+contract;
+
+enum OnlyUnitsInEnum {
+    a: (),
+    b: (),
+}
+
```
vnepveu, line 7: This test project should be merged with the `use_enum_input` test (there might be some necessary refactoring)

vnepveu (changes requested):
Some nits + one request about merging test projects

packages/fuels-core/src/abi_encoder.rs
vnepveu: `An enum is always as wide as its biggest variant` is the key sentence here, very well put -> should be in the below comment

packages/fuels-core/src/abi_decoder.rs
vnepveu: Smart! This should be documented somewhere.

packages/fuels-abigen-macro/tests/test_projects/enums_with_unit_types/src/main.sw
segfault-magnet: I've joined this one and the 'use_enum_input' into a new one that has more robust checks in place.

I've found that if you just return what you received as an arg you can have improperly encoded enums just returned to you and then you wouldn't test anything since you'd be in an echo chamber. This approach fixes that.

packages/fuels-core/src/encoding_utils.rs
segfault-magnet: Rust and your IDE won't allow you to mix these two up. Since the argument type is part of the fn signature, putting it in the function name is redundant.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1408,29 +1408,6 @@ async fn workflow_struct_inside_enum() {
     assert_eq!(result.value, 8888);
 }
 
-#[tokio::test]
-async fn workflow_use_enum_input() {
```
vnepveu, line 1412: you shouln't remove that test

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1477,32 +1454,6 @@ async fn test_logd_receipts() {
     assert!(result.logs.is_empty());
 }
 
-#[tokio::test]
```
vnepveu, line 1480: this one either

vnepveu (changes requested):
Fix CI. I don't understand where the `enum_input` Sway test project comes into play, and so I don't understand why the `unit_type_enum` and `use_enum_input` were removed.

segfault-magnet:
> Fix CI. I don't understand where the `enum_input` Sway test project comes into play, and so I don't understand why the `unit_type_enum` and `use_enum_input` were removed.

Hey @vnepveu , "use_enum_input" sway project was used by the "unit_type_enums" and "workflow_use_enum_input" harness tests. They didn't test whether the contract understood the enums as we did. And mostly didn't check whether we understood the enums as the contract did. Both are important for proper enum functioning.

Both those tests were merged into one as per suggestion, which is now called 'enum_as_input'.

The new e2e test checks that a standard enum, as well as a 'unit' one, can be passed into and received from a contract function -- and that nothing was mangled in the process.

Since these are e2e tests it is difficult to single out only one thing to test with them. For example:

The 'contract understood our enums as we did, and we understood its enums as it did' basically means that the coding/decoding of enums is working properly. There is already a test for that but it is specifically geared towards guarding against a regression -- i.e. it has extra parameters around the enum so that they might get mangled in the case of a regression in forc which we can detect.

The new 'use enum as input' test has a bit of overlap with the encoding test but it tests our handling of code generation as well -- that we can handle generating code for both standard and unit type enums when they appear as direct inputs/outputs to a function.

vnepveu:
My question specifically is "why is the test project still here if it's not used anywhere"

segfault-magnet:
> My question specifically is "why is the test project still here if it's not used anywhere"

Here is the list of all tests projects in my branch:
https://github.com/segfault-magnet/fuels-rs/tree/segfault_magnet/352-enum-encoding/packages/fuels-abigen-macro/tests/test_projects

It doesn't contain 'use_enum_input'. 

I've added 'enum_as_input' and 'enum_encoding' both of which are used in the harness.rs.

vnepveu:
It contains `enum_as_input`, got confused with all the renaming

vnepveu:
ok sry got confused.

vnepveu (dismissed):

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -1,252 +1,288 @@
...
+        let result = DecodeResult {
+            token: Token::String(decoded.to_string()),
+            bytes_read: padded_len(encoded_str),
+        };
+
+        Ok(result)
+    }
+
+    fn decode_b256(data: &[u8]) -> Result<DecodeResult, CodecError> {
+        let b256: Bits256 = peek(data, 32)?.try_into().unwrap();
```
digorithm, line 123: For the sake of consistency and proper error handling, what do you think about turning this `try_into().unwrap()` into a proper `try_into()?` converting the error to `CodecError`? 

I can imagine this being very annoying to debug if there's something fishy in `data` and we hit this `unwrap()`.

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -1,252 +1,288 @@
...
+            token: Token::B256(b256),
+            bytes_read: 32,
+        };
+
+        Ok(result)
+    }
+
+    fn decode_byte(data: &[u8]) -> Result<DecodeResult, CodecError> {
+        // Grab last byte of the word and compare it to 0x00
+        let byte = *peek_word(data)?.last().unwrap();
```
digorithm, line 135: Same comment about error handling here.

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -1,252 +1,288 @@
...
+            token: Token::Byte(byte),
+            bytes_read: 8,
+        };
+
+        Ok(result)
+    }
+
+    fn decode_bool(data: &[u8]) -> Result<DecodeResult, CodecError> {
+        // Grab last byte of the word and compare it to 0x00
+        let b = peek_word(data)?.last().unwrap() != &0u8;
```
digorithm, line 147: Same comment about error handling here.

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -1,252 +1,288 @@
...
+
+        Ok(DecodeResult {
+            token,
+            bytes_read: enum_width * WORD_SIZE,
+        })
+    }
+
+    fn decode_discriminant(data: &[u8]) -> Result<u32, CodecError> {
+        let discriminant = peek_word(data)?;
+        Ok(u32::from_be_bytes(discriminant[4..8].try_into().unwrap()))
```
digorithm, line 235: Same comment about error handling here.

packages/fuels-core/src/encoding_utils.rs
```diff
@@ -0,0 +1,109 @@
+use crate::constants::ENUM_DISCRIMINANT_WORD_WIDTH;
+use crate::{EnumVariants, ParamType, WORD_SIZE};
+
+// Calculates how many WORDs are needed to encode an enum
```
digorithm, line 4:
```suggestion
// Calculates how many WORDs are needed to encode an enum.
```

(I'm so sorry 😂 )

packages/fuels-core/src/abi_decoder.rs
segfault-magnet: Sure! This was just copied&pasted from the original impl.

packages/fuels-core/src/encoding_utils.rs
segfault-magnet: :laughing:

packages/fuels-core/src/abi_decoder.rs
digorithm: Yeah, that makes sense, then let's do `expect("something something. This should never happen")` so that at least we know where to look at if the underlying impl ever breaks.

segfault-magnet: There are no unwraps left in the production code in this file. Can you see if you're ok with the solution?

digorithm (approved):
Great work! :shipit:

PR #355, opened 2022-06-01
feat: introduce the option of using `fuel-core` binary instead of `fuel-core` library

Salka1988:
I made new functions that will use fuel-core.bin instead of fuel-core which is in dependencies.
I added a new Config class that I will use with the new functions to make the necessary JSON files and thus enable the same functionality we had with the fuel-core.

In order to use the fuel-core-bin it is necessary to set the evn. variable:

export FUEL_CORE_BIN = "<your absolute path to fuel-core.bin> fuel-core"

export FUEL_CORE_CONFIG = "<your absolute path to config.json> /config.json"

The next steps are:
Using the "cfg" flag (eg "use-fuel-core-bin") I will extract all methods, dependencies, tests, etc. that will be compiled / used only if the flag is entered as a parameter.

PR #363, opened 2022-06-06
fix(examples): add Config from fuel-core to prelude

Br1ght0ne:
Should fix some build problems on #355. Investigating why CI was green in the first place.
Re @Salka1988

### DIFF
--- a/packages/fuels/src/lib.rs
+++ b/packages/fuels/src/lib.rs
@@ -61,6 +61,7 @@ pub mod prelude {
     pub use super::core::parameters::*;
     pub use super::core::{Detokenize, InvalidOutputType};
     pub use super::core::{Token, Tokenizable};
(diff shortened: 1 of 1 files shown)


digorithm:
@Br1ght0ne Do you mind giving me a bit more context as to why `pub use super::node::service::Config;` is needed?

Br1ght0ne:
@digorithm Examples are using `Config::local_node()` and no other imports other than `fuels::prelude::*`. `Config` is unavailable from `fuels` directly, so those tests shouldn't have been green even when initially added.

The book uses `use fuels::node::service::{Config, FuelService}` when instantiating a client. If importing `Config` directly through this path is desirable, I can amend the example code.

digorithm (approved):
I see -- makes sense to me!

Issue #211, opened 2022-04-13
Introduce API for bundling many contract calls in a single transaction


digorithm:
Inspired by the problems discussed here https://github.com/FuelLabs/fuels-ts/issues/216. 

The SDK should have a convenient API for bundling many calls (i.e. appending `CALL` instructions to the `script`, etc.) into a single transaction. 

A use case can be found in the issue above.

PR #325, opened 2022-05-26
add script_runner method in fuels-test-helpers

Salka1988:
I added script_runner method in fuels-test-helpers so it can be used in SDK testing.

Issue #184, opened 2022-03-27
Actually forward gas on call


adlerjohn:
Actually forward the amount of gas on contract calls

m3talsmith:
@digorithm I'm happy to take this one on.

digorithm:
Awesome, go for it!

adlerjohn:
This issue is to forward an amount of gas with a call, not all remaining gas (which is the current behavior). To do this, you need to load the amount of gas into a register.

PR #364, opened 2022-06-06
Ahmedm/multi call refactor

### LINKED ISSUE #200: feat: SDK should offer an option to see the transaction before using `call` or `simulate` on it
*  We should build  a new struct that gets returned with a `.build()`, which includes the tx and also a method `.call()` which does the call on the tx that's already built.

MujkicA:
Closes #200

This PR is prep work for the multi call feature tracked in #211. 
It refactors the build process of the script that the SDK creates to execute contract calls. This is done by:
1. Renaming `ContractCall` to `ContractCallHandler` and creating a new `ContractCall` struct which holds all the data relevant to a contract call while `ContractCallHandler` is a wrapper to operate with it.
2. Moving script creation methods to `script.rs`
3. Splitting up script creation methods to separate concerns

The above changes satisfy the requirements of #200. An example workflow is provided in `harness.rs`.

### DIFF
--- a/packages/fuels-abigen-macro/tests/harness.rs
+++ b/packages/fuels-abigen-macro/tests/harness.rs
@@ -47,7 +47,7 @@ async fn compile_bindings_from_contract_file() {
     // Currently this prints `0000000003b568d4000000000000002a000000000000000a`
     // The encoded contract call. Soon it'll be able to perform the
     // actual call.
-    let contract_call = contract_instance.takes_ints_returns_bool(42);
+    let call_handler = contract_instance.takes_ints_returns_bool(42);
 
     // Then you'll be able to use `.call()` to actually call the contract with the
     // specified function:
@@ -56,8 +56,8 @@ async fn compile_bindings_from_contract_file() {
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!("000000009593586c000000000000002a", encoded);
@@ -95,12 +95,12 @@ async fn compile_bindings_from_inline_contract() {
     //`SimpleContract` is the name of the contract
     let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
-    let contract_call = contract_instance.takes_ints_returns_bool(42_u32);
+    let call_handler = contract_instance.takes_ints_returns_bool(42_u32);
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!("000000009593586c000000000000002a", encoded);
@@ -137,12 +137,12 @@ async fn compile_bindings_array_input() {
     let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let input: Vec<u16> = vec![1, 2, 3, 4];
-    let contract_call = contract_instance.takes_array(input);
+    let call_handler = contract_instance.takes_array(input);
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!(
@@ -182,12 +182,12 @@ async fn compile_bindings_bool_array_input() {
     let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
     let input: Vec<bool> = vec![true, false, true];
-    let contract_call = contract_instance.takes_array(input);
+    let call_handler = contract_instance.takes_array(input);
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!(
@@ -226,12 +226,12 @@ async fn compile_bindings_byte_input() {
     // `SimpleContract` is the name of the contract
     let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
-    let contract_call = contract_instance.takes_byte(10u8);
+    let call_handler = contract_instance.takes_byte(10u8);
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!("00000000a4bd3861000000000000000a", encoded);
@@ -267,12 +267,12 @@ async fn compile_bindings_string_input() {
     // `SimpleContract` is the name of the contract
     let contract_instance = SimpleContract::new(null_contract_id(), wallet);
 
-    let contract_call = contract_instance.takes_string("This is a full sentence".into());
+    let call_handler = contract_instance.takes_string("This is a full sentence".into());
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!(
@@ -316,12 +316,12 @@ async fn compile_bindings_b256_input() {
 
     let arg = hasher.finalize();
 
-    let contract_call = contract_instance.takes_b256(arg.into());
+    let call_handler = contract_instance.takes_b256(arg.into());
 
     let encoded = format!(
         "{}{}",
-        hex::encode(contract_call.encoded_selector),
-        hex::encode(contract_call.encoded_args)
+        hex::encode(call_handler.contract_call.encoded_selector),
+        hex::encode(call_handler.contract_call.encoded_args)
     );
 
     assert_eq!(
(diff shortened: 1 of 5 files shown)


packages/fuels-contract/src/contract.rs
```diff
@@ -487,35 +257,86 @@ impl Contract {
...
+pub struct ContractCallHandler<D> {
+    pub contract_call: ContractCall,
+    pub fuel_client: FuelClient,
+    pub datatype: PhantomData<D>,
+}
+
+impl<D> ContractCallHandler<D>
 where
-    D: Detokenize,
+    D: Detokenize + std::fmt::Debug,
```
vnepveu, line 331: See remark above

packages/fuels-contract/src/contract.rs
```diff
@@ -487,35 +257,86 @@ impl Contract {
...
 }
 
-impl<D> ContractCall<D>
+#[derive(Debug)]
+#[must_use = "contract calls do nothing unless you `call` them"]
+/// Helper for managing a transaction before submitting it to a node
+pub struct ContractCallHandler<D> {
+    pub contract_call: ContractCall,
+    pub fuel_client: FuelClient,
+    pub datatype: PhantomData<D>,
```
vnepveu, line 326: what is this `PhantomData` (I know it's not from your changes but it should have been documented)

packages/fuels-contract/src/contract.rs
```diff
@@ -563,30 +384,22 @@ where
...
+        let receipts = if simulate {
+            script.simulate(&self.fuel_client).await?
+        } else {
+            script.call(&self.fuel_client).await?
+        };
+        tracing::debug!(target: "receipts", "{:?}", receipts);
 
-        let (decoded_value, receipts) = Self::get_decoded_output(receipts, &self.output_params)?;
-        Ok(CallResponse::new(D::from_tokens(decoded_value)?, receipts))
+        self.build_response(receipts)
```
vnepveu, line 398: If the `build_response` function returns a response, the method should be called `get_response` or something like that. Otherwise it looks like a builder method.

packages/fuels-contract/src/script.rs
```diff
@@ -22,7 +35,213 @@ impl Script {
         Self { tx }
     }
 
-    // Calling the contract executes the transaction, and is thus state-modifying
+    pub async fn from_call(call: &ContractCall) -> Self {
```
vnepveu, line 38: docstring (even if the name is self-documenting, a summary of what the whole function does is in order here)

vnepveu, line 38: `from_contract_call`

packages/fuels-contract/src/script.rs
```diff
@@ -22,7 +35,213 @@ impl Script {
...
+        );
+        call.wallet.sign_transaction(&mut tx).await.unwrap();
+
+        Script::new(tx)
+    }
+
+    /// Given the necessary arguments, create a script that will be submitted to the node to call
+    /// the contract. The script is the actual opcodes used to call the contract, and the script
+    /// data is for instance the function selector. (script, script_data) is returned as a tuple
+    /// of hex-encoded value vectors
```
vnepveu, line 149: This docstring is out of date

packages/fuels-contract/src/script.rs
```diff
@@ -22,7 +35,213 @@ impl Script {
...
+        (instructions, offset)
+    }
+
+    /// The script data consists of the following items in the given order:
+    /// 1. Asset ID to be forwarded (AmountId::LEN)
+    /// 2. Amount to be forwarded (1 * WORD_SIZE)
+    /// 3. Contract ID (ContractID::LEN);
+    /// 4. Function selector (1 * WORD_SIZE);
+    /// 5. Calldata offset (optional) (1 * WORD_SIZE)
+    /// 6. Encoded arguments (optional) (variable length)
```
vnepveu, line 170: update dostring

packages/fuels-contract/src/script.rs
```diff
@@ -22,7 +35,213 @@ impl Script {
...
+    fn get_script_data(calls: Vec<&ContractCall>, offset: usize) -> Vec<u8> {
```
vnepveu, line 171: I think `get_script_data_from_calls` or something would be a better name.

packages/fuels-contract/src/script.rs
```diff
@@ -22,7 +35,213 @@ impl Script {
...
+        script_data
+    }
+
+    /// Returns the VM instructions for calling a contract method
+    /// We use the Opcode to call a contract: `CALL` pointing at the
+    /// following registers;
+    ///
+    /// 0x10 Script data offset
+    /// 0x11 Gas price
+    /// 0x12 Coin amount
```
vnepveu, line 208: I think you forgot to add a previous comment  about an issue tracking this (0x11 Gas price TODO: https://github.com/FuelLabs/fuels-rs/issues/184)

vnepveu (changes requested):
Great work!
- It would be good to document the new workflow both in the harness and *The Book* 
- I feel like at some place there is maybe too much refactoring going on, or maybe not enough "separation" between functions: the `Script` struct becomes a little too "powerful" maybe 
- Documentation could be improved.

Overall great refactoring process, will make the multicall implementation much easier!

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1477,3 +1479,34 @@ async fn test_wallet_balance_api() {
...
+
+    let call_handler = contract_instance.initialize_counter(42);
+
+    let script = Script::from_call(&call_handler.contract_call).await;
+    assert!(script.tx.is_script());
+
+    let receipts = script.call(client).await.unwrap();
+
+    let response = call_handler.build_response(receipts).unwrap();
+    assert_eq!(response.value, 42);
```
vnepveu, line 1511: This kind of workflow is worse from a user standpoint. We now need to write 3 lines just to get a response and receipts out.

vnepveu (changes requested):
I just realized that the impact on usability is pretty high. We should discuss how to keep both the usability (which is at the limit of what a user should do) and the necessary refactoring.

packages/fuels-abigen-macro/tests/harness.rs
MujkicA: The old workflow is retained, notice nothing is changed for the old tests except for calling it `call_handler`. This is optional to satisfy #200.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -979,7 +980,7 @@ async fn test_provider_launch_and_connect() {
 }
 
 #[tokio::test]
-async fn test_contract_calling_contract() {
+async fn test_call_handlering_contract() {
```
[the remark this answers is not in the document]
MujkicA: Oh, this happened because I used automatic match and replace, hence the `handlering`

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1477,3 +1479,34 @@ async fn test_wallet_balance_api() {
         );
     }
 }
+
+#[tokio::test]
+async fn test_transaction_script_workflow() {
```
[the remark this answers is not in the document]
MujkicA: The method introduced in #325 takes a script binary, initializes the transaction and performs the call. The new workflow introduces a way for the user to interact with the contents of the transaction before submitting the call.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -115,7 +115,7 @@ impl Abigen {
         } else {
             (
                 quote! {
-                    use fuels::contract::contract::{Contract, ContractCall};
+                    use fuels::contract::contract::{Contract, ContractCall, ContractCallHandler};
```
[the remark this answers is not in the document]
MujkicA: Good catch!

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: Oh ok, wasn't sure!

packages/fuels-contract/src/contract.rs
MujkicA: Rust complains if `D` is not used in a data field so it is just a placeholder to make the compiler happy. I think docs are unnecessary since the Rust docs cover the use of `PhantomData`

MujkicA:
@vnepveu good review. While updating some of the comments I figured it's better to move `wallet` and `tx_parameters` from `ContractCall` to `ContractCallHandler`. Otherwise these fields are duplicated when we have multi calls since they only apply per handler not per call. I would postpone the doc update until multi calls are merged and or we settle on approach for the docs overhaul.

packages/fuels-contract/src/contract.rs
```diff
@@ -319,16 +319,18 @@ impl ContractCall {
 
 #[derive(Debug)]
 #[must_use = "contract calls do nothing unless you `call` them"]
-/// Helper for managing a transaction before submitting it to a node
+/// Helper that handles submitting a call to a client and formatting the formatting the response
```
vnepveu, line 322: typo

packages/fuels-contract/src/script.rs
```diff
@@ -14,9 +14,19 @@ use fuel_gql_client::{
...
+use fuels_signers::{LocalWallet, Signer};
+
+#[derive(Default)]
+/// Specifies offsets of Opcode::CALL parameters stored in the script
+/// data from which they can be loaded into registers
+struct CallParamOffsets {
+    pub asset_id_offset: usize,
+    pub amount_offset: usize,
+    pub call_data_offset: usize,
+}
```
vnepveu, line 29: vry nice

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1560,6 +1559,46 @@ async fn test_wallet_balance_api() {
     }
```
vnepveu, line 1559: please rework the comments l45-60 as they are outdated, especially with your new additions

vnepveu, line 1559: change the function name l983

vnepveu (changes requested):
Some nits left, great work!

packages/fuels-abigen-macro/tests/harness.rs
MujkicA: I believe the ''clean code" approach is to not have these kinds of comments in code at all as they are not maintainable.
This information should be provided by the book (maybe is already).

vnepveu: Agreed!

vnepveu (approved):
LGTM!

vnepveu (approved):

Issue #326, opened 2022-05-26
Using an `Address` as a type inside a locally defined struct errors when passed into a function


Braqzen:
The following

```
contract;

use std::address::Address;

struct User {
    address: Address,
    weight: u64
}

abi MyContract {
    fn test_function(user: User) -> bool;
}

impl MyContract for Contract {
    fn test_function(user: User) -> bool {
        true
    }
}
```

produces
![1](https://user-images.githubusercontent.com/103777923/170499457-7824d257-20b3-4e78-809d-a1d43ccc0c5a.png)

The issue is the `User` struct contains an `Address`.
Defining a local struct that contains a boolean and replacing the `Address` for that local struct does not cause issues. This seems specific to the `Address` and not structs in general.

Note: Passing in `Address` as the parameter type instead of `User` works as expected.

When changing the code above make sure to run `forc clean` each time before you `forc test` otherwise issues sometimes do not pop up.

Voxelot:
Related to https://github.com/FuelLabs/fuels-rs/issues/323

PR #366, opened 2022-06-06
Add support for Sway-native types wrapped in custom types

digorithm:
Closes https://github.com/FuelLabs/fuels-rs/issues/326.

This PR adds support for Sway-native types (`ContractId`, `AssetId`, `Address`) inside custom types being passed as input/output in contract calls. Example:

```Rust
let instance = MyContract::new(id.to_string(), wallet.clone());

let user = User {
    weight: 10,
    address: Address::zeroed(),
};
let result = instance.wrapped_address(user).call().await.unwrap();

assert_eq!(result.value.address, Address::zeroed());
```


packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -273,14 +272,14 @@ pub fn expand_custom_enum(name: &str, prop: &Property) -> Result<TokenStream, Er
             #( #enum_variants ),*
         }
 
-        impl #enum_ident {
-            pub fn param_types() -> Vec<ParamType> {
+        impl Parameterize for #enum_ident {
```
vnepveu, line 275: Why is the `Parameterize` trait needed for general enums and structs?

vnepveu (changes requested):
Some nits and questions

packages/fuels-abigen-macro/tests/test_projects/sway_native_types/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+name = "sway_native_types"
+authors = ["Fuel Labs"]
```
MujkicA, line 3: This seems to be the standard in all other tomls
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

MujkicA (approved):

packages/fuels-core/src/code_gen/custom_types_gen.rs
vnepveu: Ah right, I had forgotten that we were implementing them outside of traits!

vnepveu (approved):
:shipit:

PR #368, opened 2022-06-07
Fix examples cargo TOML

digorithm:
Context: https://github.com/FuelLabs/fuels-rs/runs/6780831010?check_suite_focus=true

The CI is trying to publish one of the examples -- it shouldn't. Plus some extra metadata.

### DIFF
--- a/examples/contracts/Cargo.toml
+++ b/examples/contracts/Cargo.toml
@@ -1,7 +1,13 @@
 [package]
 name = "fuels-example-contracts"
-version = "0.1.0"
+version = "0.0.0"
 edition = "2021"
+publish = false
+authors = ["Fuel Labs <contact@fuel.sh>"]
+homepage = "https://fuel.network/"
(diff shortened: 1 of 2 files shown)


adlerjohn (approved):

vnepveu (approved):

vnepveu:
Wait @adlerjohn  when do we alphabetize `Cargo.toml` fields? Is it only in deps?

digorithm:
> Ideally fields should be formatted as per https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/cargo.md

> Sort key names alphabetically within each section, with the exception of the [package] section.

hah!

adlerjohn:
@digorithm well, specifically, immediately following:

> Put the [package] section at the top of the file; put the name and version keys in that order at the top of that section, followed by the remaining keys other than description in alphabetical order, followed by the description at the end of that section.

PR #372, opened 2022-06-07
Add parameterize trait

tjsharp1:
Need Parameterize trait now too.


digorithm (approved):
Ah yeah, nice catch!

tjsharp1:
> Ah yeah, nice catch!

Yeah, it kinda slipped through. I guess those changes were upstream.

Voxelot (approved):

Issue #359, opened 2022-06-02
Generate additional test for the Json ABI cli tool


hal3e:
Add additional tests to cover cases when a wrong value string is passed and check if the right error is returned.

For example, when encoding a struct with a values string of `"((42, true)"` we should check if an `InvalidData` error with the message `"struct value string has excess opening brackets"` is returned.

PR #373, opened 2022-06-08
feat: extend abi json cli tests to cover expected errors

hal3e:
This PR should close https://github.com/FuelLabs/fuels-rs/issues/359

I have added tests to cover expected tokenization errors for the built-in types, tuples, arrays and structs.

I have also added a new error for the `b256` type, when the hex encoding doesn't have the required length.

### DIFF
--- a/packages/fuels-core/src/json_abi.rs
+++ b/packages/fuels-core/src/json_abi.rs
@@ -1861,7 +1861,7 @@ mod tests {
     }
 
     #[test]
-    fn tokenize_errors_built_in_types() {
+    fn tokenize_uint_types_expected_error() {
         let abi = ABIParser::new();
 
         // We test only on U8 as it is the same error on all other unsigned int types
@@ -1874,6 +1874,11 @@ mod tests {
             "Parse integer error: invalid digit found in string",
             error_message
         );
+    }
+
+    #[test]
+    fn tokenize_bool_expected_error() {
+        let abi = ABIParser::new();
 
         let error_message = abi
             .tokenize(&ParamType::Bool, "True".to_string())
@@ -1884,6 +1889,11 @@ mod tests {
             "Parse boolean error: provided string was not `true` or `false`",
             error_message
         );
+    }
+
+    #[test]
+    fn tokenize_b256_invalid_length_expected_error() {
+        let abi = ABIParser::new();
 
         let value = "d57a9c46dfcc7f18207013e65b44e4cb4e2c2298f4ac457ba8f82743f31e90b".to_string();
         let error_message = abi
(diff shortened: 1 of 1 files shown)


packages/fuels-core/src/json_abi.rs
```diff
@@ -211,6 +211,11 @@ impl ABIParser {
             ParamType::Bool => Ok(Token::Bool(trimmed_value.parse::<bool>()?)),
             ParamType::Byte => Ok(Token::Byte(trimmed_value.parse::<u8>()?)),
             ParamType::B256 => {
+                if trimmed_value.len() != 64 {
```
leviathanbeak, line 214: Would putting 64 behind a `const`ant provide more cleaner and descriptive code?

hal3e: Thanks for the input @leviathanbeak. I added the constant.

packages/fuels-core/src/json_abi.rs
```diff
@@ -1848,4 +1855,264 @@ mod tests {
...
+                error_message
+            );
+
+            let error_message = abi
+                .tokenize_tuple("(0, [0,0,0], 0)", &tuple_params)
+                .unwrap_err()
+                .to_string();
+
+            assert_eq!(
+                "Invalid data: last tuple value is missing a matching param",
```
vnepveu, line 1982: `has more elements than declared parameter types` or something would make more sense, wdyt?

vnepveu (changes requested):
didn't realize it at the previous PR but this is a huge debugging QOL improvement! great work!

packages/fuels-core/src/json_abi.rs
hal3e: I agree that we should change it. However, I would not use the word _elements_ as it could be even more confusing. My proposal: we remove the word _last_ and use just `"Invalid data: tuple value is missing a matching parameter type"`

vnepveu: `missing a matching parameter` doesn't convey that basically the user has more parameter types than value, "matching" is too vague. `tuple value contains more elements than the parameter types provided` is good I think.

packages/fuels-core/src/json_abi.rs
```diff
@@ -1848,4 +1859,264 @@ mod tests {
         let expected_encode = "00000000e33a11ce0000000000000001000000000000002a";
         assert_eq!(encoded, expected_encode);
     }
+
+    #[test]
+    fn tokenize_errors_built_in_types() {
```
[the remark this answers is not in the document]
hal3e: @segfault-magnet thanks for your input. I have rearranged the test to adhere to the AAA pattern and I have changed the test names.

segfault-magnet (approved):

vnepveu (approved):
LGTM!

PR #379, opened 2022-06-09
fix: improve fn signature generation for tuples

digorithm:
This fixes the following case:

```Rust
fn tuple_with_b256(p: (b256, u8)) -> (b256, u8);
```

```Rust
let id = *ContractId::zeroed();
let my_b256_u8_tuple: ([u8; 32], u8) = (id, 10);

let response = instance
    .tuple_with_b256(my_b256_u8_tuple)
    .call()
    .await
    .unwrap();

assert_eq!(response.value, my_b256_u8_tuple);
```

Previously, the generated function looked like this:

```
 pub fn tuple_with_b256(&self, value: (b256, u8)) -> ContractCall<()>
 ```
 
 But `b256` isn't a Rust type, this needs to be properly expanded to `[32; u8]`. 

This PR improves how the components inside a tuple are being expanded, fixing this issue.

### DIFF
--- a/packages/fuels-abigen-macro/tests/harness.rs
+++ b/packages/fuels-abigen-macro/tests/harness.rs
@@ -1259,6 +1259,17 @@ async fn test_tuples() {
         .unwrap();
 
     assert_eq!(response.value, my_enum_tuple);
+
+    let id = *ContractId::zeroed();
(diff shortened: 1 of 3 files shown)


hal3e:
Looks good to me.

hal3e (approved):

PR #385, opened 2022-06-10
fix: check file type

### LINKED ISSUE #383: Check the file type in `load_sway_contract`
# Problem

Currently, `load_sway_contract` doesn't care whether you're passing a filepath to a compiled binary or something else:

```Rust
pub fn load_sway_contract(binary_filepath: &str) -> Result<CompiledContract> {
    Self::load_sway_contract_with_salt(binary_filepath, Salt::from([0u8; 32]))
}

pub fn load_sway_contract_with_salt(
    binary_filepath: &str,
    salt: Salt,
) -> Result<CompiledContract> {
    let bin = std::fs::read(binary_filepath)?;
    Ok(CompiledContract { raw: bin, salt })
}
```

This means that any file could be interpreted as a `Vec<u8>` by `let bin = std::fs::read(binary_filepath)?` and that `Vec<u8>` would be deployed, even if it's not a binary. 

A user could, for instance, accidentally pass in the JSON ABI file and no warnings/errors would pop up, just very weird behavior later on (this actually happened).

# Solution

A simple solution would be to make sure the base file in the `binary_filepath` is indeed a binary file.

matt-user:
closes #383


Br1ght0ne (dismissed):
Consider converting `binary_filepath` to `Path` with [`Path::new`](https://doc.rust-lang.org/std/path/struct.Path.html#method.new) and using [`Path::extension`](https://doc.rust-lang.org/std/path/struct.Path.html#method.extension).

EDIT: Also consider [`Result::is_err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.is_err) (if you don't want to check for a specific error) instead of `#[should_panic]` attribute.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -671,6 +672,38 @@ async fn type_safe_output_values() {
...
+    )
+    .await;
+    assert!(deploy_result.is_err());
+
+    // Should panic as we are passing in a JSON instead of BIN
+    let deploy_result = Contract::deploy_with_salt(
+        "tests/test_projects/contract_output_test/out/debug/contract_output_test-abi.json",
+        &wallet,
+        TxParameters::default(),
+        Salt::from([1u8; 32]),
```
adlerjohn, line 701: If you don't need a specific salt (looks like you don't), you could just use the default salt. cc @vnepveu @digorithm to confirm

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -671,6 +672,38 @@ async fn type_safe_output_values() {
     assert!(response.value.bar);
 }
 
+#[tokio::test]
+async fn panics_on_non_binary_file() {
+    // Generates the bindings from the an ABI definition inline.
+    // The generated bindings can be accessed through `SimpleContract`.
```
vnepveu, line 678: Remove those comments (they're duplication)

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -671,6 +672,38 @@ async fn type_safe_output_values() {
...
+    abigen!(
+        MyContract,
+        "packages/fuels-abigen-macro/tests/test_projects/contract_output_test/out/debug/contract_output_test-abi.json"
+    );
+
+    let wallet = launch_provider_and_get_single_wallet().await;
+
+    println!("here");
```
vnepveu, line 686:
```suggestion

```

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: Yes you can use the default, since the idea is to just test the different path, it doesn't matter that it's the same salt as the one used in `Contract::deploy`

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -671,6 +672,38 @@ async fn type_safe_output_values() {
...
+
+    // Should panic as we are passing in a JSON instead of BIN
+    let deploy_result = Contract::deploy_with_salt(
+        "tests/test_projects/contract_output_test/out/debug/contract_output_test-abi.json",
+        &wallet,
+        TxParameters::default(),
+        Salt::from([1u8; 32]),
+    )
+    .await;
+    assert!(deploy_result.is_err());
```
vnepveu, line 704: A better way to test is to use the `#[should-panic(xxx)]` proc macro. There are examples in the harness.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -671,6 +672,38 @@ async fn type_safe_output_values() {
     assert!(response.value.bar);
 }
 
+#[tokio::test]
+async fn panics_on_non_binary_file() {
```
vnepveu, line 676: Those tests are unit tests so you should probably put them in `contract.rs`

vnepveu (changes requested):
Very good! Nice Rust contribution :smile: Left some requests for change.

packages/fuels-abigen-macro/tests/harness.rs
matt-user: Should I put the tests inside of contract.rs itself?  Or should I put the tests inside of a tests/ folder similar to how fuels-abigen-macro is structured?

vnepveu: We generally put unit tests in the file containing the function they are unit-testing! So you can go ahead and put it directly in `contract.rs`.

packages/fuels-contract/tests/harness.rs
```diff
@@ -0,0 +1,44 @@
...
+        "tests/test_projects/contract_output_test/out/debug/contract_output_test-abi.json",
+        &wallet,
+        TxParameters::default(),
+    )
+    .await
+    .unwrap();
+}
+
+#[tokio::test]
+#[should_panic]
```
vnepveu, line 26: The `should_panic` tag should check for the _right_ kind of panic (ie the right error message)! So we are sure we are failing for the right reasons :smile_cat:

packages/fuels-contract/tests/harness.rs
```diff
@@ -0,0 +1,44 @@
+use fuels::prelude::{launch_provider_and_get_single_wallet, Contract, TxParameters};
```
vnepveu, line 1: You should put these tests in `contract.rs`. These are "unit tests" whereas we consider the "harness" naming to be more E2E tests.

packages/fuels-contract/Cargo.toml
```diff
@@ -27,3 +27,7 @@ strum_macros = "0.21"
 thiserror = { version = "1.0.26", default-features = false }
 tokio = "1.12"
 tracing = "0.1"
+
+[dev-dependencies]
+fuels = { path = "../fuels" }
```
vnepveu, line 32: try to not import `fuels` but rather just the `fuels-test-helper`, because I think it's creating circular deps. Even if it's not, It's better to use the smallest import set anyway.

vnepveu (changes requested):
Some changes left then lgtm

packages/fuels-contract/src/contract.rs
```diff
@@ -155,9 +156,12 @@ impl Contract {
         wallet: &LocalWallet,
         params: TxParameters,
     ) -> Result<ContractId, Error> {
-        let compiled_contract = Contract::load_sway_contract(binary_filepath).unwrap();
+        let compiled_contract = Contract::load_sway_contract(binary_filepath);
```
leviathanbeak, line 159: would this work and be cleaner?

```suggestion
        let compiled_contract = Contract::load_sway_contract(binary_filepath)?;
        Self::deploy_loaded(&(compiled_contract), wallet, params).await
```

packages/fuels-contract/src/contract.rs
```diff
@@ -167,10 +171,12 @@ impl Contract {
...
     ) -> Result<ContractId, Error> {
-        let compiled_contract =
-            Contract::load_sway_contract_with_salt(binary_filepath, salt).unwrap();
+        let compiled_contract = Contract::load_sway_contract_with_salt(binary_filepath, salt);
 
-        Self::deploy_loaded(&compiled_contract, wallet, params).await
+        match compiled_contract {
+            Ok(_) => Self::deploy_loaded(&(compiled_contract.unwrap()), wallet, params).await,
+            Err(e) => Err(e),
+        }
```
leviathanbeak, line 179: same as above

packages/fuels-contract/src/contract.rs
matt-user: That's much cleaner, thanks for the suggestion!

packages/fuels-contract/src/contract.rs
```diff
@@ -427,3 +429,40 @@ where
...
+            "tests/test_projects/contract_output_test/out/debug/contract_output_test-abi.json",
+            &wallet,
+            TxParameters::default(),
+        )
+        .await
+        .unwrap();
+    }
+
+    #[tokio::test]
+    #[should_panic(expected = "called `Result::unwrap()` on an `Err` value: InvalidData(\"json\")")]
```
vnepveu, line 454: nice

vnepveu (approved):
LGTM! Congrats on your (first?) Rust PR!

Issue #375, opened 2022-06-08
`Option` return type causes `abigen` macro to panic with `not implemented`


bitzoic:
The following contract will cause an error:

```
contract;

use std::{
    address::Address,
    constants::NATIVE_ASSET_ID,
    identity::Identity,
    option::Option,
};

abi MyContract {
    fn test_ouput() -> Option<Identity>;
}

impl MyContract for Contract {
    fn test_ouput() -> Option<Identity> {
        Option::Some(Identity::Address(~Address::from(NATIVE_ASSET_ID)))
    }
}

```

It is believed that the return Option type is causing this problem.
The harness file is the default file.
This is seen both in v0.14 and v0.15 SDK.

![optionError](https://user-images.githubusercontent.com/54727135/172711985-bc02e61f-96b7-454f-8edc-41efff20ebb3.png)

Update(@digorithm ):

This is due to the lack of support for nested enums in the SDK, see https://github.com/FuelLabs/fuels-rs/blob/53e4fa099804a92bcf382b5f5bb939cd02f8accb/packages/fuels-core/src/code_gen/custom_types_gen.rs#L194-L197. 

This needs to be properly implemented and timely addressed since it's now a blocker.

PR #389, opened 2022-06-13
feat: add support for nested enums

hal3e:
Should close https://github.com/FuelLabs/fuels-rs/issues/375

The function `pub fn expand_custom_enum(name: &str, prop: &Property) -> Result<TokenStream, Error>` can now handle nested enums. 
I have updated the corresponding unit test and I have added a new `test_project` in `fuels-abigen-macro` and tested the new behavior.

### DIFF
--- a/packages/fuels-abigen-macro/tests/harness.rs
+++ b/packages/fuels-abigen-macro/tests/harness.rs
@@ -1629,3 +1629,32 @@ async fn test_transaction_script_workflow() {
     let response = call_handler.get_response(receipts).unwrap();
     assert_eq!(response.value, 42);
 }
+
+#[tokio::test]
+async fn nested_enums() {
+    abigen!(
+        MyContract,
+        "packages/fuels-abigen-macro/tests/test_projects/nested_enums/out/debug/nested_enums-abi.json"
+    );
+
+    let wallet = launch_provider_and_get_single_wallet().await;
+
+    let id = Contract::deploy(
+        "tests/test_projects/nested_enums/out/debug/nested_enums.bin",
+        &wallet,
+        TxParameters::default(),
+    )
+    .await
+    .unwrap();
+
+    let instance = MyContract::new(id.to_string(), wallet.clone());
+    let expected = Option::Some(Identity::Address(Address::zeroed()));
+
+    let result = instance.some_address().call().await.unwrap();
+
+    assert_eq!(result.value, expected);
+
+    let result = instance.none().call().await.unwrap();
+
+    assert_eq!(result.value, Option::None());
+}
--- a/packages/fuels-core/src/code_gen/custom_types_gen.rs
+++ b/packages/fuels-core/src/code_gen/custom_types_gen.rs
@@ -192,8 +192,34 @@ pub fn expand_custom_enum(name: &str, prop: &Property) -> Result<TokenStream, Er
         match param_type {
             // Case where an enum takes another enum
             ParamType::Enum(_params) => {
-                // TODO: Support nested enums
-                unimplemented!()
+                let inner_enum_name =
+                    &extract_custom_type_name_from_abi_property(component, Some(CustomType::Enum))?
+                        .to_class_case();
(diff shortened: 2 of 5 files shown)


packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1629,3 +1629,32 @@ async fn test_transaction_script_workflow() {
     let response = call_handler.get_response(receipts).unwrap();
     assert_eq!(response.value, 42);
 }
+
+#[tokio::test]
+async fn nested_enums() {
```
vnepveu, line 1634: this test doesn't test for using nested enums as input, but only as output. To check for encoding, please take inspiration from the `enum_input` tests in PR #361 (it's actually sometimes not enough to just get them in then out).

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1629,3 +1629,32 @@ async fn test_transaction_script_workflow() {
...
+        &wallet,
+        TxParameters::default(),
+    )
+    .await
+    .unwrap();
+
+    let instance = MyContract::new(id.to_string(), wallet.clone());
+    let expected = Option::Some(Identity::Address(Address::zeroed()));
+
+    let result = instance.some_address().call().await.unwrap();
```
vnepveu, line 1653: use `?`

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1629,3 +1629,32 @@ async fn test_transaction_script_workflow() {
...
+
+    assert_eq!(result.value, expected);
+
+    let result = instance.none().call().await.unwrap();
```
vnepveu, line 1657: same

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -192,8 +192,34 @@ pub fn expand_custom_enum(name: &str, prop: &Property) -> Result<TokenStream, Er
         match param_type {
             // Case where an enum takes another enum
             ParamType::Enum(_params) => {
```
vnepveu, line 194: Unless I am mistaken, I don't think this supports deeply nested enums, but rather just depth=2. Can you please provide tests to show that it's supported, or implement the support for "arbitrarily" (n=3 is fine for tests) deep nested enums.

vnepveu (changes requested):
Very good! Just a question about nesting depth.

packages/fuels-abigen-macro/tests/harness.rs
hal3e: Thanks for the great review. I have updated both the unit and the harness tests. We check now for input and output of nested enums with depth 3, as was done in https://github.com/FuelLabs/fuels-rs/pull/361.

packages/fuels-abigen-macro/tests/harness.rs
hal3e: As far as I can tell, all test use this pattern. I would not like to change it only in my test. If we want to change it in all test, I would gladly do it and I suggest we make a new issue.

packages/fuels-core/src/code_gen/custom_types_gen.rs
hal3e: I have added the test.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1706,3 +1706,52 @@ async fn enum_as_input() {
...
+        "tests/test_projects/nested_enums/out/debug/nested_enums.bin",
+        &wallet,
+        TxParameters::default(),
+    )
+    .await
+    .unwrap();
+
+    let instance = MyContract::new(id.to_string(), wallet.clone());
+
+    let expected_enum = EnumLevel3::El2(EnumLevel2::El1(EnumLevel1::Num(42)));
```
vnepveu, line 1729: Very clear and simple, great!

vnepveu (approved):
This is great! LGTM

PR #392, opened 2022-06-14
fix: do not modify custom type names

digorithm:
# Problem

Suppose the following two structs in Sway:

```Rust
pub struct Campaign {
    author: Identity,
    asset: ContractId,
    beneficiary: Identity,
    claimed: bool,
    deadline: u64,
    state: u64, 
    target_amount: u64,
    total_pledge: u64,
}

pub struct Campaigns {
    active: [u64; 1],
    completed: [u64; 1],
}
```

Given the current state of the SDK, this would lead to this error:

image

That's because `Campaigns` was being turned into `Campaign` due to `Inflector`'s `.to_class_case()`, and then conflicting with the original `Campaign`. 

Fiddling with the casing chosen by the Sway user is bad and not wanted, whatever the Sway user decides (and the _compiler_ allows) should be what is generated by the SDK, otherwise, there will be a discrepancy between what the user chose and what the SDK is generating, which isn't good.

# Solution

This PR simply removes the mutation of casing altogether. No more `.to_class_case()`. A very few tests were updated to use proper casing and not to expect casing mutation at all. The other tests don't break with the removal of `.to_cass_case()` and cases where the user defines structs such as `Campaign` and `Campaigns` are expected to work.

### DIFF
--- a/packages/fuels-abigen-macro/tests/test_projects/enum_as_input/src/main.sw
+++ b/packages/fuels-abigen-macro/tests/test_projects/enum_as_input/src/main.sw
@@ -1,15 +1,15 @@
 contract;
 
 enum StandardEnum {
-    one: b256,
-    two: u32,
-    three: bool,
+    One: b256,
+    Two: u32,
+    Three: bool,
 }
 
 enum UnitEnum {
-    one: (),
-    two: (),
-    three: (),
+    One: (),
+    Two: (),
+    Three: (),
 }
 
 abi EnumTesting {
@@ -22,11 +22,11 @@ abi EnumTesting {
 
 impl EnumTesting for Contract {
     fn get_standard_enum() -> StandardEnum {
-        StandardEnum::two(12345)
+        StandardEnum::Two(12345)
     }
     fn check_standard_enum_integrity(arg: StandardEnum) -> bool {
         match arg {
-            StandardEnum::two(value) => {
+            StandardEnum::Two(value) => {
                 value == 12345u32
             },
             _ => {
@@ -36,11 +36,11 @@ impl EnumTesting for Contract {
     }
 
     fn get_unit_enum() -> UnitEnum {
-        UnitEnum::two()
+        UnitEnum::Two()
     }
     fn check_unit_enum_integrity(arg: UnitEnum) -> bool {
         match arg {
-            UnitEnum::two(_) => {
+            UnitEnum::Two(_) => {
                 true
             },
             _ => {
--- a/packages/fuels-core/src/code_gen/custom_types_gen.rs
+++ b/packages/fuels-core/src/code_gen/custom_types_gen.rs
@@ -13,8 +13,7 @@ use quote::quote;
 /// Transforms a custom type defined in [`Property`] into a [`TokenStream`]
 /// that represents that same type as a Rust-native struct.
 pub fn expand_custom_struct(prop: &Property) -> Result<TokenStream, Error> {
-    let struct_name = &extract_custom_type_name_from_abi_property(prop, Some(CustomType::Struct))?
-        .to_class_case();
+    let struct_name = &extract_custom_type_name_from_abi_property(prop, Some(CustomType::Struct))?;
     let struct_ident = ident(struct_name);
     let components = prop
         .components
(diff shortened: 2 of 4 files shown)


packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -741,32 +734,6 @@ pub fn hello_world(
...
-            name: String::from("unused"),
-            type_field: String::from("enum babies"),
-            components: None,
-        };
-        let enum_name = Some(&enum_prop);
-        let result = expand_input_param(&def, "unused", &enum_type, &enum_name);
-        // Notice the removed plural!
-        assert_eq!(result.unwrap().to_string(), "Baby");
-    }
-    #[test]
```
matt-user, line 769: Why did you remove the #[test] (macro?) from test_expand_input_param_struct_wrong_name()?

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -574,11 +568,11 @@ mod tests {
     fn test_expand_custom_struct_with_struct() {
         let p = Property {
             name: String::from("unused"),
-            type_field: String::from("struct cocktail"),
+            type_field: String::from("struct Cocktail"),
```
matt-user, line 571: Did you rename all of the type fields to match the expected ClassCase types for the tests, now that there is no longer a conversion to ClassCase?

packages/fuels-abigen-macro/tests/test_projects/enum_as_input/src/main.sw
```diff
@@ -1,15 +1,15 @@
...
+    Three: bool,
 }
 
 enum UnitEnum {
-    one: (),
-    two: (),
-    three: (),
+    One: (),
+    Two: (),
+    Three: (),
```
matt-user, line 12: Why rename these enum types?  Is it because the tests expect them to be in ClassCase?

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -741,32 +734,6 @@ pub fn hello_world(
         assert_eq!(result.unwrap().to_string(), ":: std :: vec :: Vec < u64 >");
     }
     #[test]
-    fn test_expand_input_param_custom_type() {
```
matt-user, line 744: Why delete the test case?  Couldn't you rename the variable types as you did in other places?

matt-user (commented):
Not so much a review... just curious about the code.

packages/fuels-core/src/code_gen/functions_gen.rs
digorithm: Uh... weird diff. It's still there, see whole file: https://github.com/FuelLabs/fuels-rs/blob/e07c117d2de67f66a1e67c3959b0d1834b1e6f3f/packages/fuels-core/src/code_gen/functions_gen.rs#L736

packages/fuels-abigen-macro/tests/test_projects/enum_as_input/src/main.sw
digorithm: It's mostly because I believe that that's the conventional way to name your enum variants in Sway. I'm simply matching them against what the test expects, yes. 

The whole point of this PR is that what you write in Sway is what you get on the SDK side.

packages/fuels-core/src/code_gen/functions_gen.rs
digorithm: Oh no, that was by mistake (had it deleted while debugging some stuff), thanks for catching that!

packages/fuels-core/src/code_gen/custom_types_gen.rs
digorithm: Yup, naming these structs just as you would when writing Sway. I could keep it as `cocktail` but then the expected should then be `cocktail` and not `Cocktail`, as we're not mutating it anymore.

vnepveu (approved):
Good catch! LGTM

Salka1988 (dismissed):

Br1ght0ne (dismissed):

MujkicA (approved):

MujkicA (approved):

Issue #230, opened 2022-04-27
fix: figure out why `Enum::new_from_tokens` receives `vec` with more than 1 element sometimes


vnepveu:
See #213

PR #397, opened 2022-06-15
Segfault magnet/230 enum tokenization

segfault-magnet:
closes: #230 
### The Issue

A minimum example that was added to the e2e tests:
```sway
contract;

pub struct SomeStruct {
    par1: u32,
}

pub struct AllStruct {
    some_struct: SomeStruct,
}

abi MyContract {
    fn get_struct() -> AllStruct;
    fn check_struct_integrity(arg: AllStruct) -> bool;
}

    fn get_struct() -> AllStruct {
        AllStruct {
            some_struct: SomeStruct {
                par1: 12345u32,
            },
        }
    }
    fn check_struct_integrity(arg: AllStruct) -> bool {
        arg.some_struct.par1 == 12345u32
    }
}
```

If you proceeded to call `get_struct` you'd get:
```
'Failed to run `new_from_tokens()` for custom SomeStruct struct (tokens have wrong order and/or wrong types): InvalidOutputType("Expected `u32`, got Struct([U32(12345)])")'
```

The original error from the issue can be reproduced as well but I decided to use u32 to make the minimal example simpler.

The token representation of `AllStruct` is as follows:
```
[Struct([Struct([U32(12345)])])]
```
When we try to instantiate AllStruct the process goes through
```rust
impl<T: Tokenizable> Detokenize for T {
    fn from_tokens(mut tokens: Vec<Token>) -> Result<Self, InvalidOutputType> {
        let token = match tokens.len() {
            0 => Token::Struct(vec![]),
            1 => tokens.remove(0),
            _ => Token::Struct(tokens),
        };

        Self::from_token(token)
    }
}
```

Which strips away the first `Struct` and forwards 
```
[Struct([U32(12345)])]
```

 to `::from_token` in `AllStruct`.
 
The problem is that the `::from_token` implementations are not suited to do any further stripping away of `Struct` wrappers and they just delegate what they receive to their subcomponents. Here the `Struct` token in whole is given to `U32::from_token` which causes an issue.

### About the solution

We have multiple traits and methods that are connected with tokenization. 

Tokenizable - which can tokenize and detokenize
Detokenize - which can just detokenize
Parameterize - which can also detokenize, and does some other stuff as well

I've removed `Detokenize` and the detokenizing logic from `Parameterize`.

The only thing needed for a type to be able to be tokenized or detokenized is to implement `Tokenizable`.

I've adapted the code generation to make the Tokenizable implementation self-sufficient without it delegating to `Parameterized`. 

I've also changed the interface to accept *a* Token rather than a slice of Tokens, as one type can *only* ever be constructed from one token. Having a slice that contained the right token at the start, and a bunch of other ones you don't need only confused the issue.

I've also laid most of the groundwork for #358 , since the 'ability' to have multiple function outputs was probably the reason why the interface accepted a slice of Tokens.

Although there was always a check that we either have 0 or 1 return arguments, it is now explicit through the usage of 'Option' rather than a `Vec`.

Almost all `.clone()`s have been eliminated thanks to the change in the interface -- the detokenization process now consumes the tokens given to it.


adlerjohn (commented):
Merge conflict

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -149,13 +149,9 @@ fn expand_fn_outputs(outputs: &[Property]) -> Result<TokenStream, Error> {
             }
         }
         // Recursively expand the outputs
```
digorithm, line 151: Maybe remove this comment now that we can finally not care about multiple return values 😂 
```suggestion
```

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -549,57 +525,6 @@ pub fn hello_world(
...
-
-        let some_struct = Property {
-            name: "unused".to_string(),
-            type_field: String::from("struct Carmaker"),
-            components: Some(two_empty_components),
-        };
-        let result = expand_fn_outputs(&[some_struct.clone(), some_struct]);
-        assert_eq!(result.unwrap().to_string(), "(Carmaker , Carmaker)")
-    }
-
```
digorithm, line 602: 🔥

packages/fuels-core/src/errors.rs
```diff
@@ -75,13 +74,25 @@ impl From<ParseError> for Error {
...
-    }
-}
 impl From<std::io::Error> for Error {
     fn from(err: std::io::Error) -> Error {
         Error::ContractCallError(err.to_string(), vec![])
     }
 }
+
+#[derive(Debug)]
+pub struct InstantiationError(pub String);
```
digorithm, line 84: Nice new error, made things much clearer on the error handling side of this process!

digorithm (changes requested):
I absolutely love what you did here! Feels good to stop supporting multiple return values. This is also a _major_ internal improvement over some original ideas that I brought from `ethers-rs`. 

I just left a micro nitpick. But other than that, LGTM!

digorithm (approved):
:shipit:

PR #399, opened 2022-06-16
Multi-call

### LINKED ISSUE #211: Introduce API for bundling many contract calls in a single transaction

MujkicA:
Closes #211

This PR introduces the `MultiContractCallHandler` for bundling contract calls into a single transaction. It provides a similar interface to `ContractCallHandler`.

```rust
let call_handler_1 = contract_instance.initialize_counter(42);
let call_handler_2 = contract_instance.get_array([42; 2].to_vec());

let mut multi_call_handler = MultiContractCallHandler::new(wallet.clone());

multi_call_handler
    .add_call(call_handler_1)
    .add_call(call_handler_2);

let (counter, array): (u64, Vec<u64>) = multi_call_handler.call().await.unwrap().values;
```
Alternatively, the response can be fetched like this if we want to have access to receipts and logs  in the `MultiCallResponse`:
```rust
let response = multi_call_handler.call::<(u64, Vec<u64>)>().await.unwrap();
```

The harness is extended with a test for the regular work flow and one for the script workflow (exposes script). Another test is added to `script.rs` for verifying that script data i built correctly.


packages/fuels-contract/src/contract.rs
```diff
@@ -410,17 +430,125 @@ where
...
+        Script::from_contract_calls(
+            self.contract_calls.as_ref().unwrap().iter().collect(),
+            &self.tx_parameters,
+            &self.wallet,
+        )
+        .await
+    }
+
+    /// Call contract methods on the node, in a state-modifying manner.
+    pub async fn call<D: Tokenizable + Debug>(&self) -> Result<MultiCallResponse<D>, Error> {
```
digorithm, line 501: It may look simple, but this line took us days of debugging and redesign to get right 😂

MujkicA: Also the trick with bundling the output tokens of all the separate calls into a `Token::Tupple` so that the generic `D` effectively covers multiple types to avoid the need for dynamic dispatch

docs/src/getting-started/basics.md
```diff
@@ -247,6 +247,29 @@ If you already have a deployed contract and want to call its methods using the S
...
+
+### Output values
+To get the output values of the bundled calls, you need to provide explicit type annotations when saving the result of `call()` or `simulate()` to a variable:
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:multi_call_values}}
+```
+You can also interact with the `CallResponse` by moving the type annotation to the invoked method:
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:multi_call_response}}
+```
```
adlerjohn, line 271:
````suggestion
With `ContractMultiCallHandler` you can execute multiple contract calls within a single transaction. To achieve this, you first prepare all the contract calls that you want to bundle:

```rust,ignore
{{#include ../../../examples/contracts/src/lib.rs:multi_call_prepare}}
```
At this point you can also choose to set call parameters, variable outputs or external contracts for every contract call, as long as you don't execute it with `call()` or `simulate()`.
Next, you provide the prepared calls to your `ContractMultiCallHandler` and optionally configure transaction parameters:

```rust,ignore
{{#include ../../../examples/contracts/src/lib.rs:multi_call_build}}
```
Note that any transaction parameters configured on separate contract calls are disregarded in favor of the parameters provided to `ContractMultiCallHandler`.

### Output values
To get the output values of the bundled calls, you need to provide explicit type annotations when saving the result of `call()` or `simulate()` to a variable:

```rust,ignore
{{#include ../../../examples/contracts/src/lib.rs:multi_call_values}}
```
You can also interact with the `CallResponse` by moving the type annotation to the invoked method:

```rust,ignore
{{#include ../../../examples/contracts/src/lib.rs:multi_call_response}}
```
````

**Review by adlerjohn — COMMENTED** [review]
nit

File: packages/fuels-contract/src/script.rs [hunk 3]
```diff
@@ -259,6 +178,84 @@ impl Script {
...
+    async fn get_transaction_inputs_outputs(
+        calls: Vec<&ContractCall>,
+        wallet: &LocalWallet,
+    ) -> (Vec<Input>, Vec<Output>) {
+        let mut inputs: Vec<Input> = vec![];
+        let mut outputs: Vec<Output> = vec![];
+
+        // Get all unique contract ids
+        let mut contract_ids = HashSet::new();
+        calls.iter().for_each(|call| {
```
segfault-magnet, line 192: A more functional approach, not sure how readable though:
```rust
  let contract_ids: HashSet<ContractId> = calls
            .iter()
            .flat_map(|call| {
                let mut ids = call
                    .external_contracts
                    .as_ref()
                    .cloned()
                    .unwrap_or_default();
                ids.insert(call.contract_id);
                ids
            })
            .collect();
```

Also you can do without the `clone()` in `call.external_contracts.clone()` by doing `&call.external_contracts` since .extend does its own cloning inside.

packages/fuels-contract/src/script.rs
```diff
@@ -259,6 +178,84 @@ impl Script {
...
+                contract_id,
+            );
+            inputs.push(self_contract_input);
+
+            let external_contract_output = Output::contract(idx as u8, zeroes, zeroes);
+            outputs.push(external_contract_output);
+        }
+
+        // Get all unique asset ids
+        let mut asset_ids = HashSet::new();
```
segfault-magnet, line 217: Also a more functional approach, if you like it:
```rust
let asset_ids: HashSet<AssetId> = calls
            .iter()
            .map(|call| call.call_parameters.asset_id)
            .chain(iter::once(AssetId::default()))
            .collect();
```

packages/fuels-contract/src/script.rs
```diff
@@ -259,6 +178,84 @@ impl Script {
...
+        asset_ids.insert(AssetId::default());
+        calls.iter().for_each(|call| {
+            asset_ids.insert(call.call_parameters.asset_id);
+        });
+
+        let mut spendables = vec![];
+        for asset_id in asset_ids.iter() {
```
segfault-magnet, line 224: I'd split this into two loops going over asset_ids.

segfault-magnet (commented):
Whenever I see somebody do something smart I wonder 'would I have thought of that?' 

Let's just say I've been doing a lot of wondering :D 

Great work.

packages/fuels-contract/src/contract.rs
```diff
@@ -410,17 +413,116 @@ where
...
+            self.contract_calls.as_ref().unwrap().iter().collect(),
+            &self.tx_parameters,
+            &self.wallet,
+        )
+        .await
+    }
+
+    /// Call contract methods on the node, in a state-modifying manner.
+    pub async fn call<D: Tokenizable + Debug>(&self) -> Result<CallResponse<D>, Error> {
+        Self::call_or_simulate(self, false).await
```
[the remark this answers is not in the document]
MujkicA: At first I liked this but after changing it, I got second thoughts. Boolean flags that alter behavior are usually a bad sign. But IMO, not duplicating the 3 lines is useful because of the debug log and the functional-metal-jacket obfuscates stuff.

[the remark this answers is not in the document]
segfault-magnet: I'm ok with being pragmatic just as long as we're making an informed decision :)  We can always revisit if we find every other fn starting to have a 'doThisIfEnabledThatIfDisabled' flag

Issue #343, opened 2022-05-28
style: use `response` instead of `result` as variable names


vnepveu:
`result` as variable name is confusing

PR #400, opened 2022-06-16
refactor: rename `result` variables to `response`

thdaraujo:
This PR closes https://github.com/FuelLabs/fuels-rs/issues/343 by changing variable names from `result` to `response` for contract calls to make the intent more clear.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -16,16 +16,16 @@ jobs:
       - name: Setup mdBook
         uses: peaceiris/actions-mdbook@v1
         with:
-          mdbook-version: '0.4.7'
+          mdbook-version: "0.4.15"
 
       - run: mdbook build docs
 
-      - name: Deploy latest
+      - name: Deploy master
         uses: peaceiris/actions-gh-pages@v3
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs/book
-          destination_dir: latest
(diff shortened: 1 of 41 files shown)


packages/fuels-abigen-macro/tests/test_projects/auth_testing_contract/src/main.sw
```diff
@@ -14,12 +14,12 @@ impl AuthTesting for Contract {
     }
 
     fn check_msg_sender(expected_id: Address) -> bool {
-        let result: Result<Identity, AuthError> = msg_sender();
+        let response: Result<Identity, AuthError> = msg_sender();
```
adlerjohn, line 17: Correct me if I'm wrong @vnepveu, but I think the original (poorly-documented) reasoning of the issue was to change "result" to "response" for contract calls, since it's over the network (local, but still).

This one here isn't a response, because there's no network request. It's actually a result.

digorithm: Yup, that's what I mentioned to @thdaraujo in other channels; this refactor should only be related to variables receiving the response from a network call. So, everything with a `.call()` or `.transfer()` in this case.

digorithm (dismissed):
LGTM -- thanks for the help @thdaraujo!

digorithm (approved):

adlerjohn (approved):

Issue #390, opened 2022-06-14
Enums with unit type variants are only 8 bytes wide


mohammadfawaz:
https://github.com/FuelLabs/sway/pull/1922 changes the memory layout of enums so that if all the variants are unit types, then the size of the enum is 8 bytes, i.e., only the tag needs to be propagated. This was done in conformance with the updated spec: https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md#sum-types-enums
The SDK should follow the same memory layout for both input and output values.

mohammadfawaz:
@segfault-magnet told me that he was still following the current behavior of the compiler which is using 16 bytes instead of 8 byte the case above. This is going to change in `v0.16.0` to follow the spec.

PR #409, opened 2022-06-17
Encoding of an enum with all unit variants (#390)

segfault-magnet:
closes: #390 
### The issue
The sway compiler, since the release of Forc 16.0.0, includes an optimization with regards to enums that only have unit types for variants.

Such enums are to be encoded in only one word that contains their discriminant.

This PR also contains the necessary fixes to our Sway files thanks to @digorithm 

### The solution 
The cleanest way of looking at this is to say that enums with only unit variants will not have their unit variant encoded at all. This is preferable to wording like 'units are 0B in enums, 1 B elsewhere'. Although the 'elsewhere' part can be debated since the SDK doesn't support using Units anywhere else. This phrasing was translated into code.

Although the changes are covered by unit tests, there is an e2e test that will safeguard us against regressions in the sway compiler -- the test will only pass if the aforementioned type of enums is encoded in one word.

I've also taken the liberty of refactoring ABIEncoder, since it was unnecessarily storing the encoded function selector while it was only necessary for it to compute it.

The flow of 'construct an encoder, give it data, get the encoding back' was wrapped in an associated function which can now be called `ABIEncoder::encode(...)` since this is the only way the encoder is used.

### DIFF
--- a/packages/fuels-core/src/abi_encoder.rs
+++ b/packages/fuels-core/src/abi_encoder.rs
@@ -75,11 +75,14 @@ impl ABIEncoder {
 
         self.encode_discriminant(discriminant);
 
-        let param_type = Self::type_of_chosen_variant(discriminant, variants)?;
-
-        self.add_enum_padding(variants, param_type);
-
-        self.encode(slice::from_ref(token_within_enum))?;
+        // The sway compiler has an optimization for enums which have only Units
+        // as variants -- such an enum is encoded only by encoding its
+        // discriminant.
+        if !variants.only_units_inside() {
+            let param_type = Self::type_of_chosen_variant(discriminant, variants)?;
+            self.add_enum_padding(variants, param_type);
+            self.encode(slice::from_ref(token_within_enum))?;
+        }
 
         Ok(())
     }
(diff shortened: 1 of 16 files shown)


adlerjohn (commented):
CI failing

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -2,98 +2,152 @@ use crate::constants::{ENUM_DISCRIMINANT_WORD_WIDTH, WORD_SIZE};
...
+        (&mut output[4..]).copy_from_slice(&result[..4]);
+        output
     }
 
-    pub fn new_with_fn_selector(signature: &[u8]) -> Self {
-        Self {
-            function_selector: Self::encode_function_selector(signature),
-            encoded_args: Vec::new(),
+    /// Encodes `Token`s in `args` following the ABI specs defined
+    /// [here](https://github.com/FuelLabs/fuel-specs/blob/master/specs/protocol/abi.md)
```
digorithm, line 28: Do you mind using the permalink here? i.e. https://github.com/FuelLabs/fuel-specs/blob/1be31f70c757d8390f74b9e1b3beb096620553eb/specs/protocol/abi.md

digorithm (changes requested):
Tiny nits, other than that LGTM. Nice refactor and test 🔥

Br1ght0ne (dismissed):
Other than what @digorithm said, LGTM 🎉

digorithm (approved):
:shipit:

PR #413, opened 2022-06-20
release: bump versions to 0.16.0

digorithm:


adlerjohn:
Is #404 breaking? If so would be nice to include it here just to avoid forcing the next release to be breaking.

digorithm:
It is breaking! I'll wait for it to be merged then.

adlerjohn (approved):

Issue #414, opened 2022-06-20
abigen! cannot find types of nested structs


matt-user:
main.sw
```
contract;

use std::contract_id::ContractId;

pub struct CallData {
    memory_address: MemoryAddress,
    num_coins_to_forward: u64,
    asset_id_of_coins_to_forward: ContractId,
    amount_of_gas_to_forward: u64,
}

struct MemoryAddress {
    contract_id: ContractId,
    function_selector: u64,
    function_data: u64,
}

abi NestedStructs {
    fn dummy_func(call_data: CallData) -> bool;
}

impl NestedStructs for Contract {
    fn dummy_func(call_data: CallData) -> bool {
        true
    }
}
```
harness.rs
```
use fuels::{prelude::*, tx::ContractId};
use fuels_abigen_macro::abigen;

// Load abi from json
abigen!(NestedStructs, "out/debug/nested_structs-abi.json");

async fn get_contract_instance() -> (NestedStructs, ContractId) {
    // Launch a local network and deploy the contract
    let wallet = launch_provider_and_get_single_wallet().await;

    let id = Contract::deploy("./out/debug/nested_structs.bin", &wallet, TxParameters::default())
        .await
        .unwrap();

    let instance = NestedStructs::new(id.to_string(), wallet);

    (instance, id)
}

#[tokio::test]
async fn can_get_contract_id() {
    let (_instance, _id) = get_contract_instance().await;

    // Now you have an instance of your contract you can use to test each function
    let mem_address = nestedstructs_mod::MemoryAddress {
        contract_id: 0,
        function_selector: 0,
        function_data: (500, 0),
    };

    let call_data = nestedstructs_mod::CallData {
        memory_address: mem_address,
        num_coins_to_forward: 0,
        asset_id_of_coins_to_forward: _id,
        amount_of_gas_to_forward: 20000,
    };
}
```

failed to resolve: use of undeclared type `MemoryAddress`
use of undeclared type `MemoryAddress`rustc[E0433](https://doc.rust-lang.org/error-index.html#E0433)
cannot find type `MemoryAddress` in this scope
not found in this scope

digorithm:
Thanks for finding this and for the easy minimal repro code @matt-user. Pretty silly bug with a very simple fix!

PR #416, opened 2022-06-21
fix: custom types containing reserved keyword

digorithm:
Fixes https://github.com/FuelLabs/fuels-rs/issues/414.

Custom types in the JSON ABI containing the strings `Address` or `ContractID` were being ignored and not generated. That's because we don't re-generate Sway-native types, but the check for it was buggy: we checked if it contained, not if it's an exact match.

For instance, `struct Address` in your JSON ABI would be correctly ignored. `struct MemoryAddress` would be _incorrectly_ ignored because it contains `Address` in it. 

This PR changes the check logic and fixes this issue.

### DIFF
--- a/packages/fuels-abigen-macro/tests/test_projects/nested_structs/src/main.sw
+++ b/packages/fuels-abigen-macro/tests/test_projects/nested_structs/src/main.sw
@@ -1,5 +1,7 @@
 contract;
 
+use std::contract_id::ContractId;
+
 pub struct SomeStruct {
     par1: u32,
 }
@@ -8,9 +10,23 @@ pub struct AllStruct {
     some_struct: SomeStruct,
 }
 
+pub struct CallData {
+    memory_address: MemoryAddress,
+    num_coins_to_forward: u64,
+    asset_id_of_coins_to_forward: ContractId,
+    amount_of_gas_to_forward: u64,
+}
+
+struct MemoryAddress {
+    contract_id: ContractId,
+    function_selector: u64,
+    function_data: u64,
+}
+
 abi MyContract {
     fn get_struct() -> AllStruct;
     fn check_struct_integrity(arg: AllStruct) -> bool;
+    fn nested_struct_with_reserved_keyword_substring(call_data: CallData) -> CallData;
 }
 
 impl MyContract for Contract {
@@ -23,5 +39,9 @@ impl MyContract for Contract {
     }
     fn check_struct_integrity(arg: AllStruct) -> bool {
         arg.some_struct.par1 == 12345u32
+    }
+
+    fn nested_struct_with_reserved_keyword_substring(call_data: CallData) -> CallData {
+        call_data
     }
 }
--- a/packages/fuels-core/src/code_gen/abigen.rs
+++ b/packages/fuels-core/src/code_gen/abigen.rs
@@ -5,6 +5,7 @@ use crate::code_gen::custom_types_gen::{
     expand_custom_enum, expand_custom_struct, extract_custom_type_name_from_abi_property,
 };
 use crate::code_gen::functions_gen::expand_function;
+use crate::constants::{ADDRESS_SWAY_NATIVE_TYPE, CONTRACT_ID_SWAY_NATIVE_TYPE};
 use crate::errors::Error;
 use crate::json_abi::ABIParser;
 use crate::source::Source;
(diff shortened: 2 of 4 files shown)


vnepveu (approved):
LGTM, super fast :shipit:

packages/fuels-abigen-macro/tests/test_projects/nested_structs/src/main.sw
```diff
@@ -8,9 +10,23 @@ pub struct AllStruct {
...
 
+pub struct CallData {
+    memory_address: MemoryAddress,
+    num_coins_to_forward: u64,
+    asset_id_of_coins_to_forward: ContractId,
+    amount_of_gas_to_forward: u64,
+}
+
+struct MemoryAddress {
+    contract_id: ContractId,
```
matt-user, line 21: How would is_sway_native_type handle an Identity here?  Is the ABI able to understand that is it either an Address or a ContractId?

matt-user (commented):
Looks good! Just have a question.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -198,6 +199,23 @@ impl Abigen {
...
     }
 
+    // Checks whether the given type field is a Sway-native type.
+    // It's expected to come in as `"struct T"` or `"enum T"`.
+    // `T` is a Sway-native type if it matches exactly one of
+    // the reserved strings, such as "Address" or "ContractId".
+    fn is_sway_native_type(type_field: &str) -> bool {
+        let split: Vec<&str> = type_field.split_whitespace().collect();
+
+        if split.len() > 2 {
```
segfault-magnet, line 209: Another way of writing this would have been:
```rust
match split.as_slice() {
   [_, CONTRACT_ID_SWAY_NATIVE_TYPE] | [_, ADDRESS_SWAY_NATIVE_TYPE] => true,
   _ => false
}
```

packages/fuels-abigen-macro/tests/test_projects/nested_structs/src/main.sw
digorithm: No, IIRC `Identity` was recently added, not sure how it's behaving now. Do you mind testing it out? If it fails similarly as it did with `Address` before, I can quickly add it to the list of reserved keyword 😄

matt-user:
```
contract;

use std::identity::Identity;

pub struct CallData {
    memory_address: MemoryIdentity,
    num_coins_to_forward: u64,
    asset_id_of_coins_to_forward: Identity,
    amount_of_gas_to_forward: u64,
}

struct MemoryIdentity {
    contract_id: Identity,
    function_selector: u64,
    function_data: u64,
}

abi NestedStructs {
    fn dummy_func(call_data: CallData) -> bool;
    fn nested_struct_with_reserved_keyword_substring(call_data: CallData) -> CallData;
}

impl NestedStructs for Contract {
    fn dummy_func(call_data: CallData) -> bool {
        true
    }
    fn nested_struct_with_reserved_keyword_substring(call_data: CallData) -> CallData {
        call_data
    }
}
```
and
```
use fuels::{prelude::*, tx::ContractId};

// Load abi from json
abigen!(NestedStructs, "out/debug/nested_structs-abi.json");

async fn get_contract_instance() -> (NestedStructs, ContractId) {
    // Launch a local network and deploy the contract
    let wallet = launch_provider_and_get_single_wallet().await;

    let id = Contract::deploy("./out/debug/nested_structs.bin", &wallet, TxParameters::default())
        .await
        .unwrap();

    let instance = NestedStructs::new(id.to_string(), wallet);

    (instance, id)
}

#[tokio::test]
async fn can_get_contract_id() {
    let (_instance, _id) = get_contract_instance().await;

    // Now you have an instance of your contract you can use to test each function
    let mem_address = nestedstructs_mod::MemoryIdentity {
        contract_id: nestedstructs_mod::Identity::ContractId(ContractId::zeroed()),
        function_selector: 0,
        function_data: 0,
    };

    let call_data = nestedstructs_mod::CallData {
        memory_address: mem_address,
        num_coins_to_forward: 0,
        asset_id_of_coins_to_forward: nestedstructs_mod::Identity::ContractId(_id),
        amount_of_gas_to_forward: 20000,
    };

    let actual = _instance
        .nested_struct_with_reserved_keyword_substring(call_data.clone())
        .call()
        .await
        .unwrap()
        .value;

    assert_eq!(actual, call_data);
}
```
Work!

PR #419, opened 2022-06-22
feat!: add `transaction_parameters` to config json

Salka1988:
Replaced hardcoded ``transaction_parameters`` with ConsensusParameters::DEFAULT in config json.


packages/fuels-test-helpers/src/lib.rs
```diff
@@ -188,19 +190,26 @@ pub async fn setup_test_client(
     let config_with_coins: Value =
         serde_json::from_str(result.as_str()).expect("Failed to build config_with_coins JSON");
 
+    let transaction_parameters = serde_json::to_value(ConsensusParameters::DEFAULT).expect("Failed to build transaction_parameters JSON");
```
segfault-magnet, line 193: Maybe we should limit the proliferation of serde? Keep the data in domain types until the last moment when you actually need them to be serialized.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -159,6 +160,7 @@ pub async fn setup_test_client(
     let srv = FuelService::new_node(config).await.unwrap();
     let client = FuelClient::from(srv.bound_address);
 
+
```
MujkicA, line 163:
```suggestion
```

packages/fuels-test-helpers/src/lib.rs
MujkicA: Also, maybe `transaction_parameters` is not ideal as a variable name in this context since `ConsensusParameters` could be confused with `TxParameters`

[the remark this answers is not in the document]
Voxelot: Yes, default params are just another form of hard coding and don't respect what the node may actually be using.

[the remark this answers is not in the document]
Voxelot: Looking closer at the context here, it seems like this is just setting the default params for the test node used in the harness? Since we're spawning a new node here, using default params for tests seems fine. Although I think the other comments from @segfault-magnet about using the actual ConsensusParams type instead of JSON is important though.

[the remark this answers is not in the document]
digorithm: That's true, this is in `setup_test_client()`. So it's 100% fine to use the defaults in this case.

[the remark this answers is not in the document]
Salka1988: Now we have ``consensus_parameters_config: Option<ConsensusParameters>`` as parameter when starting node so we can configure it as we want.

digorithm (changes requested):
Many JSON ABI files and their respective binaries were committed; they shouldn't be committed.

Also, some context for other people: although the title and description PR isn't very descriptive, this PR is also moving the test harness from under `fuels-abigen-macro` to the root crate: `fuels`. This is something we've wanted to do, but due to a bunch of other limitations, we couldn't. `fuels-abigen-macro` not being part of `fuels` and having to be explicitly imported was one of these limitations.

Voxelot:
> Also, some context for other people: although the title and description PR isn't very descriptive, this PR is also moving the test harness from under fuels-abigen-macro to the root crate: fuels. This is something we've wanted to do, but due to a bunch of other limitations, we couldn't. fuels-abigen-macro not being part of fuels and having to be explicitly imported was one of these limitations.

seems like this should be split into a separate PR?

packages/fuels-test-helpers/src/node.rs
```diff
@@ -171,7 +173,35 @@ impl<'de> DeserializeAs<'de, BlockHeight> for HexNumber {
...
+    let result = serde_json::to_string(&coin_configs).expect("Failed to stringify coins vector");
+
+    let coins: Value =
+        serde_json::from_str(result.as_str()).expect("Failed to build config_with_coins JSON");
+
+    let consensus_parameters =
+        serde_json::to_value(consensus_parameters_config.unwrap_or_default())
+            .expect("Failed to build transaction_parameters JSON");
+
     let config = json!({
```
Voxelot, line 205: Seems like we should maybe move fuel-core's `ChainConfig` into the interfaces crate so you could avoid this manual json serialization without taking a dependency on fuel-core.

Voxelot:
Is there a plan to make a more user-friendly auto-magic way to construct TxParameters using the new chain-info API?

digorithm:
> > Also, some context for other people: although the title and description PR isn't very descriptive, this PR is also moving the test harness from under fuels-abigen-macro to the root crate: fuels. This is something we've wanted to do, but due to a bunch of other limitations, we couldn't. fuels-abigen-macro not being part of fuels and having to be explicitly imported was one of these limitations.
> 
> seems like this should be split into a separate PR?

Yes, I agree. It should have -- I guess things got really entangled in this one. We should be more cognizant of keeping the scope of PRs small and concise.

digorithm:
> Is there a plan to make a more user-friendly auto-magic way to construct TxParameters using the new chain-info API?

Not yet, what would that look like? We can file an issue and put it on the roadmap very soon 😁.

packages/fuels-test-helpers/src/node.rs
digorithm: That'd be nice, my only concern is increasing operational overhead by introducing another small-ish repo.

Voxelot:
> Not yet, what would that look like? We can file an issue and put it on the roadmap very soon grin.

I believe @luizstacio already implemented most of this behavior on the typescript SDK, including gas limit estimation, gas pricing, etc.

packages/fuels-contract/src/contract.rs
```diff
@@ -181,9 +181,19 @@ impl Contract {
...
+            .get_provider()
+            .expect("Can't get wallet provider?")
+            .client;
+
+        let chan_info = client.chain_info().await?;
+
         wallet.sign_transaction(&mut tx).await?;
 
-        match wallet.get_provider().unwrap().client.submit(&tx).await {
+        let _ = tx.validate_without_signature(0, &chan_info.consensus_parameters.into());
```
MujkicA, line 194: We should open a new Issue to use the actual block height here once we can query it on the SDK side

MujkicA, line 194: The returned `Result` is ignored, errors should be propagated

packages/fuels-test-helpers/src/script.rs
```diff
@@ -34,6 +41,12 @@ pub async fn run_compiled_script(binary_filepath: &str) -> Result<Vec<Receipt>,
 
     let script = get_script(script_binary);
 
+    let chan_info = client.chain_info().await?;
+
+    let _ = script
```
MujkicA, line 46: Same as above

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -94,15 +101,19 @@ pub async fn launch_custom_provider_and_get_wallets(
 pub async fn setup_test_provider(
```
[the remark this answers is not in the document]
Salka1988: done!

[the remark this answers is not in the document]
digorithm: Agreed. I really dislike having to slap a `None` everywhere, plus it adds very little context on what it means.

packages/fuels-contract/src/contract.rs
```diff
@@ -182,9 +182,19 @@ impl Contract {
     ) -> Result<ContractId, Error> {
         let (mut tx, contract_id) =
             Self::contract_deployment_transaction(compiled_contract, wallet, params).await?;
+
+        let client = &wallet
+            .get_provider()
+            .expect("Can't get wallet provider?")
```
digorithm, line 188: `Can't get wallet provider?` I don't know, you tell me 😂
```suggestion
            .expect("Can't get wallet provider")
```

packages/fuels-contract/src/script.rs
```diff
@@ -280,6 +280,10 @@ impl Script {
 
     /// Execute the transaction in a state-modifying manner.
     pub async fn call(self, fuel_client: &FuelClient) -> Result<Vec<Receipt>, Error> {
+        let chan_info = fuel_client.chain_info().await?;
```
digorithm, line 283: Making an additional network trip by calling `.chain_info()`) at every contract call is very wasteful; We're doubling the number of trips per call! 

It's very unlikely that the `chain_info` will change in a short period of time. For instance, when calling `do_something()` `n` times in a row, with some processing time in between them.

I'm not suggesting caching at this layer though -- that'd be overkill. But here's a simple potential solution that's easy to implement: 

1. Extend `Script.call(self, fuel_client: &FuelClient)` to take in the chain_info => `Script.call(self, fuel_client: &FuelClient, chain_info: ChainInfo)`;
2. At the `ContractCallHandler` level, extend it (the struct itself) to include a `pub chain_info: ChainInfo`
3. In `Contract::method_hash()` , call `.chain_info()` to get the chain info just at the time of contract handler creation, then pass it to the contract handler creation part (` Ok(ContractCallHandler {...`) because now `ContractCallHandler` holds a `ChainInfo`;
4. At the call level, no more network trips necessary will happen. `chain_info` checking will happen only at contract call handler creation, which isn't very often, but often enough to possibly catch any change in the chain config.

digorithm (changes requested):

packages/fuels-contract/src/script.rs
Voxelot: Would it be helpful if we made a separate API query just for fetching the consensus params? Since chain info includes a lot of other transient details such as latest block and peer count.

digorithm: It would, but I wouldn't make it a top priority or anything like that. Definitely a nice to have though! 

My suggestion above is just to avoid extra network latency/trips when you're making sequential calls to the same methods 😅

packages/fuels-contract/src/contract.rs
Salka1988: :smile:

packages/fuels-contract/src/script.rs
Salka1988: If I implement as you stated, then I have to set the ``method_hash`` function to be ``async``. This further means that this will also have to be ``async``: 

``
     Ok (quote! {
         # doc
         pub fn #name(&self #input) -> #result {
             Contract::method_hash(&self.wallet.get_provider().expect("Provider not set up"), self.contract_id, &self.wallet,
                 #tokenized_signature, #output_param, #arg).expect("method not found (this should never happen)")
         }
     })
``

If I turn this function into an ``async`` function, then every contract function will become ``async``, which requires that ``await`` be called on each one.

I implemented this :

`` let consensus_parameters: ConsensusParameters = 
   wallet.get_provider()?.client.chain_info().await?.consensus_parameters.into();

   let instance = MyContract::new(id.to_string(), wallet.clone(), consensus_parameters);
``

If you think this is ok, I will push branch.

digorithm: Turning `method_hash` into async would be an inconvenience, yes. But I believe you can use `Tokio`'s `block_on` around `chain_info().await` to block it, removing the need to turn the caller function into async. See examples in https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on

Salka1988: :partying_face: I will read this and implement it.

Salka1988:
Ill put change this PR to draft cause it need new fuel-core release to be functional. We'll need a fuel-core patch release with the latest hotfix from fuel-vm.

Voxelot:
@Salka1988 should be unblocked now

vnepveu (changes requested):
Remove the `sway_native_types` gitignore file

vnepveu (approved):
:shipit: Nice!

digorithm (approved):
Nice!

PR #425, opened 2022-06-25
add fix 

Salka1988:


packages/fuels-test-helpers/src/lib.rs
```diff
@@ -188,11 +188,10 @@ pub async fn setup_test_client(
...
 
-    let srv_address = if let Some(node_config) = node_config {
-        node_config.addr
+    let srv_address = if node_config.is_some() && node_config.as_ref().unwrap().addr.port() != 0 {
+        node_config.unwrap().addr
     } else {
-        let free_port = pick_unused_port().expect("No ports free");
-        SocketAddr::new("127.0.0.1".parse().unwrap(), free_port)
+        get_socket_address()
     };
```
leviathanbeak, line 195: if-let chains are not yet supported but another approach would be 

```rust
let srv_address = match node_config {
        Some(config) if config.addr.port() != 0 => config.addr,
        _ => get_socket_address(),
    };
```

pattern matching ftw

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -188,11 +188,9 @@ pub async fn setup_test_client(
     let config_with_coins: Value =
         serde_json::from_str(result.as_str()).expect("Failed to build config_with_coins JSON");
 
-    let srv_address = if let Some(node_config) = node_config {
-        node_config.addr
-    } else {
-        let free_port = pick_unused_port().expect("No ports free");
-        SocketAddr::new("127.0.0.1".parse().unwrap(), free_port)
+    let srv_address = match node_config {
+        Some(config) if config.addr.port() != 0 => config.addr,
```
hal3e, line 192: What would happen if the `config.addr.port` is not 0 but is already used ?

Salka1988: I think the user should handle that. He will get a message ``Error: Address already in use``. @digorithm what do you think?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -188,11 +188,10 @@ pub async fn setup_test_client(
...
         serde_json::from_str(result.as_str()).expect("Failed to build config_with_coins JSON");
 
-    let srv_address = if let Some(node_config) = node_config {
-        node_config.addr
-    } else {
-        let free_port = pick_unused_port().expect("No ports free");
-        SocketAddr::new("127.0.0.1".parse().unwrap(), free_port)
+    let srv_address = match node_config {
+        Some(config) if config.addr.port() != 0 && is_free(config.addr.port()) => config.addr,
+        Some(config) if !is_free(config.addr.port()) => panic!("Error: Address already in use"),
```
leviathanbeak, line 193: not sure why would this be useful since this would panic fail anyway if the port was already taken

packages/fuels-test-helpers/src/lib.rs
Salka1988: @hal3e  and I analyzed this situation "What would happen if the config.addr.port is not 0 but is already used ?". Due to the way we run nodes using fuel-core.bin we can't catch `` Error: Address already in use`` coming from the server. The error would only be printed on the terminal but did not affect the test flow. Therefore, I added another condition that solved that bug.

packages/fuels-test-helpers/src/lib.rs
Salka1988: https://github.com/FuelLabs/fuels-rs/pull/425#discussion_r906842007

digorithm (approved):
LGTM!

Issue #410, opened 2022-06-17
Initial version for an API to configure block height of a transaction


digorithm:
# Problem

Recently, some people expressed the need for a way to configure the target block height for a transaction, i.e., configuring _when_ a transaction should be executed. 

In Ethereum/Solidity, one can do something like this:

```js
contract.methodName({ blockNumber: blockNumber }).then((result) => {
    console.log(result);
});
```

In order to execute the call at `blockNumber`. 

In addition to setting the block height for a transaction, we would also like a way to quickly get the current block height, which is currently but you have to bypass the SDK's API and use `fuel-core`'s `Client` directly. This could be simplified by creating a shallow wrapper around the `Client`'s block API in the SDK's `Provider` struct.

So, three APIs would be nice to have:

- [ ] Get the current block height using the SDK's API
- [ ] `tx.set_block_height(block_height: u64)` to set the target block height for the transaction `tx`
- [ ] `add_to_block_height(blocks_to_add: u64)` force `blocks_to_add` blocks in front of the current block height

# Blockers

Given the current state of the client (`fuel-core`), it's not possible yet to _properly_ do the setting of block height in a transaction. For this to be possible, it needs the txpool to be fully released. This will enable transactions to sit in the txpool queue until they can be included at a specific block. But currently, it doesn't work because blocks are only produced by submitting new transactions.

# Design

Even though there's this big blocker, we can create the initial APIs for these functionalities and make it work by using a "brute-force" design; an inefficient way to get this to work while we wait for the client to deliver the txpool and some other features to support the proper way of doing this.

It would work as follows. `tx.set_block_height(block_height: u64)` would, internally, create $n$ dummy transactions where $n$ is $blockHeight - currentBlockHeight$. Because 1 transaction executed == one block produced, running these $n$ transactions before running the actual `tx` would make the `tx` be executed at the desired target `block_height`. For `get_block_height`, getting the current block height using the `Client`'s API is relatively trivial, so that would be a simple wrapper.

When the blockers from the client are gone, i.e., txpool released and support for respecting the `block_height` / `maturity` of a tx before including it in the pool, we can do follow-up work to re-work the internals of these new APIs (`set_block_height`, etc.) by doing it in a proper way, using the new `Client`'s API for that.

adlerjohn:
Is that actually how ethers-rs does it? I was under the assumption you could use the `evm_mine` RPC endpoint (wrapped in an SDK function of course) to advance block height.

[...]

PR #428, opened 2022-06-25
Segfault magnet/410 transaction target block height

segfault-magnet:
closes: #410 
blocked by: FuelLabs/fuel-core#432

To enable users to use/test `maturity` in tests this PR will:
* actually propagate the `maturity` from the transaction parameters into the `submit` gql request
* add a `add_blocks` helper function that will add the requested number of blocks to the chain.

Since a block can, theoretically (in the future), contain multiple transactions, the implementation of `add_block` will run a single script (noop) transaction and wait for the block to be generated. 

Doing it in parallel [wasn't possible originally ](https://github.com/FuelLabs/fuel-core/pull/442) is not future-proof -- we'd be relying on the fact that `fuel-core` currently only packs one transaction inside a block. 

I don't know where to put `add_blocks` without doing a major refactor. The current approach has it as an unassociated function that accepts a wallet and the number of blocks you want to add.

I'd like to avoid mixing this helper with anything production-grade since this, although now theoretically runnable 'in prod', will definitely not be in the future as soon as the new `txpool` is out.

Ideally, I'd like Wallets to not `own` providers and all the `setup provider and get wallet` to actually return a wallet and a `TestProvider`.

The `TestProvider` will own a `Provider` and delegate all `prod-grade` functions to it while extending the interface with test-only helpers so that the add_blocks will look like:
```rust
let (wallet, provider) : (LocalWallet, TestProvider) = setup_test_environment(...); 

provider.add_blocks(10);
```
or something like that but this requires major refactoring.


Cargo.toml
```diff
@@ -20,3 +20,6 @@ members = [
     "scripts/build-test-projects",
     "tools/fuels-abi-cli",
 ]
+
+[patch.crates-io]
+fuel-vm = { git = 'https://github.com/FuelLabs/fuel-vm', branch = 'Voxelot/maturity-validation'}
```
segfault-magnet, line 25: This is just so I can see the CI go green once the issue is fixed.

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -0,0 +1,87 @@
...
+/// * `wallet`: A `Wallet` that must contain a provider and spendable coins.
+/// * `amount`: By how much to increase the block height.
+///
+/// returns: Result<(), Error>
+pub async fn add_blocks(wallet: &LocalWallet, amount: usize) -> Result<(), Error> {
+    let provider = wallet.get_provider()?;
+    for _ in 0..amount {
+        let height_before_transaction = provider.latest_block_height().await?;
+
+        let transaction = no_op_signed_transaction(wallet).await?;
```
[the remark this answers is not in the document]
vnepveu: If we don't, maybe this should be an issue in `fuel-core`

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -0,0 +1,87 @@
...
+    };
+
+    let height_increased = retry_until(has_block_height_increased, 5, Duration::from_millis(100))
+        .await
+        .map_err(|err| Error::NetworkError(err.to_string()))?;
+
+    Ok(height_increased)
+}
+
+fn generate_no_op_script(inputs: Vec<Input>, outputs: Vec<Output>) -> Transaction {
```
vnepveu, line 72: instead of this function, use the `Transaction::default()` and `Script::new(tx)` constructors.

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -0,0 +1,87 @@
...
+    let outputs = vec![Output::change(wallet.address(), 0, AssetId::default())];
+
+    let mut transaction = generate_no_op_script(inputs, outputs);
+
+    wallet.sign_transaction(&mut transaction).await?;
+
+    Ok(transaction)
+}
+
+async fn check_if_block_height_increased(
```
vnepveu, line 56: better name would be `has_block_height_increased_since` (boolean function)

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -0,0 +1,87 @@
...
+    provider: &Provider,
+    height_to_compare_with: u64,
```
vnepveu, line 58: `previous_height` is a more concise argument name

packages/fuels-test-helpers/src/chains.rs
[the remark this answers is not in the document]
vnepveu: created https://github.com/FuelLabs/fuel-core/issues/455 after checking with the client team

packages/fuels-test-helpers/src/utility.rs
```diff
@@ -0,0 +1,103 @@
+use std::future::Future;
```
vnepveu, line 1: Call the file `utils` since that's how we call them in our repo AFAIK

packages/fuels-test-helpers/src/utility.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::time::Duration;
+
+pub async fn retry_until<Fut>(
+    condition: impl Fn() -> Fut,
+    max_attempts: usize,
+    between_attempts: Duration,
```
vnepveu, line 7: Call this argument `interval`, this name is a shortened version of `time_between_attempts`, which would be a good name, but too long, but its shortened version isn't clear enough.

packages/fuels-test-helpers/src/utility.rs
```diff
@@ -0,0 +1,103 @@
+use std::future::Future;
+use std::time::Duration;
+
+pub async fn retry_until<Fut>(
+    condition: impl Fn() -> Fut,
```
vnepveu, line 5: in our context, could we call this condition `predicate`? It would convey that it is actually a dynamic function that can change in the future.

packages/fuels-test-helpers/src/utility.rs
```diff
@@ -0,0 +1,103 @@
...
+        let err = response
+            .err()
+            .expect("Should have propagated the failure of the condition");
+
+        assert_eq!(err.to_string(), "Some error");
+        assert_eq!(*times_called.lock().await, 1);
+
+        Ok(())
+    }
+}
```
vnepveu, line 103: vgud tests! vry thorough

vnepveu (changes requested):
Should take more advantage of existing library functions. Otherwise great, it will be even more user friendly once we get a `mine_block` API on the node side.

docs/src/testing/chains.md
```diff
@@ -0,0 +1,10 @@
+## Increase the block height 
+
+`add_blocks` can be used to help achieve a desired block height. Useful for when you want to do any testing regarding
```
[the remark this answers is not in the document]
adlerjohn: "mining" only makes sense in the context of PoW though. "add" or "produce" is more generic.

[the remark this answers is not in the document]
vnepveu: Ah true. I think `produce` is better still, it conveys that it's added with some rules and process, whereas `add` just feels like it's legos.

packages/fuels-test-helpers/src/chains.rs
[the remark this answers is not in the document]
segfault-magnet: Great initiative! I assumed this was already underway, as mentioned in the task itself. Hopefully, we'll have an API for this soon.

packages/fuels-test-helpers/src/chains.rs
segfault-magnet: I initially did `previous_height` then I thought... what does previous mean in this context? This function doesn't know about the loop it is used in.

But after your comment I see now that `previous` could work because we are asking 'has the block height increased'. Applying the change.

packages/fuels-test-helpers/src/utils.rs
```diff
@@ -0,0 +1,101 @@
+use std::future::Future;
+use std::time::Duration;
+
+pub async fn retry_until<Fut>(
```
leviathanbeak, line 4: While this function is certainly not the focus of this PR, so not the biggest deal,
but I feel that this is too constraining, it should be either renamed to `retry_until_true` or even better made more generic,
where condition passes if the Result is Ok(T) instead of constraining it to `bool`.

What if someone wants to reuse it querying some chain API until it gets Ok Response instead ?

leviathanbeak: if made more generic then it could be maybe handled cleaner with
`tokio::time::timeout(duration_between, condition)`

packages/fuels-test-helpers/src/chains.rs
leviathanbeak: hmm this should work for you I think

```rust
let random_data = rand::thread_rng().gen::<[u8; 32]>();
Transaction::Script {
        inputs,
        outputs,
        script_data: random_data,
        ..Default::default()
}
```

packages/fuels-abigen-macro/tests/test_projects/transaction_block_height/src/main.sw
```diff
@@ -0,0 +1,10 @@
+contract;
+
+abi MyContract {
+    fn test_function();
+}
+
+impl MyContract for Contract {
+    fn test_function() {
+    }
```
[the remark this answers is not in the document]
segfault-magnet: There is a way but I've decided against returning the block height from the contract itself.

The reason is twofold:
1. Calling a contract function that will return the current block height will produce a block in and out of itself. This makes testing weird because you'd have these +2 increments due to the previous check and the actual call. And if you use the height getter contract function as the primary function to test maturity, then you'll have these weird cases where you need the thing to fail but also need to validate the height but can't since the contract failed.

2. Don't feel the need to be paranoid about the gql node API, it seems simple enough and we'd effectively be testing what the fuel-core team should test.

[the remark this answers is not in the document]
leviathanbeak: > 1. Calling a contract function that will return the current block height will produce a block in and out of itself. 

@adlerjohn is this really the case? does a contract call that just reads some data get actually stored in the block as a transaction?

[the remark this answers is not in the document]
segfault-magnet: Ok did it. There is a test now which confirms that block height from within sway and from the gql API is of the same value.

Not sure if we're the ones that should be testing this, but until a more generalized whole solution test suite gets implemented, I guess the SDK is the QA department :D

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -2006,45 +2007,104 @@ async fn contract_method_call_respects_maturity() -> anyhow::Result<()> {
...
+    let id = Contract::deploy(
+        "tests/test_projects/transaction_block_height/out/debug/transaction_block_height.bin",
+        &wallet,
+        TxParameters::default(),
+    )
+    .await?;
+    let instance = MyContract::new(id.to_string(), wallet.clone());
+
+    let block_height_from_contract = || async {
+        Ok(instance.get_current_height().simulate().await?.value) as Result<u64, Error>
```
vnepveu, line 2099: great workaround!

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -1,18 +1,15 @@
-use crate::{utility, Opcode, Transaction, REG_ONE};
+use crate::utils;
```
vnepveu, line 1: redundant with line 10

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -53,35 +50,50 @@ async fn no_op_signed_transaction(wallet: &LocalWallet) -> Result<Transaction, E
...
+        maturity,
+        receipts_root: _,
+        script,
+        script_data: _,
+        inputs: _,
+        outputs: _,
+        witnesses,
+        metadata: _,
+    } = Transaction::default()
+    {
```
vnepveu, line 83: Not really convinced by this. If you are really scared that all of a sudden we break the `Transaction::default -> Transaction::Script` assertion, you can just check that the Enum variant is the right one. My idea was to reduce the number of SLOC :smile: 
I think you can remove this function altogether! See above comment.

packages/fuels-test-helpers/src/utils.rs
```diff
@@ -55,10 +55,10 @@ mod tests {
 
     #[tokio::test]
     async fn retry_until_will_respect_delay_between_attempts() -> anyhow::Result<()> {
-        let timestamps_condition_was_called_at: Mutex<Vec<Instant>> = Mutex::new(vec![]);
+        let timestamps_predicate_was_called_at: Mutex<Vec<Instant>> = Mutex::new(vec![]);
```
vnepveu, line 58: `%s/timestamps/timestamp/g`

packages/fuels-test-helpers/src/utils.rs
segfault-magnet: > but I feel that this is too constraining, it should be either renamed to retry_until_true or even better made more generic (as the name suggests),
where condition passes if the Result is Ok(T) instead of constraining it to bool.

The name could sure use a rehaul that is for sure. But the thing is I need something that will:

1. Retry in case of transient network errors (i.e. Result::Err)
2. Have a (simple or exponential) delay between retries as to wait out the network error
3. If we managed to get through to the node but it still hasn't mined a block, then I need to try again later on.

So I need a Result::Ok *and* need the thing to be true. A kind of asynchronous fail-able predicate.

I can't get that last part if I generalize the solution to 'retry until Result::Ok' since that is only half of the condition.

> What if someone wants to reuse it querying some chain API until it gets Ok Response instead ?

I'm of the YAGNI mindset, especially for private internals. If we ever need such a thing we can write it.

> if made more generic then it could be maybe handled cleaner with
tokio::time::timeout(duration_between, request)

That was my initial approach, which looked something like this:
```rust

pub async fn retry_until_w_timeout<Fut>(
    predicate: impl Fn() -> Fut,
    timeout_after: Duration,
    interval: Duration,
) -> anyhow::Result<bool>
where
    Fut: Future<Output = anyhow::Result<bool>>,
{
    tokio::time::timeout(timeout_after, async {
        loop {
            if predicate().await? {
                return Ok(true);
            }
            tokio::time::sleep(interval).await;
        }
    })
    .await?
}
```
but ultimately I decided against it, something about the appearance of the fn signature put me off.

But now I can see a huge benefit of this. I don't know the timeout configuration of the gql client, so 'tokio::time::timeout' might be better to avoid spending `num_attempts * network_timeout` in the case of a generous timeout value or network issues (albeit this is currently only tested against a node in localhost).

packages/fuels-test-helpers/src/utils.rs
vnepveu: You renamed the tests but not the actual function argument here, any particular reason?

vnepveu (changes requested):
Starting to look really good, a few things left!

docs/src/testing/chains.md
```diff
@@ -1,10 +1,10 @@
 ## Increase the block height 
 
-`add_blocks` can be used to help achieve a desired block height. Useful for when you want to do any testing regarding
+`produce_blocks` can be used to help achieve a desired block height. Useful for when you want to do any testing regarding
 transaction maturity.
```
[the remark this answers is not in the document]
segfault-magnet: I've applied the suggestion manually since this was outdated, the line breakage wouldn't cause issues in the book since md only breaks paragraphs if you have a double line break, afaics.

packages/fuels-test-helpers/src/chains.rs
segfault-magnet: How would you go about checking if the variant is the right one if not by doing a `match`? And if you do a match on an exhaustive enum you need to handle all matches. And handling that requires panicking if we ever break `Transaction::default->Transaction::Script`.

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -53,35 +50,50 @@ async fn no_op_signed_transaction(wallet: &LocalWallet) -> Result<Transaction, E
     Ok(transaction)
```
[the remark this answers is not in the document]
segfault-magnet: The thing is I would love to do that, but Transaction is an Enum of two variants and you can't index into the fields of the struct-like variants like this. You need to know if it is a Script or a Transaction. 

And to know that you have to match. And if you match you need to handle the other case.

This would work if the variants weren't struct-like but rather contained structs in them like:
```
enum Transaction {
Script(ScriptData), 
...
}
```
Then you could match and extract the script data and do this on it since you know the exact type.

But the thing is this code is in `fuel-tx` and I didn't fancy doing breaking changes for a few lines here.

If you can find a better way to do this I would absolutely love this, cuz this is hideous.

Please reopen if you think of something.

packages/fuels-test-helpers/src/utils.rs
segfault-magnet: > You renamed the tests but not the actual function argument here, any particular reason?

Yeah, I reverted it by accident trying to be smart with git. I re-reverted. It should be updated by now.

> %s/timestamps/timestamp/g
A fellow VIM user :D 

Actually, these are timestamp**s**, plural. Predicate called more than once, and thus more than one timestamp is written thus we have timestamps of all the times the predicate was called.

If you feel otherwise please re-open.

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -1,18 +1,15 @@
...
 use rand::Rng;
 use std::time::Duration;
-use utility::retry_until;
+use utils::retry_until;
 
-/// Can be used to generate `amount` number of blocks. A block is generated by
+/// Can be used to produce `amount` number of blocks. A block is produced by
 /// submitting a no-op script for execution. Transactions are submitted one
 /// after the other. A new transaction won't be submitted until an increase of
 /// block height has been noticed.
```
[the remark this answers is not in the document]
vnepveu: I'm not a fan of creating a ticket that just refers the `fuel-core` ticket unless you think it's really necessary. Doesn't have to be a `TODO`, but we need to document that it *is* a workaround anyway.

docs/src/testing/chains.md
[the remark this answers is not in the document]
vnepveu: hehe my macro does the same, and adds John a lot of work when writing docs lol

packages/fuels-test-helpers/src/utils.rs
vnepveu: Ah yes my bad! Thought it was only one

packages/fuels-test-helpers/src/chains.rs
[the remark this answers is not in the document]
segfault-magnet: The thing I would like to achieve is for us to remember to improve this once the fuel core ticket is done.

AFAICS we can do that by either doing a TODO (which is kind of annotated in the IDE) but that has a habit of staying around without ppl noticing. 

Or we can create an issue like 'use the new tx pool API for producing blocks' which will remind us to do it by its very presence in the backlog.

Because, if I'm not mistaken, the only reminder we have now is our short-term memory and the notification you're going to get if somebody starts working on the fuel core task since you're the reporter.

packages/fuels-test-helpers/src/utils.rs
leviathanbeak: I'm just going to disagree once more and leave it be since it's just a helper function and not worth much more time.

> So I need a Result::Ok and need the thing to be true. A kind of asynchronous fail-able predicate.

I don't see it that way, you could still just write a closure that returns `Err("block height has not changed")`
there is zero reason to use `Ok(false)` in the first place - and that decision, how I see it, is the only reason why you can't make this function more generic. If you went with my suggestion you would have:

> Retry in case of transient network errors (i.e. Result::Err)
- Yes, by default
> Have a (simple or exponential) delay between retries as to wait out the network error
- Yes, as an argument
> If we managed to get through to the node but it still hasn't mined a block, then I need to try again later on.
- Yes, just return `Err("block height has not changed")` or something similar.

As for your comment on `tokio::timeout`, the way you wrote it is not correct, it should be way simpler:
```rust
    for 0..max_attempts {
        if let Ok(Ok(result) = tokio::time::timeout(duration_between, request) {
           return Ok(result)
        }
    }
    
    Err("failed to get the result after N amounts of attemprs")
   ```

Note that the way I wrote it I assumed that you took my suggestion of a more generic helper function, if not then it in your case it would be `Ok(Ok(true))`

File: packages/fuels-test-helpers/src/chains.rs (hunk 3 above)
Comment (leviathanbeak) [reply packages/fuels-test-helpers/src/chains.rs:68]: @segfault-magnet I don't get it, if this is what you need - this should work
```rust
fn generate_no_op_script(inputs: Vec<Input>, outputs: Vec<Output>) -> Transaction {
    let random_data = rand::thread_rng().gen::<[u8; 32]>();
    Transaction::Script {
        inputs,
        outputs,
        script_data: random_data,
        ..Default::default()
    }
}
```

packages/fuels-test-helpers/src/chains.rs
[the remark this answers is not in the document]
leviathanbeak: > I'm not a fan of creating a ticket that just refers the fuel-core ticket unless you think it's really necessary. Doesn't have to be a TODO, but we need to document that it is a workaround anyway.

What better way to document it than having a ticket for it, TODOs without a ticket are useless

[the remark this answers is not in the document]
vnepveu: What I mean is we already have a ticket for it in `fuel-core`. Possibly we can add in that ticket that the change should also be ported to the SDK, but I don't see the merit in having duplicate tickets accross repos.

packages/fuels-test-helpers/src/chains.rs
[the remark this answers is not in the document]
vnepveu: Didn't try anything, will maybe toy around with it in the future haha, sorry I thought this would work lol.

vnepveu (dismissed):
:shipit:

packages/fuels-test-helpers/src/chains.rs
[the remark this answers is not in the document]
segfault-magnet: The ticket in `fuel-core` is for them to provide an API. 

The ticket in `fuels-rs` would be for us to use it.

So they are related but not duplicated, but I understand the feeling. 

I'll create a ticket in fuels-rs, if you don't mind, just to have some peace of mind otherwise it's going to be taking up space as a sticky note on the wall of my office.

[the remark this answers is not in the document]
ControlCplusControlV: @segfault-magnet PR is open for `fuel-core` if you want to have the ticket tracking that [#461](https://github.com/FuelLabs/fuel-core/pull/461). You could wait for that to go through but idk how concerned you are about this getting stale and it may be easier to just merge this PR then use the new API in a later one

packages/fuels-test-helpers/src/utils.rs
segfault-magnet: Initially, I didn't like the idea of returning `Err` from the closure since the block height not being increased **yet** isn't an error per see, just a fact,  at least until all the attempts have been exhausted. To top it off I'd be mixing it real errors like 'cannot reach node' which would require more effort should you later on wish to differentiate what exactly went wrong -- did we hear back from the node but the height remained the same or did we have network issues and don't exactly know what is going on with the height?

But I decided to do a poc anyway and through it saw that I really don't care what happened -- I always do the same thing -- report it to the user as a stringified message inside InfrastructureError.

As such I don't have to differentiate between actual errors and the fake ones we created to serve the retry mechanism. 

So I've generalized the retry fn, but note that your suggestion:

> As for your comment on tokio::timeout, the way you wrote it is not correct, it should be way simpler:

> ```rust
>   for 0..max_attempts {
>         if let Ok(Ok(result) = tokio::time::timeout(duration_between, request) {
>            return Ok(result)
>         }
>     }
>     
>     Err("failed to get the result after N amounts of attempts")
> ```

doesn't work for the use case I need because the suggested snippet would only limit the maximum duration of one request -- should the request fail quickly because the client replied promptly with a response indicating the block height hadn't increased yet, we would shoot through all of our attempts without giving the node time to produce the block.

My solution looks like this:
```rust
  let mut last_err = None;

    tokio::time::timeout(abort_after, async {
        loop {
            match action().await {
                Ok(value) => break value,
                Err(error) => last_err = Some(error),
            }

            tokio::time::sleep(interval).await;
        }
    })
    .await
    .map_err(|_| RetryExhausted {
        interval,
        abort_after,
        error_from_last_attempt: last_err,
    })
```
With it, the number of attempts becomes a bit indeterministic since we would need to factor in how long the actual request took. But since I only care that it tries a couple of times, while giving the node some room to breathe, this is ok.

vnepveu (approved):

Issue #430, opened 2022-06-27
Automagically recover an ABI-JSON struct from a Vec<u8>


ra0x3:
- If I have a struct `Foo` that I `return` in my Sway contract, that struct `Foo` gets packed into the `data` property of my `Receipt::Logata` as a `Vec<u8>`
  -  For example
- However, there is no way for me to recover my `Foo` struct from this `Vec<u8>`, without manually decoding that `Vec<u8>` into a `Token`
- It would be nice if I could call something like `Foo::from_vec_u8(data: Vec<u8>)` that would recover my `Foo` struct without me having to manually do it myself
  - This is _almost_ like `Tokenizable`, but not completely

I should be able to do the following (rough example):

Using this contract abi JSON
```json
[
    {
        "type":"contract",
        "inputs":[
            {
                "name":"Foo",
                "type":"struct Foo",
                "components": [
                    {
                        "name": "id",
                        "type": "u64"
                    },
                    {
                        "name": "account",
                        "type": "b256"
                    }
                ]
            }
        ],
        "name":"takes_struct",
        "outputs":[]
    }
]
```

And then use that JSON as follows:
```rust
// With the ABI JSON including a `Foo` struct

abigen!(Contract, "path/to/my-contract.bin");

fn do_something(data: Vec<u8>) -> Foo {
    let foo: Foo = Foo::from_vec_u8(data);
    println!("Foo.id is {} and Foo.address is {}", foo.id, foo.address);
    foo
}
```

The idea is that 

- You already know the components of my `Foo` struct (because I can see that in the ABI JSON)
- So you can then just chunk the `Vec<u8>` into pieces (as defined in the ABI JSON), and recover the `Foo` struct using these chunks
  - Doesn't need to be anything fancy
  - Can be optimistic (i.e., assume the `Vec<u8>` contains the proper data) else just fail
- This would greatly improve devx, as I wouldn't have to manually keep track of which pieces of data, are in which order, in order to manually decode the `Vec<u8>` into a `Token`.
  - I could just call this one-liner `Foo::from_vec_u8` and boom 💥 , I have my `Foo` struct back
  - This is an actual issue (an issue in having to doing it manually) I'm experiencing on the indexer

Additional context

- There was a Slack convo between @digorithm and myself [here](https://fuellabs.slack.com/archives/C02CX4JVCSJ/p1655834736318249) that might offer more context

PR #431, opened 2022-06-28
Improve devx by providing shortcuts to decode bytes into contract types.

segfault-magnet:
closes: #430 

### About the solution
You can now do the following:
```rust
    let shaker_in_bytes: Vec<u8> = vec![0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 2];

    let actual: Shaker = shaker_in_bytes.try_into()?;
```
or you can do:
```rust
    let shaker_in_bytes: Vec<u8> = vec![0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 2];

    let actual: Shaker = try_from_bytes(&shaker_in_bytes)?;
```
Where `Shaker` is a type that implements `Parameterize` and `Tokenize`.

The caveat of the first approach is that it only works if the target type is generated by `abigen!` because of the orhpan rule. I've tried doing the newtype pattern but it didn't make things prettier.

So if you have, for example, a ContractId (which is a `[u8; 32]` and falls into the orphan rule group) then you need to do:
```rust
let actual: ContractId = try_from_bytes(&contract_id_bytes);
```
The same thing goes for `u8` `u16`...

I've also extended the tuple generating macro to implement `Parameterize` so that we can do this as well:
```rust
        let tuple_in_bytes: Vec<u8> = vec![0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 2];

        let the_tuple: (u64, u32) = try_from_bytes(&tuple_in_bytes)?;

        assert_eq!(the_tuple, (1, 2));
```
One thing that won't work is decoding tuples with `Strings` in them since we cannot know the string len via the current design.

I think I could have implemented `TryFrom` for tuples but I got cold feet -- can't currently wrap my head around what issues that might cause further on. Let me know if you think I should do that as well so that we may use the first form of `try_into`. 

The `abigen` was extended to implement `TryFrom` for `Vec<u8>`, `&Vec<u8>` and `&[u8]` so that we may improve the devx.


packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -64,15 +62,7 @@ pub fn expand_custom_struct(prop: &Property) -> Result<TokenStream, Error> {
...
-                // The enum we're currently looking at must have variants due to
-                // the usage of `EnumVariants`. Because of this we can safely
-                // call unwrap() since this is the same ParamType::Enum that
-                // will be used to generate the actual Enum type in Rust whose
-                // param_types() method will not return an empty Vec due to the
-                // aforementioned EnumVariants.
-                let variants = quote! {EnumVariants::new(#enum_name::param_types()).unwrap()};
-
-                param_types.push(quote! { types.push(ParamType::Enum(#variants)) });
+                param_types.push(quote! { types.push(#enum_name::param_type()) });
```
digorithm, line 65: 🔥

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -282,10 +284,10 @@ pub fn expand_custom_enum(enum_name: &str, prop: &Property) -> Result<TokenStrea
         }
 
         impl Parameterize for #enum_ident {
-            fn param_types() -> Vec<ParamType> {
+            fn param_type() -> ParamType {
                 let mut types = Vec::new();
                 #( #param_types; )*
-                types
+                ParamType::Enum(EnumVariants::new(types).unwrap())
```
digorithm, line 290: Should we have an `.expect()` with a hint instead of `.unwrap()` here? Feels like this would be a pain to debug if it ever breaks 😅

digorithm (changes requested):
Great work, awesome refactor you did along the way! 

Now, this (`try_into`) is something that users might wanna do often but it's fairly tricky to learn how to do it. Maybe we should add a small section in the docs talking about converting raw data (`Vec<u8>`) into `abigen`'d structs. Don't worry too much about where to put this in the docs because I'm already working on re-organizing it, just adding this piece of info now so that we don't forget later should be enough!

ra0x3 (dismissed):
- utAck (obviously will defer to the SDK team)
- Many thanks for getting to this @segfault-magnet  🙏

packages/fuels-core/src/code_gen/custom_types_gen.rs
segfault-magnet: Revisiting this made me see an edge case where we could have a panic in runtime. Kudos! Fixed it in the latest commit.

Also added the message.

digorithm (approved):
:shipit:

ra0x3 (approved):
🚢

PR #432, opened 2022-06-28
feat: add predicate support baseline with tests

### LINKED ISSUE #341: Support for predicates
The Fuel VM allows for a Coin or Message input to be locked behind a predicate. To spend one of these inputs, the predicate is provided as a field of the input. If `hash(predicate) == Input.owner && eval(predicate)` , the UTXO can be spent. Note that in this case, no signature is required as the predicate alone is sufficient to spend the UTXO.

The SDK will therefore need to be able to:
- attach a compiled predicate to an Input.
- attach predicate data to an input. (This is analagous to providing scriptdata to a script - the data are parameters to the main function)
- send a transaction without the need for any wallet or signer
- set an output's `owner` to the hash of a compiled predicate

camsjams:
Closes #341 

Will be enhanced once https://github.com/FuelLabs/sway/issues/1981 is completed

PR for Sway is ready for review: https://github.com/FuelLabs/sway/pull/1984


packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1910,3 +1923,437 @@ async fn nested_enums_are_correctly_encoded_decoded() {
 
     assert_eq!(response.value, expected_none);
 }
+
+#[tokio::test]
+async fn can_call_no_arg_predicate_returns_true() {
```
vnepveu, line 1928: Here I feel like you can definitely "group" tests together, because there seems to be a lot of boilerplate code being repeated.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1910,3 +1923,437 @@ async fn nested_enums_are_correctly_encoded_decoded() {
...
+    .await;
+    let connected_provider = Provider::connect(address).await.unwrap();
+    wallet.set_provider(connected_provider);
+
+    let instance = Predicate::new(predicate_code);
+    instance
+        .create_predicate(&wallet, amount_to_predicate, asset_id)
+        .await
+        .unwrap();
+    let receiver_balance_before = get_balance(&wallet, receiver_address, asset_id).await;
```
vnepveu, line 1958: You can use just `connected_provider.get_balance(asset_id)` iirc

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1910,3 +1923,437 @@ async fn nested_enums_are_correctly_encoded_decoded() {
...
+    instance
+        .spend_predicate(
+            &wallet,
+            amount_to_predicate,
+            asset_id,
+            receiver_address,
+            None,
+        )
+        .await
+        .unwrap();
```
vnepveu, line 1969: Use `?` here too, even in tests. For this, you will have to change the return type of the test to `Result<(), Error>`, see examples above in this file.

vnepveu (changes requested):
Good first shot! My review isn't as thorough as it could be because I think I gave you enough material to improve for now :smiley_cat:. Congrats on this first Rust PR!

packages/fuels-abigen-macro/tests/test_projects/predicate_address/Forc.toml
```diff
@@ -0,0 +1,5 @@
+[project]
+name = "predicate_address"
+license = "Apache-2.0"
```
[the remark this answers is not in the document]
vnepveu: Generally I think you can try to merge all of those "small" test projects into one big one!

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1910,3 +1923,437 @@ async fn nested_enums_are_correctly_encoded_decoded() {
...
+    let (_launched_provider, address) = setup_test_provider(
+        coins,
+        Some(Config {
+            predicates: true,
+            utxo_validation: true,
+            ..Config::local_node()
+        }),
+    )
+    .await;
+    let connected_provider = Provider::connect(address).await.unwrap();
```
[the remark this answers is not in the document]
vnepveu: The function you mentioned is the one I had in mind, how are you blocked by feature flags? I've been using it all the time but maybe there is something I'm not aware of.

packages/fuels-abigen-macro/tests/harness.rs
vnepveu: So our current workflow is that `unit tests` are put in the file that they are testing. And the harness is for "end to end" tests. 
For reuse of code you can look at other examples in the `harness.rs` file, basically you can skip the whole node/coins/wallet setup and do it just once at the beginning.

packages/fuels/tests/harness.rs
```diff
@@ -2265,3 +2268,355 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
...
+    let wallets = launch_custom_provider_and_get_wallets(
+        WalletsConfig::default(),
+        Some(Config {
+            predicates: true,
+            utxo_validation: true,
+            ..Config::local_node()
+        }),
+    )
+    .await;
+    let wallet = &wallets[0];
```
vnepveu, line 2288: you should do `provider = wallet.get_provider` then you can avoid doing `get_provider` all the time.

packages/fuels/tests/harness.rs
```diff
@@ -2265,3 +2268,355 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
...
+
+    let instance = Predicate::new(predicate_code);
+    instance
+        .deploy_predicate(wallet, amount_to_predicate, asset_id)
+        .await?;
+    let receiver_balance_before = wallet
+        .get_provider()?
+        .get_asset_balance(&receiver_address, asset_id)
```
vnepveu, line 2401: you can do `wallet.get_asset_balance`

packages/fuels/tests/harness.rs
```diff
@@ -2265,3 +2268,355 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
...
+            amount_to_predicate,
+            asset_id,
+            receiver_address,
+            None,
+        )
+        .await?;
+    let receiver_balance_after = wallet
+        .get_provider()?
+        .get_asset_balance(&receiver_address, asset_id)
+        .await?;
```
vnepveu, line 2311: wallet.get_asset_balance

packages/fuels/tests/harness.rs
```diff
@@ -2265,3 +2268,355 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
...
+        std::fs::read("tests/test_projects/predicate_true/out/debug/predicate_true.bin")?;
+    let receiver_address: Address = LocalWallet::new_random(None).address();
+    let wallets = launch_custom_provider_and_get_wallets(
+        WalletsConfig::default(),
+        Some(Config {
+            predicates: true,
+            utxo_validation: true,
+            ..Config::local_node()
+        }),
+    )
```
vnepveu, line 2286: basically I think your best bet is to have a wallet with the predicate address inside, this way you can avoid doing the `provider` lookup everytime.

packages/fuels/tests/harness.rs
```diff
@@ -2265,3 +2268,355 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn can_call_no_arg_predicate_returns_true() -> anyhow::Result<(), anyhow::Error> {
```
vnepveu, line 2273: returning `Result<(), Error>` should work too.

vnepveu (changes requested):
Starting to look good! 
you can still streamline the process with the providers and wallets.

hal3e:
Good work! I will add some nits from my side

hal3e:
@camsjams sorry! I wanted to suggest a change with a commit but it was applied directly. I did it on this branch :/ You can check it out and if you like it you can leave it.

hal3e (changes requested):

packages/fuels-contract/src/predicate.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+impl Predicate {
+    pub fn new(code: Vec<u8>) -> Self {
+        let address = (*Contract::root_from_code(&code)).into();
+        Self { code, address }
+    }
+
+    /// Craft a transaction used to transfer funds between two addresses.
+    pub fn build_transfer_tx(
```
MujkicA, line 26: Use `build_transfer_tx()` from `Provider`

packages/fuels-contract/src/predicate.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /// Deploys locked coins in a Predicate to the given wallet's provider.
+    ///
+    /// # Arguments
+    ///
+    /// * `wallet` - A wallet that funds this transaction
+    /// * `coin_amount_to_predicate` - The amount of locked coins as given asset id to store within Predicate
+    /// * `asset_id` - The asset id of the locked coins stored within Predicate
+    pub async fn deploy_predicate(
```
MujkicA, line 57: I dislike that we are duplicating code here from `wallet.transfer()`.

Maybe we could either move this to `wallet` and have it  be a wrapper for `wallet.transfer()` or extract `wallet.transfer()`s functionality into a provider method and have `wallet` and `predicate` use it. 
`spend_predicate` from below could also benefit from this.

This can also be done in a separate refactoring issue, just wanted to share my thoughts.

packages/fuels/tests/harness.rs
```diff
@@ -2243,6 +2250,351 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+#[cfg(feature = "fuel-core-lib")]
+async fn can_call_no_arg_predicate_returns_true() -> Result<(), anyhow::Error> {
+    let amount_to_predicate: u64 = 16;
```
MujkicA, line 2256: The tests seem to be doing some repeated setup work. You could extract that into a separate function to avoid duplication.

packages/fuels/tests/harness.rs
```diff
@@ -2243,6 +2250,351 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
...
+    let asset_id = AssetId::default();
+    let predicate_code =
+        std::fs::read("tests/test_projects/predicate_true/out/debug/predicate_true.bin")?;
+    let receiver = LocalWallet::new_random(None);
+    let receiver_address: &Bech32Address = receiver.address();
```
MujkicA, line 2261: Type annotation seems unnecessary

packages/fuels/tests/harness.rs
```diff
@@ -2243,6 +2250,351 @@ async fn can_handle_sway_function_called_new() -> anyhow::Result<()> {
...
+        WalletsConfig::default(),
+        Some(TestConfig {
+            predicates: true,
+            utxo_validation: true,
+            ..TestConfig::local_node()
+        }),
+    )
+    .await;
+    let wallet = &wallets[0];
+    let provider: &Provider = wallet.get_provider()?;
```
MujkicA, line 2272: Annotation here too

MujkicA (commented):
I've added a couple of suggestions. One thing I'm not sure about is do we need all those tests. IIUC it checks if predicates work with different types of data but the question is should that be the responsibility of the SDK vs. just confirming that we reliably transfer the data.

packages/fuels-contract/src/predicate.rs
hal3e: I would suggest a separate issue that someone from the SDK team takes - I would volunteer.  We also need to add documentation as I have not seen an open issuer for that.

simonr0204:
Is `deploy_predicate` terminology that we want to commit to? Technically nothing is deployed (and this is one of the great advantages of predicates - no state use). All predicate roots "are out" there ready to receive funds, just as all wallet addresses are (in fact, they're the same space), so nothing is deployed any more than a wallet address could be said to be when you send it coins.

adlerjohn:
Deploy isn't the best word, for the reasons you listed. You can send funds to a predicate (hash/root).

hal3e:
I will support @camsjams with this PR to get it closed faster :rocket: 

I did some refactoring and tried to implement the recent comments.

Changes:
- removed the `#[cfg(feature = "fuel-core-lib")]`. Now test work with and without `--all-features`
- updated how node config is used (predicates were not enabled with when running `fuel-core` bin
- `deploy_predicate` was renamed to `transfer_to_predicate` and  moved to `wallet`. It is now a wrapper of `transfer`
- `spend_predicate` was moved to `provider` with a wrapper function `recieve_from_predicate` in `wallet`
- tests were slightly refactored and I left all of them there.

iqdecay (changes requested):
Few nits left

packages/fuels-signers/src/wallet.rs
```diff
@@ -283,6 +283,38 @@ impl Wallet {
...
+        &self,
+        to: &Bech32Address,
+        amount: u64,
+        asset_id: AssetId,
+    ) -> Result<Vec<Receipt>, Error> {
+        let (_, receipts) = self
+            .transfer(to, amount, asset_id, TxParameters::default())
+            .await?;
+        Ok(receipts)
+    }
```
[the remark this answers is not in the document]
hal3e: Good question :smile: :. It does the same thing. The `transfer_to_predicate` used to be a method of `Predicate` were code was reused from `transfer`. An advice was to make it a wrapper of `transfer` (that is the current state). However, I also moved it to wallet as it did not make sense to keep it in `Predicate`. That said,  we could also use `transfer` directly.

On the other hand we have a wrapper for `spend_predicate` which is called `receive_from_predicate` in wallet so for me it made sense to also have `transfer_to_predicate` in wallet.

I would like to hear some opinions whether to keep the `transfer_to_predicate` function or use `transfer` directly.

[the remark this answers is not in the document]
iqdecay: IMO better to just use `transfer` directly! No use exposing two different API that do the same thing :)

packages/fuels-signers/src/provider.rs
```diff
@@ -455,6 +455,49 @@ impl Provider {
         self.client.produce_block(amount).await
     }
 
+    pub async fn spend_predicate(
```
MujkicA, line 458: IIUC we only supporting spending, but not sending to predicates so far

MujkicA (approved):
We should either add docs to this PR or create another issue. Other than that, LGTM.
Great work!

hal3e:
Here is the issue https://github.com/FuelLabs/fuels-rs/issues/535. I will add the docs after we merge this PR.

hal3e (approved):

Br1ght0ne (approved):

iqdecay (approved):
Nice! :ship:

PR #433, opened 2022-06-28
fix: make generated module public

digorithm:
This is a nice-to-have in case users want to directly access the generated `mod` in other files.


MujkicA (commented):
LGTM, except for the patch file

contract.patch
```diff
@@ -0,0 +1,142 @@
+diff --git a/packages/fuels-contract/src/contract.rs b/packages/fuels-contract/src/contract.rs
```
adlerjohn, line 1: 😬

digorithm: LOL

MujkicA (approved):

Issue #435, opened 2022-06-29
build currently broken -- CI always uses the feature `fuel-core-lib`


segfault-magnet:
If we run `cargo build --no-default-features` we'll get:
```
error[E0425]: cannot find function `is_free` in this scope
   --> packages/fuels-test-helpers/src/lib.rs:192:52
    |
192 |         Some(config) if config.addr.port() != 0 && is_free(config.addr.port()) => config.addr,
    |                                                    ^^^^^^^ not found in this scope
    |
help: consider importing this function
    |
3   | use portpicker::is_free;
    |

error[E0425]: cannot find function `is_free` in this scope
   --> packages/fuels-test-helpers/src/lib.rs:193:26
    |
193 |         Some(config) if !is_free(config.addr.port()) => panic!("Error: Address already in use"),
    |                          ^^^^^^^ not found in this scope
    |
help: consider importing this function
    |
3   | use portpicker::is_free;
    |

For more information about this error, try `rustc --explain E0425`.
error: could not compile `fuels-test-helpers` due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
```

It seems that our CI always runs `fuels-rs` with `--all-features` which renders some of the code disabled due to the conditional compilation based on the flag 'fuel-core-lib`. As such we have parts of our code that are basically unchecked by the CI.

Fix the CI and make the build pass.

Ideally, we'd have a test run with fuel-core as a lib and another one with it as a binary, for maximum regression protection.

digorithm:
> Ideally, we'd have a test run with fuel-core as a lib and another one with it as a binary, for maximum regression protection.

Yes, my only concern here is that the CI time could increase significantly 🤔 

As for the issues w.r.t feature flag, I'm pinging the `fuel-core-lib` flag creator @Salka1988.

segfault-magnet:
Perhaps  @Salka1988 can measure the impact? At the very least we need a cargo check without the feature to keep it buildable.

PR #440, opened 2022-06-29
fix: cargo build with `--no-default-features` flag

Salka1988:
add missing method is_free()


segfault-magnet (changes requested):
[From the issue](https://github.com/FuelLabs/fuels-rs/issues/435#issue-1289280616):

> It seems that our CI always runs fuels-rs with --all-features which renders some of the code disabled due to the conditional compilation based on the flag 'fuel-core-lib`. As such we have parts of our code that are basically unchecked by the CI.
> Fix the CI and make the build pass.
> 
> Ideally, we'd have a test run with fuel-core as a lib and another one with it as a binary, for maximum regression protection.


This should also be addressed since it is what allowed this to happen in the first place.

Salka1988:
> [From the issue](https://github.com/FuelLabs/fuels-rs/issues/435#issue-1289280616):
> 
> > It seems that our CI always runs fuels-rs with --all-features which renders some of the code disabled due to the conditional compilation based on the flag 'fuel-core-lib`. As such we have parts of our code that are basically unchecked by the CI.
> > Fix the CI and make the build pass.
> > Ideally, we'd have a test run with fuel-core as a lib and another one with it as a binary, for maximum regression protection.
> 
> This should also be addressed since it is what allowed this to happen in the first place.

I added a few more commands in CI and `` Total duration`` has not changed significantly.

vnepveu (dismissed):
lgtm

.github/workflows/ci.yml
```diff
@@ -116,6 +116,10 @@ jobs:
             args: --all-targets --all-features
           - command: test
             args: --all-targets --all-features
+          - command: test
+            args: --all-targets --no-default-features
+          - command: build
```
segfault-magnet, line 121: I'm not sure you need this build step -- how could cargo test something it has not built ?

If you ask it to test without default features it will be forced to build the extra code that was enabled.

So doing a build after that is redundant.

segfault-magnet (approved):
lgtm

PR #441, opened 2022-06-30
feat: add manual initial storage slots

### LINKED ISSUE #86: Add manual initial storage slots in the SDK

MujkicA:
Closes #86 

This PR enables manual initialization of storage slots via the SDK. `deploy` and `deploy_with_salt()` now accept a `Vec<StorageSlot>` which is used to calculate the contract_id on the SDK side and forwarded via the contract deployment transaction. 

I have discussed multiple options for the UI with @digorithm. The conclusion is that the above approach creates the least amount of cognitive load for the user, i.e. they don't have to remember another workflow. 

```rust
let storage_slot = ...

        "tests/test_projects/storage/out/debug/storage.bin",
        &wallet,
        TxParameters::default(),
        vec![storage_slot.clone()], // <----------
    )
    .await?;
```

Since `StorageSlot` requires the user to convert key/value to `Bytes32` I thought it would be convenient to provide a helper to create a slot from a generic key-value pair. 
`create_storage_slot` uses unsafe Rust and panics if key/value is larger than 32 byte since this is their limit as stated in the specs. 

```rust
let storage_slot = create_storage_slot("slot", 42);
```


packages/fuels-test-helpers/src/lib.rs
```diff
@@ -162,6 +163,24 @@ pub async fn setup_test_client(
     (client, srv.bound_address)
 }
 
+pub fn create_storage_slot<T: Sized, D: Sized>(key: T, value: D) -> StorageSlot {
+    if ::std::mem::size_of::<T>() > Bytes32::LEN || ::std::mem::size_of::<D>() > Bytes32::LEN {
+        panic!("StorageSlot key/value is limited to 32 bytes.");
+    }
+
+    unsafe fn to_bytes32<T: Sized>(p: &T) -> Bytes32 {
```
segfault-magnet, line 171: IIUC you're taking the raw bytes representation from memory and putting that in storage? 

I'm not sure that is necessarily compatible with how the ABI requires types to be encoded -- i.e. the endianess might be different for a start. Enum encoding and other stuff is not specified in Rust and is free to be whatever the compiler wants it to be and probably other stuff as well.

I'm not savvy with storage slots but if gets sent towards the VM then we'll have a problem. 

Otherwise, if it is just for local usage it seems like a weird way to pass data around the program :D

adlerjohn: If it's something purely for our internal testing then it shouldn't be exposed to the user as a feature.

packages/fuels-contract/src/contract.rs
```diff
@@ -156,8 +160,10 @@ impl Contract {
         binary_filepath: &str,
         wallet: &LocalWallet,
         params: TxParameters,
+        storage_slots: Vec<StorageSlot>,
```
adlerjohn, line 163: Instead of taking in a mandatory vector (which from the tests is empty 100% of the time), why not change `deploy_with_salt` to `deploy_with_parameters` (or something like that), that offloads all the complex parameter choosing to this advanced function. Then the vanilla `deploy` can be as simple as possible.

The simplest use case is users will want to deploy a contract with initial storage slots set automatically from a file, with a salt of zero. Anything else is advanced.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -162,6 +163,24 @@ pub async fn setup_test_client(
     (client, srv.bound_address)
 }
 
+pub fn create_storage_slot<T: Sized, D: Sized>(key: T, value: D) -> StorageSlot {
```
vnepveu, line 166: why not use `StorageSlot::new()`?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -162,6 +163,24 @@ pub async fn setup_test_client(
...
+    if ::std::mem::size_of::<T>() > Bytes32::LEN || ::std::mem::size_of::<D>() > Bytes32::LEN {
```
vnepveu, line 167: Reimplementing already implemented functions

vnepveu (changes requested):
Use already existing functions for storage slots, and make the storage slot vec non mandatory for vanilla deployment

packages/fuels-test-helpers/src/lib.rs
MujkicA: I decided to remove it since its benefits are negligible.

packages/fuels-test-helpers/src/lib.rs
MujkicA: `StorageSlot::new` accepts `Bytes32` while this would have worked with generic types. It has been removed however, see above discussion.

examples/contracts/src/lib.rs
```diff
@@ -79,44 +78,8 @@ mod tests {
...
-        let value = contract_instance.get_value(key).call().await?.value;
-        assert_eq!(expected_value.as_slice(), value.as_slice());
-
-        Ok(())
-    }
-
-    #[tokio::test]
-    // ANCHOR: deploy_with_salt
-    async fn deploy_with_salt() -> Result<(), Error> {
+    // ANCHOR: deploy_with_parameters
```
vnepveu, line 81: I think you removed a commit by mistake

vnepveu (changes requested):
I think you removed too many things

examples/contracts/src/lib.rs
MujkicA: After some changes the test in `examples` turned out to be the same as the one in `harness`. I moved the ANCHOR tags to `harness` and removed the example.

digorithm (approved):
LGTM, nice work on the API!

vnepveu (approved):
:shipit:

Issue #411, opened 2022-06-17
Use new `fuel-core` APIs to get transaction consensus parameters and minimum gas fee


digorithm:
Some transaction-related configurations in the SDK are either hardcoded or not checked at all. For instance, the `transaction_parameters` passed to the chain config below:

https://github.com/FuelLabs/fuels-rs/blob/054f352216f2327b36b20b884453787ce30166bc/packages/fuels-test-helpers/src/node.rs#L174-L197

There are currently no checks for these `max_*` configurations on the SDK end. We don't have it because we didn't have a way to dynamically query that from the node, [now we do](https://github.com/FuelLabs/fuel-core/pull/373), using: `client.chain_info()`. So these should be queried and checked when creating a new transaction.

Another configuration that should be dynamically checked is the minimum gas price. Which can now be queried through the `Client`'s `node_info()`.

We should use these constraints (`max_*`, `min_*`, etc.) when setting these values and properly handle errors.

MujkicA:
To get finer grained error handling as required by #249 and #345 we would need to check the transaction for validity on the SDK side. IIUC, `Transaction::validate_without_signature()` checks both the consensus parameters and validity e.g no duplicate inputs.  
We get two birds with one stone if we check  'validate_without_signature()' before submitting transactions to the VM.
One problem is that validation requires the current `block_height` to check if the required `maturity` is met, so we would need to connect it with #410 at some point.

PR #442, opened 2022-06-30
feat: add initial storage automatically

Salka1988:
add add automatic storage initialization


digorithm:
@Salka1988 Ready to review or still a draft?

packages/fuels-contract/src/contract.rs
```diff
@@ -203,7 +223,45 @@ impl Contract {
...
+    }
+
+    fn auto_init_storage(binary_filepath: &str) -> Vec<StorageSlot> {
+        let storage_path = format!(
+            "{}{}",
+            Path::new(binary_filepath)
+                .with_extension("")
+                .to_str()
+                .unwrap(),
+            "-storage_slots.json"
```
mohammadfawaz, line 241: The name of this file is not really in the spec and I'm not sure we should rely on it here. Should this maybe be a user input just like the JSON ABI file?

digorithm: I hate to ask for more inputs from the user... but I don't see a way out in this case :(

mohammadfawaz: Yeah :/ That being said, `forc new` (the new `forc init`) should and will handle that for the user, just like JSON ABI. I don't remember the last time I had to worry about pointing to a specific JSON or *.bin in an SDK test.

Salka1988: Then its semi-automatic storage initialization :). I saw that when ``forc`` generates a .json file, it only adds "-storage.json" to the name of contract, so I thought why not to use that. Ok, ill change this.

.github/workflows/ci.yml
```diff
@@ -8,19 +8,24 @@ on:
...
 env:
   CARGO_TERM_COLOR: always
   DASEL_VERSION: https://github.com/TomWright/dasel/releases/download/v1.24.3/dasel_linux_amd64
   RUSTFLAGS: "-D warnings"
   RUST_VERSION: 1.61.0
-  FORC_VERSION: 0.16.0
+  FORC_VERSION: 0.17.0
 
 jobs:
+  cancel-previous-runs:
```
adlerjohn, line 19: Why are you reverting #445?

adlerjohn (changes requested):

.github/workflows/ci.yml
Salka1988: For some reason it did not merge CI from the master. :thinking:

packages/fuels-core/src/parameters.rs
```diff
@@ -66,3 +66,32 @@ impl TxParameters {
...
+    }
+
+    pub fn with_storage_path(storage_path: Option<String>) -> Self {
+        Self {
+            storage_path,
+            manual_storage_vec: None,
+        }
+    }
+
+    pub fn with_manuel_storage(manual_storage_vec: Option<Vec<StorageSlot>>) -> Self {
```
adlerjohn, line 91:
```suggestion
    pub fn with_manual_storage(manual_storage_vec: Option<Vec<StorageSlot>>) -> Self {
```

Salka1988: :man_facepalming:  fixed

adlerjohn (approved):
lgtm, but defer to @digorithm or @vnepveu for final approval

packages/fuels-contract/src/contract.rs
```diff
@@ -160,8 +168,25 @@ impl Contract {
         binary_filepath: &str,
         wallet: &LocalWallet,
         params: TxParameters,
+        storage_configuration: StorageConfiguration,
```
MujkicA, line 171: in #411 it was discussed that `deploy` should stay as simple as possible. If the user wants advanced features like salt and storage they should use `deploy_with_parameters`. Perhaps we should move this to `deploy_with_parameters`.

packages/fuels-contract/src/contract.rs
```diff
@@ -173,7 +198,8 @@ impl Contract {
         storage_slots: Vec<StorageSlot>,
         salt: Salt,
     ) -> Result<ContractId, Error> {
-        let mut compiled_contract = Contract::load_sway_contract_with_salt(binary_filepath, salt)?;
+        let mut compiled_contract =
+            Contract::load_sway_contract_with_salt(binary_filepath, None, salt)?;
```
MujkicA, line 202: In addition to the above comment on `deploy`, it is confusing to have `deploy` do storage with manual/automatic merging while `deploy_with_parameters` only does manual storage.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -2034,3 +2066,42 @@ async fn can_use_try_into_to_construct_enum_from_bytes() -> Result<(), Error> {
...
+        StorageConfiguration::with_storage_path(
+            Some("tests/test_projects/contract_storage_test/out/debug/contract_storage_test-storage_slots.json".to_string())
+        ),
+    )
+    .await?;
+
+    println!("Foo contract deployed @ {:x}", contract_id);
+
+    let key1 =
+        Bytes32::from_str("de9090cb50e71c2588c773487d1da7066d0c719849a7e58dc8b6397a25c567c0")
```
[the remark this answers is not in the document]
Salka1988: Yes, it will be.

packages/fuels-contract/src/contract.rs
Salka1988: You're right, maybe we should. I'll move it!

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -2034,3 +2039,44 @@ async fn can_use_try_into_to_construct_enum_from_bytes() -> Result<(), Error> {
...
+async fn test_init_storage_automatically() -> Result<(), Error> {
+    abigen!(
+        MyContract,
+        "packages/fuels-abigen-macro/tests/test_projects/contract_storage_test/out/debug/contract_storage_test-abi.json"
+    );
+
+    let wallet = launch_provider_and_get_wallet().await;
+
+    // ANCHOR: automatic_storage
+    let contract_id = Contract::deploy_with_parameters(
```
mohammadfawaz, line 2053: Will the `deploy()` function not allow passing in a path to the storage slots JSON? The reason I ask is because `forc new` generates a test harness that uses `deploy()` and we're probably going to have to change that to `deploy_with_parameters()` to minimize the amount of work users have to do. This means that most tests out there will eventually be using `deploy_with_parameters()` instead of `deploy()`.

docs/src/getting-started/basics.md
```diff
@@ -48,20 +48,33 @@ Alternatively, if you want multiple instances of the same contract then use `dep
...
-Once created, the slots can be passed to `deploy_with_parameters`: 
+Once created, the slots can be used to initialize `StorageConfiguration::with_manual_storage` which was then passed to `deploy_with_parameters`: 
 
 ```rust,ignore
 {{#include ../../../packages/fuels-abigen-macro/tests/harness.rs:manual_storage}}
 ```
 
+### Initializing storage slots automatically
+
+The storage slots of a contract can be initialized automatically from auto-generated JSON file where literals containing key-value pair are transformed into `Vec<StorageSlot>`.
```
MujkicA, line 66: IMO we should give the user more context on where the JSON file is coming from. We could also omit the details of how the read values are transformed. Maybe something like this?

```suggestion
The storage slots of a contract can be initialized automatically from the JSON file generated by the sway compiler.
```

docs/src/getting-started/basics.md
```diff
@@ -48,20 +48,33 @@ Alternatively, if you want multiple instances of the same contract then use `dep
...
+
+Path to JSON file is used to initialize `StorageConfiguration::with_storage_path` which was then passed to `deploy_with_parameters`:
+
+```rust,ignore
+{{#include ../../../packages/fuels-abigen-macro/tests/harness.rs:automatic_storage}}
+```
+
+### Initializing storage slots both automatically and manually
+
+If we initialize the slots both automatically and manually, then the generated `Vec<StorageSlot>` will be merged and the `manually` generated `Vec<StorageSlot>` takes precedence.
```
MujkicA, line 76: IIUC this requires the user to use `StorageConfiguration::new()`. I think we should demonstrate this or at least mention how this can be achieved

packages/fuels-abigen-macro/tests/harness.rs
Salka1988: Now we have it in both functions!

PR #443, opened 2022-06-30
refactor: use `InstantiationError` as a variant of `Error`

vnepveu:
- This PR makes the `InstantiationError` error a variant of the `Error` enum instead of being a standalone error.


packages/fuels-core/src/lib.rs
```diff
@@ -203,27 +201,27 @@ impl Default for Token {
...
 
-pub fn try_from_bytes<T>(bytes: &[u8]) -> Result<T, crate::errors::Error>
+pub fn try_from_bytes<T>(bytes: &[u8]) -> Result<T, Error>
 where
     T: Parameterize + Tokenizable,
 {
     let token = ABIDecoder::decode_single(&T::param_type(), bytes)?;
 
-    // Not inlined so that the error conversion InstantiationError -> Error
+    // Not inlined so that the error conversion Error -> Error
```
segfault-magnet, line 217: Since InstantiationError is now part of the 'Error', I think this can now be inlined? i.e.

```rust
let token = ABIDecoder::decode_single(&T::param_type(), bytes)?;

T::from_token(token)
```

packages/fuels-core/src/lib.rs
```diff
@@ -247,10 +245,10 @@ impl Tokenizable for bool {
 }
 
 impl Tokenizable for String {
-    fn from_token(token: Token) -> Result<Self, InstantiationError> {
+    fn from_token(token: Token) -> Result<Self, Error> {
         match token {
             Token::String(data) => Ok(data),
-            other => Err(InstantiationError(format!(
+            other => Err(Error::InstantiationError(format!(
```
segfault-magnet, line 251: In the Bool tokenization this was changed to return `InvalidData` while here it returns `InstantiationError` for the same cause.


I'm not sure which one is better, but we should definitely choose one.

segfault-magnet (approved):

PR #444, opened 2022-06-30
refactor!: move the `Error` type inside `fuels-types` workspace

vnepveu:
- Move the `Error` type inside `fuels-types` workspace, to use them in the crate, which is imported by `fuel-core`.

I'm not 100% sure this is a good move though. Feel free to criticize, I'm just putting it out there for discussion.


digorithm (approved):
I see no issues with this refactor at first glance. In fact, it makes sense, from an organizational perspective; types should be in `fuel_types`, which includes error types!

Keep in mind this is a breaking change though, people could've been using the `::core::errors:Error` exported in the prelude, which won't exist once this chance is in. I updated the title and labels to reflect that so we make sure to bump the version properly.

vnepveu:
Agreed that it makes much more sense in terms of coherence, my worry was more on the added `fuels-types` dependency everywhere. Maybe this should be a `fuels-core` re-export?
Mb about missing the breaking nature of this change, it's super important!
I opened issue to track those concerns.

PR #447, opened 2022-07-05
fix: abigen enum with string or array

### LINKED ISSUE #438: `"String(4)"` is not a valid identifier
When generating bindings for this contract:
```sway
contract;

enum SomeEnum {
	SomeNumber: u32,
	SomeStr: str[4]
}

abi MyContract {
    fn test_function(my_enum: SomeEnum) -> SomeEnum;
}

    fn test_function(my_enum: SomeEnum) -> SomeEnum {
        SomeEnum::SomeStr("asdf")
    }
}
```
`abigen!` panics with error:
```
error: proc macro panicked
 --> tests/harness.rs:5:1
  |
5 | abigen!(MyContract, "out/debug/what_about_strs-abi.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: `"String(4)"` is not a valid identifier
```

MujkicA:
Closes #438 

Prior to this, abigen would try to convert `String(4)` to `Ident` when trying to generate the `Tokenizable` trait for the custom enum defined in sway.

This PR introduces a special case for enum variant generation in abigen where `String` is converted to Ident, and `String(4)` is kept to generate `ParamType::String(4)` in `Parametrize` and `method_hash`.

This PR also introduces special case handling for arrays inside of enum variants since this led to similar errors.

The above approach is similar to how we handle strings and arrays in structs.

### DIFF
--- a/packages/fuels-abigen-macro/tests/harness.rs
+++ b/packages/fuels-abigen-macro/tests/harness.rs
@@ -1996,3 +1996,40 @@ async fn can_use_try_into_to_construct_enum_from_bytes() -> Result<(), Error> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn string_and_array_inside_enum() -> Result<(), Error> {
+    abigen!(
+        MyContract,
+        "packages/fuels-abigen-macro/tests/test_projects/string_and_array_inside_enum/out/debug\
+        /string_and_array_inside_enum-abi.json"
+    );
+
+    let wallet = launch_provider_and_get_wallet().await;
+
+    let id = Contract::deploy(
+        "tests/test_projects/string_and_array_inside_enum/out/debug/string_and_array_inside_enum.bin",
+        &wallet,
+        TxParameters::default(),
+    )
+        .await?;
+
+    let instance = MyContract::new(id.to_string(), wallet.clone());
+
+    let enum_string = SomeEnum::SomeStr("asdf".to_owned());
+    let enum_array = SomeEnum::SomeArr(vec![1,2,3,4]);
+
+    let response = instance
+        .get_enum_str(enum_string.clone())
+        .call()
+        .await?;
+    assert_eq!(response.value, enum_string);
+
+    let response = instance
+        .get_enum_arr(enum_array.clone())
+        .call()
+        .await?;
+    assert_eq!(response.value, enum_array);
+
+    Ok(())
+}
--- a/packages/fuels-abigen-macro/tests/test_projects/string_and_array_inside_enum/src/main.sw
+++ b/packages/fuels-abigen-macro/tests/test_projects/string_and_array_inside_enum/src/main.sw
@@ -1,0 +1,21 @@
+contract;
+
+enum SomeEnum {
+	SomeStr: str[4],
+	SomeArr: [u64; 4]
+}
+
+abi MyContract {
+    fn get_enum_str(my_enum: SomeEnum) -> SomeEnum;
+    fn get_enum_arr(my_enum: SomeEnum) -> SomeEnum;
+}
+
+impl MyContract for Contract {
+    fn get_enum_str(my_enum: SomeEnum) -> SomeEnum {
+        my_enum
+    }
+
+    fn get_enum_arr(my_enum: SomeEnum) -> SomeEnum {
+        my_enum
+    }
+}
--- a/packages/fuels-core/src/code_gen/custom_types_gen.rs
+++ b/packages/fuels-core/src/code_gen/custom_types_gen.rs
@@ -261,19 +261,44 @@ pub fn expand_custom_enum(enum_name: &str, prop: &Property) -> Result<TokenStrea
             // Elementary type
             _ => {
                 let ty = expand_type(&param_type)?;
-                let param_type_string = ident(&param_type.to_string());
+
+                let mut param_type_string = param_type.to_string();
+
+                let param_type_string_ident_tok: proc_macro2::TokenStream =
+                    param_type_string.parse().unwrap();
+
+                param_types.push(quote! { types.push(ParamType::#param_type_string_ident_tok) });
+
+                if let ParamType::Array(..) = param_type {
+                    param_type_string = "Array".to_string();
+                }
+                if let ParamType::String(..) = param_type {
+                    param_type_string = "String".to_string();
+                }
+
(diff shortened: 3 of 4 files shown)


packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -261,19 +261,44 @@ pub fn expand_custom_enum(enum_name: &str, prop: &Property) -> Result<TokenStrea
             // Elementary type
```
vnepveu, line 261: Document that this includes the "weird" `String(x)` Sway type and the `Arrays` because it is not what someone would immediately think about when reading "elementary type"

packages/fuels-abigen-macro/tests/test_projects/string_and_array_inside_enum/src/main.sw
```diff
@@ -0,0 +1,21 @@
+contract;
```
vnepveu, line 1: refactor to include inside the `xx_inside_enum` so we have one "big project" that has all the corner cases instead of several small ones. We are starting to have to many small test projects.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -1996,3 +1996,34 @@ async fn can_use_try_into_to_construct_enum_from_bytes() -> Result<(), Error> {
...
+    let enum_string = SomeEnum::SomeStr("asdf".to_owned());
+    let enum_array = SomeEnum::SomeArr(vec![1, 2, 3, 4]);
+
+    let response = instance.get_enum_str(enum_string.clone()).call().await?;
+    assert_eq!(response.value, enum_string);
+
+    let response = instance.get_enum_arr(enum_array.clone()).call().await?;
+    assert_eq!(response.value, enum_array);
+
+    Ok(())
```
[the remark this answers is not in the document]
MujkicA: AFAICS `enum_inside_struct` doesn't modify the incoming enum. 
Also, this is not an encoding issue. It's about how abigen expands custom enums.

packages/fuels-abigen-macro/tests/test_projects/string_and_array_inside_enum/src/main.sw
vnepveu: In that particular case it just makes much more sense to have every `xxx_inside_enum` inside one project, that would be the "goto" place for that kind of issue. It makes the directory structure less hell to navigate. But you're right that it shouldn't be an optimisation criteria, but here I'm deeply conviced it makes a lot of sense.

vnepveu (changes requested):
I think I'd like to see the tests be unit tests on the encoding, because from what you said it's a purely Abigen problem. So you don't even need a sway project, just need to show that `ABI -> Rust ` translation happens well.

MujkicA:
> I think I'd like to see the tests be unit tests on the encoding, because from what you said it's a purely Abigen problem. So you don't even need a sway project, just need to show that `ABI -> Rust ` translation happens well.

Unit tests are a good idea. However we still need the E2E test since the unit test just confirms that some Rust code is generated, not that it implements the required functionality.

packages/fuels-abigen-macro/tests/harness.rs
```diff
@@ -2075,6 +1996,64 @@ async fn can_use_try_into_to_construct_enum_from_bytes() -> Result<(), Error> {
     Ok(())
 }
 
+#[tokio::test]
+async fn type_inside_enum() -> Result<(), Error> {
```
vnepveu, line 2000: Beautiful :heart_eyes:

vnepveu (approved):
LGTM! Love the new test project, super clean

packages/fuels-core/src/code_gen/custom_types_gen.rs
```diff
@@ -258,22 +259,41 @@ pub fn expand_custom_enum(enum_name: &str, prop: &Property) -> Result<TokenStrea
                 param_types.push(quote! { types.push(ParamType::Unit) });
                 args.push(quote! {(#dis, token, _) => Ok(#enum_ident::#variant_name()),});
             }
-            // Elementary type
+            // Elementary types, String, Array
             _ => {
                 let ty = expand_type(&param_type)?;
```
segfault-magnet, line 264: Tried to make this a bit cleaner:
```rust
let param_type_construction = gen_param_type_constructing_code(&param_type)?;
param_types.push(quote! { types.push(#param_type_construction) });

let ty = expand_type(&param_type)?;
enum_variants.push(quote! { #variant_name(#ty)});

args.push(quote! {(#dis, token, _) => Ok(#enum_ident::#variant_name(<#ty>::from_token(token)?)),},);

let token = gen_token_constructing_code(&param_type);
enum_selector_builder.push(quote! {
  #enum_ident::#variant_name(value) => (#dis,  #token)
 });
```
And the helpers:
```rust
fn gen_param_type_constructing_code(param_type: &ParamType) -> Result<TokenStream, LexError> {
    let variant = TokenStream::from_str(&param_type.to_string())?;
    Ok(quote! {ParamType::#variant})
}
fn gen_token_constructing_code(param_type: &ParamType) -> TokenStream {
    let init_value = if let ParamType::Array(..) = param_type {
        quote! { vec![value.into_token()] }
    } else {
        // Primitive type
        quote! { value }
    };

    let token_type = determine_token_variant(&param_type);
    quote! {Token::#token_type(#init_value)}
}
fn determine_token_variant(param_type: &ParamType) -> Ident {
    let token_variant = match param_type {
        ParamType::Array(..) => "Array".to_string(),
        ParamType::String(..) => "String".to_string(),
        _ => param_type.to_string(),
    };
    ident(&token_variant)
}
```

Even more could be achieved in terms of readability but it would require more refactoring outside the scope of your changes.

segfault-magnet: Now that you mention it, it does seem like this could be generalized. The ones above 100% work only for enums, didn't check if the other cases have anything different with regards to the parameters and tokens construction.

vnepveu: Yes that's what I was hinting at in my previous comment. It's definitely a good tech debt to fix, but it should be done in a separate PR.

Issue #365, opened 2022-06-06
Fuel Rust SDK book re-org


digorithm:
# Problem

Currently, the `fuels-rs` book, although it has plenty of information, it is poorly organized, making it really hard to find information as a user, especially as a beginner. We have been basically dumping all we know about it without much thought, which is fine for starters but has grown to be hard to use.

It also lacks _some_ sections, such as types (i.e. how do my types defined in Sway show up in rust-land? How about Sway-native types?), providers, and more fundamentals.

# Proposal

To begin addressing this, we must first rethink the outline of the documentation. Then, once we have each chapter/section defined, we can work in parallel.

Here's the initial re-org of the documentation, I'll update it as we discuss:

- [x] Getting started
  - [x] Setup
  - [x] Terminology (briefly explains contracts, providers, wallets/signers, etc.)
  - [x] Contracts (the very basics of contracts in the SDK, no advanced usage)
  - [x] Wallets and signing 
  - [x] Types (explains how `abigen!` works, at a high level, and how it generates the types coming from Sway contracts)
- [x] API (I'll will actually point to the rust docs and improve these rustdocs)
  - [x] Providers
  - [x] Wallets
  - [x] Contracts (focus on configuration stuff such as txparams, call params, contracts calling other contracts, and more)
  - [x] Utilities and helpers (more in-depth around the test helpers, logging, encoding/decoding data, etc.)
- [ ] Cookbook (more complete examples, probably coming from bigger examples in the `examples/` crate)
- [ ] Debugging (a whole chapter dedicated to some debugging practices we use)
  - [ ] Debugging abigen errors (talk about using `cargo-expand` to expand macros, etc)
  - [ ] Debugging contract call errors (network call worked, but contract code failed in the VM)
  - [ ] Debugging network call errors
- [ ] Testing

Once we settle with an outline, I'll turn these into checkboxes and we can work on each of them at the appropriate pace.

cc @vnepveu @MujkicA @segfault-magnet @Salka1988 @hal3e @Br1ght0ne @adlerjohn @SilentCicero

vnepveu:
I really like this first outline! Seems progressive enough.

adlerjohn:
Looks good!

Issue #294, opened 2022-05-18
Document in the fuels-rs book how to pass in data into `sway` functions when calling them via the `harness.rs` file


Braqzen:
When writing tests for `sway` contracts a developer might need to pass in specific types that do not exist in `Rust` e.g. the `b256` type.
That type in `Rust` is `[u8; 32]`, similarly if a `sway` function signature takes in an `enum` that argument comes to a `[u8; 16]`.

This information, among the other types, should be documented in the book starting from the basics of the Endianness to the actual values e.g. a `bool` is big endian and when it has a truth value of true its array representation is `[0, 0, 0, 0, 0, 0, 0, 1]`.

PR #453, opened 2022-07-07
The big Fuel SDK Book re-org

digorithm:
How it was: 

image

How it is now:

image

This addresses many, many things doc-related and partially resolves https://github.com/FuelLabs/fuels-rs/issues/365. 

The main issue with the current documentation is information accessibility; it's very, very hard to have an overview of the whole SDK and find specific things. This re-org helps with that in way that most important points of the SDK are exposed in the sidebar.

These changes also add many more code examples, with open opportunities to add many more. Some pages are still a WIP, but I _had_ to get this out now because it was getting too big (uh oh, the PR is already too big...).

I also added missing important information that's very sought after by newcomers, e.g., connecting to external nodes, how to use the `Provider`, and many `Contract` related techniques. Yet, there are still many more to add. We'll get there.

The sections/chapters outline is also done here and we can slowly add more to it. 

Also, closes https://github.com/FuelLabs/fuels-rs/issues/294.


docs/src/getting-started/contracts.md
```diff
@@ -0,0 +1,34 @@
...
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:use_deployed_contract}}
+```
+
+Alternatively, if you want multiple instances of the same contract then use `deploy_with_salt`
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:deploy_with_salt}}
+```
```
vnepveu, line 34: This appears to be broken

docs/src/contracts/the-fuelvm-binary-file.md
```diff
@@ -0,0 +1,52 @@
...
+        true
+    }
+}
+```
+
+After `forc build`, will have a binary file that contains:
+
+```terminal
+cat out/debug/my-test.bin
+G4]�]D`I]C�As@
```
vnepveu, line 23:
```suggestion
~$ cat out/debug/my-test.bin
~$ G4]�]D`I]C�As@
```

docs/src/contracts/the-fuelvm-binary-file.md
```diff
@@ -0,0 +1,52 @@
...
+          7   28     LW(16, 63, 1)     5d 43 f0 01
+          8   32     EQ(16, 17, 16)    13 41 14 00
+          9   36     JNZI(16, 11)      73 40 00 0b   conditionally jump to byte 44
+         10   40     RVRT(0)           36 00 00 00
+         11   44     LW(16, 63, 0)     5d 43 f0 00
+         12   48     RET(16)           24 40 00 00
+         13   52     Undefined         00 00 00 00
+         14   56     Undefined         00 00 00 01
+         15   60     Undefined         00 00 00 00
+         16   64     XOR(20, 27, 53)   21 51 bd 4b
```
vnepveu, line 49:
```suggestion
~$ forc parse-bytecode out/debug/my-test.bin
~$ half-word   byte   op                raw           notes
           0   0      JI(4)             90 00 00 04   jump to byte 16
           1   4      NOOP              47 00 00 00
           2   8      Undefined         00 00 00 00   data section offset lo (0)
           3   12     Undefined         00 00 00 34   data section offset hi (52)
           4   16     LW(63, 12, 1)     5d fc c0 01
           5   20     ADD(63, 63, 12)   10 ff f3 00
           6   24     LW(17, 6, 73)     5d 44 60 49
           7   28     LW(16, 63, 1)     5d 43 f0 01
           8   32     EQ(16, 17, 16)    13 41 14 00
           9   36     JNZI(16, 11)      73 40 00 0b   conditionally jump to byte 44
          10   40     RVRT(0)           36 00 00 00
          11   44     LW(16, 63, 0)     5d 43 f0 00
          12   48     RET(16)           24 40 00 00
          13   52     Undefined         00 00 00 00
          14   56     Undefined         00 00 00 01
          15   60     Undefined         00 00 00 00
          16   64     XOR(20, 27, 53)   21 51 bd 4b
```

docs/src/contracts/the-json-abi-file.md
```diff
@@ -0,0 +1,42 @@
...
+    "name": "test_function",
+    "outputs": [
+      {
+        "name": "",
+        "type": "bool",
+        "components": null
+      }
+    ]
+  }
+]
```
vnepveu, line 38:
```suggestion
$ cat out/debug/my-test-abi.json
$ [
    {
      "type": "function",
      "inputs": [],
      "name": "test_function",
      "outputs": [
        {
          "name": "",
          "type": "bool",
          "components": null
        }
      ]
    }
  ]
```

docs/src/calling-contracts/call-response.md
```diff
@@ -0,0 +1,25 @@
+# Call response
+
+You've probably noticed that you're often chaining `.call().await.unwrap()`. That's because:
```
vnepveu, line 3: Should we encourage using `unwrap` instead of `?`?

docs/src/calling-contracts/read-only.md
```diff
@@ -0,0 +1,15 @@
+# Read-only calls
```
vnepveu, line 1: Not sure that `Read only` is the proper way to call this. Can be decided later though.

vnepveu (changes requested):
Looking geat!!! Generally I think you can start the anchors 1 line below, and end them one line above so we don't see the name of the test function. This is a huge step forward for our docs :heart_eyes:

docs/src/calling-contracts/read-only.md
digorithm: I think so too, couldn't think of anything better for now so I kept the current naming.

docs/src/calling-contracts/call-params.md
```diff
@@ -0,0 +1,39 @@
+# Call parameters
+
+Call parameters are:
+
+1. Amount;
+2. Asset ID;
+3. Gas forwarded.
+
+You can use these to forward coins to a contract. You can configure these parameters by creating an instance of [`CallParameters`](https://github.com/FuelLabs/fuels-rs/blob/adf81bd451d7637ce0976363bd7784408430031a/packages/fuels-contract/src/parameters.rs#L15) and passing it to a chain method called `call_params`.
```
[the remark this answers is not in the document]
digorithm: Actually, we should point it to the actual docs, which will be up-to-date. I'd rather not use anchor here because it would be placing the anchor in a non-example/non-test piece of code. Point to [this](https://docs.rs/fuels/latest/fuels/core/parameters/struct.CallParameters.html#) should be fine.

docs/src/providers/querying.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## Get all coins from an address
+
+This method returns all coins from a wallet, including spent ones.
+
+```rust,ignore
+{{#include ../../../examples/providers/src/lib.rs:get_coins}}
+```
+
```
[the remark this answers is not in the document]
digorithm: I don't know, I'd rather _not_ assume what the user will or will not do with the SDK. I believe we should make all public APIs in the SDK easily visible/searchable. What if someone is trying to build a chain scrapper? They would want to know about this. And since this is in fact querying the chain, it should be in `querying.md` IMO.

docs/src/getting-started/contracts.md
```diff
@@ -0,0 +1,34 @@
+# Deploying contracts
+
+There are two main ways of working with contracts in the SDK: deploying a contract with SDK or using the SDK to connect to existing contracts.
```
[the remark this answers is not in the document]
digorithm: "already-deployed" is a mouthful IMO haha. How about just "deployed"? So it would be "using the SDK to interact with deployed contracts." We replace `connect` with `interact`, and it should be inferrable that, by using this, you're only skipping the deployment step.

docs/src/calling-contracts/call-params.md
```diff
@@ -6,7 +6,7 @@ Call parameters are:
 2. Asset ID;
 3. Gas forwarded.
 
-You can use these to forward coins to a contract. You can configure these parameters by creating an instance of [`CallParameters`](https://github.com/FuelLabs/fuels-rs/blob/adf81bd451d7637ce0976363bd7784408430031a/packages/fuels-contract/src/parameters.rs#L15) and passing it to a chain method called `call_params`.
+You can use these to forward coins to a contract. You can configure these parameters by creating an instance of [`CallParameters`](https://docs.rs/fuels/latest/fuels/core/parameters/struct.CallParameters.html#) and passing it to a chain method called `call_params`.
```
vnepveu, line 9: Ah yes nice! Much better than linking to a code line.

vnepveu (approved):
:shipit: 
LOVE IT

adlerjohn (approved):
I have some nits but I can file issues for later instead of stalling review. Big improvement.

PR #457, opened 2022-07-08
refactor!: move `ParamType` type inside `fuels-types`

vnepveu:
This makes more sense this way and is a refactoring needed for further work on the abigen macro


digorithm:
LGTM, but the branch is outdated!

digorithm (approved):
:shipit:

Issue #277, opened 2022-05-12
Sway function named 'new' leads to collision


MujkicA:
The abigen macro generates a `new` method for creating contract instances. If the ABI includes a function named `new` this will result in a naming collision.

hal3e:
I had a look at this error. Rust doesn't support overloaded functions/methods, so whatever we choose for the name a collision can still happen.

``` rust    
impl MyContract {
        pub fn new(contract_id: String, wallet: LocalWallet) -> Self {
            let contract_id = ContractId::from_str(&contract_id).unwrap();
            Self {
                contract_id,
                wallet,
            }
        }
        ///Calls the contract's `new` (0x00000000b833a3fa) function
        pub fn new(&self, input: (u64, bool)) -> ContractCall<()> {
```
@digorithm could we move this function outside of the `impl MyContract` and give it some generic name like `new_contract`?

Another option would be that `sway` doesn't allow this function name.

digorithm:
> @digorithm could we move this function outside of the impl MyContract and give it some generic name like new_contract?

I _think_ you need that to be generated here, it's how the `new` in `let instance = MyContract::new(id.to_string(), wallet);` is generated... Unless you have a different idea? 

This treatment definitely needs to be on the SDK though, Sway itself not allowing that name would be a bit weird.

segfault-magnet:
@digorithm this seems to be deadlocked.

It seems to me that whatever name we choose there is nothing stopping Sway developers from creating a function with the same name (since naming rules are very similar/identical between rust and sway). 

Another way might be to do what @hal3e suggested and create a new struct that will `impl` the API of the contract:
```rust
instance.caller.new();
```
Where caller will be a struct which will delegate back to `MyContract` through a reference if needed. It will be constructed by `MyContract` through the `SomeStruct{}` syntax and given a reference to itself, like a [method object](https://refactoring.guru/replace-method-with-method-object).

Or something like that. Somehow that feels not worth the hassle. 

I suggest we keep our own `new` as a way to initialize the contract but change the method binding name to something like _new (like safe_ident does) to avoid all identifiers already present in `MyContract`. 

We will then proceed to document this behavior so as not to come as a surprise.

@digorithm Thoughts?

PR #458, opened 2022-07-09
Collision with rust bindings if Sway fn is called `new`

segfault-magnet:
closes: #277 

### The issue
The struct in whose `impl` block we generate bindings for calling the contract functions also has an associated function called `new` that is used to initialize the struct with data needed for the actual calls -- i.e. the wallet and the contract id.

Since Rust doesn't allow for fn overload, there can only ever be one `new` inside that struct.

### Possible approaches
I'll list the ideas that came to mind on how to solve this. They are ordered starting from the one I favor most.

**Approach 1: Do nothing**
Accept that we don't support contracts that have `new` as a function.

**Approach 2: The Builder pattern**
For each contract generate two structures, for example:
`MyContract` and `MyContractBuilder`. 

`MyContract` will hold all of the bindings through which you can call your contract's functions.

`MyContractBuilder` will be used to construct an instance of `MyContract`:
```rust
let my_contract: MyContract = MyContractBuilder::new(id, wallet).build();
```

Pros: 

- Users will recognize this pattern.

Cons:

- None of the traditional reasons for using the builder pattern. It is used here just to eliminate the `new` in `MyContract`.
- Breaking change.

**Approach 3: Have `new` return something other than `Self`**
Here we create two structs `MyContract` which will only have the `new` fn, and `MyContractInstance` which will contain the sway function bindings.

`MyContract::new` will, instead of returning an instance of `MyContract`, it will return an instance of `MyContractInstance`:
```rust
let my_contract: MyContractInstance = MyContract::new(id, wallet); 
```

Pros: 

- Non breaking due to type inference. (Technically it is breaking..but less of a pain to fix)

Cons:

- Unheard of, that `new` should return something other than Self. I.e. unexpected. Maybe if we call it a factory and return a trait object :D Java flashbacks.
- Difficulty finding a good name for `MyContractInstance`.

**Approach 4: Rename collisions**
If we have a collision then just rename the rust binding of the sway fn -- i.e. call it new_():
```rust
let instance = MyContract::new(id, wallet);
let result = instance.new_().call().await?;
```

Pros:

- Non breaking

Cons:

- What if you also had a sway fn called `new_`? Would you call the original `new__` ? What if you also had `new__`...? 
- If you have Sway functions of `new` and `new_`. Will you bind the second `new_` as is and change `new` to `new__` ? Or will you bind `new` as `new_` and change `new_` to `new__`? I.e. do you first bind already valid names and skip over the invalid ones, then do another pass or just do it all at once. Hell to document and probably to use.

**Approach 5: Field inside `MyStruct`**
Something like this:
```rust
let instance = MyContract::new(id, wallet);
let rtn = instance.functions.new().call().await?.value;
```
where `functions` is a public field of some type that will hold all of the sway function bindings and have no `new` of its own.

Pros:
- ...
Cons:
- Ugly stepping stone

**Approach 6: Struct init syntax**
Have the fields of MyContract made public so you can use:
```rust
let my_contract: MyContract = MyContract { wallet: some_wallet, id: some_id };
``` 

Pros:

- Recognizable to users.

Cons:
- Users will no longer be able to pass a `String` as a contract id but would rather have to create a `ContractId` themselves:
```rust
   let contract_id = ContractId::from_str(&some_stringy_id).expect("Invalid contract id");
   let my_contract = MyContract { contract_id ... }
```
- We will no longer be able to do any validation on the parameters passed to `MyContract`.
- Users might change the underlying fields whenever they wish, possibly to invalid values.
- Loss of encapsulation
- Breaking change

**Approach 7: Dynamic function dispatch**
Don't generate bindings for reserved identifiers but rather offer a mechanism to let the user call a function without type safety:
```rust
let instance = MyContract::new(id, wallet);
let result: Token = instance.dyn_call("new", vec![some_tokenizable_value, some_other_tokenizable_value]).call().await?.value;
let real_result: SomeStruct = SomeStruct::from_token(result);
```

Pros:
* non breaking
* would work for all sorts of forbidden names, especially if there are some sway finds ok but rust doesn't

Cons:
* I mean... total overkill and probably a pain to implement.


vnepveu:
Wait so I don't understand, which approach did you go with? Shouldn't this discussion be in the issue, so we can decide before doing a PR?

digorithm:
> Wait so I don't understand, which approach did you go with? Shouldn't this discussion be in the issue, so we can decide before doing a PR?

We went with approach number 2 (you can see in the PR's code that he's using `.build()`). There was a lot of back-and-forth in mixed channels, including in the [issue itself](https://github.com/FuelLabs/fuels-rs/issues/277). 

The PR itself is the suggestion; it's short enough that we can make changes if it's not the best approach. But I believe 2 (`.build()`) is the least painful way to achieve this. And we (@segfault-magnet and I) thought about this quite a bit! But, as always, feel free to explore and suggest better approaches if applicable.

vnepveu:
Ok! I just thought that the summary of the solutions was the discussion itself, didn't realize it already happened. I think it would be better to document the suggested solutions in the issue itself? Rather than it getting lost in a PR. Will review soon.

vnepveu (approved):
LGTM! But maybe in terms of semantics (and I think it fits well with reality too!), maybe calling it `Generator` or something would be better? So people don't get confused by a `build()` method (which is a pattern in itself iirc), but call `generate()` which might be less confusing? Just spitballing.

segfault-magnet:
> LGTM! But maybe in terms of semantics (and I think it fits well with reality too!), maybe calling it `Generator` or something would be better? So people don't get confused by a `build()` method (which is a pattern in itself iirc), but call `generate()` which might be less confusing? Just spitballing.

Normally I'd be ok with Generator, but in this case, I prefer Builder because:
1. We need to generate another struct where we will move our 'constructing' functions. The same is done in the builder pattern -- i.e. you have a `SomeThingBuilder`. 
2. The Builder pattern will be recognizable to users -- they'd be like 'oh, you don't instantiate contracts directly, you use a builder' -- only because of this familiarity I've added the `contract_id` and `wallet` methods to make it look like the builder pattern even further. The only thing that is missing is to have a `builder() -> SomeContractBuilder` associated function in `SomeContract` -- we can't have that since that would bring us back to square 1 -- having a reserved name we can't use in contracts. If we do a Generate thingy we'd be doing something novel -- which isn't bad per se, but why not make it more familiar.

vnepveu (approved):
Good with me then! :shipit:

vnepveu:
Make sure to do a nice explanation in the squash commit message :smile:

PR #460, opened 2022-07-10
ci: use forc v0.18

vnepveu:


digorithm (approved):
:shipit: 

So that confirms that the changes introduced in `forc` `0.18` (type parameters in the JSON ABI) didn't break anything, right?

Issue #376, opened 2022-06-08
Add bech32 support


adlerjohn:
All addresses that an end-user sees must be bech32. SDK methods that are meant for use in anything user-facing should accept bech32 addresses.

PR #467, opened 2022-07-13
feat: add bech32 support

MujkicA:
Closses #376 

On the SDK side we want to use addresses in the bech32m format (see [BIP 0173](https://en.bitcoin.it/wiki/BIP_0173), [BIP 0350](https://en.bitcoin.it/wiki/BIP_0350)) The core client works with plain addresses so we need to convert bech32 addresses behind the scenes when querying a provider. The same applies for contract ids.

This PR adds a macro to generate the bech32 types `Bech32Address` and `Bech32ContractId`. Albeit these types have the same interface, it is important that we keep them separate because addresses and contract ids live in different spaces.

Two small tests are included just as sanity check since the `bech32` crate already covers encoding/decoding tests.
User-facing, `wallet` and `provider` methods that previously used `Address` and 'ContractId' now accept their bech32 equivalents

**Contract methods with Address/ContractId**
Contract methods that have `ContractId` and `Address` as input/output still require the user to pass the native equivalents in the SDK. We could parse this and have the SDK bindings use bech32 instead. However, I decided against this to reduce obfuscation.


packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,102 @@
+use crate::errors::Error;
+use bech32::Variant::Bech32m;
+use bech32::{FromBase32, ToBase32};
+use core::str;
+use fuel_tx::Address;
+use std::fmt::{Display, Formatter};
+
+pub const FUEL_BECH32_HRP: &str = "fuel";
```
vnepveu, line 8: I think a comment explaining what this constant is/means would be nice, I'm not sure what it is.

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,102 @@
...
+
+#[derive(Debug, Clone)]
+pub struct Bech32Address {
+    plain_addr: Address,
+    encoding: String,
+}
+
+impl Bech32Address {
+    pub fn new(hrp: &str, hash: [u8; 32]) -> Self {
```
vnepveu, line 17: I think replacing `hash` with `pubkey_hash` or something would make it more clear.

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl str::FromStr for Bech32Address {
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (_, hash_base32, _) = bech32::decode(s)?;
+
+        let hash: [u8; Address::LEN] = Vec::<u8>::from_base32(&hash_base32)
+            .unwrap()
```
vnepveu, line 52: I think you can use `?` here

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+impl Bech32Address {
+    pub fn new(hrp: &str, hash: [u8; 32]) -> Self {
+        let data_base32 = hash.to_base32();
+        let encoding = bech32::encode(hrp, &data_base32, Bech32m).unwrap();
+
+        Self {
+            plain_addr: Address::from(hash),
+            encoding,
```
vnepveu, line 23: I don't see the `encoding` field being used, can you please document what it represents.

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,102 @@
...
+use bech32::{FromBase32, ToBase32};
+use core::str;
+use fuel_tx::Address;
+use std::fmt::{Display, Formatter};
+
+pub const FUEL_BECH32_HRP: &str = "fuel";
+
+#[derive(Debug, Clone)]
+pub struct Bech32Address {
+    plain_addr: Address,
```
vnepveu, line 12: if `plain_addr` is pretty clear, I don't know what `encoding` corresponds to. For a reader unfamiliar with BECH32 like me, I feel like it's hard to understand what this type does "differently" than our old Address.

packages/fuels-signers/src/wallet.rs
```diff
@@ -407,8 +408,8 @@ impl Signer for Wallet {
         Ok(sig)
     }
 
-    fn address(&self) -> Address {
-        self.address
+    fn address(&self) -> &Bech32Address {
```
vnepveu, line 411: Why are we returning a reference type now?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -62,7 +63,7 @@ pub use wallets_config::*;
 /// pre-existing coins, with `num_asset` different asset ids. Note that one of the assets is the
 /// base asset to pay for gas.
 pub fn setup_multiple_assets_coins(
-    owner: Address,
+    owner: &Bech32Address,
```
vnepveu, line 66: why is it a reference type

packages/fuels-types/src/bech32.rs
vnepveu: What does the `hrp` correspond to? From looking up I know it means "human-readable part" but I'm not sure I understand what it does

vnepveu (changes requested):
Overall no remarks on the implementation except that I don't understand why types were switched to reference types: is that a necessity? Feel cumbersome.  Otherwise,  I'm not sure I understand everything. I don't really understand the difference with Address, and feel like it's not documented. It might be because I don't know what BECH32 is but I still think it should be loosely explained in the implementation.  
Also not sure I understood what the impact on UX is, shouldn't we just alias the `Address` from `fuel-tx` to something else, and use `Address` as our basic address? Since it's supposed to be our default address iiuc?

packages/fuels-types/src/bech32.rs
MujkicA: One benefit of bech32 is error detection.

Given that I understood it correctly,  the encoding guarantees a distance of 5 characters between every valid address. This means you can detect a typo of up to 4 characters and report it to the user.

Checkout [BIP 0173](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#rationale)

packages/fuels-signers/src/wallet.rs
MujkicA: `Address` implemented the `Copy` trait, so passing it to functions didn't transfer ownership. 
I opted for a reference to avoid copying memory.

adlerjohn (commented):
I'm not sure "bech32**address**" is the best term. This encoding could be used for both addresses and contract IDs.

MujkicA:
> I'm not sure "bech32**address**" is the best term. This encoding could be used for both addresses and contract IDs.

I didn't consider this. Do we want to use it for both?

If so, here is a few ways we could implement it:

**1.** 
Just rename it to `Bech32` and have the following methods ` to_address()`, `to_contract_id()`.
- PRO: straightforward UI
- CON: no way to know what the actual underlying data is

**2.**
Rename it, have an enum field with variants `ContentType::Address` and `ContentType::ContractId`.
- PRO: the user can check what data is encoded
-CON: one more parameter when initializing from string e.g. `Bech32::from_string("...", ContentType::Address)`

**3**
Make it hold a generic `Bech32<T>`  and implement it for the fuel_type `Address` and `ContractId`

- PRO: more rust like I guess
- CON: extra info on the UI side needed, e.g. `Bech32<Address>::from_str(...)`, and we still expose the fuel_types

cc @digorithm

digorithm:
> > I'm not sure "bech32**address**" is the best term. This encoding could be used for both addresses and contract IDs.
> 
> I didn't consider this. Do we want to use it for both?
> 
> If so, here is a few ways we could implement it:
> 
> **1.** Just rename it to `Bech32` and have the following methods ` to_address()`, `to_contract_id()`.
> 
> * PRO: straightforward UI
> * CON: no way to know what the actual underlying data is
> 
> **2.** Rename it, have an enum field with variants `ContentType::Address` and `ContentType::ContractId`.
> 
> * PRO: the user can check what data is encoded
>   -CON: one more parameter when initializing from string e.g. `Bech32::from_string("...", ContentType::Address)`
> 
> **3** Make it hold a generic `Bech32<T>` and implement it for the fuel_type `Address` and `ContractId`
> 
> * PRO: more rust like I guess
> * CON: extra info on the UI side needed, e.g. `Bech32<Address>::from_str(...)`, and we still expose the fuel_types
> 
> cc @digorithm

I like number 2. Instead of passing `ContentType::<type>` to `from_string` could we have two separate methods `new_address_from_string` and `new_contract_id_from_string`? IMO it would be easier to remember or look up the method names than thinking about passing enums as an argument.

MujkicA:
> I like number 2. Instead of passing `ContentType::<type>` to `from_string` could we have two separate methods `new_address_from_string` and `new_contract_id_from_string`? IMO it would be easier to remember or look up the method names than thinking about passing enums as an argument.

Number 2. is implemented. I changed the PR description btw.

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,154 @@
...
+pub enum ContentType {
+    Address,
+    ContractId,
+}
+
+/// An address in the bech32 format consisting of a
+/// human-readable part (HRP), '1' as a separator, the
+///  public key hash encoded with a u5 charset, and a checksum.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Bech32 {
```
segfault-magnet, line 20: A more type strong approach will allow you to write code as follows:
```rust
        let bech32 = Bech32::Address("fuel1the_hash_part".parse()?);
        println!("{}", bech32);
        let address: Address = bech32.into();

        let bech32 = Bech32::ContractId("fuel1the_hash_part".parse()?);
        println!("{}", bech32);
        let contract_id: ContractId = bech32.into();
```

the implementation might go as follows:

Have `Bech32Data` be just a simple record-like struct:
```rust
pub struct Bech32Data {
    pub hrp: String,
    pub hash: Bytes32,
}
```
Like the real-world counterpart, it has a `hrp` and the hash behind it.

Make it so that it can be converted into `ContractId` and `Address`:
```rust
impl From<Bech32Data> for Address {
    fn from(data: Bech32Data) -> Self {
        Self::new(*data.hash)
    }
}
impl From<Bech32Data> for ContractId {
    fn from(data: Bech32Data) -> Self {
        Self::new(*data.hash)
    }
}
```
Implement `FromStr` to be able to parse a string into `Bech32Data`:
```rust
impl FromStr for Bech32Data {
    type Err = Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let (hrp, pubkey_hash_base32, _) = bech32::decode(s)?;

        let pubkey_hash: [u8; Address::LEN] = Vec::<u8>::from_base32(&pubkey_hash_base32)?
            .as_slice()
            .try_into()?;

        Ok(Self {
            hrp,
            hash: Bytes32::from(pubkey_hash),
        })
    }
}
```

Don't cache the data for the needs of `Display` -- compute it every time (unless benchmarking shows this to be too prohibitive):
```rust
impl Display for Bech32Data {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        let data_base32 = self.hash.to_base32();
        let encoding = bech32::encode(&self.hrp, &data_base32, Bech32m).unwrap();

        write!(f, "{}", encoding)
    }
}
```

Don't use an additional enum for types, use types for types:
```rust
struct Bech32Address(Bech32Data);
```
Make it so that you can construct both an `Address` and `ContractId` from it:
```rust
impl From<Bech32Address> for Address {
    fn from(address: Bech32Address) -> Self {
        address.0.into()
    }
}
impl From<Bech32Address> for ContractId {
    fn from(address: Bech32Address) -> Self {
        address.0.into()
    }
}
```

it can also be parsed from a string:
```rust
impl FromStr for Bech32Address {
    type Err = Error;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        Ok(Bech32Address(s.parse()?))
    }
}
```
Also make it display-able:
```rust
impl Display for Bech32Address {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.0)
    }
}
```
Repeat the above for `Bech32ContractId`.

Then unite them in a `Bech32` enum so you can pass both around through one type:
```rust
enum Bech32 {
    Address(Bech32Address),
    ContractId(Bech32ContractId),
}
```

preferably add some qol methods for extracting the hash and hrp:
```rust
impl Bech32 {
    pub fn hash(&self) -> Bytes32 {
        match self {
            Bech32::Address(address) => address.0.hash,
            Bech32::ContractId(contract_id) => contract_id.0.hash,
        }
    }
    pub fn hrp(&self) -> &str {
        match self {
            Bech32::Address(address) => &address.0.hrp,
            Bech32::ContractId(contract_id) => &contract_id.0.hrp,
        }
    }
}
```

and finally, make it so that an `Address` and a `ContractId` can be constructed from it:
```rust

impl From<Bech32> for Address {
    fn from(bench32: Bech32) -> Self {
        Self::new(*bench32.hash())
    }
}

impl From<Bech32> for ContractId {
    fn from(bench32: Bech32) -> Self {
        Self::new(*bench32.hash())
    }
}
```

tldr the differences:
1. the bech32 is a simple data struct that resembles the real-world bech encoding (hrp + hash)
2. content type is embedded in the bech type itself
3. you use rust-idiomatic conversions instead of creating a proliferation of helper methods (construct_x_from_str, construct_x)

don't have particularly strong feelings towards this solution, just wanted to give it a try since it rubbed me the wrong way.

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,154 @@
...
+        Ok(Self {
+            hash: Bytes32::from(pubkey_hash),
+            encoding: s.to_string(),
+            content_type,
+        })
+    }
+}
+
+#[allow(clippy::from_over_into)]
+impl Into<Address> for &Bech32 {
```
segfault-magnet, line 82: Why not go for:
```rust
impl From<&Bech32> for Address { ...
```

and make clippy happy?

examples/contracts/src/lib.rs
```diff
@@ -279,7 +279,7 @@ mod tests {
             StorageConfiguration::default(),
         )
         .await?;
-        println!("Contract deployed @ {:x}", contract_id);
+        println!("Contract deployed @ {contract_id}");
```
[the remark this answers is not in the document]
MujkicA: I removed the ones from `harness.rs`, I figured these ones might have some use since the examples are in the book. Could be wrong.

packages/fuels-types/src/bech32.rs
```diff
@@ -0,0 +1,154 @@
...
+///  public key hash encoded with a u5 charset, and a checksum.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct Bech32 {
+    /// Public key hash or contract hash
+    hash: Bytes32,
+
+    /// String representation of the bech32 format
+    encoding: String,
+
+    pub content_type: ContentType,
```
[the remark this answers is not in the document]
MujkicA: I actually wanted to leave the possibility to change what the bech32 represents, but on second thought, that use-case seems rather unrealistic.

packages/fuels-types/src/bech32.rs
MujkicA: Looks better, let me give it a try

packages/fuels-types/src/bech32.rs
MujkicA: That impl would need to go upstream to `fuel-types` and it seemed odd to have `Bech32` in there since it is only used on the SDK side.

segfault-magnet: rust doesn't seem to mind having it here? just back from compiling it. can you re-check, maybe I'm doing something wrong?

examples/contracts/src/lib.rs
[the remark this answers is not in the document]
segfault-magnet: Oh, didn't see that we were in `examples`. Yeah, that's fine by me.

packages/fuels-types/src/lib.rs
```diff
@@ -3,12 +3,17 @@
 //! We declare these in a dedicated, minimal crate in order to allow for downstream projects to
 //! consume or generate these ABI-compatible types without needing to pull in the rest of the SDK.
 
+extern crate core;
```
vnepveu, line 6: are you sure you need to take the core crate (iiuc it's the `fuels-core` crate right?). 
Isn't that a circular deps? We're importing `fuels-types` everywhere including `fuels-core`.

MujkicA: that somehow landed there while merging, good catch

packages/fuels-signers/src/provider.rs
```diff
@@ -36,6 +36,12 @@ impl From<WalletError> for ProviderError {
         ProviderError::WalletError(e.to_string())
     }
 }
+
+impl From<ProviderError> for Error {
+    fn from(e: ProviderError) -> Self {
+        Error::ProviderError(e.to_string())
+    }
+}
```
vnepveu, line 44: Shouldn't this be in `error.rs` with all the other error conversion functions?

packages/fuels-types/src/bech32.rs
```diff
@@ -2,9 +2,9 @@ use crate::errors::Error;
 use bech32::Variant::Bech32m;
 use bech32::{FromBase32, ToBase32};
 use core::str;
-use std::str::FromStr;
 use fuel_tx::{Address, Bytes32, ContractId};
 use std::fmt::{Display, Formatter};
+use std::str::FromStr;
 
 pub const FUEL_BECH32_HRP: &str = "fuel";
```
vnepveu, line 9: Explain what this constant is (why is it `fuel` and what it's used for)

vnepveu (dismissed):
One question about error and another comment, otherwise lgtm!

packages/fuels-signers/src/provider.rs
MujkicA: Since `From<WalletError> for Error` is implemented in `wallet.rs` i thought this would be appropriate. 
It kind of makes sense for me this way, since `Error` is a global error type so it shouldn't be aware of local error types, rather the local error types should know how to convert to it.

vnepveu: You're right! Makes me think we should implement the other `From<LocalErrorTypes>` inside their respective files.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -124,21 +124,22 @@ impl Abigen {
...
                     use fuels::types::errors::Error as SDKError;
                     use fuels::types::param_types::{EnumVariants, ParamType};
                     use std::str::FromStr;
                 },
                 quote! {
                     pub struct #name {
-                        contract_id: ContractId,
-                        wallet: LocalWallet
+                        pub contract_id: Bech32ContractId,
+                        pub wallet: LocalWallet
```
hal3e, line 135: Did we agree to make the `contract_id` and `wallet` public ? If so we do not need the getters anymore.

hal3e: Otherwise LGTM! I like how it was implemented!

docs/src/types/bech32.md
```diff
@@ -1,9 +1,11 @@
...
+`Bech32Address` and `Bech32ContractId` enable the use of addresses and contract ids in the bech32 format. They can easily be converted to their counterparts `Address` and `ContractId`.
 
-These are the main ways of creating a `Bech32`: 
+Here are the main ways of creating a `Bech32Address`, but note that the same applies to `Bech32ContractId`: 
 
 ```rust,ignore
 {{#include ../../../examples/types/src/lib.rs:bech32}}
 ```
+
+> **Note:** when creating a `Bech32Address` from `Address` or `Bech32ContractId` from `ContractId` the HRP (Human-Readable Part) is set to **"fuel"** per default.
```
iqdecay, line 11: nice

packages/fuels-types/src/bech32.rs
```diff
@@ -6,193 +6,103 @@ use fuel_tx::{Address, Bytes32, ContractId};
...
-}
-
-impl Bech32Data {
-    pub fn new(hrp: &str, hash: Bytes32) -> Bech32Data {
-        Self {
-            hrp: hrp.to_string(),
-            hash,
+/// Generate type represented in the Bech32 format,
+/// consisting of a human-readable part (hrp) and a hash (e.g. pubkey-, contract hash)
+macro_rules! bech32type {
```
iqdecay, line 14: oooh nice macro work!

iqdecay (approved):
Nice! LGTM after conflict solving :tada:

packages/fuels-core/src/code_gen/abigen.rs
MujkicA: This was done before your recent PR. I will update it to use the getters. However, I think we need to open up this discussion at some point again. We added an instance builder to abigen to avoid collision with `new` but now we have the same problem with the getters/setter and soon with `instance::connect` too and we still kept the builder.

iqdecay (dismissed):
LGTM

packages/fuels-core/src/code_gen/abigen.rs
hal3e: I agree.

hal3e (approved):

PR #468, opened 2022-07-13
docs: follow-ups from the book re-org

digorithm:
A couple more updates, nits, etc., related to https://github.com/FuelLabs/fuels-rs/issues/365.

### DIFF
--- a/docs/src/wallets/test-wallets.md
+++ b/docs/src/wallets/test-wallets.md
@@ -26,3 +26,23 @@ You can create a test wallet containing multiple assets (including the base asse
 
 - coins: `Vec<(UtxoId, Coin)>` has num_assets * coins_per_assets coins (UTXOs)
 - asset_ids: `Vec<AssetId>` contains the num_assets randomly generated `AssetId`s (always includes the base asset)
+
+## Setting up assets
+
+The Fuel blockchain holds many different assets; you can create your asset with its unique `AssetId` or create random assets for testing purposes.
+
+You can use only one asset to pay for transaction fees and gas: the base asset, whose AssetId is `0x000...`, a 32-byte zeroed value.
+
+For testing purposes, you can configure coins and amounts for assets. You can use `setup_multiple_assets_coins`:
+
+```rust,ignore
+{{#include ../../../examples/wallets/src/lib.rs:multiple_assets_coins}}
+```
+
+>**Note** If setting up multiple assets, one of these assets will always be the base asset.
+
+If you want to create coins only with the base asset, then you can use:
+
+```rust,ignore
+{{#include ../../../examples/providers/src/lib.rs:setup_single_asset}}
+```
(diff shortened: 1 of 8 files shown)


docs/src/wallets/test-wallets.md
```diff
@@ -26,3 +26,23 @@ You can create a test wallet containing multiple assets (including the base asse
 
 - coins: `Vec<(UtxoId, Coin)>` has num_assets * coins_per_assets coins (UTXOs)
 - asset_ids: `Vec<AssetId>` contains the num_assets randomly generated `AssetId`s (always includes the base asset)
+
+## Setting up assets
+
+The Fuel blockchain holds many different assets; you can create your asset with its unique `AssetId` or create random assets for testing purposes.
+
+You can use only one asset to pay for transaction fees and gas: the base asset, whose AssetId is `0x000...`, a 32-byte zeroed value.
```
vnepveu, line 34:
```suggestion
You can use only one asset to pay for transaction fees and gas: the base asset, whose AssetId is `0x000...0`, a 32-byte zeroed value.
```

docs/src/wallets/test-wallets.md
```diff
@@ -26,3 +26,23 @@ You can create a test wallet containing multiple assets (including the base asse
 
 - coins: `Vec<(UtxoId, Coin)>` has num_assets * coins_per_assets coins (UTXOs)
 - asset_ids: `Vec<AssetId>` contains the num_assets randomly generated `AssetId`s (always includes the base asset)
+
+## Setting up assets
```
vnepveu, line 30: There is a `Setting up a test wallet with multiple assets` section right above, it feels like a duplication. What's more, the ANCHORs overlap. Maybe the above section could be re-worded instead?

vnepveu (changes requested):
Good! Just one question about a duplicate paragraph.

docs/src/wallets/test-wallets.md
digorithm: Yeah, I had this concern too. After reading `Setting up a test wallet with multiple assets` with fresh eyes, I noticed it didn't say much about setting up assets and coins, it just made _use_ of it, which is essentially a reference to something that hasn't been "defined" yet (or "talked about") and could be a source of confusion.

I could talk about setting about assets and then say

> For testing purposes, you can configure coins and amounts for assets. You can use setup_multiple_assets_coins as seen above.

And skip the code block, but that felt empty, somehow. So I decided to experiment with _some_ duplication here (I'm not really duplicating it all, just an inner section of the outer anchor), just focusing on the code for setting up multiple assets. 

I think re-wording the above section to include asset stuff would be kinda like giving more responsibilities to a section that's supposed to be only about setting up a test wallet 🤔

vnepveu: Ok with me then!

vnepveu (approved):
:ship:

Issue #470, opened 2022-07-14
abigen! fails to get b256 type when it is with another reference type as a parameter


matt-user:
If b256 is with another reference type as a parameter abigen cannot resolve the b256 type. [repro](https://github.com/matt-user/repro)

digorithm:
More context: Just checked the expanded types from the abigen, I think it’s related to tuples:

```Rust
 pub fn dummy(
            &self,
            arg_0: TestStruct,
            arg_1: [u8; 32],
        ) -> ContractCallHandler<(TestStruct, b256)> {}
```
`b256` should’ve been expanded to an array, like it does when not in a tuple.

PR #471, opened 2022-07-14
b256 in a tuple with other custom types breaks `abigen!`

segfault-magnet:
closes: #470 

To be deemed as a hotfix as we should give this more thought to whether it might crop up in other weird places.


digorithm (approved):
Thanks for the speedy hotfix! Wondering if there will be other edge cases like this one; Right now I can't think of any other non-Rust type that we have to expand here 🤔 But we'll keep an eye on it.

PR #478, opened 2022-07-17
feat: refactor `produce_blocks` API

### LINKED ISSUE #449: use API from fuel-core to mine empty blocks
In [410](https://github.com/FuelLabs/fuels-rs/issues/410) we added a helper function to increase the block height by submitting empty scripts. 

This can be done more cleanly by incorporating the API that, ATTOW, is being developed [here](https://github.com/FuelLabs/fuel-core/issues/455).

Once it is done we can switch over to using it.

iqdecay:
Closes #449.
The endpoint from `fuel-core` will probably change to `produce_blocks` in https://github.com/FuelLabs/fuel-core/pull/493 .


packages/fuels/tests/harness.rs
```diff
@@ -2160,13 +2161,21 @@ async fn contract_method_call_respects_maturity() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
-async fn contract_deployment_respects_maturity() -> anyhow::Result<()> {
+#[cfg(feature = "fuel-core-lib")]
```
segfault-magnet, line 2164: Why is this test limited to only run if we're using fuel core as a lib?

AFAIR `fuel-core` has had a patch release that includes the `fuel-vm` fixes that are needed for these tests to pass. So, as long as we're using the latest `fuel-core` binary these tests should pass. 

Or am I missing something ?

vnepveu: This limitation was introduced by the developer because he couldn't make it work lol because of some weird prelude imports 😁 You're right I should make it work regardless.

vnepveu:
Ok so it turns out that by "definition" you cannot make those tests run without the `fuel-core` lib, because you need the chain to have the `manual blocks` feature enabled, which afaik cannot be done without the lib or would require more refactoring. I'm op

hal3e:
Nice, LGTM!

hal3e (approved):

packages/fuels-test-helpers/src/chains.rs
```diff
@@ -1,46 +1,21 @@
...
-
-        provider.send_transaction(&transaction).await?;
-
-        confirm_block_created(provider, height_before_transaction).await?;
-    }
-    Ok(())
-}
-
-async fn confirm_block_created(provider: &Provider, previous_height: u64) -> Result<(), Error> {
+pub async fn confirm_blocks_created(
```
MujkicA, line 7: Since we don't use a no op script anymore, this error message is inaccurate:
**"Couldn't confirm a block generation via no-op script -- {}"**

packages/fuels/tests/harness.rs
```diff
@@ -2159,13 +2160,21 @@ async fn contract_method_call_respects_maturity() -> anyhow::Result<()> {
...
+async fn contract_deployment_respects_maturity() -> Result<(), Error> {
     abigen!(
         MyContract,
         "packages/fuels/tests/test_projects/transaction_block_height/out/debug/transaction_block_height-abi.json"
     );
 
-    let wallet = launch_provider_and_get_wallet().await;
+    let config = Config {
+        manual_blocks_enabled: true,
```
MujkicA, line 2171: Should this be documented?

MujkicA (commented):
We should consider extending the docs, other than that and the error message LGTM!

segfault-magnet:
> Ok so it turns out that by "definition" you cannot make those tests run without the `fuel-core` lib, because you need the chain to have the `manual blocks` feature enabled, which afaik cannot be done without the lib or would require more refactoring. I'm opening an issue so that nodes can be started with the `manual blocks` flag.

I'd do one of two things:
1. Wait with merging until fuel-core binaries can be started in a 'test' mode which accepts this way of producing blocks, or,
2. Create a task in fuels-rs that will track the one over on fuel-core, so that we may revisit this once they've implemented the flag.

Otherwise, this limitation will stay here until somebody remembers.

packages/fuels/tests/harness.rs
iqdecay: Yes absolutely!

iqdecay:
>     1. Wait with merging until fuel-core binaries can be started in a 'test' mode which accepts this way of producing blocks, or,
> 
>     2. Create a task in fuels-rs that will track the one over on fuel-core, so that we may revisit this once they've implemented the flag.
So the limitation here is from the SDK side, the `fuel-core` bin can already run with the flags (which is why it works without the `fuel-core-lib` feature). I have created issue #489 to track this. Otherwise, the current limitation is just that it doesn't run with the `fuel-core-lib` feature enabled. I think this is enough, or do you have something else in mind?

segfault-magnet:
> So the limitation here is from the SDK side, the `fuel-core` bin can already run with the flags (which is why it works without the `fuel-core-lib` feature). I have created issue #489 to track this. Otherwise, the current limitation is just that it doesn't run with the `fuel-core-lib` feature enabled. I think this is enough, or do you have something else in mind?

Now I'm confused.

I can see the test `contract_deployment_respects_maturity` is enabled only if we're using `fuel-core-lib`. 

Which made me think that using the binary is problematic. But then you say: 

> fuel-core bin can already run with the flags

This makes me think that it is not the binary's fault per-se but rather our wrapper around it in the SDK. But then you say

> I have created issue https://github.com/FuelLabs/fuels-rs/issues/489 to track this.

But the issue title is `allow using fuel-core-lib with custom parameters`

But fuel-core can already be used as a lib and be passed the custom flag, which is done in the test `contract_deployment_respects_maturity`... 

>  the current limitation is just that it doesn't run with the fuel-core-lib feature enabled

who doesn't run with that feature enabled? This is the feature of the SDK. Do you mean the CI tests of the SDK's `fuels` package?This was fixed as part of [this issue](https://github.com/FuelLabs/fuels-rs/issues/435).

I guess the TLDR question is:

If we can run the fuel-core binary in a mode that accepts `produce_blocks` requests *and* if we can run fuel-core as a library in a mode that accepts `produce_blocks` requests -- why is the test only being compiled if we have the library?

iqdecay:
Definitely got confused here. Fixed #489 to reflect that it's actually related to our wrapper around the binary and not the library.

iqdecay:
@segfault-magnet PR #501 fixes the issue I mentioned. It is dependent on this PR though.

segfault-magnet (approved):

PR #479, opened 2022-07-17
Set fuel-core stdin/stdout stream to null

rakita:
When running `forc test` there are some spammy console logs originating from fuel_core.

This will void `stdin` and `stdout` stream and should remove those logs (`stderr` is left for better diagnostic).


digorithm (approved):
Thanks!

vnepveu (approved):
Nice ty!

PR #480, opened 2022-07-18
add features.md and fuel-core lib doc.

Salka1988:


docs/src/providers/features.md
```diff
@@ -0,0 +1,5 @@
+# Features
+
+### Fuel-core lib
+
+The `fuel-core-lib` feature allows us to run a fuel-core node without installing the fuel-core binary on the local machine. This entails downloading all the dependencies needed to run the fuel-core node.
```
vnepveu, line 5: You should explain what dependencies it requires, maybe even put an example `Cargo.toml` file like in the `Setup` part.

vnepveu (changes requested):
This is going in the right direction, but IMO you should be documenting how to use the feature (people unfamiliar would not necessarily know how to). Also I think you could put it inside the `Running a short-lived fuel-node` as a paragraph, I think it would make more sense, what do you think? Since it's used for connecting to a node in that case.

docs/src/providers/short-lived.md
```diff
@@ -19,3 +19,13 @@ You can also use `launch_provider_and_get_wallet()`, which abstracts away the `s
 ```rust,ignore
 ```
+
+# Features
+
+### Fuel-core lib
```
digorithm, line 25:
```suggestion
## Fuel-core lib
```

docs/src/providers/short-lived.md
```diff
@@ -19,3 +19,13 @@ You can also use `launch_provider_and_get_wallet()`, which abstracts away the `s
...
+
+The `fuel-core-lib` is a feature defined in `fuels` library, and it allows us to run a fuel-core node without installing the fuel-core binary on the local machine. This entails downloading all the dependencies needed to run the fuel-core node.
```
digorithm, line 27: Clarity improvements:

```suggestion
The `fuel-core-lib` is a feature defined in the `fuels` library, allowing us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine. Using the `fuel-core-lib` feature flag entails downloading all the dependencies needed to run the fuel-core node.
```

digorithm (changes requested):

vnepveu (approved):
perfect! will be clearer for future users :1st_place_medal:

digorithm (approved):
:shipit:

PR #483, opened 2022-07-20
ci: remove the `--no-default-features` flag for running tests

vnepveu:
This is actually not useful and prevents some tests from running as they should.


digorithm (approved):
👍

Issue #475, opened 2022-07-15
Expand `WalletsConfig` to accept user-configured `AssetId`s


digorithm:
Users would like to bootstrap their wallets with certain user-configured assets/coins. Currently, it only bootstrap test wallets with the base asset and randomly-generated ones. 

Ideally, it would look like this:

```rust
WalletsConfig::new_multiple_assets([(asset_id_1, 23), (asset_id_2, 42), ...])
```

PR #484, opened 2022-07-21
feat!: add multiple custom assets

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/475

Initially, the idea was to have something like this:
`WalletsConfig::new_multiple_assets([(asset_id_1, 23), (asset_id_2, 42), ...])`
However, we have decided that the user should also be able to set the number of coins and the amount per coin. So we have the following new method:
`WalletsConfig::new_multiple_assets([(asset_id_1, 100, 23), (asset_id_2, 100, 42), ...])` where we specify both the number of coins and the amount per coin in the tuple.

I have:
- refactored the `WalletConfig` and implemented the new method
- added the underlying function to create the coins and added a unit test
- added a test with the provider to see if the assets are generated and if we can retrieve them
- added an example and updated the docs

Regarding the approach. There is an open question whether we should create the base asset automatically or leave it to the user. I went with the second option. If the user is defining custom assets then maybe he also wants to be able to specify the number of coins and amount of the base asset. I have added this to the docs. I would like to hear you opinion on this.


hal3e:
In the last commit I did some refactoring of the test and removed unnecessary lines and increased the number of wallets. I had it set to 1 for debugging.

examples/wallets/src/lib.rs
```diff
@@ -189,6 +189,43 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    #[allow(unused_variables)]
+    async fn setup_wallet_custom_assets() -> Result<(), Box<dyn std::error::Error>> {
```
vnepveu, line 194: I think you can use `Result<(), Error>` here

docs/src/wallets/test-wallets.md
```diff
@@ -27,6 +27,17 @@ You can create a test wallet containing multiple assets (including the base asse
 - coins: `Vec<(UtxoId, Coin)>` has num_assets * coins_per_assets coins (UTXOs)
 - asset_ids: `Vec<AssetId>` contains the num_assets randomly generated `AssetId`s (always includes the base asset)
 
+## Setting up a test wallet with multiple custom assets
```
vnepveu, line 30: I think you should also rename the previous section with `with multiple random assets` or something.

examples/wallets/src/lib.rs
```diff
@@ -189,6 +189,43 @@ mod tests {
...
+        let mut asset_id_2 = AssetId::zeroed();
+        asset_id_2.try_fill(&mut rng)?;
+        let coins_per_asset_2 = 10;
+        let amount_per_coin_2 = 12;
+
+        let assets = &[
+            (asset_id_base, coins_per_asset_base, amount_per_coin_base),
+            (asset_id_1, coins_per_asset_1, amount_per_coin_1),
+            (asset_id_2, coins_per_asset_2, amount_per_coin_2),
+        ];
```
vnepveu, line 220: I feel that using `Vec` here would be cleaner. Also, maybe a struct would be in order here? Something like `AssetSetup` with the 3 fields?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -91,6 +91,18 @@ pub fn setup_multiple_assets_coins(
     (coins, asset_ids)
 }
 
+/// Create a vector of of UTXOs with the provided AssetIds, num_coins and amount_per_coin
+pub fn setup_custom_assets_coins(
+    owner: Address,
+    assets: &[(AssetId, u64, u64)],
```
vnepveu, line 97: Using `Vec` is preferred afaikk

packages/fuels-test-helpers/src/wallets_config.rs
```diff
@@ -19,16 +20,18 @@ impl WalletsConfig {
...
+            )],
         }
     }
 
-    pub fn new_single(coins: Option<u64>, amount: Option<u64>) -> Self {
+    pub fn new_multiple_assets(num_wallets: u64, assets: Vec<(AssetId, u64, u64)>) -> Self {
         Self {
-            num_wallets: 1,
-            coins_per_wallet: coins.unwrap_or(DEFAULT_NUM_COINS),
-            coin_amount: amount.unwrap_or(DEFAULT_COIN_AMOUNT),
```
vnepveu, line 31: I don't think this setup function should disappear, it's quite practical to be able to just generate a new single wallet without having to specify assets.

vnepveu (changes requested):
It's looking good, I think you can use a struct and `vec!` more, also I'm not sure why you removed the `setup_single` function.

examples/wallets/src/lib.rs
hal3e: This was definitely on my mind as you need to know the order. However the requested function was a vector with tuples. If we agree on the change I will gladly implement it.

packages/fuels-test-helpers/src/wallets_config.rs
hal3e: It is basically the same as `new` but the first parameter is set to 1 and it was used once...

examples/wallets/src/lib.rs
vnepveu: Ah you're right, let's ping @digorithm to see what he thinks of a struct. the struct could be built using helper functions taki

packages/fuels-test-helpers/src/wallets_config.rs
vnepveu: Ok I thought it was more used than that. I just always want make sure that we keep very simple and easy defaults available.

hal3e:
@vnepveu thanks for the review! In the end, I have added the `AssetConfig` struct and refactored the `new_multiple_assets` function. I like how it looks now. Let's see what others say :)

docs/src/wallets/test-wallets.md
```diff
@@ -27,6 +27,17 @@ You can create a test wallet containing multiple assets (including the base asse
...
+
+You can also create assets with specific `AssetId`s, coin amounts and number of coins.
+
+```rust,ignore
+{{#include ../../../examples/wallets/src/lib.rs:custom_assets_wallet}}
+```
+
+>**Note** In this case you need to specify also the base asset and the corresponding number of
```
MujkicA, line 38:
```suggestion
>**Note** In this case, you need to specify also the base asset and the corresponding number of
```

MujkicA (dismissed):
LGTM :ok_hand:

hal3e:
Thanks @MujkicA for the review!

digorithm:
Branch outdated -- could you sync with base? 😁

docs/src/wallets/test-wallets.md
```diff
@@ -27,6 +27,17 @@ You can create a test wallet containing multiple assets (including the base asse
 - coins: `Vec<(UtxoId, Coin)>` has num_assets * coins_per_assets coins (UTXOs)
 - asset_ids: `Vec<AssetId>` contains the num_assets randomly generated `AssetId`s (always includes the base asset)
 
+## Setting up a test wallet with multiple custom assets
+
+You can also create assets with specific `AssetId`s, coin amounts and number of coins.
```
digorithm, line 32:
```suggestion
You can also create assets with specific `AssetId`s, coin amounts, and number of coins.
```
I'm deeply sorry, it had to be done 😂 

the-oxford-comma_52c855ed979ed_w1500

digorithm (changes requested):

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -91,6 +91,17 @@ pub fn setup_multiple_assets_coins(
     (coins, asset_ids)
 }
 
+/// Create a vector of of UTXOs with the provided AssetIds, num_coins and amount_per_coin
```
digorithm, line 94:
```suggestion
/// Create a vector of UTXOs with the provided AssetIds, num_coins, and amount_per_coin
```

docs/src/wallets/test-wallets.md
hal3e: Will never forget it again :D

packages/fuels-test-helpers/src/wallets_config.rs
```diff
@@ -1,34 +1,40 @@
 // These constants define the default number of wallets to be setup,
 // the number of coins per wallet and the amount per coin
+use fuels_signers::fuel_crypto::fuel_types::AssetId;
+
 pub const DEFAULT_NUM_WALLETS: u64 = 10;
 pub const DEFAULT_NUM_COINS: u64 = 1;
 pub const DEFAULT_COIN_AMOUNT: u64 = 1_000_000_000;
 
+#[derive(Debug, Clone)]
+pub struct AssetsConfig {
```
digorithm, line 10: How about calling it `AssetConfig` instead of `AssetsConfig` (notice the `s` gone)? This is the configuration for a single asset.

digorithm:
Adding the breaking label and editing the title to `feat!` so we can remember this is a breaking change (we're changing `WalletsConfig`'s pub methods) once we release it.

hal3e:
@digorithm thanks for the review!

digorithm (approved):
LGTM -- :shipit:

examples/wallets/src/lib.rs
```diff
@@ -191,33 +191,37 @@ mod tests {
...
         use fuels::prelude::*;
         use rand::Fill;
 
         let mut wallet = LocalWallet::new_random(None);
         let mut rng = rand::thread_rng();
 
-        let asset_id_base = BASE_ASSET_ID;
-        let coins_per_asset_base = 2;
-        let amount_per_coin_base = 4;
+        let asset_base = AssetConfig {
```
iqdecay, line 202: `AssetConfig` is indeed much cleaner, like it!

iqdecay (approved):
Well done! :ship:

examples/wallets/src/lib.rs
hal3e: Thanks!

Issue #485, opened 2022-07-21
Create `ContractId` and `LocalWallet` getters for the generated contract struct


digorithm:
Currently, the generated contract struct has the `ContractId` and `LocalWallet` as private fields:

```Rust
// ...
quote! {
  pub struct #name {
      contract_id: ContractId,
      wallet: LocalWallet
  }
// ...
```

It would be useful to have an `abigen`-generated getter for those fields so users could easily access them.

simonr0204:
Could we also get one for the wallet? :D

digorithm:
> Could we also get one for the wallet? 😁

Don't we already have a getter for a wallet's address? 

https://github.com/FuelLabs/fuels-rs/blob/b00a3a707c2013bf2a5dc331dfb1b4052a00b522/packages/fuels-signers/src/lib.rs#L32

Unless you mean something else 😅

Issue #486, opened 2022-07-21
Function to create new instances of a contract with different wallets connected


simonr0204:
It's common to want to deploy a contract with one wallet, and then interact with it with a different one.

Would it be possible to have something that allows you to take an existing instance, and build additional ones (connected to other wallets) off it on-the-fly? Something like:

```
// Initial instance
MyContractInstance = MyContract::new(asset_id.to_string(), wallet1.clone());

...
MyContractInstance.my_function().call().await;                      // Calling wallet is wallet1
MyContractInstance::connect(wallet2).my_function().call().await;    // Calling wallet is wallet2
```

So `connect` returns a new instance connected to `wallet2`. It's a convenient short-hand compared to creating a 2nd instance from scratch, which you currently have to do:

```
MyContractInstance2 = MyContract::new(asset_id.to_string(), wallet2.clone());
MyContractInstance2.my_function().call().await;    // Calling wallet is wallet2
```

For reference, ethers.js has this which gets used all the time: https://docs.ethers.io/v5/api/contract/contract/#Contract-connect

PR #487, opened 2022-07-21
feat: add contract_id and wallet getters

hal3e:
should close: https://github.com/FuelLabs/fuels-rs/issues/485

I think we will again hit the naming problem with this approach.
Currently, I put leading underscores for the getters.


vnepveu (approved):
lgtm!

MujkicA:
In this case, would it have been better to just expose the fields? That would also solve #486, otherwise we run into collision problem there as well.

Issue #473, opened 2022-07-15
'Instantiating a Fuel client' example in book not working


seem-less:
It seems that the example '[Instantiating a Fuel client](https://fuellabs.github.io/fuels-rs/v0.17.0/getting-started/basics.html#instantiating-a-fuel-client)' does not compile because the `node` module is not part of the `fuels` crate.

`use fuels::node::service::{Config, FuelService};`

It seems that the required types are actually in the `fuels_test_helpers` crate instead

vnepveu:
Do you mind providing a repro please? I can't seem to make it fail locally, and *normally* (though CI works in mysterious ways), the docs should be tested by CI.

seem-less:
I created a new library with `cargo` and then following the [Setup Instructions](https://fuellabs.github.io/fuels-rs/v0.17.0/getting-started/setup.html) my files look like this:

`Cargo.toml`
```
[package]
name = "rust-sdk-example"
version = "0.1.0"
edition = "2021"

[dependencies]
fuels = "0.17"
```

`src/lib.rs`
```
use fuels::prelude::*;

async fn instantiate_client() -> Result<(), Error> {
    use fuels::client::FuelClient;
    use fuels::node::service::{Config, FuelService};

    let server = FuelService::new_node(Config::local_node()).await?;
    let client = FuelClient::from(server.bound_address);
    assert!(client.health().await?);
    Ok(())
}
```

It seems that if the `fuel-core-lib` feature is enabled on the `fuels` crate, the required imports are provided by the `fuels::prelude::*`, which means that the separate import of `Config` and `FuelService` would not be required.

vnepveu:
Can you try runnig with `--features fuel-core-lib` flag please?

Salka1988:
`use fuels::node::service::{Config, FuelService};` is not exposed anymore and     
`let server = FuelService::new_node(Config::local_node()).await?;` needs fuel-core-lib flag.

I will take this issue and we can discuss it today

PR #488, opened 2022-07-21
instantiating a fuel client bug

Salka1988:
https://github.com/FuelLabs/fuels-rs/issues/473 - This was not a bug, but the implementation for the `FuelClient` for the default (fuel-cor bin) option was missing.

- I added the implementation for FuelClient


hal3e:
Could you please provide more information about this PR?

packages/fuels-test-helpers/src/node.rs
```diff
@@ -271,4 +298,32 @@ pub async fn server_health_check(client: &FuelClient) {
...
+}
+
+pub struct FuelService {
+    pub bound_address: SocketAddr,
+}
+
+impl FuelService {
+    pub async fn new_node(config: Config) -> Result<Self, AnyError> {
+        let bound_address = match config {
+            node_config if node_config.addr.port() != 0 && is_free(node_config.addr.port()) => {
```
segfault-magnet, line 316: I found this difficult to read, requiring paper to understand.

It reads like this: 
if the port of the address is not zero and is free use it
if it is zero or if it is something else but is not free, check if it is not free, if so panic
otherwise get a random address.

Perhaps a simpler way? 
```rust
        let requested_port = config.addr.port();

        let chosen_addr = if requested_port == 0 {
            get_socket_address()
        } else if is_free(requested_port) {
            config.addr
        } else {
            panic!("Error: Address already in use");
        };
```

packages/fuels-test-helpers/src/node.rs
```diff
@@ -271,4 +298,32 @@ pub async fn server_health_check(client: &FuelClient) {
...
+                node_config.addr
+            }
+            node_config if !is_free(node_config.addr.port()) => {
+                panic!("Error: Address already in use")
```
segfault-magnet, line 320: Why not just `bail!` since you're already returning a Result? Let the decision to panic, if it is made, be higher up.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -230,32 +233,45 @@ fn write_temp_config_file(config: Value) -> NamedTempFile {
...
 }
 
-pub fn spawn_fuel_service(
+pub async fn new_fuel_node(
     coins: Vec<(UtxoId, Coin)>,
     consensus_parameters_config: Option<ConsensusParameters>,
-    free_port: Port,
+    socket_addr: SocketAddr,
 ) {
+    let (tx, rx) = oneshot::channel();
```
digorithm, line 241: I think I understand why we're using tokio's `onyshot` channels here; we want to wait for a signal from the `tokio::spawn` below that it has finished setting up the node. Is that right?

If so, could you write a tiny comment line explaining this rationale? It might confuse people who are new to Rust and channel-based async stuff.

digorithm (changes requested):
Looking good! Just left a small request for comment.

packages/fuels-test-helpers/src/node.rs
Salka1988: Yes, you're right. I'll do it.

digorithm (approved):
LGTM!

iqdecay (approved):
LGTM, might be a fix to something else I had in mind, nice!

PR #491, opened 2022-07-22
Add transfer to contract

### LINKED ISSUE #477: Implement transfer to contract
The SDK needs a method for transferring assets to a contract similar to `wallet.transfer()`.

MujkicA:
Closes #477

I've added `force_transfer_to_contract()` to the existing `wallet` methods. The method call creates a `Transaction::Script` that loads relevant parameters into registers and calls the `TR` instruction. 
To be able to check contract balances, I've added wrapper methods in `provider` for the matching `FuelClient` methods.

Docs now include another subsection in the wallet section that covers contract transfers and address to address transfers (for some reason we had an example for this one but it wasn't included in the docs).

I chose `force_transfer_to_contract` for the name to match the sway counterpart.


packages/fuels-signers/src/wallet.rs
```diff
@@ -288,13 +290,58 @@ impl Wallet {
...
+    }
+
+    /// UNCONDITIONAL transfer of `amount` coins of type `asset_id` to
+    /// the contract at `to`.
+    /// Fails if amount for asset ID is larger than this wallets spendable coins.
+    /// Returns the transaction ID that was sent and the list of receipts.
+    ///
+    /// CAUTION !!!
+    ///
+    /// This will transfer coins to a contract even with no way to retrieve them
```
hal3e, line 310: I would remove the word even.

packages/fuels-signers/src/wallet.rs
```diff
@@ -288,13 +290,58 @@ impl Wallet {
...
+    /// (i.e. no withdrawal functionality on receiving contract), possibly leading
+    /// to the PERMANENT LOSS OF COINS if not used with care.
+    pub async fn force_transfer_to_contract(
+        &self,
+        to: &ContractId,
```
hal3e, line 315: I would suggest we change the variable `to` to something like: `recipient`, `destination_contract` etc. It will make also the comments above clearer.

hal3e (dismissed):
I have added some nits. Other than that LGTM! Well done!

packages/fuels-signers/src/provider.rs
```diff
@@ -183,7 +193,67 @@ impl Provider {
...
-    // TODO: add unit tests for the balance API. This is tracked in #321.
+
+    /// Craft a transaction used to transfer funds to a contract.
+    pub fn build_contract_transfer_tx(
+        &self,
+        to: ContractId,
+        amount: u64,
+        asset_id: AssetId,
+        inputs: &[Input],
+        outputs: &[Output],
```
iqdecay, line 204: use `Vec`

packages/fuels-signers/src/wallet.rs
```diff
@@ -288,13 +290,58 @@ impl Wallet {
...
+    /// UNCONDITIONAL transfer of `amount` coins of type `asset_id` to
+    /// the contract at `to`.
+    /// Fails if amount for asset ID is larger than this wallets spendable coins.
+    /// Returns the transaction ID that was sent and the list of receipts.
+    ///
+    /// CAUTION !!!
+    ///
+    /// This will transfer coins to a contract even with no way to retrieve them
+    /// (i.e. no withdrawal functionality on receiving contract), possibly leading
+    /// to the PERMANENT LOSS OF COINS if not used with care.
```
iqdecay, line 312:
```suggestion
    /// Unconditionally transfers `amount` coins of type `asset_id` to
    /// the contract at `to`.
    /// Fails if amount for `asset_id` is larger than this wallet's spendable coins.
    /// Returns the corresponding transaction ID and the list of receipts.
    ///
    /// CAUTION !!!
    ///
    /// This will transfer coins to a contract, possibly leading
    /// to the PERMANENT LOSS OF COINS if not used with care.
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -288,13 +290,58 @@ impl Wallet {
...
+    pub async fn force_transfer_to_contract(
+        &self,
+        to: &ContractId,
+        amount: u64,
```
iqdecay, line 316: Is `amount` really the number of coins? It sounds like it's balance, and not the number of coins (ie UTXOs). The wording should be improved around this.

iqdecay (changes requested):
There is some confusion between amount and balance from what I understand. Otherwise looks great!

packages/fuels-signers/src/provider.rs
segfault-magnet: Slices are more flexible, why require a vec? We don't need ownership here.

iqdecay: I get what you mean, but since we use `Vec<something>` pretty much everywhere I figured it would make more sense. But you're right, iirc it would mean that we don't have to `clone` to use that method? 
In that case, maybe we should be using slices more.

packages/fuels-signers/src/wallet.rs
MujkicA: I would like to keep it consistent with `wallet.transfer`

packages/fuels-signers/src/provider.rs
segfault-magnet: Ideally, according to the [rust API guidelines](https://rust-lang.github.io/api-guidelines/flexibility.html?highlight=intoiter#functions-minimize-assumptions-about-parameters-by-using-generics-c-generic), we should strive for the least constricting interface -- i.e. the most flexible.

> The fewer assumptions a function makes about its inputs, the more widely usable it becomes.

> Prefer
> 
> ```rust
> fn foo<I: IntoIterator<Item = i64>>(iter: I) { /* ... */ }
> ```
> over any of
> 
> 
> ```rust
> fn foo(c: &[i64]) { /* ... */ }
> fn foo(c: &Vec<i64>) { /* ... */ }
> fn foo(c: &SomeOtherCollection<i64>) { /* ... */ }
> ```
> if the function only needs to iterate over the data.
> 
> More generally, consider using generics to pinpoint the assumptions a function needs to make about its arguments.

It further goes on to list the advantages and disadvantages. Using slices here is the pragmatic solution to using the `IntoIterator` trait bound due to the verbosity of the latter. 

But slices have no inherent increase in verbosity while offering a more flexible interface accepting anything that can be converted into a slice of the given type -- aka Vecs, arrays, etc.

packages/fuels-signers/src/wallet.rs
```diff
@@ -288,13 +290,58 @@ impl Wallet {
...
+        // Build transaction and sign it
+        let mut tx = self.get_provider()?.build_contract_transfer_tx(
+            *to,
+            amount,
+            asset_id,
+            &inputs,
+            &outputs,
+            tx_parameters,
+        );
+        let _sig = self.sign_transaction(&mut tx).await?;
```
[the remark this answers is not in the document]
MujkicA: I copied this over from `wallet.transfer()` so it slipped my mind. I've updated both

packages/fuels-signers/src/wallet.rs
MujkicA: Depends IIUC, from the perspective of the wallet, `amount` will be deducted, from the contracts perspective, `balance` will be added. Since this is a `wallet` method i think its fine to keep it as `amount`

packages/fuels-signers/src/provider.rs
```diff
@@ -47,11 +48,6 @@ impl From<WalletError> for ProviderError {
     }
 }
```
iqdecay, line 50: You cut the end of the sentence

packages/fuels-signers/src/wallet.rs
```diff
@@ -294,53 +292,58 @@ impl Wallet {
...
-    /// Returns the transaction ID that was sent and the list of receipts.
+    /// Fails if amount for `asset_id` is larger than this wallet's spendable coins.
+    /// Returns the corresponding transaction ID and the list of receipts.
     ///
     /// CAUTION !!!
     ///
-    /// This will transfer coins to a contract even with no way to retrieve them
-    /// (i.e. no withdrawal functionality on receiving contract), possibly leading
+    /// This will transfer coins to a contract, possibly leading
     /// to the PERMANENT LOSS OF COINS if not used with care.
```
iqdecay, line 310: Fix the docstring

packages/fuels-signers/src/wallet.rs
iqdecay: What I mean is `amount` coins means the total balance change is going to be: 
Sum over the coins (UTXOs) of the amount of each coin.
The wording is confusing: it's not the number of UTXOs that matters, but the resulting balance. Each coin has an amount, so sending `amount` of coins, you could send 10 coins of 100 amount each, or 10 coins of amount 1 each.
-> what matters is the balance, not the number (`amount`)of UTXOs.

iqdecay (changes requested):
Still think there is confusion around the wording.

packages/fuels-signers/src/provider.rs
MujkicA: I don't know what is meant by this

packages/fuels-signers/src/wallet.rs
```diff
@@ -299,9 +299,9 @@ impl Wallet {
         Ok((tx.id().to_string(), receipts))
     }
 
-    /// Unconditionally transfers `amount` coins of type `asset_id` to
+    /// Unconditionally transfers `balance` of type `asset_id` to
     /// the contract at `to`.
-    /// Fails if amount for `asset_id` is larger than this wallet's spendable coins.
+    /// Fails if balance for `asset_id` is larger than this wallet's spendable coins.
```
iqdecay, line 304:
```suggestion
    /// Fails if balance for `asset_id` is larger than this wallet's spendable balance.
```

iqdecay: Just for clarifying: my issue wasn't with the word "amount" per itself, but the fact that the docstring said `amount of _coins_` which is plain wrong. Feel free to use `amount` if you think it's better than balance, it just has to be super clear that what counts is the total balance (ie the total sum of the amounts of the various coins) and *not* the number of UTXOs (which is what was previously described).

iqdecay (approved):

packages/fuels-signers/src/provider.rs
iqdecay: Github playing tricks on me lol

PR #492, opened 2022-07-24
refactor: move function selector methods out of `ABIParser` struct

vnepveu:
* These methods will ultimately be used inside the Sway repository, and
  they didn't use the ABIParser fields.


hal3e:
LGTM!

segfault-magnet (approved):

PR #494, opened 2022-07-24
refactor: move tokenization methods inside a `Tokenizer` struct type

iqdecay:
The tokenization process is different from the parsing and so should be separated. It will allow cleaner refactoring down the line.


packages/fuels-core/src/tokenizer.rs
```diff
@@ -0,0 +1,866 @@
...
+
+        //To parse the value string we use a two pointer/index approach.
+        //The items are comma separated and if an item is tokenized the last_item
+        //index is moved to the current position.
+        //The variable nested is incremented and decremented if a bracket is encountered,
+        //and appropriate errors are returned if the nested count is not 0.
+        //If the struct has an array inside its values the current position will be incremented
+        //until both the opening and closing bracket are inside the new item.
+        //Characters inside quotes are ignored and they are tokenized as one item.
+        //An error is return if there is an odd number of quotes.
```
digorithm, line 88:
```suggestion
        // To parse the value string, we use a two-pointer/index approach.
        // The items are comma-separated, and if an item is tokenized, the last_item
        // index is moved to the current position.
        // The variable nested is incremented and decremented if a bracket is encountered,
        // and appropriate errors are returned if the nested count is not 0.
        // If the struct has an array inside its values, the current position will be incremented
        // until the opening and closing bracket are inside the new item.
        // Characters inside quotes are ignored, and they are tokenized as one item.
        // An error is returned if there is an odd number of quotes.
```

packages/fuels-core/src/tokenizer.rs
```diff
@@ -0,0 +1,866 @@
...
+            return Err(Error::InvalidData(
+                "array value string must start and end with square brackets".into(),
+            ));
+        }
+
+        if value.chars().count() == 2 {
+            return Ok(Token::Array(vec![]));
+        }
+
+        //for more details about this algorithm, refer to the tokenize_struct method
```
digorithm, line 181:
```suggestion
        // For more details about this algorithm, refer to the tokenize_struct method.
```

packages/fuels-core/src/tokenizer.rs
```diff
@@ -0,0 +1,866 @@
...
+            return Err(Error::InvalidData(
+                "tuple value string must start and end with round brackets".into(),
+            ));
+        }
+
+        if value.chars().count() == 2 {
+            return Ok(Token::Tuple(vec![]));
+        }
+
+        //for more details about this algorithm, refer to the tokenize_struct method
```
digorithm, line 278:
```suggestion
        // For more details about this algorithm, refer to the tokenize_struct method.
```

digorithm (changes requested):
Much cleaner! Also, I'm assuming there are no changes to the core logic, we're just moving things around here, so I didn't deeply review the core logic. 

Left just some comment-related nits.

packages/fuels-core/src/tokenizer.rs
```diff
@@ -0,0 +1,866 @@
...
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+impl Tokenizer {
+    /// Takes a ParamType and a value string and joins them as a single
+    /// Token that holds the value within it. This Token is used
+    /// in the encoding process.
+    pub fn tokenize(&self, param: &ParamType, value: String) -> Result<Token, Error> {
```
segfault-magnet, line 20: Since it doesn't have any state, we can go for associated functions rather than methods.

i.e.
```rust
let result = Tokenizer::tokenize(param_type, value);
```

packages/fuels-core/src/tokenizer.rs
```diff
@@ -0,0 +1,866 @@
+use crate::Token;
+use fuels_types::errors::Error;
+use fuels_types::param_types::ParamType;
+use fuels_types::utils::has_array_format;
+use hex::FromHex;
+
+#[derive(Default)]
+pub struct Tokenizer {}
```
segfault-magnet, line 8: You can do without the parentheses:
```rust
pub struct Tokenizer;
```

segfault-magnet (approved):

Issue #437, opened 2022-06-29
Strict validation on string contents and length


segfault-magnet:
ATTOW the following is true for strings over in the sway compiler:

- a character takes up one byte.
- strings are not variable length -- they have a length that is part of their type (just as the array size is part of the array's type in Rust)

The above two points are not enforced in our Rust bindings -- we use the `String` type from the std lib in Rust which is:

1. UTF-8 encoded (this means 1 to 4 B of encoding vs 1B expected by the ABI)
2. of variable length (and we don't check the length when encoding). 

This means if our SDK users don't pay attention they can easily mangle data while calling their contracts since a string that is too long or too short will move bytes out of their expected places.

The same goes if the string should ever carry anything which doesn't fall into the ASCII subset -- although we may sometimes hit the required number of bytes some of them won't be valid ASCII.

Until string support becomes better in the sway compiler we can implement a length and ascii-only content check on our strings so that we may fail early rather than opt for undefined behavior.

PR #496, opened 2022-07-25
feat: validate string length and characters

hal3e:
closes https://github.com/FuelLabs/fuels-rs/issues/437

The approach:
- The user-facing functions stay the same and Strings are used.
- Internally, the `Token::String(String)` was changed to `Token::String(StringToken))` where `StringToken` is a struct with fields `data` and `expected_len` (naming is not perfect. I am open for suggestions :) )
- Inside the `abigen` we create the `StringToken` with the expected length and prepare the conversion from `String`
- If the user inputs a `String` with the wrong length or with non-ascii characters the encoder will return errors with appropriate messages.

- I have added corresponding test to both the `json_abi` and the `harness`

Addition:
- In the `json_abi` we cannot parse the `value` string if a non-ascii character is there. This is because we iterate over the value string using `chars` and `positions` and this is not possible with non-ascii chars as they can be encoded with more than 1 byte. For this case, I just added a new check when parsing the value string and return the `InvalidData` error if non-ascii characters are present.


packages/fuels-core/src/abi_encoder.rs
```diff
@@ -65,8 +66,19 @@ impl ABIEncoder {
         self.encode_tokens(arg_array)
     }
 
-    fn encode_string(&mut self, arg_string: &str) {
-        self.buffer.extend(pad_string(arg_string));
+    fn encode_string(&mut self, arg_string: &StringToken) {
+        if !arg_string.0.is_ascii() {
+            panic!("String parameters can only have ascii values");
```
segfault-magnet, line 71: Since we're leaving for the user the responsibility of passing appropriately sized strings to the encoder, I'm not sure we should panic in case he does not do so mostly because it doesn't seem to me to be such an unexpected thing to happen, especially since we're still using rust Strings.

A Result might be better.

If we were not building an SDK I'd say leave the panic, because it points to the error of the programmer for not guarding against this, and panicking would be a fail-early approach.

But this is a library, hate to make such decisions for our users -- don't know how important their process is for us to just kill it by panicking.

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -65,8 +66,19 @@ impl ABIEncoder {
...
+        }
+
+        if arg_string.0.len() != arg_string.1 {
```
segfault-magnet, line 74: Readability might benefit from having the tuple restructured into descriptively named variables:
```rust
let (str_data, expected_len) = arg_string;

if str_data.len() != expected_len { ...
```

Perhaps even a reason for using a struct instead of a tuple:
```rust
struct StringToken {
    data: String,
    expected_len: usize,
}

impl StringToken {
    pub fn correct_len(&self) -> bool {
        self.data.len() == self.expected_len
    }
}

// later on
if !string_arg.correct_len() {
    // ...
}

```

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -257,7 +257,11 @@ fn expand_function_arguments(
         args.push(quote! { #name: #tok });
 
         // This `name` TokenStream is also added to the call arguments
-        call_args.push(name);
+        if let ParamType::String(len) = kind {
```
segfault-magnet, line 260: If you use `kind` by reference:
```rust
        let tok = if let ParamType::Tuple(_tuple) = &kind {
```
and
```rust
        if let ParamType::String(len) = &kind {
```
then you can remove the `.clone()`

packages/fuels-core/src/lib.rs
```diff
@@ -104,12 +105,12 @@ impl Tokenizable for bool {
...
 }
 
-impl Tokenizable for String {
+impl Tokenizable for StringToken {
     fn from_token(token: Token) -> Result<Self, Error> {
         match token {
             Token::String(data) => Ok(data),
             other => Err(Error::InstantiationError(format!(
-                "Expected `String`, got {:?}",
+                "Expected `StringToken`, got {:?}",
```
segfault-magnet, line 113: Actually, you expected a `Token::String` which in turn has a `StringToken` -- i.e. the reason you're in the catch-all is not because of not having a `StringToken` but due to not having a `Token::String`.

packages/fuels-core/src/abi_encoder.rs
hal3e: I agree that at this part of the code a Result would be better and I added it. However, the panic is still going to happen because of the way the contract methods are built. The `method_hash` function from `fuels-contract` calls an `unwrap` on the encoder result: `let encoded_args = ABIEncoder::encode(args).unwrap();` and IMO this is ok as we should stop if there are any errors within the encoder.

hal3e:
I have changed the `StringToken` to a struct and updated the PR description.

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -66,19 +66,23 @@ impl ABIEncoder {
         self.encode_tokens(arg_array)
     }
 
-    fn encode_string(&mut self, arg_string: &StringToken) {
-        if !arg_string.0.is_ascii() {
-            panic!("String parameters can only have ascii values");
+    fn encode_string(&mut self, arg_string: &StringToken) -> Result<(), CodecError> {
+        if !arg_string.data.is_ascii() {
```
segfault-magnet, line 70: Looking at this I'm wondering two things:
1. Can we extract these two checks into `impl StringToken`?
2. Better yet, can we make StringToken conditionally constructible?

I'm not sure of the different ways tokens are used but I have a feeling that once you have a Token::String on your hands you're not going to change its innards but rather pass it around. 

If this is so then we can effectively treat `StringToken` as an immutable data structure -- make its innards private, have a `new` method that will return `Result<StringToken, Error>` and have the assurance that if we managed to get our hands on a `StringToken` the string inside is definitely valid.

Thoughts?

hal3e: I have implemented option 1. Regarding option 2, we can not implement it without completely changing how the contract methods are generated and I do not recommend that.

packages/fuels-core/src/lib.rs
```diff
@@ -29,6 +29,31 @@ pub type Selector = ByteArray;
 pub type Bits256 = [u8; 32];
 pub type EnumSelector = (u8, Token, EnumVariants);
 
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct StringToken {
+    pub data: String,
```
segfault-magnet, line 34: I'd make this private and create a `new` fn.

Don't see a reason to allow somebody to extract potentially invalid `data` from this struct.

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -119,7 +119,7 @@ impl ABIDecoder {
         let decoded = str::from_utf8(encoded_str)?;
 
         let result = DecodeResult {
-            token: Token::String(decoded.to_string()),
+            token: Token::String(StringToken::new(decoded.into(), *length)),
```
iqdecay, line 122: Not your code but why is the `length` argument a reference? Seems weird to me.

packages/fuels-core/src/lib.rs
```diff
@@ -29,6 +29,35 @@ pub type Selector = ByteArray;
...
+    pub fn get_encodable_str(&self) -> Result<&str, CodecError> {
+        if !self.data.is_ascii() {
+            return Err(CodecError::InvalidData(
+                "String parameters can only have ascii values".into(),
+            ));
+        }
+
+        if self.data.len() != self.expected_len {
+            return Err(CodecError::InvalidData(format!(
+                "String parameter has len {}, but should have {}",
```
iqdecay, line 52: `Provided String data has len {}, but has len {} expected`

packages/fuels-core/src/lib.rs
```diff
@@ -105,7 +134,7 @@ impl Tokenizable for bool {
     }
 }
 
-impl Tokenizable for String {
+impl Tokenizable for StringToken {
     fn from_token(token: Token) -> Result<Self, Error> {
         match token {
             Token::String(data) => Ok(data),
```
iqdecay, line 140: Here I think you should explicit that `data` is actually a tuple (iiuc), so it's less confusing.

iqdecay (changes requested):
Very good!!! Just some nits.

packages/fuels-core/src/lib.rs
hal3e: It is a `struct ` but I got your point. We could put `data @ StringToken { .. }` but IMO it is redundant as we are implementing it for `StringToken`. I changed `data` to `string_token` and I think it is clear what is done here.

iqdecay (approved):
LGTM!

PR #497, opened 2022-07-26
contract connect function

### LINKED ISSUE #486: Function to create new instances of a contract with different wallets connected

Salka1988:
Closes #486

Added a `_connect` function that allows us to change the wallet on the `Contract` instance.


packages/fuels/tests/harness.rs
```diff
@@ -2314,3 +2315,53 @@ async fn test_contract_id_and_wallet_getters() {
...
+        MyContract,
+        "packages/fuels/tests/test_projects/contract_test/out/debug/contract_test-abi.json"
+    );
+
+    let num_wallets = 2;
+    let num_coins = 3;
+    let amount = 1000000000000;
+    let config = WalletsConfig::new(Some(num_wallets), Some(num_coins), Some(amount));
+
+    let wallets = launch_custom_provider_and_get_wallets(config, None).await;
```
iqdecay, line 2331: it would be clearer to have:
```rust
first_wallet = wallets.get(1)?;
second_wallet = wallets.get(2)?;
```

packages/fuels/tests/harness.rs
```diff
@@ -2314,3 +2315,53 @@ async fn test_contract_id_and_wallet_getters() {
...
+    assert!(gas_used > 0);
+
+    let gas_used_two = contract_instance
+        .connect(wallets.get(1).unwrap().clone())
+        .initialize_counter(42)
+        .call()
+        .await?
+        .gas_used;
+
+    assert!(gas_used_two > 0);
```
iqdecay, line 2364: here, you're not actually testing something that would "prove" that the transaction did occur with the second wallet. You're just testing that it spent some gas. What you can do is probably use non-zero gas price, and check that the wallet_1 balance is unchanged, while wallet_2 balance is diminished (base asset balance I mean).

iqdecay (changes requested):
Nice functionality, but I'm not sure about the naming and the design of the test :)

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -144,6 +144,19 @@ impl Abigen {
                         pub fn _get_wallet(&self) -> LocalWallet {
                             self.wallet.clone()
                         }
+
+                        pub fn connect(&mut self, new_wallet: LocalWallet) -> Self {
```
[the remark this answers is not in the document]
Salka1988: can we do that here?

MujkicA:
I've agreed with @Salka1988 to finalise this in his stead.

segfault-magnet (dismissed):

MujkicA:
I've changed the new method name to `_with_wallet()` for readability and transparency.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -140,6 +140,13 @@ impl Abigen {
                         pub fn _get_wallet(&self) -> LocalWallet {
                             self.wallet.clone()
                         }
+
+                        pub fn _with_wallet(&self, mut wallet: LocalWallet) -> Result<Self, SDKError> {
+                           let provider = self.wallet.get_provider()?;
+                           wallet.set_provider(provider.clone());
+
+                           Ok(Self { contract_id: self.contract_id.clone(), wallet: wallet })
+                        }
```
digorithm, line 149: We should document somewhere that the `Provider` inside `wallet` will be ignored and superseded by the `Provider` originally used to create the contract instance. 

Users could try and create a wallet that holds a different `Provider`, pass it to `_with_wallet()` and get some weird errors because that `Provider` has been ignored.

I don't think this should be a use case at all (i.e., testing stuff with a single instance on two different wallets on two different providers), but we should at least cover this in the docs somewhere.

digorithm (changes requested):
Looking great! Just a small doc-related nit.

packages/fuels-core/src/code_gen/abigen.rs
MujkicA: Docs have been updated. Let me know if it is clear enough.

digorithm (approved):
Docs looking good! :shipit:

iqdecay (approved):
vry gud!

Issue #345, opened 2022-05-28
feat: the SDK should differentiate `invalid transactions` and `reverting transactions` 


iqdecay:
There are 3 states:
1. The transaction is invalid or the node is offline,
2. The transaction is valid, but reverts,
3. The transaction is valid, and executes to completion.

With #324, even if logging and debugging was improved, we are still bundling `1.` and `2.` together, whereas we want them to be differentiated.

MujkicA:
If #411 is implemented to use `Transaction::validate_without_signature()`, the first case should be covered.
As for the second case: 
When a contract call fails, we receive `TransactionStatus::Failure` where the field `program_state` is set to the variant `Revert` if a a RVRT instruction was run.

PR #498, opened 2022-07-27
Improved error-handling

### LINKED ISSUE #249: Improve the quality of error-handling
While implementing "load_sway_contract_with_salt", I realized that in the case of the error, in my case TransactionError, I was missing a context that would explain to me in more detail why a particular problem arose (e.g "Input already exists").
Since that error happens in the fuel-core, there was nothing I could do.

Could we perhaps improve / extend the error handling to make it easier for both us and the users to use the code?

MujkicA:
Closes #249, #345
A previous PR already added tx validation checks on the SDK side. This PR focuses on making our error types and handling cleaner.

**Problem:**
- We convert `ProviderError` -> `WalletError` -> `Error::WalletError`. Communication errors with a node bubble up as a `WalletError` since the wallet always fetches spendable coins first before transactions.
- We convert `std::io::Error` -> 'Error::ContractCallError' because the gql client interface works with `std::io::Error`. This means that any failure to communicate with a node will be reported as a call error instead of a generic communication error. Further, this will also cause errors originating from e.g. file reads to bubble up to a call error.
- Multiple error types that are used in only one place, even though they don't provide additional info or convey the same info as another type

**Solution:**
To fully abstract the gql client errors, we convert `std::io::Error` -> `ProviderError` within the context of the `Provider` wrapper. This also means that we have to use `Provider` instead of `FuelClient` for exchanging data with a node so that errors can be propagated properly.
After refactoring this leaves us with 3 main error types related to transactions and node interaction:
- ProviderError (network and request related issues)
- ValidationError (transaction format validity)
- RevertTransactionError (reverted transaction)

I've made it so that `deploy()` and `wallet.transfer()` also query the node for the transaction status. This incurs additional network request however. I'm not sure about this change and would appreciate some input.
I've also consolidated `MissingData` and `InvalidName` to `InvalidData` since they were only used in one place.

The purpose of the following errors is still questionable to me:
`InfrastructureError`
`CompilationError`, `InvalidType` - does the user care for this kind of granularity? can they be replaced with `InvalidData`?


packages/fuels-signers/src/provider.rs
```diff
@@ -67,10 +64,26 @@ impl Provider {
...
-        let tx_id = self.client.submit(tx).await?;
+    pub async fn send_transaction(&self, tx: &Transaction) -> Result<Vec<Receipt>, Error> {
+        let (status, receipts) = self.submit_with_feedback(tx).await?;
+
+        match status {
+            TransactionStatus::Failure { reason, .. } => {
+                Err(Error::RevertTransactionError(reason, receipts))
+            }
+            _ => Ok(receipts),
+        }
```
segfault-magnet, line 75:
```suggestion
        if let TransactionStatus::Failure { reason, .. } = status {
            Err(Error::RevertTransactionError(reason, receipts))
        } else {
            Ok(receipts)
        }
```

Looks simpler?

packages/fuels-signers/src/provider.rs
```diff
@@ -226,28 +247,31 @@ impl Provider {
...
         &self,
         request: PaginationRequest<String>,
-    ) -> std::io::Result<PaginatedResult<TransactionResponse, String>> {
-        self.client.transactions(request).await
+    ) -> Result<PaginatedResult<TransactionResponse, String>, ProviderError> {
+        Ok(self.client.transactions(request).await?)
     }
 
-    // - Get transaction(s) by owner
+    /// Get transaction(s) by owner
```
segfault-magnet, line 265:
```suggestion
    // Get transaction(s) by owner
```

packages/fuels-signers/src/provider.rs
```diff
@@ -226,28 +247,31 @@ impl Provider {
...
     pub async fn get_transactions_by_owner(
         &self,
         owner: &str,
         request: PaginationRequest<String>,
-    ) -> std::io::Result<PaginatedResult<TransactionResponse, String>> {
-        self.client.transactions_by_owner(owner, request).await
+    ) -> Result<PaginatedResult<TransactionResponse, String>, ProviderError> {
+        Ok(self.client.transactions_by_owner(owner, request).await?)
```
segfault-magnet, line 271:
```suggestion
        self.client
            .transactions_by_owner(owner, request)
            .await
            .map_err(Into::into)
 ```
 
 I believe I saw you do this in a previous PR. It is more transparent but the current version seems more... ergonomic? 
 
 Dunno just wanted to point it out.

File: packages/fuels/tests/harness.rs [hunk 4]
```diff
@@ -2314,3 +2316,39 @@ async fn test_contract_id_and_wallet_getters() {
...
+        StorageConfiguration::default(),
+    )
+    .await;
+
+    match response {
+        Err(ProviderError(_)) => {}
+        _ => {
+            panic!("This should be impossible");
+        }
+    };
```
segfault-magnet, line 2351:
```suggestion
    assert!(matches!(response, Err(ProviderError(_))));
```

packages/fuels-signers/src/provider.rs
MujkicA: Potato potato 😁

segfault-magnet: No u didn't :D It has one indent less! A universal banana measurement.

examples/contracts/src/lib.rs
```diff
@@ -333,11 +332,16 @@ mod tests {
                 let receipts: Vec<Receipt> = call_response.receipts;
                 // Do things with logs and receipts
             }
-
-            // The transaction is invalid or node is offline
-            // OR
+            // The transaction is malformed
+            Err(ValidationError(e)) => {
+                println!(": {}", e);
```
iqdecay, line 337:
```suggestion
                println!("Transaction is malformed (ValidationError): {}", e);
```

packages/fuels-contract/src/contract.rs
```diff
@@ -439,7 +434,7 @@ pub struct ContractCallHandler<D> {
     pub contract_call: ContractCall,
     pub tx_parameters: TxParameters,
     pub wallet: LocalWallet,
-    pub fuel_client: FuelClient,
+    pub provider: Provider,
```
iqdecay, line 437: All of this is sooo much cleaner, I love it. It had been bugging me for a while lol

packages/fuels-signers/src/provider.rs
```diff
@@ -15,29 +16,25 @@ use fuel_gql_client::{
...
+impl From<ProviderError> for Error {
+    fn from(e: ProviderError) -> Self {
+        Error::ProviderError(e.to_string())
     }
 }
+
 /// Encapsulates common client operations in the SDK.
 /// Note that you may also use `client`, which is an instance
-/// of `FuelClient`, directly, which providers a broader API.
+/// of `FuelClient`, directly, which provides a broader API.
```
iqdecay, line 37: Should we keep this comment? I feel like it's a little misguiding, we are explicitely doing these changes so we stop using the `FuelClient` directly. wdyt?

packages/fuels-signers/src/provider.rs
```diff
@@ -67,10 +64,25 @@ impl Provider {
...
 
-        self.client.receipts(&tx_id.0.to_string()).await
+        if let TransactionStatus::Failure { reason, .. } = status {
+            Err(Error::RevertTransactionError(reason, receipts))
+        } else {
+            Ok(receipts)
+        }
+    }
+
+    async fn submit_with_feedback(
```
iqdecay, line 77: cool function

packages/fuels-signers/src/wallet.rs
```diff
@@ -73,21 +73,14 @@ pub enum WalletError {
...
     Parsing(#[from] std::array::TryFromSliceError),
     #[error("No provider was setup: make sure to set_provider in your wallet!")]
     NoProvider,
-    #[error("Provider error: {0}")]
-    ProviderError(#[from] ProviderError),
     /// Keystore error
     #[error(transparent)]
     KeystoreError(#[from] KeystoreError),
-    #[error("invalid mnemonic word count (expected 12, 15, 18, 21, 24, found `{0}`")]
-    InvalidMnemonicWordCount(usize),
```
iqdecay, line 90: Why did you remove it? Even if it is not used currently, it would probably be in the future? Or is it to "merge it" with `MnemonicError`?

packages/fuels/src/lib.rs
```diff
@@ -66,4 +66,5 @@ pub mod prelude {
     pub use super::test_helpers::*;
     pub use super::tx::Salt;
     pub use super::types::errors::Error;
+    pub use super::types::errors::Error::*;
```
iqdecay, line 69: Is this really needed?

packages/fuels-types/src/errors.rs
```diff
@@ -47,20 +45,18 @@ pub enum Error {
     Utf8Error(#[from] Utf8Error),
     #[error("Compilation error: {0}")]
     CompilationError(String),
-    #[error("Network error: {0}")]
-    NetworkError(String),
```
iqdecay, line 51: NetworkError I'm not sure should disappear, what happens if for instance the node actually works, but your localhost has badly configured firewall or whatever? It's semantically very different IMO.

iqdecay (changes requested):
Looking very good, definitely a nice overhaul! Some questions about deletions.

packages/fuels-signers/src/provider.rs
MujkicA: Since we don't fully wrap the client I figured we should still leave the user a way to interact with it. 
@segfault-magnet also made a good point about the possibility that the client could add new functionalities before the SDK catches up so the user would still have some way to access those until our next release.

packages/fuels-signers/src/wallet.rs
MujkicA: I assumed that it was used at some point and now lost its purpose, the YAGNI way would be to remove it, but I don't mind keeping it. Its your call.

iqdecay: Yeah I guess merging it into `MnemonicError` is fine either way, trash it!

packages/fuels-types/src/errors.rs
MujkicA: I looked into this a bit more. The gql client always returns a generic error type of variant [Other](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.Other) which means we cannot differentiate between actual network errors or other client related errors by parsing the error type. I prefer to return a generic `ClientRequestError` over parsing error messages.

iqdecay (dismissed):
LGTM!

iqdecay (approved):

PR #499, opened 2022-07-27
Fix deunified build ignoring workspace package

Voxelot:
While the deunified build created a number of parallel jobs, the package field wasn't properly passed to certain jobs as expected, leading to redundant checks.


iqdecay (approved):
Honestly have no idea what it means but LGTM! :smile:

Voxelot:
this is still a wip, need to fix workspace package name resolution since they don't match the directory names.

Voxelot:
fixed in https://github.com/FuelLabs/fuels-rs/pull/657

PR #501, opened 2022-07-28
feat: launch node with `manual_blocks_enabled` without features

### LINKED ISSUE #489: feat: allow starting the node binary with custom parameters
For instance starting the node with `manual_blocks_enabled` so the `produce_blocks` API can be used.

iqdecay:
This PR enables starting a node with the `manual_blocks_enabled` even if the `fuel-core-lib` feature is not enabled.


packages/fuels-signers/src/wallet.rs
```diff
@@ -46,9 +46,9 @@ type W = English;
 ///
 ///   // Recover address that signed the message
 ///   let message = Message::new(message);
-///   let recovered_address = signature.recover(&message).unwrap();
+///   let recovered_address = signature.recover(&message)?;
 ///
-///   assert_eq!(wallet.address(), recovered_address);
+///   assert_eq!(wallet.address().to_string(), recovered_address.to_string());
```
MujkicA, line 51: nit: compare `.hash()` instead of string

MujkicA: nit: also change the function return type to our own error type instead of std

packages/fuels-test-helpers/src/node.rs
```diff
@@ -27,12 +27,14 @@ use tokio::process::Command;
...
 pub struct Config {
     pub addr: SocketAddr,
+    pub manual_blocks_enabled: bool,
 }
 
 impl Config {
     pub fn local_node() -> Self {
         Self {
             addr: SocketAddr::new(Ipv4Addr::new(127, 0, 0, 1).into(), 0),
+            manual_blocks_enabled: false,
```
MujkicA, line 37: Would it be better to just have this set to true for our usual test setup methods?

MujkicA (commented):
Added some nits. Also, #489 mentions parameters in general while this only handles `manual_blocks_enabled `. What is the plan for the remaining params?

iqdecay:
Yeah I realized that the plans for `parameters in general` are harder because it comes down to adding a complete config of things. In that sense, it does not completely close #489.

packages/fuels-test-helpers/src/node.rs
iqdecay: I'm not sure, since we are mostly not using this feature,it's pretty "potent/dangerous", and the error message is pretty clear.

MujkicA (approved):
We should keep track of the remaining features, some are also being exposed by #432. Other than that LGTM!

Issue #505, opened 2022-07-28
fix: ABI with array inputs and outputs raise an error.


QuinnLee:
I wrote some test for fuels-rs [here](https://github.com/FuelLabs/fuels-rs/pull/503/files) 

Generating the abi with cargo run --bin build-test-projects works, but on cargo test I get this error
```
error: proc macro panicked
    --> packages/fuels/tests/harness.rs:2310:5
     |
2310 | /     abigen!(
2311 | |         TestContract,
2312 | |         "packages/fuels/tests/test_projects/array_contract_inputs/out/debug/array_contract_inputs-abi.json"
2313 | |     );
     | |_____^
     |
     = help: message: called `Result::unwrap()` on an `Err` value: InvalidType("Expected parameter type `str[n]`, found `[str[3]; 3]`")
```
I'm on forc 0.19.0

QuinnLee:
oops closed the wrong one

PR #506, opened 2022-07-29
fix: string in array parsing

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/505

There was a bug in the parsing logic of arrays and strings.
I have taken the test_project created by @QuinnLee and removed the part with primitive types inside arrays, as we have tested them already.

### DIFF
--- a/packages/fuels-types/src/parse_param.rs
+++ b/packages/fuels-types/src/parse_param.rs
@@ -17,12 +17,11 @@ impl TryFrom<&Property> for ParamType {
                 if prop.type_field == "()" {
                     return Ok(ParamType::Unit);
                 }
-                if prop.type_field.contains('[') && prop.type_field.contains(']') {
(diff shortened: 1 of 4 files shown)


QuinnLee:
The memory flow issue might be a compiler issue.  I created it https://github.com/FuelLabs/sway/issues/2410 to keep track of it.  Feel free to update

hal3e:
As we have fixed the parsing issue in the SDK, I  would suggest we close the PR and check the function after it gets fixed by the compiler team.

digorithm (approved):
Good stuff! 

Do you mind also creating an issue on `fuels-rs` for re-enabling the skipped test once the compiler issue is resolved?

hal3e:
Thanks! I added the new issue (https://github.com/FuelLabs/fuels-rs/issues/509).

Issue #509, opened 2022-07-30
enable skipped test in harness.rs


hal3e:
In the PR https://github.com/FuelLabs/fuels-rs/pull/506 we skipped the test on
`fn take_array_string_return_single_element(a: [str[3];3]) -> str[3];` as it results in a compiler error.
We should wait until https://github.com/FuelLabs/sway/issues/2410 is resolved and then enable it.

iqdecay:
Good thinking to track it in an issue!

PR #512, opened 2022-08-02
fix: enable skipped test for strings in arrays

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/509

The compiler error was fixed and we can now run the full test.


digorithm (approved):
:shipit:

Issue #513, opened 2022-08-02
Wallets created with helper functions should be deterministic


simonr0204:
Currently, when using helper functions to launch a node with a number of wallets (e.g. `launch_custom_provider_and_get_wallets`), the wallets are created with a random seed and are non-deterministic.

It would be helpful for these to be deterministic so that they can be hard-coded into Sway programs if necessary. 

I think this just requires these methods to use `LocalWallet::new_from_private_key` (where the private key would basically go from 0 to however many wallets are requested) under the hood instead of `LocalWallet::new_random`

PR #517, opened 2022-08-04
refactor: generate wallets with deterministic keys

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/513

Test wallets are generated using deterministic keys.

### DIFF
--- a/packages/fuels-test-helpers/src/signers.rs
+++ b/packages/fuels-test-helpers/src/signers.rs
@@ -60,9 +60,23 @@ pub async fn launch_custom_provider_and_get_wallets(
     wallet_config: WalletsConfig,
     provider_config: Option<Config>,
 ) -> Vec<LocalWallet> {
-    let mut wallets: Vec<LocalWallet> = (1..=wallet_config.num_wallets)
-        .map(|_i| LocalWallet::new_random(None))
-        .collect();
+    let mut wallets: Vec<LocalWallet> = vec![];
+
+    if !wallet_config.private_keys.is_empty() {
+        wallets.extend(
+            wallet_config
+                .private_keys
+                .iter()
+                .map(|pk| LocalWallet::new_from_private_key(*pk, None)),
+        );
+    }
+
+    // If the requested `num_wallets` is graeter then the number of the provided `private_keys`,
+    // generate the remaining wallets with random keys.
+    wallets.extend(
+        (wallet_config.private_keys.len()..wallet_config.num_wallets as usize)
+            .map(|_i| LocalWallet::new_random(None)),
+    );
 
     let mut all_coins: Vec<(UtxoId, Coin)> = Vec::with_capacity(wallet_config.num_wallets as usize);
     for wallet in &wallets {
@@ -101,12 +115,13 @@ pub async fn setup_test_provider(
 
 #[cfg(test)]
 mod tests {
-    use super::*;
-    use crate::{launch_custom_provider_and_get_wallets, WalletsConfig};
+    use crate::{launch_custom_provider_and_get_wallets, AssetConfig, WalletsConfig};
     use fuels_core::constants::BASE_ASSET_ID;
-    use fuels_signers::fuel_crypto::fuel_types::AssetId;
+    use fuels_signers::fuel_crypto::{fuel_types::AssetId, SecretKey};
+    use fuels_signers::Signer;
     use fuels_types::errors::Error;
     use rand::Fill;
+    use std::str::FromStr;
 
     #[tokio::test]
     async fn test_wallet_config() -> Result<(), Error> {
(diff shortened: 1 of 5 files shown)


Br1ght0ne (commented):
LGTM, I really like the `extend` approach. Will defer approval to @FuelLabs/sdk-rust.

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -60,9 +60,23 @@ pub async fn launch_custom_provider_and_get_wallets(
...
+    if !wallet_config.private_keys.is_empty() {
+        wallets.extend(
+            wallet_config
+                .private_keys
+                .iter()
+                .map(|pk| LocalWallet::new_from_private_key(*pk, None)),
+        );
+    }
+
+    // If the requested `num_wallets` is graeter then the number of the provided `private_keys`,
```
segfault-magnet, line 74:
```suggestion
    // If the requested `num_wallets` is greater then the number of the provided `private_keys`,
```

iqdecay:
Will review later but I'm usually not a fan at all of using private keys "in plain text" as I suppose it's being done here. Seems like the best way to allow a user to footgun.

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -60,9 +60,23 @@ pub async fn launch_custom_provider_and_get_wallets(
     wallet_config: WalletsConfig,
     provider_config: Option<Config>,
 ) -> Vec<LocalWallet> {
-    let mut wallets: Vec<LocalWallet> = (1..=wallet_config.num_wallets)
-        .map(|_i| LocalWallet::new_random(None))
-        .collect();
+    let mut wallets: Vec<LocalWallet> = vec![];
```
segfault-magnet, line 63: A more functional approach: 

```rust
    let mut wallets: Vec<_> = wallet_config
        .private_keys
        .iter()
        .map(|pk| LocalWallet::new_from_private_key(*pk, None))
        .chain(iter::repeat_with(|| LocalWallet::new_random(None)))
        .take(wallet_config.num_wallets as usize)
        .collect();

    let all_coins = wallets
        .iter()
        .flat_map(|wallet| {
            setup_custom_assets_coins(wallet.address(), wallet_config.assets.clone())
        })
        .collect::<Vec<_>>();
```

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -182,4 +197,88 @@ mod tests {
...
+            SecretKey::from_str(
+                "5f70feeff1f229e4a95a7056e8b4d80d0b24b565674860cc213bdb07127ce1b1",
+            )?,
+            SecretKey::from_str(
+                "705ffeeff1f229e4a95e1056e8b4d80d0b24b565674860cc213bdb07127ce1b1",
+            )?,
+        ];
+
+        let config =
+            WalletsConfig::new(Some(num_wallets), None, None).with_private_keys(private_keys);
```
segfault-magnet, line 215: Some separation between arrange, act and assert will make this test more readable.

simonr0204:
My 2 cents on this is that I think test wallet addresses (So anything created with a `WalletConfig`) should be deterministic by default. 

I think we should discourage users from bringing their own private keys, especially when working on a test net. I can't think of a case where having a _specific_ address is necessary, so much as having a deterministic one. There's only really a use case for users' own private keys when interacting with a public network, but in that case `WalletConfig` can't be used at all. 

Happy to go with the consensus on this, but since I filed the original issue I thought I should weigh in 🙂 .

hal3e:
Just to be on the same page here. Until now we would create test wallets with `LocalWallet::new_random()` which underneath would call `new_from_private_key()` where the private key would be randomly generated. The user does not have to bring his private keys (I do not even what _his_ private keys here would mean). This is just a way to get deterministic wallets if you need them. And I do not thing we need them to be deterministic for all tests.

simonr0204:
I just mean that with this set up, when setting `private_keys` in   `.with_private_keys(private_keys)`

most of the time I would probably end up putting `0x0000000000000000000000000000000000000000000000000000000000000001` through `0x000000000000000000000000000000000000000000000000000000000000000n` where `n` is however many wallets I need., in which case we could just do this under the hood, and then wallets would always be deterministic. 

In my mind it's clear the advantages that deterministic wallet addresses bring. Are there any downsides you can see that I'm overlooking to having them by default?

hal3e:
I implemented the deterministic wallet generation. I have also refactored the `WalletsConfig` and added a missing example in the docs for custom assets.

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -60,16 +61,25 @@ pub async fn launch_custom_provider_and_get_wallets(
     wallet_config: WalletsConfig,
     provider_config: Option<Config>,
 ) -> Vec<LocalWallet> {
-    let mut wallets: Vec<LocalWallet> = (1..=wallet_config.num_wallets)
-        .map(|_i| LocalWallet::new_random(None))
+    let mut wallets: Vec<_> = (1..=wallet_config.num_wallets())
+        .map(|wallet_counter| {
+            let bytes32: Vec<u8> = [0; 24] // [u8; 24] is padding as u64.to_be_bytes() is [u8,8]
```
segfault-magnet, line 66: You can do without the comment:
```rust
    const PADDING_BYTES: usize = std::mem::size_of::<Address>() - std::mem::size_of::<u64>();
    let bytes32: Vec<u8> = [0; PADDING_BYTES]
```

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -60,16 +61,25 @@ pub async fn launch_custom_provider_and_get_wallets(
...
+                .into_iter()
+                .chain(wallet_counter.to_be_bytes().into_iter())
```
segfault-magnet, line 68: This is beyond beautiful. Kudos!

packages/fuels-test-helpers/src/wallets_config.rs
```diff
@@ -37,6 +37,14 @@ impl WalletsConfig {
             assets,
         }
     }
+
+    pub fn num_wallets(&self) -> u64 {
+        self.num_wallets
+    }
+
+    pub fn assets(&self) -> &Vec<AssetConfig> {
```
segfault-magnet, line 45: Maybe return a slice here?

iqdecay (dismissed):
Looks really clean, well done! :ship:

hal3e:
> Looks really clean, well done! ship

Thanks!

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -60,16 +61,26 @@ pub async fn launch_custom_provider_and_get_wallets(
...
+
+    let mut wallets: Vec<_> = (1..=wallet_config.num_wallets())
+        .map(|wallet_counter| {
+            secret_key[PADDING_BYTES..].copy_from_slice(&wallet_counter.to_be_bytes());
+
+            return LocalWallet::new_from_private_key(
+                SecretKey::try_from(secret_key.as_slice())
+                    .expect("This should never happen as we provide a [u8; SIZE_SECRET_KEY] array"),
+                None,
+            );
```
segfault-magnet, line 76:
```suggestion
            LocalWallet::new_from_private_key(
                SecretKey::try_from(secret_key.as_slice())
                    .expect("This should never happen as we provide a [u8; SIZE_SECRET_KEY] array"),
                None,
            )
```

segfault-magnet (approved):

Br1ght0ne (approved):

digorithm (approved):
Great stuff! :shipit:

PR #519, opened 2022-08-04
deps: update `coins-bip` versions

iqdecay:
Update `coins-bips` deps versions.


iqdecay:
Closing because even though CI fails, it is not for the reason I'm trying to diagnose right now.

PR #520, opened 2022-08-04
Use `fuel-crypto` for secret key generation from mnemonic

digorithm:
This removes the use of `bip32/39` dependencies from the SDK and moves the mnemonic creation logic to `fuel-crypto`.


hal3e (approved):
LGTM

iqdecay (approved):
:ship:

PR #523, opened 2022-08-05
Cookbook examples

### LINKED ISSUE #508: Document how to run a _custom_ short-lived Fuel node
We currently have documented how to run a short-lived Fuel node, but only with default values. We should extend it to show how to create a custom node using `setup_test_client`, which involves using `ConsensusParameters` to configure a custom chain.

MujkicA:
Closes #508

This PR adds the following examples to the cookbook section of the docs:

1. Starting a client with custom consensus parameters
2. Liquidity pool deposit and withdraw
3. Custom transfer transaction for all assets

Also included is a wallet helper method that creates outputs for a transfer in an effort to abstract away some fuels-tx types.


Salka1988 (commented):
LGTM!

hal3e (approved):

PR #524, opened 2022-08-06
Document setup coins

### LINKED ISSUE #507: Document `setup_multiple_assets_coins` and `setup_single_asset_coins` limitations w.r.t. number of coins
Currently, we have a known performance issue with regards to the number of coins being created in test helpers such as [setup_single_asset_coins](https://github.com/FuelLabs/fuels-rs/blob/c7fd123165d6e847a9f61a029d9b6fe854ab4b0b/packages/fuels-test-helpers/src/lib.rs#L109) and [setup_multiple_assets_coins](https://github.com/FuelLabs/fuels-rs/blob/c7fd123165d6e847a9f61a029d9b6fe854ab4b0b/packages/fuels-test-helpers/src/lib.rs#L60). When a large `num_of_coins` is passed to these functions, it hangs while it loops over the creation of each coin. 

While this problem isn't solved, we should at the very least document this limitation, so users are aware that a large `num_of_coins` may cause delays in their tests.

MujkicA:
Closes #507

I've determined some values for the number of coins and assets that lead to reasonable runtimes and added them as a recommended limit to the docs. 
I've also added a small performance improvement to `setup_multiple_assets_coins`

### DIFF
--- a/docs/src/wallets/test-wallets.md
+++ b/docs/src/wallets/test-wallets.md
@@ -57,3 +57,5 @@ If you want to create coins only with the base asset, then you can use:
 ```rust,ignore
 {{#include ../../../examples/providers/src/lib.rs:setup_single_asset}}
 ```
+
+>**Note** Choosing a large number of coins and assets for `setup_multiple_assets_coins`, or number of coins for `setup_single_asset_coins`, can lead to considerable runtime for these methods. This will be improved in the future but for now we recommend using upto **1_000_000** coins, or **1000** coins and assets simultaneously.
--- a/docs/src/wallets/test-wallets.md
+++ b/docs/src/wallets/test-wallets.md
@@ -58,4 +58,4 @@ If you want to create coins only with the base asset, then you can use:
 {{#include ../../../examples/providers/src/lib.rs:setup_single_asset}}
 ```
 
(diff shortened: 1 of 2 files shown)


docs/src/wallets/test-wallets.md
```diff
@@ -57,3 +57,5 @@ If you want to create coins only with the base asset, then you can use:
 ```rust,ignore
 {{#include ../../../examples/providers/src/lib.rs:setup_single_asset}}
 ```
+
+>**Note** Choosing a large number of coins and assets for `setup_multiple_assets_coins`, or number of coins for `setup_single_asset_coins`, can lead to considerable runtime for these methods. This will be improved in the future but for now we recommend using upto **1_000_000** coins, or **1000** coins and assets simultaneously.
```
digorithm, line 61:
```suggestion
>**Note** Choosing a large number of coins and assets for `setup_multiple_assets_coins` or `setup_single_asset_coins` can lead to considerable runtime for these methods. This will be improved in the future but for now, we recommend using up to **1_000_000** coins, or **1000** coins and assets simultaneously.
```

digorithm (changes requested):
Nice -- just left a tiny grammar nit!

digorithm (approved):
:shipit:

Issue #514, opened 2022-08-02
Interface for gas limit estimation


digorithm:
We need a simple method for gas limit estimation. It could be something similar to what was done in https://github.com/FuelLabs/fuels-ts/pull/400. See the linked PR for more context and detail.

PR #527, opened 2022-08-11
feat: add transaction cost estimation

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/514

This PR is heavily based on https://github.com/FuelLabs/fuels-ts/pull/400

The `provider` was updated and the following methods were added / changed:
  - `get_transaction_cost()` returns `TransactionCost` struct
  - `get_gas_used()` returns the estimated gas used from the `dry_run` receipt
  - `chargeableByteSize()` return byte size of the tx - byte size of witness
  - `send_transaction()` was updated. Before we submit the transaction, we call `get_transaction_cost` and check if `gas_price`, `byte_price` and `gas_limit` meet the required amounts of `min_gas_price`, `min_byte_price` and `gas_used`. This prevents transactions from failing after submitting and spending coins. 
 
The `ContractCallHandler` and `ContractMultiCallHandler` got a wrapper function
  - `get_transaction_cost()` which call the `provider's` function
  
I have added test and updated the docs.

### DIFF
--- a/docs/src/calling-contracts/cost-estimation.md
+++ b/docs/src/calling-contracts/cost-estimation.md
@@ -1,6 +1,6 @@
 # Estimating contract call cost
 
-With with the function `get_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for a specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the estimation:
+With with the function `estimate_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for a specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the estimation:
 
 
 ```rust,ignore
--- a/docs/src/calling-contracts/cost-estimation.md
+++ b/docs/src/calling-contracts/cost-estimation.md
@@ -1,0 +1,29 @@
+# Estimating contract call cost
+
+With with the function `get_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for the specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the call cost:
+
+
+```rust,ignore
+TransactionCost {
+    min_gas_price: u64,
+    min_byte_price: u64,
+    byte_price: u64,
+    gas_price: u64,
+    gas_used: u64,
+    byte_size: u64,
+    fee: u64, // where fee is the sum of the gas and byte fees
+}
+```
+
+Below are examples how to get the estimate transaction cost from single and multi call transactions.
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:contract_call_cost_estimation}}
+```
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:multi_call_cost_estimation}}
+```
+
+The transaction cost estimation can be used to set the gas limit for the actual call, or to show the user the estimated cost.
+
--- a/docs/src/calling-contracts/cost-estimation.md
+++ b/docs/src/calling-contracts/cost-estimation.md
@@ -1,6 +1,6 @@
 # Estimating contract call cost
 
-With with the function `get_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for the specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the call cost:
+With with the function `get_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for a specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the estimation:
 
 
 ```rust,ignore
@@ -11,11 +11,11 @@ TransactionCost {
     gas_price: u64,
     gas_used: u64,
     byte_size: u64,
-    fee: u64, // where fee is the sum of the gas and byte fees
+    total_fee: f64, // where total_fee is the sum of the gas and byte fees
 }
 ```
 
-Below are examples how to get the estimate transaction cost from single and multi call transactions.
+Below are examples that show how to get the estimated transaction cost from single and multi call transactions.
 
 ```rust,ignore
 {{#include ../../../examples/contracts/src/lib.rs:contract_call_cost_estimation}}
@@ -25,5 +25,5 @@ Below are examples how to get the estimate transaction cost from single and mult
 {{#include ../../../examples/contracts/src/lib.rs:multi_call_cost_estimation}}
 ```
 
-The transaction cost estimation can be used to set the gas limit for the actual call, or to show the user the estimated cost.
+The transaction cost estimation can be used to set the gas limit for an actual call, or to show the user the estimated cost.
 
--- a/examples/contracts/src/lib.rs
+++ b/examples/contracts/src/lib.rs
@@ -75,6 +75,41 @@ mod tests {
 
         assert_eq!(52, response.value);
         // ANCHOR_END: use_deployed_contract
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn contract_call_cost_estimation() -> Result<(), Error> {
+        use fuels::prelude::*;
+
+        abigen!(
+            MyContract,
+            "packages/fuels/tests/test_projects/contract_test/out/debug/contract_test-abi.json"
+        );
+
+        let wallet = launch_provider_and_get_wallet().await;
+
+        let contract_id = Contract::deploy(
+            "../../packages/fuels/tests/test_projects/contract_test/out/debug/contract_test.bin",
+            &wallet,
+            TxParameters::default(),
+            StorageConfiguration::default(),
+        )
+        .await?;
+
+        // ANCHOR: contract_call_cost_estimation
+        let contract_instance = MyContractBuilder::new(contract_id.to_string(), wallet).build();
+
+        let tolerance = 0.0;
+        let transaction_cost = contract_instance
+            .initialize_counter(42) // Build the ABI call
+            .get_transaction_cost(Some(tolerance)) // Get estimated transaction cost
+            .await?;
+        // ANCHOR_END: contract_call_cost_estimation
+
+        assert_eq!(transaction_cost.gas_used, 290);
+
         Ok(())
     }
 
@@ -445,8 +480,16 @@ mod tests {
         let response = multi_call_handler.call::<(u64, Vec<u64>)>().await?;
         // ANCHOR_END: multi_call_response
 
+        // ANCHOR: multi_call_cost_estimation
+        let tolerance = 0.0;
+        let transaction_cost = multi_call_handler
+            .get_transaction_cost(Some(tolerance)) // Perform the network call
+            .await?;
+        // ANCHOR_END: multi_call_cost_estimation
+
         assert_eq!(counter, 42);
         assert_eq!(array, [42; 2]);
+        assert_eq!(transaction_cost.gas_used, 710);
 
         Ok(())
     }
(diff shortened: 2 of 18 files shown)


hal3e:
After further investigating why some test failed with `--all-features`, I found out that we were not using all of the node `Config` settings when running the `fuel-core` as a command. In addition, when not running the command we were always using `utxo-validation` irrelevant of what was set in the `Config`. I will add an update for this in the next commit.

packages/fuels-signers/src/provider.rs
```diff
@@ -399,6 +455,74 @@ impl Provider {
...
+        Ok(TransactionCost {
+            min_gas_price: min_gas_price.0,
+            min_byte_price: min_byte_price.0,
+            byte_price,
+            gas_price,
+            gas_used,
+            byte_size,
+            fee: gas_fee + byte_fee,
+        })
+    }
```
segfault-magnet, line 497: Whenever possible try to separate 'calculations' (pure functions) from 'actions' (side-effect inducing functions). Doing so will aid in testing. The thing about actions is that whichever function touches it - it itself becomes an action (impure). Something like the following is just a 5 min go at it. You can now easily unit-test the calculating logic without having to mock the heck out of the provider. 
```suggestion
/// Get estimated cost of a transaction
    pub async fn get_transaction_cost(
        &self,
        tx: &Transaction,
        tolerance: Option<f64>,
    ) -> Result<TransactionCost, Error> {
        let NodeInfo {
            min_gas_price,
            min_byte_price,
            ..
        } = self.node_info().await?;

        let tolerance = tolerance.unwrap_or(DEFAULT_GAS_ESTIMATION_TOLERANCE);
        let tmp_tx = Self::generate_dry_run_tx(tx);
        let gas_used = self.get_gas_used_with_tolerance(&tmp_tx, tolerance).await?;

        Self::calculate_tx_cost(tx, tmp_tx, min_gas_price, min_byte_price, gas_used)
    }

    fn generate_dry_run_tx(tx: &Transaction) -> Transaction {
        let mut dry_run_tx = tx.clone();
        dry_run_tx.set_gas_limit(MAX_GAS_PER_TX);
        dry_run_tx.set_gas_price(0);
        dry_run_tx.set_byte_price(0);
        dry_run_tx
    }

    fn calculate_tx_cost(
        tx: &Transaction,
        dry_run_tx: Transaction,
        min_gas_price: U64,
        min_byte_price: U64,
        gas_used: u64,
    ) -> Result<TransactionCost, Error> {
        let gas_price = std::cmp::max(tx.gas_price(), min_gas_price.0);
        let byte_price = std::cmp::max(tx.byte_price(), min_byte_price.0);

        // Simulate the contract call with MAX_GAS_PER_TX to get the complete gas_used

        let byte_size = Self::get_chargable_byte_size(dry_run_tx);
        // GAS_PRICE_FACTOR is a chain_config of the  node. Because of the different decimal precision in
        // FuelVM and EVM we need to scale the price down
        let gas_fee = ((gas_used as f64 / GAS_PRICE_FACTOR as f64) * gas_price as f64) as u64;
        let byte_fee = ((byte_size as f64 / GAS_PRICE_FACTOR as f64) * byte_price as f64) as u64;

        Ok(TransactionCost {
            min_gas_price: min_gas_price.0,
            min_byte_price: min_byte_price.0,
            byte_price,
            gas_price,
            gas_used,
            byte_size,
            fee: gas_fee + byte_fee,
        })
    }
```

packages/fuels-signers/src/provider.rs
```diff
@@ -399,6 +455,74 @@ impl Provider {
...
+            .iter()
+            .rfind(|r| matches!(r, Receipt::ScriptResult { .. }));
+
+        if let Some(script_result) = script_receipt {
+            return script_result
+                .gas_used()
+                .expect("could not retrieve gas used from ScriptResult");
+        }
+        0
+    }
```
segfault-magnet, line 519:
```suggestion
    fn get_gas_used(&self, receipts: &[Receipt]) -> u64 {
        receipts
            .iter()
            .rfind(|r| matches!(r, Receipt::ScriptResult { .. }))
            .map(|script_result| {
                script_result
                    .gas_used()
                    .expect("could not retrieve gas used from ScriptResult")
            })
            .unwrap_or(0)
    }

```

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -185,27 +184,25 @@ pub async fn setup_test_client(
     node_config: Option<Config>,
     consensus_parameters_config: Option<ConsensusParameters>,
 ) -> (FuelClient, SocketAddr) {
-    let (srv_address, manual_blocks_enabled) = match node_config {
-        Some(config) if config.addr.port() != 0 && is_free(config.addr.port()) => {
-            (config.addr, config.manual_blocks_enabled)
-        }
+    let config = match node_config {
```
segfault-magnet, line 187: This [comment](https://github.com/FuelLabs/fuels-rs/pull/488#discussion_r930118897) might apply here.

segfault-magnet (dismissed):
lgtm

packages/fuels-signers/src/provider.rs
```diff
@@ -399,6 +455,117 @@ impl Provider {
...
+
+    fn generate_dry_run_tx(tx: &Transaction) -> Transaction {
+        let mut dry_run_tx = tx.clone();
+        // Simulate the contract call with MAX_GAS_PER_TX to get the complete gas_used
+        dry_run_tx.set_gas_limit(MAX_GAS_PER_TX);
+        dry_run_tx.set_gas_price(0);
+        dry_run_tx.set_byte_price(0);
+        dry_run_tx
+    }
```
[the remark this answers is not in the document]
segfault-magnet: Functions are not just for code reuse but for code organization as well. Having this chunk named offers cognitive relief when reading the caller of this function who isn't interested in the details of the tx construction. As for the naming, I have no opinion on that, it seems ok. It being private limits the scope of the potential confusion.

Also, private functions are candidates for inlining since all of their calls are known at the time of compilation.

docs/src/calling-contracts/cost-estimation.md
```diff
@@ -0,0 +1,29 @@
...
+
+Below are examples how to get the estimate transaction cost from single and multi call transactions.
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:contract_call_cost_estimation}}
+```
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:multi_call_cost_estimation}}
+```
```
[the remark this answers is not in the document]
hal3e: I would like to keep it. I added a little bit more info and because these two are the only examples IMO it make sense to have them there.

packages/fuels-signers/src/provider.rs
```diff
@@ -60,16 +77,44 @@ impl Provider {
...
+                "gas_price({}) is lower than the required min_gas_price({})",
+                tx.gas_price(),
+                min_gas_price
+            )));
+        } else if min_byte_price > tx.byte_price() {
+            return Err(Error::ProviderError(format!(
+                "byte_price({}) is lower than the required min_byte_price({})",
+                tx.byte_price(),
+                min_byte_price
+            )));
```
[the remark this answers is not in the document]
hal3e: Either we error before or the node will consume gas and error out on insufficient gas. The same think was done in `ts-sdk`. Maybe I add a comment explaining why we are doing it ?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -176,28 +175,31 @@ pub async fn setup_test_client(
...
+    let config = node_config.unwrap_or_else(Config::local_node);
+    let requested_port = config.addr.port();
+
+    let bound_address = if requested_port == 0 {
+        get_socket_address()
+    } else if is_free(requested_port) {
+        config.addr
+    } else {
+        panic!("Error: Address already in use");
     };
```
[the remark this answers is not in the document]
hal3e: It used to be a match but IMO now it is much clearer.

docs/src/calling-contracts/cost-estimation.md
```diff
@@ -1,6 +1,6 @@
 # Estimating contract call cost
 
-With with the function `get_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for the specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the call cost:
+With with the function `get_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for a specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the estimation:
```
iqdecay, line 3:
```suggestion
With with the function `estimate_transaction_cost(tolerance: Option<f64>)` provided by `ContractCallHandler` and `ContractMultiCallHandler`, you can get a cost estimation for a specific call. The return type, `TransactionCost`, is a struct that contains relevant information for the estimation:
```

examples/contracts/src/lib.rs
```diff
@@ -480,15 +480,51 @@ mod tests {
...
+
+        multi_call_handler
+            .add_call(call_handler_1)
+            .add_call(call_handler_2);
+
         let tolerance = 0.0;
         let transaction_cost = multi_call_handler
-            .get_transaction_cost(Some(tolerance)) // Perform the network call
+            .estimate_transaction_cost(Some(tolerance)) // Get estimated transaction cost
             .await?;
```
iqdecay, line 525: perfect size of snippet :)

examples/contracts/src/lib.rs
```diff
@@ -480,15 +480,51 @@ mod tests {
...
         // ANCHOR_END: multi_call_cost_estimation
 
-        assert_eq!(counter, 42);
-        assert_eq!(array, [42; 2]);
```
iqdecay, line 449: I think maybe keep those?

packages/fuels-signers/src/provider.rs
```diff
@@ -557,15 +560,15 @@ mod tests {
...
-        assert_eq!(transaction_cost.min_gas_price, 0);
-        assert_eq!(transaction_cost.min_byte_price, 0);
+        assert_eq!(transaction_cost.min_gas_price, 42);
+        assert_eq!(transaction_cost.min_byte_price, 42);
         assert_eq!(transaction_cost.gas_price, 10_000);
         assert_eq!(transaction_cost.byte_price, 10_000);
         assert_eq!(transaction_cost.gas_used, 270);
         assert_eq!(transaction_cost.byte_size, 120);
-        assert_eq!(transaction_cost.fee, 0.0039);
+        assert_eq!(transaction_cost.total_fee, 0.0039);
```
iqdecay, line 571: `gas_used` and `total_fee`, since they are computed, should be be computed alongside and displayed in the test, so we can check that the computation is the one we expect.

packages/fuels/tests/harness.rs
```diff
@@ -2588,24 +2597,24 @@ async fn contract_call_fee_estimation() -> Result<(), Error> {
...
-    assert_eq!(transaction_cost.gas_used, 348);
-    assert_eq!(transaction_cost.byte_size, 704);
-    assert_eq!(transaction_cost.fee, 0.01052);
+    assert_eq!(estimated_transaction_cost.min_gas_price, 0);
+    assert_eq!(estimated_transaction_cost.min_byte_price, 0);
+    assert_eq!(estimated_transaction_cost.gas_price, 10_000);
+    assert_eq!(estimated_transaction_cost.byte_price, 10_000);
+    assert_eq!(estimated_transaction_cost.gas_used, 348);
+    assert_eq!(estimated_transaction_cost.byte_size, 704);
+    assert_eq!(estimated_transaction_cost.total_fee, 0.01052);
```
iqdecay, line 2617: Same remark here: we are not checking the math here, only the fact that it equals to some value. We should test that the math is the one we expect.

iqdecay (changes requested):
Niiice! We should just be checking that we indeed perform the right computations and make those appear, otherwise LGTM :smile:

examples/contracts/src/lib.rs
hal3e: We are not calling the contract functions.. We are just estimating the cost so there is no "output" to compare to.. We do assert the `gas_used` below

iqdecay (approved):
Nice! :ship:

PR #531, opened 2022-08-16
fix: remove spam from `fuel-core` when running tests

### LINKED ISSUE #495: test: remove the spam from `fuel-core` when testing without `fuel-core-lib` feature
* The fix in #479 does not work if `fuel-core-lib` feature is not enabled (which makes sense!)

iqdecay:
* This PR closes #495 by adding the `silent` option for running a Fuel node, which is enabled by default.

### DIFF
--- a/packages/fuels-test-helpers/src/node.rs
+++ b/packages/fuels-test-helpers/src/node.rs
@@ -266,13 +266,12 @@ pub async fn new_fuel_node(
         }
 
         let mut command = Command::new("fuel-core");
+        command.stdin(Stdio::null());
         if silent {
-            command
-                .stdin(Stdio::null())
-                .stdout(Stdio::null())
-                .stderr(Stdio::null());
+            command.stdout(Stdio::null()).stderr(Stdio::null());
         }
-        let mut running_node =command.args(args)
+        let mut running_node = command
+            .args(args)
             .kill_on_drop(true)
             .spawn()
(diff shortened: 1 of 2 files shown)


digorithm (approved):

packages/fuels-test-helpers/src/node.rs
```diff
@@ -264,10 +265,15 @@ pub async fn new_fuel_node(
             args.push("--manual_blocks_enabled");
         }
 
-        let mut running_node = Command::new("fuel-core").args(args)
+        let mut command = Command::new("fuel-core");
+        if silent {
+            command
+                .stdin(Stdio::null())
```
segfault-magnet, line 271: Stdin should always be closed since we're not aiming to have an interactive experience with the spawned fuel-core process.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -264,10 +265,15 @@ pub async fn new_fuel_node(
...
+                .stdout(Stdio::null())
+                .stderr(Stdio::null());
+        }
+        let mut running_node =command.args(args)
```
segfault-magnet, line 275: It might be github, but this ws is weird

packages/fuels-test-helpers/src/node.rs
digorithm: ^ Good point!

Salka1988 (approved):

segfault-magnet (approved):

PR #532, opened 2022-08-18
feat: script for checking docs validity

Salka1988:
Added a script that checks the validity of the docs.


scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    let output_to_string =
+        String::from_utf8(output_files.stdout).expect("failed to parse command output");
+    let split = output_to_string.split('\n');
+    let vec_of_anchors = split
+        .into_iter()
+        .filter(|s| !s.is_empty())
+        .collect::<Vec<&str>>();
+
+    for i in vec_of_anchors {
+        let parts = i.split(':').collect::<Vec<&str>>();
```
segfault-magnet, line 30: This can be simplified considerably if you use regexes to match and extract anchor data.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+        if *parts.get(1).expect("error in parsing vec_of_anchors") == "ANCHOR" {
+            stack.push(*parts.get(2).expect("error"));
+            // println!("{:?}", stack);
+        } else {
+            if stack.is_empty() {
```
segfault-magnet, line 35: Consider converting each line into a structure containing all the data you need and then zipping adjacent elements together to form begin, end pairs:
```rust
let anchor_tags = process_output.lines().map(transform_here_to_a_struct).collect::<Vec<_>>();

for (begin, end) in anchor_tags.tuples() {
... 
}
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,82 @@
+use std::process::Stdio;
+
+fn main() {
+    let rg_project = std::process::Command::new("rg")
```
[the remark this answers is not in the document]
Salka1988: i deleted all dependencies and now i'm using "grep"

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+        .arg("--case-sensitive")
+        .arg("-g")
+        .arg("!/scripts/check-docs/src/main.rs")
+        .arg("ANCHOR")
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failed rg command");
+
+    let output_files = std::process::Command::new("sed")
```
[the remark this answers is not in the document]
Salka1988: done

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,82 @@
+use std::process::Stdio;
+
+fn main() {
+    let rg_project = std::process::Command::new("rg")
+        .arg("--case-sensitive")
+        .arg("-g")
+        .arg("!/scripts/check-docs/src/main.rs")
+        .arg("ANCHOR")
+        .stdout(Stdio::piped())
```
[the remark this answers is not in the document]
Salka1988: I excluded this.

scripts/check-docs/src/main.rs
Salka1988: I think regex is not a part of std lib.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+
+use std::collections::{HashMap, HashSet};
+use std::env;
+use std::path::Path;
+
+fn main() {
+    let mut stack: Vec<&str> = vec![];
+    let mut valid_anchors: HashMap<&str, HashSet<&str>> = HashMap::new();
+
+    let rg_project = std::process::Command::new("grep")
```
Br1ght0ne, line 11: L19 mentions `"rg command"`, but here `grep` is called. Should be consistent.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+        .arg("-I")
+        .arg("-H")
+        .arg("-R")
+        .arg("--exclude-dir=scripts")
+        .arg("ANCHOR")
+        .arg(".")
```
Br1ght0ne, line 17:
```suggestion
        .args(["-I", "-H", "-R", "--exclude-dir=scripts"])
        .args(["ANCHOR", "."])
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,120 @@
...
+            }
+            false => {
+                let mut new_hash_set = HashSet::new();
+                new_hash_set.insert(*parts.get(2).unwrap());
+                valid_anchors.insert(*parts.get(0).unwrap(), new_hash_set);
+            }
+        }
+    }
+
+    let rg_docs = std::process::Command::new("grep")
```
Br1ght0ne, line 71: Same here for `rg`/`grep` and `arg/args`.

Br1ght0ne (commented):
Not a total review.

scripts/check-docs/src/main.rs
Br1ght0ne: I see https://github.com/FuelLabs/fuels-rs/pull/532#discussion_r951077312, so you should just change the `expect` message to `grep`.

scripts/check-docs/src/main.rs
Salka1988: I commented it before pushing :D

scripts/check-docs/src/main.rs
Salka1988: I'm not sure about this :thinking: . Lets talk.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+use std::collections::{HashMap, HashSet};
+use std::env;
+use std::path::Path;
+
+fn main() {
+    let mut stack: Vec<&str> = vec![];
+    let mut valid_anchors: HashMap<&str, HashSet<&str>> = HashMap::new();
+
+    let rg_project = std::process::Command::new("grep")
+        .args(["-I", "-H", "-R", "--exclude-dir=scripts", "ANCHOR", "."])
```
segfault-magnet, line 12: Two points:
1. (more important) consider all environments in which you expect this script to work. For example, on macs a different grep is shipped (BSD version) than on most linux disros (gnu grep) and they don't have the same flags. The only thing you can be certain they both have is the posix [subset of features](https://pubs.opengroup.org/onlinepubs/009695399/utilities/grep.html).

2. Prefer writing the long versions of flags in scripts to increase readability.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            &_ => {
+                panic!(
+                    "Invalid syntax for ANCHOR or ANCHOR_END:  \"{}\"",
+                    *parts.get(1).expect("error")
+                );
+            }
+        }
+
+        match valid_anchors.contains_key(*parts.get(0).unwrap()) {
```
segfault-magnet, line 57: `match` here only increases verbosity, prefer if/else.

scripts/check-docs/src/main.rs
segfault-magnet: Why does it need to be? A requirement somewhere prohibiting dependencies?

scripts/check-docs/src/main.rs
Salka1988: 1. They seem to be the same. At least the ones we use.
2. I think the readability is the same. :shrug:

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use std::collections::{HashMap, HashSet};
+use std::env;
+use std::path::Path;
+
+fn main() {
+    let mut stack: Vec<&str> = vec![];
+    let mut valid_anchors: HashMap<&str, HashSet<&str>> = HashMap::new();
+
+    let rg_project = std::process::Command::new("grep")
```
hal3e, line 11:
```suggestion
    let grep_project = std::process::Command::new("grep")
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+        .args(["-I", "-H", "-R", "--exclude-dir=scripts", "ANCHOR", "."])
+        .output()
+        .expect("failed rg command");
```
hal3e, line 14:
```suggestion
        .expect("failed grep command");
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    let output_to_string =
+        String::from_utf8(rg_project.stdout).expect("failed to parse command output");
```
hal3e, line 17:
```suggestion
        String::from_utf8(grep_project.stdout).expect("failed to parse command output");
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+        .filter(|s| !s.is_empty())
+        .map(|g| g.replace("./", "").replace(' ', "").replace("//", ""))
+        .collect::<Vec<_>>();
+
+    for i in &vec_of_anchors {
+        let parts = i.split(':').collect::<Vec<_>>();
+
+        match *parts.get(1).expect("error in parsing vec_of_anchors") {
+            "ANCHOR" => {
+                stack.push(*parts.get(2).expect("error"));
```
hal3e, line 32:
```suggestion
                stack.push(*parts.get(2).expect("anchor name/tag is missing"));
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+            }
+            "ANCHOR_END" => {
+                if stack.is_empty() {
+                    panic!(
+                        "ANCHOR of \"{}\" is missing or is wrong",
+                        *parts.get(2).expect("error")
```
hal3e, line 38:
```suggestion
                        *parts.get(2).expect("anchor name/tag is missing")
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+                    );
+                }
+                let first_from_stack = stack.pop().unwrap();
+                if first_from_stack != *parts.get(2).expect("error") {
+                    panic!(
+                        "ANCHOR_END of \"{}\" is missing or is wrong",
```
hal3e, line 44:
```suggestion
                        "ANCHOR_END of \"{}\" is missing corresponding ANCHOR or name/tag is wrong",
```

scripts/check-docs/src/main.rs
Salka1988: I was guided by this - https://github.com/FuelLabs/fuels-rs/pull/532#discussion_r949829467.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    if !stack.is_empty() {
+        panic!(
+            "ANCHOR_END of \"{}\" is missing or is wrong",
+            stack.pop().unwrap()
+        );
+    }
+
+    let rg_docs = std::process::Command::new("grep")
```
hal3e, line 74:
```suggestion
    let grep_docs = std::process::Command::new("grep")
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+        .args(["-I", "-H", "-R", "--exclude-dir=scripts", "{{#include", "."])
+        .output()
+        .expect("failed rg command");
```
hal3e, line 77:
```suggestion
        .expect("failed grep command");
```

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+
+    let output_docs_to_string =
+        String::from_utf8(rg_docs.stdout).expect("failed to parse command output");
```
hal3e, line 80:
```suggestion
        String::from_utf8(grep_docs.stdout).expect("failed to parse command output");
```

scripts/check-docs/src/main.rs
segfault-magnet: "-I -H -R" is as readable as "--binary-files=without-match --with-filename --dereference-recursive" ?

hal3e (commented):
Nice work!. Below are some nits.

In some error messages I wrote `name/tag`. I do not know what is appropriate here so I suggest to use one of them.

scripts/check-docs/src/main.rs
segfault-magnet: Cargo will handle rust dependencies for you. [Regex](https://docs.rs/regex/latest/regex/) is a rust crate.  Therefore cargo will make sure you have it.

It will not scan the code to look for all binaries you've called and proceed to call your package manager to install them. That is why 'rg' was not an ideal choice.

Salka1988: We'll talk.

examples/cookbook/src/lib.rs
```diff
@@ -4,7 +4,7 @@ mod tests {
 
     #[tokio::test]
     async fn liquidity() -> Result<(), Error> {
-        // ANCHOR: liquidity
+        //// ANCHOR: liquidity
```
segfault-magnet, line 7: Why are these four `/` a thing?

packages/fuels-bridge/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+use std::error::Error;
+mod types;
+
+pub async fn get_all_messages() -> Result<(), Error> {
+    unimplemented!();
```
segfault-magnet, line 5: ?

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+    let includes = parse_includes(text_mentioning_include)?;
+
+    let (include_errors, additional_warnings) = validate_includes(includes, valid_anchors);
+
+    report_warnings(&additional_warnings);
+
+    if !anchor_errors.is_empty() || !include_errors.is_empty() {
+        report_errors("anchors", &anchor_errors);
+        report_errors("includes", &include_errors);
+        panic!("Finished with errors");
```
segfault-magnet, line 25: You might as well `bail!` here.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+) -> (Vec<Error>, Vec<Error>) {
+    let (pairs, errors): (Vec<_>, Vec<_>) = includes
+        .into_iter()
+        .map(|include| {
+            let mut maybe_anchor = valid_anchors.iter().find(|anchor| {
+                anchor.file == include.anchor_file && anchor.name == include.anchor_name
+            });
+
+            match maybe_anchor.take() {
+                Some(anchor) => Ok((include, anchor.clone())),
```
segfault-magnet, line 55: It would seem you're not actually using the first part of the pair. Perhaps just collect used anchors?

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+#[derive(Debug)]
+struct Include {
+    anchor_name: String,
+    anchor_file: PathBuf,
+    include_file: PathBuf,
+    line_no: usize,
+}
+
+fn parse_includes(text_w_includes: String) -> anyhow::Result<Vec<Include>, Error> {
+    let apply_regex = |regex: Regex| {
```
segfault-magnet, line 85: You can inline this. Doesn't increase readability being factored out like this.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+        let (begin, end) = match matches_by_name.as_slice() {
+            [single_match] => Ok((start, (*single_match).clone())),
+            [] => Err(anyhow!("Couldn't find a matching end anchor for {start:?}")),
+            multiple_ends => Err(anyhow!("Found too many matching anchor ends for anchor: {start:?}. The matching ends are: {multiple_ends:?}")),
+        }?;
+
+        match check_validity_of_anchor_pair(&begin, &end) {
+            None => Ok((begin, end)),
+            Some(_) => {
+                let err_msg = check_validity_of_anchor_pair(&begin, &end).iter().map(|e|e.to_string()).collect::<Vec<_>>().join("\n");
```
segfault-magnet, line 141: You're calling the validating function twice. Perhaps just use the result ignored in `Some(_)`. That should be the returned error.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+    let begins = apply_regex(Regex::new(r"^(.+):(\d+):\s*//\s*ANCHOR\s*:\s*(\S+)")?)?;
+    let ends = apply_regex(Regex::new(r"^(.+):(\d+):\s*//\s*ANCHOR_END\s*:\s*(\S+)")?)?;
+
+    Ok((begins, ends))
+}
+
+fn search_for_patterns_in_project(pattern: &str) -> anyhow::Result<String> {
+    let grep_project = std::process::Command::new("grep")
+        .args([
+            "-I",
```
segfault-magnet, line 211: Will be more readable if you use the long versions of the flags.

packages/fuels-bridge/src/lib.rs
Salka1988: ah...leaked from another branch  :sweat_smile:

scripts/check-docs/src/main.rs
Salka1988: It doesn't matter, let it stay like this.

examples/cookbook/src/lib.rs
Salka1988: fixed

segfault-magnet (commented):
You've factored out the code nicely. Might lend itself to some good unit testing.

Shouldn't be hard to write a few -- testing out the anchor/include extraction and the match-finding process.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,335 @@
...
+
+                let anchor_file = the_path.canonicalize();
+                if anchor_file.is_err() {
+                    panic!(
+                        "{the_path:?} when canonicalized gives error {:?}",
+                        anchor_file.err().unwrap()
+                    )
+                }
+
+                let anchor_file = anchor_file?;
```
hal3e, line 107:
```suggestion
                let anchor_file = the_path.canonicalize().unwrap_or_else(|err| {
                    panic!("{the_path:?} when canonicalized gives error {:?}", err)
                });
```

scripts/check-docs/src/test_include_data.md
```diff
@@ -0,0 +1,18 @@
...
+
+```rust,ignore
+{{#include ./test_anchor_data.rs:test_with_more_forward_slashes}}
+```
+
+```rust,ignore
+{{#include ./test_anchor_data.rs:no_existing_anchor}}
+```
+
+
```
hal3e, line 18:
```suggestion
```

hal3e (commented):
LGTM! Awesome work. Just some small nits.

docs/src/cookbook/deposit-and-withdraw.md
```diff
@@ -28,9 +28,9 @@ Having launched a provider and created the wallet, we can deploy our contract an
 
 With the preparations out of the way, we can finally deposit to the liquidity pool by calling `deposit()` via the contract instance. Since we have to transfer assets to the contract, we create the appropriate `CallParameters` and chain them to the method call. To receive the minted liquidity pool asset, we have to append a variable output to our contract call.
 
-```rust,ignore
-{{#include ../../../examples/cookbook/src/lib.rs:liquidity_deposit}}
-```
+[//]: # (```rust,ignore)
+[//]: # ({{#include ../../../examples/cookbook/src/lib.rs:liquidity_deposit}})
+[//]: # (```)
```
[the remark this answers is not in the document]
Salka1988: The script does not allow us to have an "include" for which there is no "ANCHOR". I'll uncomment "Anchors" and "Include" and you can solve the rest of the code when you're done.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,329 @@
...
+
+#[derive(Debug, Clone, Hash, Eq, PartialEq)]
+struct Anchor {
+    line_no: usize,
+    name: String,
+    file: PathBuf,
+}
+
+#[allow(dead_code)]
+enum TestEnum {
```
segfault-magnet, line 178: You can put this inside the code guarded by the `cfg` for it not to be dead in production code.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,329 @@
...
+    let ends = apply_regex(Regex::new(
+        r"^(.+):(\d+):\s*(?:/{2,}|/\*)\s*ANCHOR_END\s*:\s*([\w_-]+)\s*(?:\*/)?",
+    )?)?;
+
+    Ok((begins, ends))
+}
+
+fn search_for_patterns_in_project(pattern: &str) -> anyhow::Result<String> {
+    let grep_project = std::process::Command::new("grep")
+        .arg("-I")
```
segfault-magnet, line 217: You can use longer versions of these flags for readability. '-IHRn' is only readable to sys admins.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,329 @@
...
+
+#[cfg(test)]
+mod tests {
+    use crate::{
+        extract_starts_and_ends, filter_valid_anchors, parse_includes, validate_includes, TestEnum,
+    };
+    use anyhow::bail;
+
+    fn test_search_for_patterns(pattern: &str, test_file: &str) -> anyhow::Result<String> {
+        let grep = std::process::Command::new("grep")
```
segfault-magnet, line 242: This is duplicated code. If you wanted to test the `search_for_patterns` you should call it here.

segfault-magnet: Also, this is testing just that the command ran. 

It should rather be testing that you've got the flag combination correct. That the stdout it spat out contains everything you expect it to.

segfault-magnet: Also, this is much more of an integration test -- running a process and depending on external files. Like spinning up a db and expecting it to have prepopulated data.

It is [convention](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) to place such tests inside a separate `tests` directory.

scripts/check-docs/src/main.rs
```diff
@@ -0,0 +1,329 @@
...
+            TestEnum::Anchor(anchor_vec) => anchor_vec.iter().any(|anchor| anchor.name == str),
+            TestEnum::Include(include_vec) => {
+                include_vec.iter().any(|include| include.anchor_name == str)
+            }
+            TestEnum::Errors(err_vec) => err_vec.iter().any(|err| err.to_string().starts_with(str)),
+        }
+    }
+
+    #[test]
+    fn test_anchors() -> anyhow::Result<()> {
```
segfault-magnet, line 271: This is also an integration test, tying everything together. Per [convention](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html), you might want to separate it.

scripts/check-docs/src/main.rs
Salka1988: > This is duplicated code. If you wanted to test the `search_for_patterns` you should call it here

Didn't want to change `search_for_patterns_in_project` so i made `test_search_for_patterns` with additional parameter test_file. I should maybe change its name. In this way, it checks whether `Anchors` and `Includes` are well imported, connected and whether it reacts to errors that may arise during writing anchors and includes in the docs.

Salka1988: > Also, this is much more of an integration test -- running a process and depending on external files. Like spinning up a db and expecting it to have prepopulated data.
> 
> It is [convention](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) to place such tests inside a separate `tests` directory.

Maybe i should do that. But I saw that many of our tests are written in the same file as the code, so I did so.

segfault-magnet (approved):
Great work!

hal3e (approved):

examples/contracts/src/lib.rs
```diff
@@ -248,7 +249,6 @@ mod tests {
             .await?;
 
         // ANCHOR_END: tx_parameters_default
-        // ANCHOR: tx_parameters
```
digorithm, line 251: I'm out of the loop -- why are we removing this anchor here?

Salka1988: It's duplicated. If you look at the "Master" branch, you will see that ANCHOR: tx_parameters and ANCHOR_END: tx_parameters are repeated twice on lines 231 - 241 and 251 - 261. I didn't know which one to keep, so I deleted one so that someone could notice and decide which one to keep. :thinking:

digorithm: Nice, makes sense!

digorithm:
The branch is outdated -- sync with master, please!

digorithm (approved):
Good work -- very interesting approach you took here! 😄

PR #540, opened 2022-08-25
Distinguish between a `Wallet`'s `Locked` and `Unlocked` states

mitchmindtree:
This PR was motivated by work @kayagokalp and I are doing on the
following:
https://github.com/FuelLabs/sway/issues/2413

We noticed that in order to perform any operations using a `Wallet`, the
private key must be known, even if those operations were simply
retrieving asset balances or a list of transactions.

This PR aims to draw a clearer distinction between Wallet methods that
require access to a private key and those that do not. The idea is to
encourage users of the API to be conscious of when their wallet is in an
unlocked state (i.e. their private key resides in memory) and when it
does not.

To achieve this, we turn `Wallet` into a session type, parameterised
over its state which can be either `Locked` or `Unlocked`.

The default state is `Locked` in the case that the parameter is not
specified for `Wallet` or `LocalWallet`, meaning that if users require
holding the wallet in an unlocked state they must specify the type
explicitly, e.g. `Wallet<Unlocked>`. I think this explicitness can
assist us in maintaining awareness of the state of the wallet that we
require in our downstream APIs.

Users should generally prefer using the `Wallet` in a locked state (the
default) and should ideally only `unlock` the wallet for the duration
that they need to perform operations requiring the private key. Users
can now construct a `Wallet` in a `Locked` state, e.g.
`Wallet::from_address(public_addr, provider)` and access the various
useful read-only operations without knowing the private key or having it
stored in memory.

As a follow-up to this work, we should take a closer look at the usage
of `LocalWallet` in downstream crates (including `fuels-contract`) to
see if we can refactor these to use the wallet in a `Locked` state more
often, only moving into the `Unlocked` state when absolutely necessary.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -12,6 +12,7 @@
   - [Managing wallets](./getting-started/managing-wallets.md)
     - [Creating a wallet from a private key](./wallets/private-keys.md)
     - [Creating a wallet from mnemonic phrases](./wallets/mnemonic-wallet.md)
+    - [Wallet Access](./wallets/access.md)
     - [Encrypting and storing wallets](./wallets/encrypting-and-storing.md)
     - [Checking balances and coins](./wallets/checking-balances-and-coins.md)
     - [Setting up test wallets](./wallets/test-wallets.md)
--- a/docs/src/getting-started/managing-wallets.md
+++ b/docs/src/getting-started/managing-wallets.md
@@ -7,8 +7,6 @@ You can use wallets for many important things, for instance:
 3. Signing messages and transactions;
 4. Paying for network fees when sending transactions or deploying smart contracts.
 
-The SDK gives you many different ways to create wallets. Let's explore these different ways.
-
-The following sub-chapters walk you through the different ways of creating wallets.
+The SDK gives you many different ways to create and access wallets. Let's explore these different approaches in the following sub-chapters.
 
 > **Note:** Keep in mind that you should never share your private/secret key. And in the case of wallets that were derived from a mnemonic phrase, never share your mnemonic phrase. If you're planning on storing the wallet on disk, do not store the plain private/secret key and do not store the plain mnemonic phrase. Instead, use `Wallet::encrypt` to encrypt its content first before saving it to disk.
--- a/docs/src/wallets/access.md
+++ b/docs/src/wallets/access.md
@@ -49,5 +49,5 @@ is stored in memory.
 When designing APIs that accept a `Wallet` as an input, we should think
 carefully about the kind of access that we require. API developers should aim to
 minimise their usage of `Unlocked` wallets in order to ensure private keys are
-stored in memory no longer than necessary in order to reduce the surface area
-for attacks and vulnerabilities in downstream libraries and applications.
+stored in memory no longer than necessary to reduce the surface area for attacks
+and vulnerabilities in downstream libraries and applications.
--- a/docs/src/wallets/access.md
+++ b/docs/src/wallets/access.md
@@ -1,0 +1,53 @@
+# Wallet Access
+
+The kinds of operations we can perform with a `Wallet` instance depend on
+whether or not we have access to the wallet's private key.
+
+In order to differentiate between `Wallet` instances that know their private key
+and those that do not, we use the `Unlocked` and `Locked` states respectively.
+
+## Wallet States
+
+An `Unlocked` wallet (i.e. a wallet of type `Wallet<Unlocked>`) is a wallet
+whose private key is known and stored internally in memory. A wallet must be in
+the `Unlocked` state in order to perform operations that involve [signing
+transactions](./signing.md).
+
+A `Locked` wallet (i.e. a wallet of type `Wallet<Locked>`) is a wallet whose
+private key is *not* known or stored in memory. Instead, a locked wallet only
+knows its public address. A locked wallet cannot be used to sign transactions,
+however it may still perform a whole suite of useful operations including
+listing transactions, assets, querying balances, and so on.
+
+If a wallet's type is unspecified its state is `Locked` by default. That is, the
+type `Wallet` (without providing any type arguments) is equivalent to
+`Wallet<Locked>`. This default is aimed at encouraging users to prefer the
+safer `Locked` state by default and to ensure that users remain conscious and
+aware when holding their wallet's private key in memory.
+
+## Transitioning States
+
+A `Locked` wallet can be unlocked by providing the private key:
+
+```rust,ignore
+let unlocked_wallet = locked_wallet.unlock(private_key);
+```
+
+An `Unlocked` wallet can be locked using the `lock` method:
+
+```rust,ignore
+let locked_wallet = unlocked_wallet.lock();
+```
+
+Most `Wallet` constructors that create or generate a new wallet are returned in
+the `Unlocked` state. Consider `lock`ing the wallet after the new private key
+has been handled in order to reduce the scope in which the wallet's private key
+is stored in memory.
+
+## Design Guidelines
+
+When designing APIs that accept a `Wallet` as an input, we should think
+carefully about the kind of access that we require. API developers should aim to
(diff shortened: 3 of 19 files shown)


digorithm:
I _really_ like this idea, and the API seems very friendly and nice, from a quick glance. I'm still reviewing the details though. One thing is that we should definitely document the new terms introduced here (`Locked` and `Unlocked`) and what they mean, etc., in the SDK book, somewhere in [here](https://github.com/FuelLabs/fuels-rs/tree/master/docs/src/wallets).

mitchmindtree:
OK, this should be good to go!

@digorithm thanks for the feedback, and good point about updating the docs! I've added a new "Wallet Access" chapter that should cover the new behaviour and terminology.

kayagokalp (approved):
Excited for this, lgtm 🚀

Br1ght0ne (dismissed):

segfault-magnet (commented):
Love the type safety. 

Beautifully done. I see only one potential downside to way this is implemented:

If I have a function:
```rust
fn something(a_wallet: &LocalWallet<Locked>) {
}
```
But only have an `unlocked` wallet at my disposal I'd have to call `lock` on it -- destroying it in the process:
```rust
something(&unlocked_wallet.lock()); 
// unlocked_wallet cannot be used anymore
```
so that would mean users would have to either keep locked&unlocked versions or clone on demand:
```rust
let locked_wallet = unlocked_wallet.clone().lock();
something(&locked_wallet); 
// can still use unlocked_wallet
```

This can be solved by implementing AsRef:
```rust
impl AsRef<Wallet<Locked>> for Wallet<Unlocked> {
    fn as_ref(&self) -> &Wallet<Locked> {
    }
}
```

Which would enable users to do:
```rust
something(unlocked_wallet.as_ref()); 
```

or would enable us to do:
```rust
fn something<T: impl AsRef<LocalWallet<Locked>>>(locked_wallet: &T){
}
```
which could then be called as
```rust
something(&unlocked_wallet);
```
But implementing AsRef isn't straightforward since a `Wallet<Unlocked>` doesn't really contain a `Wallet<Locked>` which you can just return by reference. So that would mean that you would have to have `Wallet<Unlocked>` delegate everything that doesn't need the private key to an enclosed Wallet<Locked>...

But this isn't such a big problem since you can still say:
```rust
fn something<T>(a_wallet: &T){
}
```
and would get only the functionality that is common between locked and unlocked wallets -- i.e. effectively only what a locked wallet can do. It's a bit more roundabout way of saying `Wallet<Locked>`. Also, the generic T will corrupt everything we might pass the wallet to, requiring that the function be generic over T as well -- since you don't actually get to call `lock` on `Wallet<T>`. This also might cause problems down the line if we should get a third specialization but we'll worry about that when we come to it.

That rubs me a bit wrong but I can live with it seeing as this is a very elegant solution.

digorithm:
> Love the type safety.
> 
> Beautifully done. I see only one potential downside to way this is implemented:
> 
> If I have a function:
> 
> ```rust
> fn something(a_wallet: &LocalWallet<Locked>) {
> }
> ```
> 
> But only have an `unlocked` wallet at my disposal I'd have to call `lock` on it -- destroying it in the process:
> 
> ```rust
> something(&unlocked_wallet.lock()); 
> // unlocked_wallet cannot be used anymore
> ```
> 
> so that would mean users would have to either keep locked&unlocked versions or clone on demand:
> 
> ```rust
> let locked_wallet = unlocked_wallet.clone().lock();
> something(&locked_wallet); 
> // can still use unlocked_wallet
> ```
> 
> This can be solved by implementing AsRef:
> 
> ```rust
> impl AsRef<Wallet<Locked>> for Wallet<Unlocked> {
>     fn as_ref(&self) -> &Wallet<Locked> {
>     }
> }
> ```
> 
> Which would enable users to do:
> 
> ```rust
> something(unlocked_wallet.as_ref()); 
> ```
> 
> or would enable us to do:
> 
> ```rust
> fn something<T: impl AsRef<LocalWallet<Locked>>>(locked_wallet: &T){
> }
> ```
> 
> which could then be called as
> 
> ```rust
> something(&unlocked_wallet);
> ```
> 
> But implementing AsRef isn't straightforward since a `Wallet<Unlocked>` doesn't really contain a `Wallet<Locked>` which you can just return by reference. So that would mean that you would have to have `Wallet<Unlocked>` delegate everything that doesn't need the private key to an enclosed Wallet...
> 
> But this isn't such a big problem since you can still say:
> 
> ```rust
> fn something<T>(a_wallet: &T){
> }
> ```
> 
> and would get only the functionality that is common between locked and unlocked wallets -- i.e. effectively only what a locked wallet can do. It's a bit more roundabout way of saying `Wallet<Locked>`. Also, the generic T will corrupt everything we might pass the wallet to, requiring that the function be generic over T as well -- since you don't actually get to call `lock` on `Wallet<T>`. This also might cause problems down the line if we should get a third specialization but we'll worry about that when we come to it.
> 
> That rubs me a bit wrong but I can live with it seeing as this is a very elegant solution.

That's a good point, but I personally don't see why we would have more functions specifically defined _only_ for `Locked` wallets other than the current proposed ones (created `Locked` and `lock()`). Shouldn't any behavior that should be acceptable by the `Locked` wallets should also be acceptable by `Unlocked` ones, therefore defined inside the generic `T` impl? 

I spent a whopping whole 3 minutes (😂 ) trying to think of a use case that goes against this but couldn't find it. I say we move on as-is, and if such a use case comes up, we could introduce more special-casing to handle it, as you mentioned.

mitchmindtree:
Keep in mind that functions like `fn something` can take a wallet in any state too in the case that we don't mind what state it's in e.g.:

```rust
fn something<T>(wallet: &Wallet<T>) {
}
```

The only method that `Wallet<Locked>` implements that `Wallet<T>` does not is the `lock` method, so it's almost always worth taking a `&Wallet<T>` in the case that you don't require signing anything.

---

Thinking on this a little further, perhaps a nicer abstraction might be:


```rust
pub struct Wallet {
    address: Bech32Address,
    provider: Option<Provider>,
}

pub struct WalletUnlocked {
    wallet: Wallet,
    private_key: SecretKey,
}
```

This way, we could have `WalletUnlocked` `deref` to its inner `Wallet` type, meaning `Wallet` methods can always be called on `WalletUnlocked` (though not vice versa).

It would also allow us to call functions like `fn something(wallet: &Wallet) { ... }` with both unlocked  and locked wallets, e.g. `something(&unlocked_wallet)`, `something(&wallet)` would both be valid. This way we also wouldn't have to worry about the generic state input type when we want our functions to generalise over wallet states.

I might have a go at this now as I imagine it'll be quite a bit less fancy, more ergonomic and simpler to create APIs for!

mitchmindtree:
I've pushed two more commits with a simpler solution as described in the second half of [this comment](https://github.com/FuelLabs/fuels-rs/pull/540#issuecomment-1229066451).

The new solution still uses session types to clearly distinguish wallet access, but we do so in a manner that does not require a generic state type parameter over a single Wallet type. This should make it easier to create ergonomic, safe-by-default APIs, e.g.

```rust
fn list_assets(wallet: &Wallet) { ... }
```
The `list_assets` function can be called with instances of either `Wallet` or `WalletUnlocked`, as `WalletUnlocked` derefs to its inner `Wallet`.

This should help to avoid the potential API design footgun highlighted by @segfault-magnet [above](https://github.com/FuelLabs/fuels-rs/pull/540#pullrequestreview-1086914187), where a user may have "accidentally" designed their APIs only accepting a locked wallet (due to `Locked` being the default state type), which would halve caused issues for users that need to retain access to their wallet in an unlocked state.

mitchmindtree:
OK, just rebased onto master and addressed the associated clippy errors!

kayagokalp (approved):

kayagokalp (approved):

Br1ght0ne (approved):
I was planning to suggest a `WalletState` trait for `T` bounds, but a wrapped type looks a lot better. `Deref` is the Rust-iest way for this I can think of. Lovely, :shipit:

PR #545, opened 2022-08-27
Renamed `packages` to `crates`

xgreenx:
Hi=) I know that this repository is not my area of responsibility. I will soon create a small change to be compatible with the `fuel-core` release `0.10.1`. I understand that you try to keep the structure from the `ether.js`. But maybe it would be better to use the naming `crates` instead of `packages`(It is more suitable for Rust)? Because for example, for me, it was not clear that crates were in the `packages` folder.

It's only an idea, and your decision is final=)


segfault-magnet:
Thank you for showing an interest in `fuels-rs`, it is most welcome :)

As for the renaming:
If you open any `Cargo.toml` of each of the subfolders inside the `packages` dir you'll see a `[package]` TOML header under which the description of the current **package** is present.

A package can contain one or more crates.

So the `packages` folder does in fact contain `package`s which in turn contain crates.

From [the rust book](https://doc.rust-lang.org/book/ch07-01-packages-and-crates.html):

> A package is a bundle of one or more crates that provides a set of functionality. A package contains a Cargo.toml file that describes how to build those crates. Cargo is actually a package that contains the binary crate for the command-line tool you’ve been using to build your code. The Cargo package also contains a library crate that the binary crate depends on. Other projects can depend on the Cargo library crate to use the same logic the Cargo command-line tool uses.

For this reason, I'd rather we leave the current naming.

Regardless of the decision taken by the codeowners, it might be more productive to sound out the rust SDK contributor team before doing a PR. Just to be aligned and not potentially have wasted work. You can reach us either by opening an issue as a suggestion or over on Slack.

xgreenx:
> Regardless of the decision taken by the codeowners, it might be more productive to sound out the rust SDK contributor team before doing a PR. Just to be aligned and not potentially have wasted work. You can reach us either by opening an issue as a suggestion or over on Slack.

I will do it next time!=) BTW, renaming is not hard to do with Cntr+R=)

> If you open any Cargo.toml of each of the subfolders inside the packages dir you'll see a [package] TOML header under which the description of the current package is present.

Yea, it is because it is only one way to publish your code to [crates.io](crates.io). Maybe it is only my understanding of `crate` and `package`. But if someone says:
- "crate" I treat as "Some library that I can import and use"
- "package" I treat as "Some binary that I can install and run"

There is no difference right now, and people use those two naming in the same context. But I often see "crate" about libraries and "package" about binaries=)

But it is my interpretation of those words=) If you think that "packages" is more descriptive, you can close the PR. I only shared my thoughts<3

iqdecay:
> If you open any Cargo.toml of each of the subfolders inside the packages dir you'll see a [package] TOML header under which the description of the current package is present.

and the Rust book extract have convinced me that `packages` is the right naming. 
I still really appreciate suggesting the change :smile:  
Closing.

PR #555, opened 2022-08-31
feat!: Add messages support

Salka1988:


examples/wallets/src/lib.rs
```diff
@@ -267,10 +267,10 @@ mod tests {
         let assets = vec![asset_base, asset_1, asset_2];
 
         let coins = setup_custom_assets_coins(wallet.address(), &assets);
-        let (provider, _socket_addr) = setup_test_provider(coins, None).await;
+        let (provider, _socket_addr) = setup_test_provider(coins, None, None).await;
         wallet.set_provider(provider);
         // ANCHOR_END: custom_assets_wallet
-        // ANCHOR: custom_assets_wallet_short
+        // ANCHOR: custom_assets_wallet_short,             None
```
pixelcircuits, line 273: What is this "None"?

packages/fuels-contract/src/contract.rs
```diff
@@ -337,6 +337,9 @@ impl Contract {
             witnesses,
         );
 
+        // dbg!("Contract deploy");
+        // dbg!(tx.clone());
```
pixelcircuits, line 341: Are these lines needed anymore?

packages/fuels-signers/src/provider.rs
```diff
@@ -67,7 +67,7 @@ impl Provider {
     /// use fuels::prelude::*;
     /// async fn foo() -> Result<(), Box<dyn std::error::Error>> {
     ///   // Setup local test node
-    ///   let (provider, _) = setup_test_provider(vec![], None).await;
+    ///   let (provider, _) = setup_test_provider(vec![], None,             None).await;
```
pixelcircuits, line 70: What's with the extra white space?

packages/fuels-contract/src/contract.rs
Salka1988: nah, i forgot them... Ill delete that

packages/fuels-signers/src/wallet.rs
```diff
@@ -509,6 +566,8 @@ impl WalletUnlocked {
             }
         };
 
+        dbg!(&tx);
```
pixelcircuits, line 569: Just checking if you think we need to keep this

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -146,13 +158,32 @@ pub async fn setup_test_client(
...
+            data: message.data,
+            da_height: message.da_height,
+        })
         .collect();
 
     // Setup node config with genesis coins and utxo_validation enabled
 
+    // message can has coin ... kaze nam dali je coin ili neka data koja je prikacena na njega.
+    // message change outputs -> Burned change...
+    // Message can have only base asset...
```
pixelcircuits, line 181: Are these comments necessary anymore?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -127,13 +128,24 @@ pub fn setup_single_asset_coins(
     coins
 }
 
+pub fn setup_single_message(owner: &Bech32Address, amount_per_message: u64) -> Vec<Message> {
```
pixelcircuits, line 131: It would be nice if this function allowed more parts to be specified for more detailed tests down the line. I made a similar function for testing that looked like this:
``` rust
pub fn setup_single_message(
    sender: Address,
    recipient: Address,
    amount: Word,
    nonce: Word,
    data: Vec<u8>,
) -> Message {
    Message {
        sender: sender,
        recipient,
        owner: recipient,
        nonce,
        amount,
        data: data,
        da_height: DaBlockHeight::default(),
        fuel_block_spend: None,
    }
}
```

Also, you'll notice it uses `recipient` for both the message `recipient` and `owner` to prepare for when `owner` will eventually be dropped in another issue. I do like how you are returning an array though and I think you should keep that.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -127,13 +128,24 @@ pub fn setup_single_asset_coins(
...
+}
+
 // Setup a test client with the given coins. We return the SocketAddr so the launched node
 // client can be connected to more easily (even though it is often ignored).
 #[cfg(feature = "fuel-core-lib")]
 pub async fn setup_test_client(
     coins: Vec<(UtxoId, Coin)>,
     node_config: Option<Config>,
     consensus_parameters_config: Option<ConsensusParameters>,
+    messages: Option<Vec<Message>>,
```
pixelcircuits, line 148: This might be picky, but can we move this param to be right after `coins`? I feel like the function makes more sense if they're next to each other.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -189,8 +227,32 @@ impl<'de> DeserializeAs<'de, BlockHeight> for HexNumber {
 
 pub fn get_node_config_json(
     coins: Vec<(UtxoId, Coin)>,
+    input_messages: Option<Vec<Message>>,
```
pixelcircuits, line 230: Why `input_messages` instead of just `messages`? similar to how it's just `coins`?

packages/fuels-test-helpers/src/node.rs
```diff
@@ -247,12 +318,13 @@ pub async fn new_fuel_node(
     coins: Vec<(UtxoId, Coin)>,
     consensus_parameters_config: Option<ConsensusParameters>,
     config: Config,
+    input_messages: Option<Vec<Message>>,
```
pixelcircuits, line 321: Again, why `input_messages` instead of just `messages`? And why not put it right after `coins` like you did with `get_node_config_json`

packages/fuels/tests/harness.rs
```diff
@@ -3596,3 +3597,36 @@ async fn testnet_hello_world() -> Result<(), Error> {
...
+    let contract_instance_connected =
+        MyContractBuilder::new(contract_id.to_string(), wallet.clone()).build();
+
+    let response = contract_instance_connected
+        .initialize_counter(42) // Build the ABI call
+        .call()
+        .await?;
+    assert_eq!(42, response.value);
+
+    Ok(())
```
pixelcircuits, line 3631: Could we add a check by calling `get_messages` or something similar to verify that the message we initialized the test client with exists?

packages/fuels-test-helpers/src/signers.rs
```diff
@@ -98,15 +99,16 @@ pub async fn launch_custom_provider_and_get_wallets(
...
 /// async fn test_provider() -> Result<(), Box<dyn std::error::Error>> {
-///   let (_provider, _address) = setup_test_provider(vec![], None).await;
+///   let (_provider, _address) = setup_test_provider(vec![], None, None).await;
 ///   Ok(())
 /// }
 /// ```
 pub async fn setup_test_provider(
     coins: Vec<(UtxoId, Coin)>,
     node_config: Option<Config>,
+    messages: Option<Vec<Message>>,
```
pixelcircuits, line 109: Should we move `messages` to be right after `coins`? And maybe either `coins` should be optional too, or we just have people specify `vec![]` if they don't want to provide either coins or messages?

packages/fuels-signers/src/wallet.rs
```diff
@@ -53,7 +56,7 @@ pub struct Wallet {
 ///
 /// async fn foo() -> Result<(), Error> {
 ///   // Setup local test node
-///   let (provider, _) = setup_test_provider(vec![], None).await;
+///   let (provider, _) = setup_test_provider(vec![], None,             None).await;
```
pixelcircuits, line 59: Fix whitespace

packages/fuels-signers/src/wallet.rs
```diff
@@ -536,7 +593,7 @@ impl WalletUnlocked {
     ///   coins_1.extend(coins_2);
     ///
     ///   // Setup a provider and node with both set of coins
-    ///   let (provider, _) = setup_test_provider(coins_1, None).await;
+    ///   let (provider, _) = setup_test_provider(coins_1, None,            None).await;
```
pixelcircuits, line 596: Fix whitespace

packages/fuels-core/src/constants.rs
```diff
@@ -4,6 +4,7 @@ use fuel_types::AssetId;
 pub const DEFAULT_GAS_LIMIT: u64 = 1_000_000;
 pub const DEFAULT_GAS_PRICE: u64 = 0;
 pub const DEFAULT_MATURITY: u64 = 0;
+pub const DEFAULT_SPEND_MESSAGE: bool = false;
```
pixelcircuits, line 7: Is this used anywhere? It looks like everywhere that builds a `Message` struct is actually defaulting `fuel_block_spend` to `None`

packages/fuels-signers/src/provider.rs
```diff
@@ -410,4 +410,30 @@ impl Provider {
...
+
+    pub async fn get_messages(&self, from: &Bech32Address) -> Result<Vec<Message>, ProviderError> {
+        let mut messages: Vec<Message> = vec![];
+
+        let mut cursor = None;
+
+        loop {
+            let pagination = PaginationRequest {
+                cursor: cursor.clone(),
+                results: 9999,
```
Voxelot, line 422: This isn't a very friendly query, typically pages of results shouldn't exceed 100 imo. I'd recommend making this a constant.

packages/fuels/tests/harness.rs
Salka1988: I added a function that partially compares Vec<Message>. It can also be improved if there is a need

pixelcircuits: Which function?

Salka1988: `compare_messages(messages_from_provider, messages) ` if you want to check if  `get_messages` is returning  same messages as the messages that we used to init. client. But maybe we don't need this?

packages/fuels-signers/src/wallet.rs
```diff
@@ -222,6 +225,46 @@ impl Wallet {
...
     }
 
+    pub async fn get_messages(&self) -> Result<Vec<schema::message::Message>, Error> {
+        Ok(self.get_provider()?.get_messages(&self.address).await?)
+    }
+
+    pub async fn get_inputs_for_messages(&self, witness_index: u8) -> Result<Vec<Input>, Error> {
+        let as_u8_slice = |v: &[i32]| -> &[u8] {
+            let element_size = mem::size_of::<i32>();
+            unsafe { from_raw_parts(v.as_ptr() as *const u8, v.len() * element_size) }
```
digorithm, line 235: Can we add a quick comment on the safety guarantees around this `unsafe` block? You should explain why it's safe or when it wouldn't be safe. 

Here's an example of this type of comment from the `fuel-vm`: https://github.com/FuelLabs/fuel-vm/blob/e823401c1f087d47ddab4ec863d34ea6c6c1b3af/src/interpreter/blockchain.rs#L249-L250

packages/fuels-signers/src/wallet.rs
```diff
@@ -467,13 +510,26 @@ impl WalletUnlocked {
             .inputs()
             .iter()
             .any(|input| matches!(input, Input::CoinSigned { .. }));
+
         if !is_using_coins && new_base_amount == 0 {
             new_base_amount = MIN_AMOUNT;
         }
 
+        // This is a temporary solution till we get update on coins_to_spend function
```
digorithm, line 518: Can we put a note in the `add_fee_coins` docstring about using `InputMessages` to cover fees in case there aren't enough regular assets?

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -202,6 +241,23 @@ pub async fn setup_test_client(
     (client, bound_address)
 }
 
+// Feel free to expand this or suggest better solution
```
digorithm, line 244:
```suggestion
```

Should be ok to remove this comment if we're ok with it?

packages/fuels/tests/harness.rs
digorithm: We should still check this variant here inside _this_ specific test, no?

digorithm (changes requested):

packages/fuels/tests/harness.rs
pixelcircuits: Oh, sorry I didn't see line 3636. I guess it could be useful to keep it in its own function if other tests want to also check it for some reason? Or would that be out of scope for other tests to check?

packages/fuels-signers/src/wallet.rs
```diff
@@ -222,6 +225,46 @@ impl Wallet {
             .map_err(Into::into)
     }
 
+    pub async fn get_messages(&self) -> Result<Vec<schema::message::Message>, Error> {
+        Ok(self.get_provider()?.get_messages(&self.address).await?)
+    }
+
+    pub async fn get_inputs_for_messages(&self, witness_index: u8) -> Result<Vec<Input>, Error> {
+        let as_u8_slice = |v: &[i32]| -> &[u8] {
```
segfault-magnet, line 233: Albeit it does look safe I'd rather we use safe code that can achieve the same effect, unless the performance impact is measurable:
```rust
        let to_u8_bytes = |v: &[i32]| { v.iter().flat_map(|e| e.to_be_bytes()).collect::<Vec<_>>() };
```
You'll have a few allocations but you'll sleep _safe_

packages/fuels-signers/src/wallet.rs
```diff
@@ -222,6 +225,46 @@ impl Wallet {
...
+                message.amount.0,
+                0,
+                message.owner.into(),
+                witness_index,
+                vec![],
+            );
+            inputs.push(input_messages);
+        }
+
+        Ok(inputs)
```
segfault-magnet, line 265: functional nit
```suggestion
        let inputs = messages
            .into_iter()
            .map(|message| {
                let message_id = Input::compute_message_id(
                    &message.sender.clone().into(),
                    &message.recipient.clone().into(),
                    message.nonce.into(),
                    &message.owner.clone().into(),
                    message.amount.0,
                    &to_u8_bytes(&message.data),
                );

                Input::message_signed(
                    message_id,
                    message.sender.into(),
                    message.recipient.into(),
                    message.amount.0,
                    0,
                    message.owner.into(),
                    witness_index,
                    vec![],
                )
            })
            .collect();

        Ok(inputs)

```

packages/fuels-signers/src/wallet.rs
```diff
@@ -467,13 +510,26 @@ impl WalletUnlocked {
...
         if !is_using_coins && new_base_amount == 0 {
             new_base_amount = MIN_AMOUNT;
         }
 
+        // This is a temporary solution till we get update on coins_to_spend function
         let new_base_inputs = self
             .get_asset_inputs_for_amount(BASE_ASSET_ID, new_base_amount, witness_index)
-            .await?;
+            .await
+            .unwrap_or(self.get_inputs_for_messages(witness_index).await?);
```
segfault-magnet, line 522: the argument to unwrap_or is eagerly evaluated -- i.e. you're always making this network call regardless of whether you had asset inputs for the amount.

Move the call behind a condition of some sort.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -202,6 +241,23 @@ pub async fn setup_test_client(
...
+pub fn compare_messages(
+    messages_from_provider: Vec<OtherMessage>,
+    used_messages: Vec<Message>,
+) -> bool {
+    used_messages
+        .iter()
+        .zip(&messages_from_provider)
+        .all(|(a, b)| {
+            a.sender == b.sender.0 .0
```
segfault-magnet, line 253: the whitespace seems weird.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -189,8 +227,32 @@ impl<'de> DeserializeAs<'de, BlockHeight> for HexNumber {
...
+        "type": "LocalTest"
+      },
+      "initial_state": {
+        "coins": coins,
+        "messages": messages
+      },
+      "transaction_parameters": consensus_parameters
+    });
+
+    config
```
segfault-magnet, line 252:
```suggestion
    json!({
      "chain_name": "local_testnet",
      "block_production": "Instant",
      "parent_network": {
        "type": "LocalTest"
      },
      "initial_state": {
        "coins": coins,
        "messages": messages
      },
      "transaction_parameters": consensus_parameters
    })
```

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -202,6 +241,23 @@ pub async fn setup_test_client(
...
 }
 
+// Feel free to expand this or suggest better solution
+pub fn compare_messages(
+    messages_from_provider: Vec<OtherMessage>,
+    used_messages: Vec<Message>,
+) -> bool {
+    used_messages
+        .iter()
+        .zip(&messages_from_provider)
```
segfault-magnet, line 251:
```suggestion
    messages_from_provider: &[OtherMessage],
    used_messages: &[Message],
) -> bool {
    zip(used_messages, messages_from_provider)
```

packages/fuels-signers/src/wallet.rs
segfault-magnet: Although this probably needs some thinking through...

If you're on a little endian platform (which you probably are) and have some numbers also in LE and want to have the underlying bytes then you'd call `to_le_bytes` 

Thankfully there is `to_ne_bytes` for native endianess.

packages/fuels-test-helpers/src/lib.rs
Salka1988: Yes it is. Its CLION formatted :)

packages/fuels-signers/src/wallet.rs
Salka1988: Nah! It is intentionally left like this because it will change when a new version of the coins_to_spend function comes out.
Some other things will also be changed.

packages/fuels-signers/src/wallet.rs
Salka1988: I think Ahmed's solution is better, so I chose it. https://github.com/FuelLabs/fuels-rs/pull/555#discussion_r967521873 There is no need to use this unsafe method.

packages/fuels/tests/harness.rs
Salka1988: I put it in the scope.

pixelcircuits (approved):
LGTM!

packages/fuels-signers/src/wallet.rs
```diff
@@ -222,6 +223,41 @@ impl Wallet {
...
     }
 
+    pub async fn get_messages(&self) -> Result<Vec<schema::message::Message>, Error> {
+        Ok(self.get_provider()?.get_messages(&self.address).await?)
+    }
+
+    pub async fn get_inputs_for_messages(&self, witness_index: u8) -> Result<Vec<Input>, Error> {
+        let to_u8_bytes = |v: &[i32]| v.iter().flat_map(|e| e.to_ne_bytes()).collect::<Vec<_>>();
+
+        let messages = self.get_provider()?.get_messages(&self.address).await?;
```
Br1ght0ne, line 233:
```suggestion
        let messages = self.get_messages().await?;
```

packages/fuels/tests/harness.rs
```diff
@@ -3596,3 +3602,62 @@ async fn testnet_hello_world() -> Result<(), Error> {
...
+#[tokio::test]
+async fn test_input_message() -> Result<(), Error> {
+    abigen!(
+        MyContract,
+        "packages/fuels/tests/test_projects/contract_test/out/debug/contract_test-abi.json"
+    );
+
+    let compare_messages =
+        |messages_from_provider: Vec<OtherMessage>, used_messages: Vec<Message>| -> bool {
+            zip(&used_messages, &messages_from_provider).all(|(a, b)| {
```
Br1ght0ne, line 3615: nit: I would write it as `iter::zip` here (and `use std::iter`), was confused for a brief moment.

packages/fuels/tests/harness.rs
```diff
@@ -3596,3 +3602,62 @@ async fn testnet_hello_world() -> Result<(), Error> {
...
+                a.sender == b.sender.0 .0
+                    && a.recipient == b.recipient.0 .0
+                    && a.owner == b.owner.0 .0
+                    && a.nonce == b.nonce.0
+                    && a.amount == b.amount.0
```
Br1ght0ne, line 3620: Maybe we could extract this comparison logic somewhere outside of a single test? I don't see it being used elsewhere, but it's still not very intuitive (for me).

Br1ght0ne (commented):
Not sure about the greatness of my review, still, lots of good things about this.

packages/fuels/tests/harness.rs
Salka1988: It was outside, but since we don't use it anywhere else, I put it inside.

packages/fuels/tests/harness.rs
Salka1988: ok

hal3e (dismissed):
Well done!

Br1ght0ne (approved):

packages/fuels-signers/src/wallet.rs
segfault-magnet: I'm not sure when or whether this new version will come out. What I do know is that you're doing the second call regardless of whether the first one succeeded. This can and should be rewritten to do the second one only if the first one failed.

If it didn't involve network calls then it would be much of a problem.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -146,13 +169,26 @@ pub async fn setup_test_client(
             amount: coin.amount,
             asset_id: coin.asset_id,
         })
+        .collect::<Vec<_>>();
+
+    let message_config = messages
+        .into_iter()
+        .map(|message| MessageConfig {
```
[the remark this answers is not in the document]
Salka1988: Then we should discuss it with the fuel-core team. :thinking:

digorithm (approved):
LGTM!

PR #556, opened 2022-09-01
Use estimation to cover transaction fees

### LINKED ISSUE #536: Use gas estimation when requiring spendable coins of the base asset.
When requesting spendable coins from the node to construct the contract-calling script, don't use the huge default but try to incorporate the work done [here](https://github.com/FuelLabs/fuels-rs/issues/514) to estimate the transaction cost and request only the minimal amount needed to cover the gas price.

Potentially include a tolerance to account for gas price fluctuations.

MujkicA:
Closes #536 

As described in the issue, we use default amounts for coins that cover the transaction fees when constructing transactions for deploying, calling or transferring.
This PR adds the method `add_fee_coins()` to `wallet`. The method takes a transaction, estimates its cost and modifies the inputs accordingly.

We need a fully built transaction to estimate, but transaction inputs can't be fully determined before estimation.
Since this is a chicken and egg problem, we estimate with whatever inputs are present prior to the estimation.
There is a non-zero chance that this estimation fails and the transaction terminates because there wasn't enough coins to cover the fee. However, the circumstances under which this could occur are negligible.

This PR also fixes predicates since they weren't accounting for transaction fees when spending the predicate.


Salka1988 (approved):

packages/fuels-signers/src/wallet.rs
```diff
@@ -336,6 +420,97 @@ impl WalletUnlocked {
...
+    ) -> Result<(), Error> {
+        let consensus_parameters = self
+            .get_provider()?
+            .chain_info()
+            .await?
+            .consensus_parameters;
+        let transaction_fee = TransactionFee::checked_from_tx(&consensus_parameters.into(), &*tx)
+            .expect("Error calculating TransactionFee");
+
+        let (base_asset_inputs, remaining_inputs): (Vec<Input>, Vec<Input>) =
```
segfault-magnet, line 444:
```suggestion
        let (base_asset_inputs, remaining_inputs): (Vec<_>, Vec<_>) =
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -336,6 +420,97 @@ impl WalletUnlocked {
...
+            tx.inputs().iter().cloned().partition(|input| {
+                matches!(input, Input::CoinSigned { .. })
+                    && *input.asset_id().unwrap() == BASE_ASSET_ID
+            });
+
+        let base_inputs_sum: u64 = base_asset_inputs
+            .iter()
+            .map(|input| input.amount().unwrap())
+            .sum();
+        // either the inputs were setup incorecctly, or the passed base_asset_amount is wrong
```
segfault-magnet, line 454:
```suggestion
        // either the inputs were setup incorrectly, or the passed base_asset_amount is wrong
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -336,6 +420,97 @@ impl WalletUnlocked {
...
+            .iter()
+            .any(|input| matches!(input, Input::CoinSigned { .. }));
+        if !is_using_coins && new_base_amount == 0 {
+            new_base_amount = MIN_AMOUNT;
+        }
+
+        let new_base_inputs = self
+            .get_asset_inputs_for_amount(BASE_ASSET_ID, new_base_amount, witness_index)
+            .await?;
+        let adjusted_inputs: Vec<Input> = remaining_inputs
```
segfault-magnet, line 477:
```suggestion
        let adjusted_inputs: Vec<_> = remaining_inputs
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -336,6 +420,97 @@ impl WalletUnlocked {
...
+            .into_iter()
+            .chain(new_base_inputs.into_iter())
+            .collect();
+
+        let is_base_change_present = tx.outputs().iter().cloned().any(|output| {
```
segfault-magnet, line 482:
```suggestion
        let is_base_change_present = tx.outputs().iter().any(|output| {
```

segfault-magnet (commented):
Needs unit testing.

packages/fuels-signers/src/wallet.rs
```diff
@@ -899,4 +904,138 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn add_fee_coins_empty_transaction() -> Result<(), Error> {
+        let wallet_config = add_fee_coins_wallet_config(1);
+        let wallet = launch_custom_provider_and_get_wallets(wallet_config, None)
+            .await
+            .pop()
+            .unwrap();
+        let mut tx = Wallet::build_transfer_tx(&[], &[], TxParameters::default());
+        wallet.add_fee_coins(&mut tx, 0, 0).await?;
```
segfault-magnet, line 975: I'd put a line break between the three AAA's of the test for visibility.

packages/fuels-signers/src/wallet.rs
```diff
@@ -899,4 +904,138 @@ mod tests {
...
+                    zero_utxo_id,
+                    wallet.address().into(),
+                    20,
+                    BASE_ASSET_ID,
+                    TxPointer::default(),
+                    0,
+                    0,
+                )
+            })
+            .collect();
```
segfault-magnet, line 1030: Another way of writing it which doesn't require ignoring the counter:
```rust
 let mut expected_inputs = repeat(Input::coin_signed(
            zero_utxo_id,
            wallet.address().into(),
            20,
            BASE_ASSET_ID,
            TxPointer::default(),
            0,
            0,
        ))
        .take(3)
        .collect::<Vec<_>>();
```
where `repeat` is `std::iter::repeat`. It generates an infinite iterator that repeats the given element from which you only take 3 elements.

packages/fuels/tests/harness.rs
```diff
@@ -1723,35 +1724,33 @@ async fn test_arrays_with_custom_types() -> Result<(), Error> {
     assert_eq!(states[1], response.value[1]);
     Ok(())
 }
+// TODO: Enable test
```
segfault-magnet, line 1727: Is this TODO as in I'll do it later in my PR or is there something stopping you from doing it in this PR?

IIRC these are the ones that required the new sway stdlib to work? Isn't that out by now? Seem to remember @digorithm enabling some tests.

segfault-magnet (approved):

hal3e (approved):
LGTM! Well done!

docs/src/calling-contracts/cost-estimation.md
```diff
@@ -26,3 +26,4 @@ Below are examples that show how to get the estimated transaction cost from sing
 
 The transaction cost estimation can be used to set the gas limit for an actual call, or to show the user the estimated cost.
 
+> **Note:** whenever you perform an action that starts a transaction (contract deployment, contract call, asset transfer), the SDK will automatically estimate the fee behind the scenes and prepare the transaction accordingly. A side-effect of this is that transactions require the wallet to own at least an amount of 1 of the base asset, even if the gas cost is set to 0 via `TxParameters`.
```
digorithm, line 29: > require the wallet to own at least an amount of 1 of the base asset

Just double checking; this means at least one coin but not necessarily with its _value_ being 1? e.g. if the base asset is ETH, you'd need at least 1 ETH "coin" but the value could be something reasonable, like `0.001`?

packages/fuels-signers/src/wallet.rs
```diff
@@ -391,35 +567,92 @@ impl WalletUnlocked {
...
-        let mut tx = self
-            .get_provider()?
-            .build_transfer_tx(&inputs, &outputs, tx_parameters);
+        let mut tx = Wallet::build_transfer_tx(&inputs, &outputs, tx_parameters);
+
+        if asset_id == AssetId::default() {
+            self.add_fee_coins(&mut tx, amount, 0).await?;
+        } else {
+            self.add_fee_coins(&mut tx, 0, 0).await?;
+        };
```
digorithm, line 576: Can we add a comment here explaining why we decide between `amount` and `0`? Very confusing without proper context.

packages/fuels-signers/src/wallet.rs
```diff
@@ -459,14 +692,20 @@ impl WalletUnlocked {
...
             &inputs,
             &outputs,
             tx_parameters,
         );
+        let base_amount = if asset_id == AssetId::default() {
+            balance
+        } else {
+            0
+        };
+        self.add_fee_coins(&mut tx, base_amount, 0).await?;
```
digorithm, line 708: Same here:

> Can we add a comment here explaining why we decide between amount and 0? Very confusing without proper context.

packages/fuels-signers/src/wallet.rs
```diff
@@ -336,6 +420,98 @@ impl WalletUnlocked {
...
+    // the existing transaction inputs because the selected coins may exceed
+    // the required amount to avoid dust. Therefore we require it as an argument.
+    //
+    // Requires contract inputs to be at the start of the transactions inputs vec
+    // so that their indexes are retained
+    pub async fn add_fee_coins(
+        &self,
+        tx: &mut Transaction,
+        previous_base_amount: u64,
+        witness_index: u8,
```
digorithm, line 434: Just some thoughts on usability here. 

From a UX perspective, I dislike having to leak `witness_index` at this level. For most users that just want to `add_fee_coins()` to their transactions, I believe this will sound very confusing. I'm seeing lots of `add_fee_coins(&mut tx, 0, 1)`, which is very unreadable :(.

I understand it's hard _not_ to need `previous_base_amount`. But how about not needing `witness_index` here? Possible? How hard? Would it be worth the work to improve the interface here? How often will this function be directly used by the users?

Is the decision between `add_fee_coins(&mut tx, amount, 0)` and `add_fee_coins(&mut tx, amount, 1)` only when it comes to deployment vs. regular calls? If so, would it be better to have one specific for deployment (e.g. `add_deployment_fee_coins()`) and one "regular" one (e.g. `add_fee_coins()` as-is)? 

Restricting the third param, which is pretty open right now (could be `0`, `1`, or any typo possible), by removing the third param and turning this decision into a simple English-based decision between two methods should be a better experience.

Thoughts?

digorithm (dismissed):

docs/src/calling-contracts/cost-estimation.md
MujkicA: IIUC `amount` is an integer so the smallest I can assume here is 1. I don't know how that will convert to ETH or some other asset, but the smallest we can represent is always an amount of 1.

packages/fuels-signers/src/wallet.rs
MujkicA: > How often will this function be directly used by the users?
Not often I assumed, since users rarely hand-craft their own transactions. And if they do, do they still need this?

> Restricting the third param, which is pretty open right now (could be 0, 1, or any typo possible), by removing the third param and turning this decision into a simple English-based decision between two methods should be a better experience.

I agree with this. However, I don't know all the possible scenarios for witness index. In our current code it is either 1 when deploying or 0 in all other cases. What are the other possibilities and are they worth considering?
I guess it's fine if we keep it as more of an internal helper.

docs/src/calling-contracts/cost-estimation.md
digorithm: Yeah, right. I guess we can worry about this later then!

packages/fuels-signers/src/wallet.rs
digorithm: > Not often I assumed, since users rarely hand-craft their own transactions.

Ah, this is a key point. For a sec, I thought this would be required in simple contract calls (like most examples in our main test suite), but this isn't the case. So, for the majority of the use cases, it's still simple, and no need to know about `add_fee_coins()`.

If they're hand-crafting txs, most likely than not they'll know and think about witness index stuff, so `add_fee_coins()` will be at the same abstraction level, which is all good 😃.

I believe we can keep as-is, just add the comments, and also turn this method's top-level comment into a proper docstring (`\\\`), please!

Salka1988 (approved):
LGTM!

PR #559, opened 2022-09-01
Use flat abigen only and drop support for the old JSON ABI

digorithm:
This PR forces the Abigen workflow to use only the new `FlatAbigen` (now renamed to just `Abigen`).

A couple of things to pay attention to:

1. Many unit tests related to the old abigen are now disabled (commented out). They weren't being used under the new abigen, and now that we've completely removed the old JSON ABI types, they can't stay around. So they must be skipped and reworked for the new JSON ABI (the correct types, etc.). This (lack of unit tests for the new abigen) has a relatively low impact because it's being extensively tested at the integration level in the main `harness.rs` file. But, we should attend to this later for extra safety/confidence;
2. The encoder/decoder CLI is also disabled for now; this hasn't seen much use other than for debugging, and, just like the tests, it needs to be reworked for the new JSON ABI format, which isn't too complicated but we can't afford to do it right now. 

Other than these points, this is mostly deleting the old `Abigen`-related code and renaming the `FlatAbigen`-related code, making it the only path for the `abigen!` macro.


packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -1,66 +1,48 @@
...
-
-    custom_structs: HashMap<String, Property>,
+use super::custom_types_gen::{expand_custom_enum, expand_custom_struct};
+use super::functions_gen::expand_function;
 
-    custom_enums: HashMap<String, Property>,
+// This is a temporary file that duplicates most of the logic in `abigen.rs`.
+// This is necessary to support both the old and new JSON ABI formats.
+// Once we drop the support for the old one, we can remove this file and rename
+// `flat_abigen.rs` to `abigen.rs`.
```
segfault-magnet, line 18:
```suggestion
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -1,66 +1,48 @@
...
+
+    types: HashMap<usize, TypeDeclaration>,
 }
 
 impl Abigen {
     /// Creates a new contract with the given ABI JSON source.
     pub fn new<S: AsRef<str>>(contract_name: &str, abi_source: S) -> Result<Self, Error> {
-        let source = Source::parse(abi_source).unwrap();
-        let mut parsed_abi: JsonABI = serde_json::from_str(&source.get().unwrap())?;
+        // Support for new JSON ABI file format.
```
segfault-magnet, line 38:
```suggestion
```

tools/fuels-abi-cli/src/main.rs
```diff
@@ -1,187 +1,185 @@
...
-    }
-
-    Ok(result)
-}
+//     Ok(result)
+// }
 
 fn main() -> anyhow::Result<()> {
-    println!("{}", execute(std::env::args())?);
+    // println!("{}", execute(std::env::args())?);
```
segfault-magnet, line 182: We might be better off erroring out. But I'm not sure of the audience of these CLI tools. Might be confusing to just have it run and do nothing :)

segfault-magnet (commented):
LGTM.

segfault-magnet (approved):

PR #566, opened 2022-09-07
feat: add setup_contract_test proc macro

hal3e:
I have added a new macro `setup_contract_test!` and now a simple test can look like this:

```rust
#[tokio::test]
async fn simple_test() -> Result<(), Error> {
    setup_contract_test!(
        contract_instance,
        wallet,
        "packages/fuels/tests/test_projects/contract_test"
    );

        .initialize_counter(42)
        .call()
        .await?;

    assert_eq!(42, response.value);
    Ok(())
}
```
The macro can be used multiple times in the same function if the variable names are different. The same contract can be deployed multiple times as it deploys with salt.

If however, you want to use the same wallet with multiple macros you have to set the first wallet name to `shared_wallet` and the rest to `None` like this:

```rust
setup_contract_test!(
    contract_instance,
    shared_wallet,
    "packages/fuels/tests/test_projects/contract_test"
);

setup_contract_test!(
    contract_instance2,
    None,
    "packages/fuels/tests/test_projects/contract_test"
);

setup_contract_test!(
    contract_instance3,
    None,
    "packages/fuels/tests/test_projects/contract_test"
);
```

This is useful if you call one contract from another as the contracts need to be deployed on the same provider.

A test inside a `forc` project can look like this.

```rust
use fuels::prelude::*;

#[tokio::test]
async fn can_get_contract_id() -> Result<(), Error> {
    setup_contract_test!(contract_instance, wallet, ".");

    let response = contract_instance.test_function().call().await?;

    assert!(response.value);

    Ok(())
}
```


hal3e:
The last commit should be enough to make the `get_contract_instance!` macro only available within `fuels-rs`.

I have tested it by creating a new cargo project and importing `fuels` with
```toml
fuels = { git = "https://github.com/FuelLabs/fuels-rs/", branch = "hal3e/get-contract-instance-macro" }
````

**Comment by hal3e** [comment]
Normally, a test specific setup function for the `harness` would be place in `tests/some_name/mod.rs`, but because this is a proc macro we can not use it from within the same crate.

**Comment by hal3e** [comment]
Changing the name is a great idea and I have no problem with exposing this to the users. I will refactor the import a bit and make it work inside a `forc` project.

I have only one doubt. Currently, the macro makes 3 variables available: 
- `contract_instance` the name can be specified,
- `wallet` - name is hardcoded
- `contract_id` name is also hardcoded

Should we have something like this where we only give the `path` as input and then we use the hardcoded names.
```rust
setup_contract_test!(PATH)
```
or all the names can be specified:

```rust
setup_contract!(
    contract_instance,
    contract_id,
    wallet,
    PATH
)
```
or something else...

digorithm:
> I have changed the name of the macro to `setup_contract_test` and I have changed what `PATH` is expected. Now the PATH is a `forc` project folder. For the harness file you would call the macro with
> 
> ```rust
>     setup_contract_test!(
>         contract_instance,
>         "packages/fuels/tests/test_projects/contract_test"
>     );
> ```
> 
> Now that the `PATH` is a `forc` project. Calling it from within one becomes:
> 
> ```rust
>     setup_contract_test!(contract_instance, ".");
> ```
> 
> However, the test setup code of the `forc` project would have to change a bit. Previously we had:
> 
> ```rust
> use fuels::{prelude::*, tx::ContractId};
> 
> // Load abi from json
> abigen!(MyContract, "out/debug/setup_test-abi.json");
> 
> async fn get_contract_instance() -> (MyContract, ContractId) {
>     // Launch a local network and deploy the contract
>     let mut wallets = launch_custom_provider_and_get_wallets(
>         WalletsConfig::new(
>             Some(1),             /* Single wallet */
>             Some(1),             /* Single coin (UTXO) */
>             Some(1_000_000_000), /* Amount per coin */
>         ),
>         None,
>     )
>     .await;
>     let wallet = wallets.pop().unwrap();
> 
>     let id = Contract::deploy(
>         "./out/debug/setup_test.bin",
>         &wallet,
>         TxParameters::default(),
>         StorageConfiguration::with_storage_path(Some(
>             "./out/debug/setup_test-storage_slots.json".to_string(),
>         )),
>     )
>     .await
>     .unwrap();
> 
>     let instance = MyContractBuilder::new(id.to_string(), wallet).build();
> 
>     (instance, id.into())
> }
> 
> #[tokio::test]
> async fn can_get_contract_id() {
>     let (_instance, _id) = get_contract_instance().await;
> 
>     // Now you have an instance of your contract you can use to test each function
> }
> ```
> 
> Now we have:
> 
> ```rust
> use fuels::prelude::*;
> 
> #[tokio::test]
> async fn test_contract() -> Result<(), Error> {
>    setup_contract_test!(contract_instance, ".");
> 
>    let response = contract_instance
>        .test_function()
>        .call()
>        .await?;
> 
>    dbg!(response.value);
>    assert!(response.value);
>    Ok(())
> }
> ```
> 
> **Note** that the macro can not be called outside of a function because of the `let` statements.

RE: the naming and parameters of the macro (`setup_contract_test!(contract_instance, ".");`), that's the sweet spot, I believe. I like it!

RE: Changing `forc` output. That's expected; we can get this in first, then push a PR on `forc` with these changes. This is a _massive_ boilerplate code reduction! The whole `get_contract_instance` from `forc` can intimidate new users.  

One might say we could be hiding too much (contract IDs, wallet, provider, etc.), but I think, in this case, it's worth it to get the user to write tests quickly. As usual, documentation will be key here: we have to let users know that, with that `contract_instance`, they can get the underlying test wallet and the contract ID in case they need those. We could even generate a comment with these auto-generated test files. Like:

```rust
use fuels::prelude::*;

#[tokio::test]
async fn test_contract() -> Result<(), Error> {
   setup_contract_test!(contract_instance, ".");
   
   // Use contract_instance._get_wallet() and contract_instance._get_contract_id()
   // to get the underlying test wallet and the contract ID, respectively.
       .test_function()
       .call()
       .await?;

   dbg!(response.value);
   assert!(response.value);
   Ok(())
}
```

hal3e:
I have refactored the macro to name all the exposed varibles. Now you can call the macro with:
```rust
    setup_contract_test!(
        contract_instance,
        wallet,
        contract_id,
        "packages/fuels/tests/test_projects/contract_test"
    );
```
and inside a `forc` project with:
```rust
    setup_contract_test!(contract_instance, wallet, contract_id, ".");
```
Now it is possible to call the macro multiple times inside a function if you change the variable names.

hal3e:
I have refactored the macro and removed the `contract_id` as we can get it from the instance. Now the macro can be called multiple times with the same contract as it deploys with salt. And we can have a shared wallet. Check the PR description for more info

packages/fuels-abigen-macro/src/lib.rs
```diff
@@ -30,6 +34,113 @@ pub fn wasm_abigen(input: TokenStream) -> TokenStream {
...
+        )
+    } else {
+        let wallet_name = Ident::new(&args.wallet_name, Span::call_site());
+        (
+            wallet_name.clone(),
+            quote! {let #wallet_name = launch_provider_and_get_wallet().await;}.into(),
+        )
+    };
+
+    let contract_deploy_token_stream: TokenStream = quote! {
```
segfault-magnet, line 120: I'm not sure you need the explicit type here. Should be ok without it.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -109,8 +109,8 @@ impl Abigen {
                     impl #name {
                         #contract_functions
 
-                        pub fn _get_contract_id(&self) -> &Bech32ContractId {
-                            &self.contract_id
+                        pub fn _get_contract_id(&self) -> Bech32ContractId {
+                            self.contract_id.clone()
```
MujkicA, line 113: Do we need to return an owned value? IMO its better to stick to the ref and push the decision to clone on the user.

MujkicA (commented):
Looks good, just a minor nit on my side.

packages/fuels-abigen-macro/src/lib.rs
```diff
@@ -30,6 +34,102 @@ pub fn wasm_abigen(input: TokenStream) -> TokenStream {
...
+    let abs_forc_dir = Path::new(&args.project_path)
+        .canonicalize()
+        .unwrap_or_else(|_| {
+            panic!(
+                "Unable to canonicalize forc project path: {}. Make sure the path is valid!",
+                &args.project_path
+            )
+        });
+
+    let forc_project_name = abs_forc_dir.file_name().unwrap().to_str().unwrap();
```
digorithm, line 58: Nit to avoid `unwrap` in order to help us debugger down the line 😄 

```suggestion
    let forc_project_name = abs_forc_dir.file_name().expect("failed to get filename").to_str().expect("failed to convert filename to string");
```

digorithm (changes requested):
Awesome work! Very excited to release this macro.

Left just a nit, feel free to merge right away after it.

One more thing: we have to document this macro somewhere in the SDK book. We can either do it here now or open an issue and address it later, totally up to you!

hal3e:
Good point! I will add the docs in this PR.

packages/fuels-abigen-macro/src/lib.rs
```diff
@@ -30,6 +34,102 @@ pub fn wasm_abigen(input: TokenStream) -> TokenStream {
...
+        .join(["out/debug/", forc_project_name, ".bin"].concat())
+        .to_str()
+        .expect("could not join path for bin file")
+        .to_string();
+
+    let storage_path = abs_forc_dir
+        .join(["out/debug/", forc_project_name, "-storage_slots.json"].concat())
+        .to_str()
+        .expect("could not join path for storage file")
+        .to_string();
```
segfault-magnet, line 76:
```suggestion
    let compiled_file_path = |suffix: &str, desc: &str| {
        abs_forc_dir
            .join(["out/debug/", forc_project_name, suffix].concat())
            .to_str()
            .unwrap_or_else(|| panic!("could not join path for {desc} file"))
            .to_string()
    };

    let abi_path = compiled_file_path("-abi.json", "the ABI file");
    let bin_path = compiled_file_path(".bin", "the binary file");
    let storage_path = compiled_file_path("-storage_slots.json", "The storage slots file");
```

hal3e: Much cleaner! Love it!

segfault-magnet (dismissed):
Love the initiative. Keep 'em coming.

Salka1988 (dismissed):
Nice work Halil! :muscle:

examples/contracts/src/lib.rs
```diff
@@ -73,6 +73,25 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn setu_contract_test_example() -> Result<(), Error> {
```
MujkicA, line 77:
```suggestion
    async fn setup_contract_test_example() -> Result<(), Error> {
```

MujkicA (dismissed):

docs/src/contracts/the-setup-contract-test-macro.md
```diff
@@ -0,0 +1,18 @@
+# The setup_contract_test! macro
+
+When deploying contracts with the `abigen!` macro, as shown in the previous sections, the user has the ability to change the default configuration parameters, lunch several providers, create multiple wallets, create specific assets, etc. However, it is often the case that we want to test only the contract methods and we want to simply deploy the contract with the default configuration parameters. The `setup_contract_test!` macro does exactly that. When expanded, the `setup_contract_test!` macro will: run the abigen, launch a local provider, setup one wallet and deploy the selected contract. The setup code that you have seen in previous sections gets reduced to:
```
segfault-magnet, line 3:
```suggestion
When deploying contracts with the `abigen!` macro, as shown in the previous sections, the user can:
* change the default configuration parameters, 
* launch several providers, 
* create multiple wallets, 
* create specific assets, etc. 

However, it is often the case that we want to test only the contract methods and we want to simply deploy the contract with the default configuration parameters. The `setup_contract_test!` macro does exactly that. When expanded, the `setup_contract_test!` macro will: 
1. run the abigen, 
2. launch a local provider, 
3. setup one wallet 
4. deploy the selected contract. 
The setup code that you have seen in previous sections gets reduced to:
```

segfault-magnet: Albeit you probably shouldn't commit the suggestion as I've typed it, cuz I've wrapped it to fit a ~100char column. Although the final output is the same, I recall somebody having an issue with the line breaks. Can't remember the details.

In any case the suggestion was `lunch` vs `launch` and the enumerated and unenumerated lists.

PR #568, opened 2022-09-09
chore: port and enable flat JSON ABI tests

### LINKED ISSUE #560: Port unit tests using the old JSON ABI types to the new flat JSON ABI
https://github.com/FuelLabs/fuels-rs/pull/559 disables many unit tests related to the old JSON ABI. For instance: https://github.com/FuelLabs/fuels-rs/pull/559/files#diff-417218521933a693aac563b25e4c2417453e15033f72a145d669350451ce5470R473

All these tests (you can find them all by grepping for `"TODO: Move tests using the old abigen to the new one."`) need to be ported to the new JSON ABI. This is a matter of keeping the logic of the tests the same, just moving the types from the old JSON ABI to the new ones, e.g., `Property` => `TypeDeclaration`.

Br1ght0ne:
Close #560.

Notes:

- [`nested_enum_w_no_variants_cannot_be_constructed`](https://github.com/FuelLabs/fuels-rs/blob/oleksii/port-abi-tests/packages/fuels-core/src/code_gen/custom_types.rs#L90) needs further debugging
- some functions used in [`functions_gen.rs` tests](https://github.com/FuelLabs/fuels-rs/blob/oleksii/port-abi-tests/packages/fuels-core/src/code_gen/functions_gen.rs#L616) were removed, so I'm uncertain if these tests can be preserved

---
Draft progress:

- [x] ~~`fuels-types/parse_param`~~
- [x] ~~`fuels-core/json_abi`~~
- [x] `fuels-core/code_gen/functions_gen`
- [x] `fuels-core/code_gen/custom_types`
- [x] `fuels-core/tokenizer`


packages/fuels-core/src/tokenizer.rs
```diff
@@ -81,7 +81,8 @@ impl Tokenizer {
     /// And attempts to return a `Token::Struct()` containing the inner types.
     /// It works for nested/recursive structs.
     pub fn tokenize_struct(value: &str, params: &[ParamType]) -> Result<Token, Error> {
-        if !value.starts_with('(') || !value.ends_with(')') {
+        dbg!((value, params));
```
MujkicA, line 84: Leftover from debugging?

hal3e (commented):
I had a look at the `nested_enum_w_no_variants_cannot_be_constructed` function and it seems that the new implementation only goes one level deep and makes the `TheEmptyEnum` with its components. With the new implementation, the error would be caught when we try to expand the `SomeEmptyEnum` later on. 
As @MujkicA said I would also advise to go over the test with @segfault-magnet as he has recently changed how things get resolved. 

That aside, maybe move the tests regarding `enums` and `structs` to the respective files in `custom_types/` or `custom_types/utils.rs`

packages/fuels-core/src/code_gen/custom_types.rs
```diff
@@ -19,477 +19,639 @@ pub use utils::{
...
+            "#,
+        )?
+        .to_string();
+
+        assert_eq!(actual, expected);
+        Ok(())
+    }
+
+    #[test]
+    fn top_lvl_enum_w_no_variants_cannot_be_constructed() -> anyhow::Result<()> {
```
segfault-magnet, line 84: A lot has changed since these tests were written.

Due to the recursive nature of the type resolving, I'd say we're safe enough if we just cover that **a** enum cannot be expanded if it has no variants.

The same should hold for structs.

So maybe to ditch these 'nested enums/structs' tests and replace them with two tests. One for no-variant enums and another for no-field structs.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -600,6 +613,9 @@ mod tests {
         Ok(())
     }
 
+    // TODO(oleksii): as expand_input_param function is gone, should all these
```
segfault-magnet, line 616: Expanding function inputs now depends on the correct functioning of 

`Component::new`, `resolve_type` and `impl From<&ResolvedType> for TokenStream`.

A `Component` contains the function argument name (transformed to snake case if need be) and the `ResolvedType` of the argument. The behavior regarding the transformation of the argument name is covered by tests.

`resolve_type` works with the raw JSON until it results in a `ResolvedType` which, through `From<&ResolvedType> for TokenStream` can be converted into a valid Rust type signature -- e.g. `SomeStruct<u8>`. 

`resolved_type` should be unit tested to ensure we're resolving all the different types available (see ParamTypes). 

For some reason unknown to me, I have not written unit tests for `resolved_type` -- so its continued functioning depends on our e2e harness.

The `From<&ResolvedType> for TokenStream` should be tested as well.

So I'd focus my efforts there, either in this PR or another one.

Br1ght0ne: Added a new issue #635. Thanks!

segfault-magnet (approved):

PR #571, opened 2022-09-12
Make gas_forwarded use gas_limit as default

### LINKED ISSUE #569: When forward gas_limit is set to None it should use the transaction gas_limit
When executing the following contract call with transaction gas_limit, the contract call returns an `OutOfGas` issue. To be able to execute, the users also need to set gas_limit to be forwarded on the contract call.

but the behavior I was expecting was, When the user doesn't forward a gas_limit or, in this case, passes a None, the contract calls should not pass any value for the VM, enabling in this way the VM to pass all the available gas to the contract call.

Is this a expected behavior?

## OutOfGas example
```rs
let response = exchange_instance
        .remove_liquidity(1, 1, 1000)
        .call_params(CallParameters::new(
            Some(lp_amount_received),
            Some(lp_asset_id.clone()),
            None
        ))
        .tx_params(TxParameters {
            gas_price: 0,
            gas_limit: 100_000_000,
            maturity: 0,
        })
        .append_variable_outputs(2)
        .call()
        .await
        .unwrap();
```

## Working example
```rs
let response = exchange_instance
        .remove_liquidity(1, 1, 1000)
        .call_params(CallParameters::new(
            Some(lp_amount_received),
            Some(lp_asset_id.clone()),
            Some(100_000_000)
        ))
        .tx_params(TxParameters {
            gas_price: 0,
            gas_limit: 100_000_000,
            maturity: 0,
        })
        .append_variable_outputs(2)
        .call()
        .await
        .unwrap();
```

MujkicA:
Closes #569 

This PR adjusts `CallParameters` so that `gas_forwarded` is stored as an `Option`. This let's us decide on using the `gas_limit` instead if it wasn't specified by the user.

I've also moved a small part of the docs related to using estimates for inputs to the tx parameters section. It seems more fitting since the estimate make use of the gas_limit.

### DIFF
--- a/docs/src/calling-contracts/call-params.md
+++ b/docs/src/calling-contracts/call-params.md
@@ -37,3 +37,5 @@ The `gas_forwarded` parameter defines the limit for the actual contract call as 
 ```rust,ignore
 {{#include ../../../examples/contracts/src/lib.rs:call_params_gas}}
 ```
+
+If you don't set the call parameters, or use `CallParameters::default()` the transaction gas limit will be forwarded instead.
--- a/docs/src/calling-contracts/call-params.md
+++ b/docs/src/calling-contracts/call-params.md
@@ -38,4 +38,4 @@ The `gas_forwarded` parameter defines the limit for the actual contract call as 
 {{#include ../../../examples/contracts/src/lib.rs:call_params_gas}}
 ```
 
-If you don't set the call parameters, or use `CallParameters::default()` the transaction gas limit will be forwarded instead.
+If you don't set the call parameters or use `CallParameters::default()`, the transaction gas limit will be forwarded instead.
--- a/packages/fuels-core/src/parameters.rs
+++ b/packages/fuels-core/src/parameters.rs
@@ -1,5 +1,6 @@
 use crate::constants::{
-    BASE_ASSET_ID, DEFAULT_FORWARDED_GAS, DEFAULT_GAS_LIMIT, DEFAULT_GAS_PRICE, DEFAULT_MATURITY,
+    BASE_ASSET_ID, DEFAULT_CALL_PARAMS_AMOUNT, DEFAULT_GAS_LIMIT, DEFAULT_GAS_PRICE,
+    DEFAULT_MATURITY,
 };
 use fuel_tx::{AssetId, StorageSlot};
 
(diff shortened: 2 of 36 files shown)


docs/src/calling-contracts/call-params.md
```diff
@@ -37,3 +37,5 @@ The `gas_forwarded` parameter defines the limit for the actual contract call as
 ```rust,ignore
 {{#include ../../../examples/contracts/src/lib.rs:call_params_gas}}
 ```
+
+If you don't set the call parameters, or use `CallParameters::default()` the transaction gas limit will be forwarded instead.
```
digorithm, line 41:
```suggestion
If you don't set the call parameters or use `CallParameters::default()`, the transaction gas limit will be forwarded instead.
```

Grammar nit!

digorithm (dismissed):
Very good! Just left a tiny nit -- feel free to merge after!

packages/fuels-core/src/parameters.rs
```diff
@@ -14,25 +15,30 @@ pub struct TxParameters {
...
-            amount: amount.unwrap_or(0),
+            amount: amount.unwrap_or(DEFAULT_CALL_PARAMS_AMOUNT),
             asset_id: asset_id.unwrap_or(BASE_ASSET_ID),
-            gas_forwarded: gas_forwarded.unwrap_or(DEFAULT_FORWARDED_GAS),
+            gas_forwarded,
         }
     }
+
+    /// Returns gas_forwarded if set or default if None
+    pub fn gas_forwarded_or(&self, default: u64) -> u64 {
```
segfault-magnet, line 31: Thinking out loud, no action needed:

Maybe we could have just returned the option and had the users call `unwrap_or` on it. Or just have this as an exception and not do this too often. Wouldn't like to do this for all options out there.

With the 'return Option approach' maybe the comment wouldn't be warranted since an option implies that something might be missing and thus you've decided on the `or` part.

Technically somebody might question, reading only the name of the method, what conditions need to be met for this method to not give the gas forwarded but rather the default passed in. The comment was needed to further explain that the gas forwarded might not have been set.

segfault-magnet (approved):

packages/fuels-core/src/parameters.rs
MujkicA: It was bit of desperate attempt to provide more context as to what is happening since the logic that defaults to gas_limit is deeply embedded into the call script generation.
But looking at it now, it doesn't achieve much so we might as well go for `unwrap_or `

segfault-magnet (approved):

Issue #448, opened 2022-07-05
feat: support type arguments in abigen and the function selector


mohammadfawaz:
Spec change: https://github.com/FuelLabs/fuel-specs/pull/382

This also requires changes to the type `Property` in `fuels-types` to add `typeArguments` which is needed by the Sway compiler.

Edit by @digorithm:

First, we need to achieve feature parity with the new JSON ABI format, this is tracked here: https://github.com/FuelLabs/fuels-rs/issues/525. 

Once this is done, I'll proceed with supporting generics and type arguments.

Issue #439, opened 2022-06-29
Handle `Unit`s in structs


segfault-magnet:
If we try to generate bindings for this (valid) contract:
```
contract;

abi MyContract {
    fn test_function2() -> Super;
}

pub struct Super {
    a: (),
    b: u32,
}

impl MyContract for Contract {
    fn test_function2() -> Super {
        Super {
            a: (),
            b: 12345u32,
        }
    }
}
```
we'll get this unreadable mess:
```
error: expected type, found `,`
 --> tests/harness.rs:5:1
  |
5 | abigen!(MyContract, "out/debug/something-abi.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected type
  |
  = note: this error originates in the macro `abigen` (in Nightly builds, run with -Z macro-backtrace for more info)

error: expected type, found `>`
 --> tests/harness.rs:5:1
  |
5 | abigen!(MyContract, "out/debug/something-abi.json");
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | expected type
  | while parsing this struct
```

Since the support for `Unit` as a field parameter in Sway isn't going away, we should either support it in Rust (since we can), 

or reject it with a more readable message.

Setting a low priority on this one since I can't possibly see what use would a Unit be as a field type.

digorithm:
@segfault-magnet isn't this already solved?

digorithm:
Awesome, thanks for covering all logical paths 😂

segfault-magnet:
Fails under master, passes under generics. Will add the keyword.

PR #573, opened 2022-09-12
feat: Implement generic support

segfault-magnet:
closes: #538 
closes: https://github.com/FuelLabs/fuels-rs/issues/448
closes: #439 

Generic support with the help of @digorithm.

### About the solution

In order for generics to be clean and (hopefully) robust I needed to do some potentially controversial changes.

Change 1:
Sway arrays are not represented by std::vec::Vec in the SDK anymore. They are a native rust array of the same size as defined in the ABI.

Reason: 
In order to not have a bunch of edge-cases when generating code, the size of the sway array needed to be part of the type, which wasn't possible with std::Vec. 

The impact of this should be lessened IMO by the introduction of vectors (which will actually be a std::vec::Vec).

Change 2:
Sway strings are not represented by std::String in the SDK anymore. There is a new type called SizedAsciiString which has a const generic LEN equal to the one used in the ABI. 

Once the sway team stops having the size of the string as part of its type, we'll remove this as well.

Reason:
Same as before.

Change 3:
`b256` is now represented by a `Bits256` wrapping struct.

Reason:
To differentiate `b256` from an ordinary array containing 32 `u8`s. The difference needs to be present due to the function selector treating [u8;32] and `b256` differently. 

The main change to how we generate code is reflected in that we now don't pry into what a type might be but rather require that all types implement Parameterized and Tokenizable.

So if you want to get the param type of a struct, just call ::param_type on all underlying fields of the struct and wrap that into a `ParamType::Struct`.

The same for enums.


hal3e:
Awesome work! :tada:  I really like the idea of separating `arrays` from `vec` and giving `strings` and `b256` new types. It makes the code much cleaner!

packages/fuels/tests/test_projects/generics/src/main.sw
```diff
@@ -0,0 +1,108 @@
...
+}
+
+enum EnumWGeneric<N> {
+    a: u64,
+    b: N,
+}
+
+struct MegaExample<T, U> {
+    a: ([U; 2], T),
+    b: ([EnumWGeneric<StructWTupleGeneric<StructWArrayGeneric<PassTheGenericOn<T>>>>; 1], u32),
```
digorithm, line 29: mother-of-god

packages/fuels-core/src/code_gen/mod.rs
```diff
@@ -4,4 +4,6 @@ pub mod abigen;
 pub mod bindings;
```
digorithm, line 4: Same comment about `mod.rs`: 

> I believe we've migrated from the mod.rs filename pattern. Not a recommended pattern anymore; see here: https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames.

I know this was already here, but since we're making massive refactors here, might as well.

packages/fuels-core/src/code_gen/resolved_type.rs
```diff
@@ -0,0 +1,115 @@
...
+            .flatten()
+            .map(|array_type| resolve_type(&array_type, &types))
+            .collect::<Result<Vec<_>, _>>()
+    };
+
+    let base_type = types.get(&type_application.type_id).unwrap();
+    let param_type = ParamType::from_type_declaration(base_type, types)?;
+
+    let (type_name, generic_params) = match &param_type {
+        ParamType::Generic(name) => Ok::<_, Error>((name.parse().unwrap(), vec![])),
```
digorithm, line 67: Could make our lives easier if we're debugging something related to this:

```suggestion
        ParamType::Generic(name) => Ok::<_, Error>((name.parse().expect("Failed to parse generic param name"), vec![])),
```

packages/fuels-core/src/lib.rs
```diff
@@ -1,4 +1,6 @@
+#![allow(warnings)] // temp remove me later
```
digorithm, line 1: Is this still needed?

packages/fuels-types/src/parse_param.rs
```diff
@@ -29,12 +29,20 @@ impl ParamType {
                     // Try to parse tuple (T, T, ..., T)
                     return ParamType::parse_tuple_param(prop, types);
                 }
+                if prop.type_field.contains("generic") {
```
digorithm, line 32: Paranoid sanity check: are we sure `type_field` containing `"generic"` is always because of the `generic <NAME>`? Would it be possible that there's a `type_field` that, for whatever reason, contains `generic` and it's not an actual generic within our context?

packages/fuels-types/src/parse_param.rs
```diff
@@ -29,12 +29,20 @@ impl ParamType {
...
+                    return Ok(Self::parse_generic_param(prop));
+                }
                 // Try to parse a free form enum or struct (e.g. `struct MySTruct`, `enum MyEnum`).
                 ParamType::parse_custom_type_param(prop, types)
             }
         }
     }
 
+    fn parse_generic_param(prop: &TypeDeclaration) -> ParamType {
+        let name = prop.type_field.split("generic").collect::<Vec<&str>>()[1].trim();
```
digorithm, line 42: Friendly boundary check / error handling:

```suggestion
        let name = prop
            .type_field
            .split("generic")
            .collect::<Vec<&str>>()
            .get(1)
            .expect("generic param string not properly formatted. This should never happen.")
            .trim();
```

digorithm (changes requested):
This is crazy good! Overall it looks excellent, just left a couple of nits, clarity-related stuff, and error-handling stuff.

packages/fuels-core/src/abi_types/common.rs
```diff
@@ -0,0 +1,51 @@
+use crate::{Parameterize, Token, Tokenizable};
```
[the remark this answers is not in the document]
segfault-magnet: Good point. I needz me a name that screams "all your parameterize/tokenize impls for native rust types belongs to us". 

But I'll just call it array for now.

packages/fuels-core/src/code_gen/custom_types_gen/common.rs
```diff
@@ -0,0 +1,316 @@
+use crate::code_gen::resolved_type::{resolve_type, ResolvedType};
```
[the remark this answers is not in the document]
segfault-magnet: Stuff used by both struct and enum gens, common to them both. 

Perhaps `utils`? It is a private mod, shouldn't be confused with other utils out there.

packages/fuels-core/src/code_gen/custom_types_gen/common.rs
```diff
@@ -0,0 +1,316 @@
...
+        types: &HashMap<usize, TypeDeclaration>,
+        snake_case: bool,
+    ) -> anyhow::Result<Component> {
+        let field_name = if snake_case {
+            component.name.to_snake_case()
+        } else {
+            component.name.to_owned()
+        };
+
+        let component1 = Component {
```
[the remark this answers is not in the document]
segfault-magnet: Refactoring remnant, good catch!

packages/fuels-core/src/code_gen/custom_types_gen/mod.rs
```diff
@@ -0,0 +1,495 @@
+mod common;
```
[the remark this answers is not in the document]
segfault-magnet: I don't know how to hide the internal structure with that approach? 

Wouldn't want anyone outside the module to be burdened by the fact I've split struct and enum gen into separate files for easier management. Or that I have one file with some stuff common to both of them? 

I'll try something out with the new approach, see how it fits.

MujkicA (commented):
Impressive, very nice. Let's see vector support.
But seriously, good stuff, the tests are intense :satisfied:

packages/fuels-core/src/code_gen/custom_types_gen/mod.rs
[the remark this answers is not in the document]
segfault-magnet: Converted to the modern approach.

packages/fuels-types/src/parse_param.rs
segfault-magnet: Great catch! I've rewritten this, take a look when you can

packages/fuels-types/src/parse_param.rs
segfault-magnet: Good point. I've taken this into consideration in the rewrite of this part, take a look when you can.

digorithm (approved):
LGTM, S-tier work! So damn excited to get this in! 🚀

digorithm (approved):

hal3e (commented):
Finally, it got merged :tada: Working on the Options and Results! Great work!

Issue #421, opened 2022-06-22
Multiple functions with different input `Option` types in results in error


bitzoic:
When two functions are taking inputs of types of `Option` with different types, defining the `Option` in harness causes an error. 

It seems that only the first `Option` is used. In the following example, a function with an input of `Option` of type `u64` is defined first. Attempting to create an `Option` of type `u64` in the harness does not result in an error. However, creating an `Option` of type `Identity` will result in an error. 

When the function definitions are flipped, the opposite is true. Creating the `Option` of type `Identity` does not result in an error however creating the `Option` of type `u64` does. 

Example Contract:

```
abi MyContract {
    fn test_function(option: Option<u64>);
    fn second_function(option: Option<Identity>);
}

impl MyContract for Contract {

    fn test_function(option: Option<u64>) {
        // Do something here
    }
    fn second_function(option: Option<Identity>) {
        // Do something here
    }
}
```

Examples harness file:

```
abigen!(MyContract, "out/debug/test_project-abi.json");

mod test {

    use super::*;

    async fn test_function(contract: &MyContract) -> CallResponse<()> {
        let option = Option::Some(1);

        contract.test_function(option).call().await.unwrap()
    }

    async fn second_function(contract: &MyContract, wallet: &LocalWallet) -> CallResponse<()> {
        let identity = Identity::Address(wallet.address());
        let option = Option::Some(identity);

        contract.test_function(option).call().await.unwrap()
    }
}
```

Corresponding error:
![optionError](https://user-images.githubusercontent.com/54727135/175093334-42877e97-a148-453b-a278-3c952a589092.png)

Issue #420, opened 2022-06-22
Multiple `Option` types must be defined explicitly in abi function definition


bitzoic:
Having multiple different `Option`s in a Sway contract where one `Option` is defined explicitly in a function and the other is not causes an error. 

In this example, we have: 
1. A `struct` that contains an `Option` of type `Identity` 
2. Another `Option` of type `u64`

Only the `Option` of type `u64` is explicitly defined in a function, the `test_function` function. When attempting to create a `TestStruct` in the harness file to pass as an input to `input_function`, the `Option` of type `Identity` is not recognized. 

While this example only shows `Option` of type `u64` returned, the same error occurs when the `Option` of type `u64` is an input variable. Thus, the `Option` must be explicitly defined in a function.

Example Contract:

```
struct TestStruct {
    option_identity: Option<Identity>
}

abi MyContract {
    fn input_function(myStruct: TestStruct);
    fn test_function() -> Option<u64>;
}

impl MyContract for Contract {

    // Here we are taking in a struct that has an Option defined within itself, 
    // but the Option is not defined in the function definition

    fn input_function(myStruct: TestStruct) {
        // Do something here
    }

    // Here we are returning an Option and the Option is defined in the 
    // function definition

    fn test_function() -> Option<u64> {
        Option::Some(1)
    }
}
```

Corresponding harness.rs file:

```
abigen!(MyContract, "out/debug/test_project-abi.json");

mod test {

    use super::*;

    // This function works as intended

    async fn test_function(contract: &MyContract) -> Option {
        contract.test_function().call().await.unwrap().value
    }

    // This function will throw us an error

    async fn input_function(contract: &MyContract, wallet: &LocalWallet) -> CallResponse<()> {
        let identity = Identity::Address(wallet.address());

        let my_struct = TestStruct {
            option_identity: Option::Some(identity) // This line is the problem
        };

        contract.input_function(my_struct).call().await.unwrap()
    }
}
```

The corresponding error:
![optionError](https://user-images.githubusercontent.com/54727135/175077915-ca49d2fe-c8e1-4e4b-8ec1-703063e47dff.png)

digorithm:
This should be resolved by https://github.com/FuelLabs/fuels-rs/issues/415.

bitzoic:
> This should be resolved by #415.

I believe #415 is the interaction between Rust's `Option` and Sway's `Option` while this is purely Sway's `Option`. The same applies to #421

digorithm:
[...]

Issue #415, opened 2022-06-21
Implement equivalency between Rust's `Option` and Sway's `Option`


digorithm:
# Problem

Currently, dealing with an `Option` returned from a Sway contract call is slightly awkward. That's because we generated an enum named `Option` with two variants: `::None()` and `::Some<T>`. This leads to two problems.

The first problem is having `Option`-related code like this:

```Rust
// In sway
abi MyContract {
    fn get_some_address() -> Option<Identity>;
    fn get_none() -> Option<Identity>;
}
```

And then, in Rust:

```Rust
let expected_some_address = Option::Some(Identity::Address(Address::zeroed()));

let response = instance.get_some_address().call().await.unwrap();

assert_eq!(response.value, expected_some_address);

let expected_none = Option::None();

let response = instance.get_none().call().await.unwrap();

assert_eq!(response.value, expected_none);
```

Notice how, instead of `Some(Identity::Address(Address::zeroed()))` we have to explicitly use `abigen!`'s generated `Option::Some`. That's because that `Option` isn't a Rust `Option`, it's the enum `Option` generated by the SDK.

The second problem gets a bit more awkward. What if your Rust code needs Rust's `Option`? Then the `Option`s will conflict, and you'll have to resolve the conflict by using, whenever you want Rust's `Option`, `core::option::Option<T>`.

Here's a concrete example:

In Sway:

```Rust
pub struct Asset {
    amount: u64,
    contract_id: ContractId,
    nft_id: Option<u64>, // Notice the Sway's `Option`.
}
```

Then, in Rust:

```Rust
// Load abi from json
abigen!(EnglishAuction, "out/debug/english-auction-abi.json");
abigen!(MyAsset, "tests/artifacts/asset/out/debug/asset-abi.json");

struct Metadata {
    asset: Option<MyAsset>, // Notice the Rust's `Option`.
    auction: EnglishAuction,
    wallet: LocalWallet,
}
```

This will lead to this error:

<img width="687" alt="image" src="https://user-images.githubusercontent.com/3484025/174695357-ac596735-ec3e-48e0-96de-fd24b5afda69.png">

This means that the Rust compiler is using the `abigen!`-generated `Option`, which doesn't take a generic param `<u64>`. But the user here really wants to use Rust's `Option`. 

To solve this, you have to do this:

```Rust
struct Metadata {
    asset: core::option::Option<MyAsset>,
    auction: EnglishAuction,
    wallet: LocalWallet,
}
```

[...]

PR #574, opened 2022-09-13
feat: add options and results support

hal3e:
Closes: https://github.com/FuelLabs/fuels-rs/issues/421, https://github.com/FuelLabs/fuels-rs/issues/420, https://github.com/FuelLabs/fuels-rs/issues/415

This PR adds support for native rust `Options` and `Results` inside the SDK. We can now work with Sway contracts like:

```rust
struct TestStruct {
    option: Option<Address>
}

enum TestEnum {
    EnumOption: Option<Address>
}

abi MyContract {
    fn get_some_u64() -> Option<u64>;
    fn get_some_address() -> Option<Address>;
    fn get_some_struct() -> Option<TestStruct>;
    fn get_some_enum() -> Option<TestEnum>;
    fn get_some_tuple() -> Option<(TestStruct, TestEnum)>;
    fn get_none() -> Option<Address>;
    fn input_primitive(s: Option<u64>) -> bool;
    fn input_struct(s: Option<TestStruct>) -> bool;
    fn input_enum(e: Option<TestEnum>) -> bool;
    fn input_none(none: Option<Address>) -> bool;
}
```
and

```rust
struct TestStruct {
    option: Option<Address>
}

enum TestEnum {
    EnumOption: Option<Address>
}

pub enum TestError {
    NoAddress: str[5],
}

abi MyContract {
    fn get_ok_u64() -> Result<u64, TestError>;
    fn get_ok_address() -> Result<Address, TestError>;
    fn get_ok_struct() -> Result<TestStruct, TestError>;
    fn get_ok_enum() -> Result<TestEnum, TestError>;
    fn get_ok_tuple() -> Result<(TestStruct, TestEnum), TestError>;
    fn get_error() -> Result<Address, TestError>;
    fn input_ok(ok_address: Result<Address, TestError>) -> bool;
    fn input_error(test_error: Result<Address, TestError>) -> bool;
}
```

### DIFF
--- a/packages/fuels-core/src/code_gen/abigen.rs
+++ b/packages/fuels-core/src/code_gen/abigen.rs
@@ -1,9 +1,6 @@
 use std::collections::HashMap;
 
 use crate::code_gen::bindings::ContractBindings;
-use crate::constants::{
-    ADDRESS_NATIVE_TYPE, CONTRACT_ID_NATIVE_TYPE, OPTION_NATIVE_TYPE, RESULT_NATIVE_TYPE,
-};
 use crate::source::Source;
 use crate::utils::ident;
 use fuels_types::errors::Error;
@@ -216,16 +213,22 @@ impl Abigen {
         Ok(structs)
     }
 
-    // Checks whether the given type field is a Native type.
+    // Checks whether the given type field is a native type.
     // It's expected to come in as `"struct T"` or `"enum T"`.
-    // `T` is a VM or Rust native type if it matches exactly one of
+    // `T` is a native `high-level language` or Rust type if it matches exactly one of
     // the reserved strings, such as "Address", "ContractId", "Option" or "Result"
     pub fn is_native_type(type_field: &str) -> bool {
+        const CONTRACT_ID_NATIVE_TYPE: &str = "ContractId";
+        const ADDRESS_NATIVE_TYPE: &str = "Address";
+        const OPTION_NATIVE_TYPE: &str = "Option";
+        const RESULT_NATIVE_TYPE: &str = "Result";
+
         let split: Vec<&str> = type_field.split_whitespace().collect();
 
         if split.len() > 2 {
             return false;
         }
+
         split[1] == CONTRACT_ID_NATIVE_TYPE
             || split[1] == ADDRESS_NATIVE_TYPE
             || split[1] == OPTION_NATIVE_TYPE
--- a/packages/fuels-core/src/code_gen/abigen.rs
+++ b/packages/fuels-core/src/code_gen/abigen.rs
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 
 use crate::code_gen::bindings::ContractBindings;
-use crate::constants::{ADDRESS_SWAY_NATIVE_TYPE, CONTRACT_ID_SWAY_NATIVE_TYPE};
+use crate::constants::{
+    ADDRESS_NATIVE_TYPE, CONTRACT_ID_NATIVE_TYPE, OPTION_NATIVE_TYPE, RESULT_NATIVE_TYPE,
+};
 use crate::source::Source;
 use crate::utils::ident;
 use fuels_types::errors::Error;
@@ -198,10 +200,10 @@ impl Abigen {
                 continue;
             }
 
-            // Skip custom type generation if the custom type is a Sway-native type.
-            // This means ABI methods receiving or returning a Sway-native type
+            // Skip custom type generation if the custom type is a native type.
+            // This means ABI methods receiving or returning a native type
             // can receive or return that native type directly.
-            if Abigen::is_sway_native_type(&prop.type_field) {
+            if Abigen::is_native_type(&prop.type_field) {
                 continue;
             }
 
@@ -214,17 +216,20 @@ impl Abigen {
         Ok(structs)
     }
 
-    // Checks whether the given type field is a Sway-native type.
+    // Checks whether the given type field is a Native type.
     // It's expected to come in as `"struct T"` or `"enum T"`.
-    // `T` is a Sway-native type if it matches exactly one of
-    // the reserved strings, such as "Address" or "ContractId".
-    fn is_sway_native_type(type_field: &str) -> bool {
+    // `T` is a VM or Rust native type if it matches exactly one of
+    // the reserved strings, such as "Address", "ContractId", "Option" or "Result"
+    pub fn is_native_type(type_field: &str) -> bool {
         let split: Vec<&str> = type_field.split_whitespace().collect();
 
         if split.len() > 2 {
             return false;
         }
-        split[1] == CONTRACT_ID_SWAY_NATIVE_TYPE || split[1] == ADDRESS_SWAY_NATIVE_TYPE
+        split[1] == CONTRACT_ID_NATIVE_TYPE
+            || split[1] == ADDRESS_NATIVE_TYPE
+            || split[1] == OPTION_NATIVE_TYPE
+            || split[1] == RESULT_NATIVE_TYPE
     }
 
     fn abi_enums(&self) -> Result<TokenStream, Error> {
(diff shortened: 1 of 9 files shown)


Salka1988 (approved):
LGTM

packages/fuels-core/src/lib.rs
```diff
@@ -367,6 +367,77 @@ impl Tokenizable for fuel_tx::AssetId {
     }
 }
 
+impl<T> Tokenizable for Option<T>
+where
+    T: Parameterize + Tokenizable,
+{
+    fn from_token(token: Token) -> Result<Self, Error> {
```
digorithm, line 374: This is such a simple and brilliant way of solving the `abigen` for `Option`/`Result` 🔥

digorithm (changes requested):
This is _so_ good. What a clean approach, really good job! Just left a tiny nit w.r.t. error handling. Feel free to merge after that.

packages/fuels-core/src/lib.rs
```diff
@@ -427,6 +498,29 @@ impl Parameterize for u64 {
     }
 }
 
+impl<T> Parameterize for Option<T>
+where
+    T: Parameterize + Tokenizable,
+{
+    fn param_type() -> ParamType {
+        let types = vec![ParamType::Unit, T::param_type()];
+        let variants = EnumVariants::new(types).expect("should never happen");
```
segfault-magnet, line 507: I'd explain in the expect message that since `!types.is_empty()` holds and `EnumVariants::new` fails only if it had no variants, this should never happen.

packages/fuels-types/src/lib.rs
```diff
@@ -17,6 +17,8 @@ pub mod utils;
 // Since it would have the format `struct foo` or `enum bar`, there is a whitespace.
 pub const STRUCT_KEYWORD: &str = "struct ";
```
segfault-magnet, line 18: Although these were once used elsewhere, none of the four now have a usage outside the one place inside this file. 

I'd inline these constants to their only place of usage and, in doing so, clarify their meaning and make the code more readable.

packages/fuels/tests/harness.rs
```diff
@@ -3847,3 +3847,195 @@ async fn test_gas_forwarded_defaults_to_tx_limit() -> Result<(), Error> {
...
+    let expected_none = None;
+    let response = contract_instance.input_none(expected_none).call().await?;
+
+    assert!(response.value);
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_rust_result_can_be_decoded() -> Result<(), Error> {
```
segfault-magnet, line 3957: I find myself joining decoding and encoding into one test case, seeing as these are e2e and it is less of a problem to do so. 

A sway function that accepts the argument validates its content, and returns the same argument but constructed by sway code so as not to have the bytecode just return your original encoded bytes right back at you.

But that is up to you. With the new macro, these are going to get a lot shorter, then we wouldn't pay the penalty of test code size but only of the time spent executing -- which still hasn't grown unwieldy -- so I'm ok if it remains the way it is.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -1,7 +1,9 @@
 use std::collections::HashMap;
 
 use crate::code_gen::bindings::ContractBindings;
-use crate::constants::{ADDRESS_SWAY_NATIVE_TYPE, CONTRACT_ID_SWAY_NATIVE_TYPE};
+use crate::constants::{
+    ADDRESS_NATIVE_TYPE, CONTRACT_ID_NATIVE_TYPE, OPTION_NATIVE_TYPE, RESULT_NATIVE_TYPE,
```
segfault-magnet, line 5: I don't find much benefit in constants that are likely to be used in only one place. Inlining them helps with clarity and if the time comes to change them, you only change them at the one place they're used.

packages/fuels/tests/harness.rs
hal3e: I have checked the generics examples you did. I like them! However, I would keep these separate as the function names directly describe what is done and the code is a bit _clearer_

packages/fuels/tests/test_projects/options/src/main.sw
```diff
@@ -0,0 +1,90 @@
...
+        if let Option::Some(u) = input {
+            return u == 36;
+        }
+        false
+    }
+
+    fn input_struct(input: Option<TestStruct>) -> bool {
+        if let Option::Some(s) = input {
+            if let Option::Some(a) = s.option {
+                return a == ~Address::from(ZERO_B256);
```
iqdecay, line 67: I think this test should be with an address different from the 0 address, just as a sanity check.

packages/fuels/tests/test_projects/options/src/main.sw
```diff
@@ -0,0 +1,90 @@
...
+            }
+        }
+        false
+    }
+
+    fn input_enum(input: Option<TestEnum>) -> bool {
+        if let Option::Some(test_enum) = input {
+            if let TestEnum::EnumOption(option) = test_enum {
+                if let Option::Some(a) = option {
+                    return a == ~Address::from(ZERO_B256);
```
iqdecay, line 77: same remark here

iqdecay (changes requested):
Supeeer clean, I love it! Some questions around naming (as usual with me lol), and also I'm always wary of testing things only with the zero values, which might act as default and thus would recommend changing for non-zero values.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -214,17 +213,26 @@ impl Abigen {
...
 
-    // Checks whether the given type field is a Sway-native type.
+    // Checks whether the given type field is a native type.
     // It's expected to come in as `"struct T"` or `"enum T"`.
-    // `T` is a Sway-native type if it matches exactly one of
-    // the reserved strings, such as "Address" or "ContractId".
-    fn is_sway_native_type(type_field: &str) -> bool {
+    // `T` is a native `high-level language` or Rust type if it matches exactly one of
+    // the reserved strings, such as "Address", "ContractId", "Option" or "Result"
+    pub fn is_native_type(type_field: &str) -> bool {
```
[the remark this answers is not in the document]
hal3e: There is a issue to remove all sway wording in the sdk. So `is_sway_native_type` is not an option IMO. I am open for other suggestions.

packages/fuels-types/src/lib.rs
```diff
@@ -13,11 +13,6 @@ pub mod param_types;
 pub mod parse_param;
 pub mod utils;
 
-// Both those constants are used to determine if a type field represents an `Enum` or a `Struct`.
-// Since it would have the format `struct foo` or `enum bar`, there is a whitespace.
-pub const STRUCT_KEYWORD: &str = "struct ";
-pub const ENUM_KEYWORD: &str = "enum ";
-
```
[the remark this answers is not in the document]
hal3e: I actually agree with @segfault-magnet on this. IMO it is better now. We only use it once and use it exactly where they are needed.

packages/fuels/tests/test_projects/options/src/main.sw
hal3e: We are testing that in other parts of the code but I agree. For sanity let's change it to something :smile:

packages/fuels-core/src/code_gen/abigen.rs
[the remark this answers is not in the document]
iqdecay: would "fuel-native" types work?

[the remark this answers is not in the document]
digorithm: As @hal3e said, we're removing Sway wording; the SDK is an API to the VM and should be language agnostic. 

In a way, these keywords (option, result, etc) are also native to the VM. So I'm ok with the function being called as it is 🙂

iqdecay (approved):
Agreed with all answers, let's :ship: it!

digorithm (approved):

PR #575, opened 2022-09-13
Replace script_with_data_offset! with GTF instructions

Salka1988:


packages/fuels-signers/src/wallet.rs
```diff
@@ -312,32 +313,16 @@ impl Wallet {
         .flatten()
         .collect();
 
-        // This script loads:
-        //  - a pointer to the contract id,
-        //  - the actual amount
-        //  - a pointer to the asset id
-        // into the registers 0X10, 0x11, 0x12
-        // and calls the TR instruction
```
Voxelot, line 320: Are these comments still relevant? Why remove them?

Salka1988: You're right, I forgot to return them.

digorithm (approved):
Nice, much cleaner. LGTM!

Issue #576, opened 2022-09-13
Add StringToken import to WASM


tjsharp1:
ABIGen for WASM has a compile error because StringToken is missing.

PR #577, opened 2022-09-13
Add StringToken import to WASM

tjsharp1:
Was getting compile errors while upgrading to the latest SDK. StringToken import was missing. #576


segfault-magnet (commented):
Can you add an e2e test in the wasm harness which reproduces this issue? 

If the new generics get merged I think StringTokens might not appear at all in the generated code. Would be nice to not have a regression once we remove it.

PR #578, opened 2022-09-16
refactor: expose the default derivation path to downstream crates

### LINKED ISSUE #551: Expose the default derivation path to downstream crates
Currently downstream creates, which need to use the prefix of the default derivation path (like `forc-wallet`) are duplicating it. But we could expose the default derivation path from SDK (currently it is located [here](https://github.com/FuelLabs/fuels-rs/blob/ac2dc6eca8e1cc7db3d09c517cffea6d3c55d272/packages/fuels-signers/src/wallet.rs#L21)) so that it can be reused. Additionally, we may want to change the prefix to something like `m/44'/1179993420'/` so that it is a little more reusable.

Related to https://github.com/FuelLabs/forc-wallet/issues/28#issuecomment-1231181032

kayagokalp:
closes #551.

### DIFF
--- a/packages/fuels-signers/src/wallet.rs
+++ b/packages/fuels-signers/src/wallet.rs
@@ -22,7 +22,7 @@ use rand::{CryptoRng, Rng};
 use std::{collections::HashMap, fmt, ops, path::Path};
 use thiserror::Error;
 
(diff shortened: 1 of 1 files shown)


iqdecay (approved):
:ship:

kayagokalp:
Oh just realized single approval is enough here sorry for pinging again!

PR #582, opened 2022-09-20
Parse logged types

### LINKED ISSUE #562: Parse logged types
Once https://github.com/FuelLabs/sway/pull/2689 is merged, it will enable the SDK to parse concrete types logged in Sway programs. Which is something we've been long waiting for! 

The SDK should support this by parsing the `loggedTypes` arrays that will show up in the JSON ABI file. 

From the above PR, the relevant points that are going to be needed to implement this support in the SDK:

> Update the JSON ABI generation to also include a loggedTypes array. This is an array of JsonLoggedType which contains a unique ID per log as well as a JsonTypeApplication. Types that are logged are added to the types array which is shared with the contract functions.
> Introduce an IR instruction log which takes a value to log, its type, and an ID (has to be u64). This ID is basically the ID of the log that is emitted in the JSON file

MujkicA:
Closes #562 

This PR adds `_logs_with_type::<T>()` to the existing contract instance methods. It requires receipts and returns a `Vec<T>` containing all the logged values of type `T`.
It also adds `_print_logs()` for outputting all the logged values to stdout.

The feature flag `--generate-logged-types` is added to the build test script so that we can test, but needs to be removed as soon as forc starts generating logged types without it.
 
As mentioned in the issue, ABI support for logging generic types is limited. However, it is likely that once added, it will work on the SDK side with little to no changes.
Another future upgrade we can expect is to further filter the logs based on the name of the logged variables once they are included in the ABI. 

I will add the follow-up issues once we agree on the changes.


packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -122,7 +128,48 @@ impl Abigen {
                            let provider = self.wallet.get_provider()?;
                            wallet.set_provider(provider.clone());
 
-                           Ok(Self { contract_id: self.contract_id.clone(), wallet: wallet })
+                           Ok(Self { contract_id: self.contract_id.clone(), wallet: wallet, logs_lookup: self.logs_lookup.clone() })
+                        }
+
+                        pub fn _logs_with_type<D: Tokenizable + Parameterize>(&self, receipts: &[Receipt]) -> Result<Vec<D>, SDKError> {
+                            let param_type = D::param_type();
```
segfault-magnet, line 135: I'd extract everything from this method into a separate function in our codebase and delegate this call to it.

For example:
```rust
// in fuel-core for instance
pub fn something_ahmedm<T: Parameterize + Tokenizable>(
    logs_lookup: &[ResolvedLog],
    receipts: &[Receipt],
) -> Result<Vec<T>, Error> { // ...

// in the abigen
pub fn _logs_with_type<D: Tokenizable + Parameterize>(&self, receipts: &[Receipt]) -> Result<Vec<D>, SDKError> {
          something_ahmedm(&self.logs_lookup, receipts)
}
```

Because:
1. You'll get IDE support since this isn't part of a `quote!{}`
2. You can now unit test this logic.
3. If our clients decide to `abigen!` once and dump the result to a file, we'll still have control over the implementation, and porting bug fixes should be as easy as updating the SDK version.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -122,7 +128,48 @@ impl Abigen {
...
+                                .filter_map(|l| {
+                                    if l.param_type == param_type {
+                                        Some(l.log_id)
+                                    } else {
+                                        None
+                                    }
+                                })
+                                .collect();
+
+                            let decode_log_content = |content: &[u8]| {
```
segfault-magnet, line 149: there is a function in `fuel-core` that does this. It is called `try_from_bytes`. Wherever this closure is called, that function can be called instead. It is already in the set of `use`s added to the `abigen`ed code.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -122,7 +128,48 @@ impl Abigen {
...
+                                let dec =
+                                    ABIDecoder::decode_single(&D::param_type(), content)?;
+                                D::from_token(dec)
+                            };
+
+                            let decoded_logs: Vec<D> = receipts
```
segfault-magnet, line 155: I'd risk needless allocations and try and make this more readable:
```rust
    let decoded_logs = receipts
        .iter()
        .filter_map(|r| match r {
            Receipt::LogData { rb, data, .. } if target_ids.contains(rb) => Some(data.clone()),
            Receipt::Log { ra, rb, .. } if target_ids.contains(rb) => {
                Some(ra.to_be_bytes().to_vec())
            }
            _ => None,
        })
        .map(|data| try_from_bytes(&data))
        .collect::<Result<Vec<_>, _>>()?;
```

Doing an `except` rather than handling `Result`s here is probably better, failing fast and all that... 

But then again it is hard to decide for our users what should kill their app.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -260,6 +312,32 @@ impl Abigen {
     pub fn get_types(abi: &ProgramABI) -> HashMap<usize, TypeDeclaration> {
         abi.types.iter().map(|t| (t.type_id, t.clone())).collect()
     }
+
+    pub fn resolve_logs(&self) -> Result<Vec<TokenStream>, Error> {
```
segfault-magnet, line 316: This doesn't seem to return Result::Err anywhere. 

Maybe return just a `Vec<TokenStream>` ?

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -260,6 +312,32 @@ impl Abigen {
...
+                        let id = l.log_id;
+
+                        quote! { ResolvedLog {
+                                log_id: #id,
+                                param_type: #param_type_call
+                            }
+                        }
+                    })
+                    .collect()
+            });
```
segfault-magnet, line 337: I find `flattening` options of containers more pleasable than handling the collection separately, YMMV:
```rust
        self.abi
            .logged_types
            .as_ref()
            .into_iter()
            .flatten()
            .map(|logged_type| {
                let resolved_type = resolve_type(&logged_type.application, &self.types)
                    .expect("Failed to resolve log type");
                let param_type_call = single_param_type_call(&resolved_type);
                let id = logged_type.log_id;

                quote! { ResolvedLog {
                        log_id: #id,
                        param_type: #param_type_call
                    }
                }
            })
            .collect()
```

packages/fuels-core/src/code_gen/custom_types/utils.rs
```diff
@@ -145,22 +145,27 @@ pub fn extract_custom_type_name_from_abi_property(prop: &TypeDeclaration) -> Res
...
-                .iter()
-                .map(TokenStream::from)
-                .collect::<Vec<_>>();
-            if parameters.is_empty() {
-                quote! { <#type_name>::param_type() }
-            } else {
-                quote! { #type_name::<#(#parameters),*>::param_type() }
-            }
-        })
+        .map(|Component { field_type, .. }| single_param_type_call(field_type))
```
segfault-magnet, line 148: I would have called it `param_type_call` singular. 

But then again those plural 's'-es can kind of hide in plain sight, so this is probably more in your face can't mistake it for the other one.

packages/fuels/tests/harness.rs
```diff
@@ -3671,6 +3632,7 @@ async fn test_rust_result_can_be_decoded() -> Result<(), Box<dyn std::error::Err
     Ok(())
 }
 
+/*
```
segfault-magnet, line 3635: Isn't this a bit early to open the comment block? I think you caught a bit of the tests for `Result`s usage in sway contracts?

packages/fuels/tests/harness.rs
```diff
@@ -3708,3 +3670,91 @@ async fn test_rust_result_can_be_encoded() -> Result<(), Box<dyn std::error::Err
 
     Ok(())
 }
+
+#[tokio::test]
```
segfault-magnet, line 3674: What do you think about temporarily adding the flag into our CI and uncommenting these tests?

But then again SDK devs would need to remember to compile the sway projects with the flag when developing. Perhaps if you add it to the 'rust-build-sway-projects` script and in the readme as well? We can remove it as soon as this is not hidden behind a feature flag.

packages/fuels-core/src/code_gen/custom_types/utils.rs
MujkicA: Yup, that was my thought process too

packages/fuels-core/src/code_gen/abigen.rs
MujkicA: Nice. I was trying to think of a way to make this look better, guess I was to stingy with the allocations.

packages/fuels/tests/harness.rs
MujkicA: Not opposed to it if the team agrees.

Salka1988: :+1:

packages/fuels/tests/test_projects/logged_types/src/main.sw
```diff
@@ -0,0 +1,62 @@
...
+        
+        let f: u64 = 64;
+        let u: b256 = 0xef86afa9696cf0dc6385e2c407a6e159a1103cefb7e2ae0636fb33d3cb2a9e4a;
+        let e: str[4] = "Fuel";
+        let l: [u8; 3] = [1u8, 2u8, 3u8];
+
+        __log(f);
+        __log(u);
+        __log(e);
+        __log(l);
```
hal3e, line 44: nice variable names :smile:

hal3e (approved):
Good work!

segfault-magnet (dismissed):

Salka1988 (commented):
Nice!LGTM!

packages/fuels/tests/harness.rs
```diff
@@ -3721,6 +3683,109 @@ async fn test_rust_result_can_be_encoded() -> Result<(), Box<dyn std::error::Err
...
+async fn test_print_logs() -> Result<(), Error> {
+    setup_contract_test!(
+        contract_instance,
+        wallet,
+        "packages/fuels/tests/test_projects/logged_types"
+    );
+
+    // ANCHOR: print_logs
+    let response = contract_instance.produce_multiple_logs().call().await?;
+    contract_instance._print_logs(&response.receipts);
```
segfault-magnet, line 3783: Is the `_print_logs` result deterministic? 

Besides not erroring out, that it actually prints something might be a reasonable assertion.

I'd give them a string and let them do whatever with it rather than print it for them.

Useless if the output is anything other than `stdout`. Besides should it be `stdout` or `stderr`... can't reasonably say as you don't know how the user is going to use it -- as a defacto useful program output or as a diagnostic message.

I'd give them a String :D

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -262,6 +286,131 @@ impl Abigen {
...
+            ids_with_data
+            .iter()
+            .for_each(|(id, data)|{
+                let param_type = id_to_param_type.get(id).expect("Failed to find log id.");
+
+                if false {
+                    //this serves as a placeholder so that the proc
+                    //macro can generate the branches as "else if..."
+                }
+                #(#branches)*
```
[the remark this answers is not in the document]
MujkicA: Nice! Didn't think of that.

packages/fuels/tests/harness.rs
MujkicA: Yup, a String is more reasonable than printing it out for them.

mohammadfawaz (approved):
Very nice. Thanks! Once generics are supported, I will be able to remove the flag completely.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -254,6 +275,141 @@ impl Abigen {
...
+                    param_type_call,
+                    resolved_type_name,
+                }
+            })
+            .collect()
+    }
+}
+
+pub fn generate_print_logs(resolved_logs: &[ResolvedLog]) -> TokenStream {
+    // if not logs are present, return an empty string vec
```
hal3e, line 304:
```suggestion
    // if logs are not present, return an empty string vec
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -254,6 +275,141 @@ impl Abigen {
...
+    if resolved_logs.is_empty() {
+        return quote! {
+            pub fn _print_logs(&self, receipts: &[Receipt]) -> Vec<String> {
```
hal3e, line 307:
```suggestion
            pub fn print_logs(&self, receipts: &[Receipt]) -> Vec<String> {
```

docs/src/calling-contracts/logs.md
```diff
@@ -0,0 +1,25 @@
...
+{{#include ../../../packages/fuels/tests/harness.rs:produce_logs}}
+```
+
+You can also get a vector of all the logged values as strings using `_print_logs()`:
+
+```rust, ignore
+{{#include ../../../packages/fuels/tests/harness.rs:print_logs}}
+```
+
+Due to possible performance hits, it is not recommended to use `_print_logs()` outside of a debugging scenario.
```
hal3e, line 23:
```suggestion
Due to possible performance hits, it is not recommended to use `print_logs()` outside of a debugging scenario.
```

docs/src/calling-contracts/logs.md
```diff
@@ -0,0 +1,25 @@
...
+
+Whenever you log a value within a contract method, the resulting log entry is added to the log receipt and the variable type is recorded in the contract's ABI. The SDK lets you parse those values into Rust types.
+
+Consider the following contract method:
+
+```rust,ignore
+{{#include ../../../packages/fuels/tests/test_projects/logged_types/src/main.sw:produce_logs}}
+```
+
+You can access the logged values in Rust by calling `_logs_with_type::<T>` from a contract instance, where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
```
hal3e, line 11:
```suggestion
You can access the logged values in Rust by calling `logs_with_type::<T>` from a contract instance, where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -127,13 +141,21 @@ impl Abigen {
...
                         }
 
+                        #print_logs
+
+                        pub fn methods(&self) -> #methods_name {
+                            #methods_name {
+                                contract_id: self.contract_id.clone(),
+                                wallet: self.wallet.clone(),
+                            }
+                        }
```
hal3e, line 158: cosmetics :smile: I would put `#print_logs` at the end.
```suggestion
                        pub fn methods(&self) -> #methods_name {
                            #methods_name {
                                contract_id: self.contract_id.clone(),
                                wallet: self.wallet.clone(),
                            }
                        }

                        #print_logs                        
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -254,6 +275,141 @@ impl Abigen {
     pub fn get_types(abi: &ProgramABI) -> HashMap<usize, TypeDeclaration> {
         abi.types.iter().map(|t| (t.type_id, t.clone())).collect()
     }
+
+    /// Reads the parsed logged types from the ABI and creates
+    /// ResolvedLogs
```
hal3e, line 280:
```suggestion
    /// Reads the parsed logged types from the ABI and creates ResolvedLogs
```

hal3e (commented):
Great work!

packages/fuels-core/src/code_gen/abigen.rs
hal3e: I think this passed because we are not using it anywhere. Maybe it would make sense to add it to a test ?

docs/src/calling-contracts/logs.md
```diff
@@ -1,25 +1,25 @@
...
 
-You can access the logged values in Rust by calling `_logs_with_type::<T>` from a contract instance, where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
+You can access the logged values in Rust by calling `logs_with_type::<T>` from a contract instance, where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
 
 ```rust,ignore
 {{#include ../../../packages/fuels/tests/harness.rs:produce_logs}}
 ```
 
-You can also print all the logged values to stdout using `_print_logs()` :
+You can also get a vector of all the logged values as strings using `_print_logs()`:
```
segfault-magnet, line 17:
```suggestion
You can also get a vector of all the logged values as strings using `print_logs()`:
```
I see you're renaming these all over the place. Perhaps you missed this one?

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -315,10 +300,19 @@ impl Abigen {
 }
 
 pub fn generate_print_logs(resolved_logs: &[ResolvedLog]) -> TokenStream {
+    // if logs are not present, return an empty string vec
+    if resolved_logs.is_empty() {
+        return quote! {
+            pub fn print_logs(&self, receipts: &[Receipt]) -> Vec<String> {
```
segfault-magnet, line 306: Can we dedupe this fn signature? 

Perhaps as a closure:
```rust
let generate_method = |body: TokenStream| {
quote! {
pub fn print_logs(&self, receipts: &[Receipt]) -> Vec<String> {
 #body
}
}

if resolved_logs.is_empty() {
 return generate_method(quote!{vec![]});
}
// ... 

Sry for the formatting. Typed it off the top of my head, didn't check it.
```

segfault-magnet (dismissed):

packages/fuels/tests/harness.rs
```diff
@@ -3696,6 +3660,154 @@ async fn test_rust_result_can_be_encoded() -> Result<(), Box<dyn std::error::Err
...
+
+    let contract_methods = contract_instance.methods();
+    let response = contract_methods.initialize_counter(42).call().await?;
+    let logs = contract_instance.print_logs(&response.receipts);
+
+    assert!(logs.is_empty());
+
+    Ok(())
+}
+
```
digorithm, line 3810: Excellent tests!

digorithm (changes requested):
This looks amazing. I have one issue with the semantics, though. Whenever I call something with `print` in it, I expect it to automatically print something to stdout/stderr. 

`print_logs()` doesn't print anything; it returns a `Vec<String>`. The name/behavior is misleading. 

I propose we rename `print_logs()` to `fetch_logs()`/`get_logs()`/`resolve_logs()` and keep the same behavior (return a `Vec<String>` and let the user decide what to do with it.)

digorithm (approved):
Looking good :shipit:

PR #583, opened 2022-09-20
add script interface

Salka1988:


packages/fuels/tests/harness.rs
```diff
@@ -3716,3 +3717,17 @@ async fn test_rust_result_can_be_encoded() -> Result<(), Box<dyn std::error::Err
...
+async fn test_script_interface() -> Result<(), Error> {
+    let path_to_bin = "../fuels/tests/test_projects/logging/out/debug/logging.bin";
+
+    let return_val =
+        run_compiled_script(path_to_bin, TxParameters::new(None, Some(2500), None)).await?;
+
+    let correct_hex =
+        hex::decode("ef86afa9696cf0dc6385e2c407a6e159a1103cefb7e2ae0636fb33d3cb2a9e4a");
+
+    assert_eq!(correct_hex?, return_val[0].data().unwrap());
```
hal3e, line 3731:
```suggestion
        hex::decode("ef86afa9696cf0dc6385e2c407a6e159a1103cefb7e2ae0636fb33d3cb2a9e4a")?;

    assert_eq!(correct_hex, return_val[0].data().unwrap());
```

hal3e (commented):
LGTM! Left a tiny nit.

packages/fuels/tests/harness.rs
Salka1988: I deleted this test from harness.rs, but within the script.rs tests I adopted your suggestions. Thanx!

Br1ght0ne (approved):

packages/fuels-contract/src/script.rs
```diff
@@ -46,6 +47,8 @@ impl Script {
         Self { tx }
     }
 
+    //builder pattern -> data & opcode, set_inputs and outputs, call after build.
+
```
iqdecay, line 51: Reformulate, maybe use a sentence with an example as we do in other places

packages/fuels-contract/src/script.rs
```diff
@@ -812,3 +815,110 @@ mod test {
...
+            receipts_root: Default::default(),
+            script: vec![],
+            script_data: vec![],
+            inputs: vec![],
+            outputs: vec![],
+            witnesses: vec![],
+            metadata: None,
+            asset_id: AssetId::default(),
+            amount: 0,
+        }
```
iqdecay, line 850: I _think_ this can be replaced with an `impl Default`?

packages/fuels-signers/src/wallet.rs
```diff
@@ -322,7 +322,7 @@ impl Wallet {
         let script = vec![
             Opcode::gtf(0x10, 0x00, GTFArgs::ScriptData),
             Opcode::ADDI(0x11, 0x10, ContractId::LEN as u16),
-            Opcode::LW(0x13, 0x11, 0),
```
iqdecay, line 325: Document the change for the LW opcode

packages/fuels-test-helpers/src/script.rs
```diff
@@ -1,48 +1,53 @@
...
 
 #[allow(dead_code)]
 #[cfg(not(feature = "fuel-core-lib"))]
-pub async fn run_compiled_script(binary_filepath: &str) -> Result<Vec<Receipt>, Error> {
+pub async fn run_compiled_script(
+    binary_filepath: &str,
+    tx_params: TxParameters,
+) -> Result<Vec<Receipt>, Error> {
     let script_binary = std::fs::read(binary_filepath)?;
+    let server = FuelService::new_node(Config::local_node()).await.unwrap();
```
iqdecay, line 38: I think `local_node` would be more appropriate :)

packages/fuels-test-helpers/src/script.rs
```diff
@@ -1,48 +1,53 @@
...
-
-    let script = get_script(script_binary);
+    let script = get_script(script_binary, tx_params);
 
-    script.call(provider).await
+    script.call(&provider).await
 }
 
-fn get_script(script_binary: Vec<u8>) -> Script {
+fn get_script(script_binary: Vec<u8>, tx_params: TxParameters) -> Script {
```
iqdecay, line 46: I think naming this `build_script` would be better. Indeed, what you're doing is bulding a `Script` object from both the script binary and some tx parameters, that we will on a per-tx basis.

packages/fuels/tests/harness.rs
```diff
@@ -3716,3 +3720,88 @@ async fn test_rust_result_can_be_encoded() -> Result<(), Box<dyn std::error::Err
...
+    .flatten()
+    .collect();
+
+    let script = vec![
+        Opcode::gtf(0x10, 0x00, GTFArgs::ScriptData),
+        Opcode::ADDI(0x11, 0x10, ContractId::LEN as u16),
+        Opcode::LW(0x11, 0x11, 0),
+        Opcode::ADDI(0x12, 0x11, WORD_SIZE as u16),
+        Opcode::TR(0x10, 0x11, 0x12),
+        Opcode::RET(REG_ONE),
```
iqdecay, line 3781: I would be nice to document what this script does

packages/fuels/tests/harness.rs
```diff
@@ -3716,3 +3720,88 @@ async fn test_rust_result_can_be_encoded() -> Result<(), Box<dyn std::error::Err
...
+        .set_maturity(tx_parameters.maturity)
+        .set_script(script)
+        .set_script_data(script_data)
+        .set_inputs(inputs.to_vec())
+        .set_outputs(outputs.to_vec())
+        .set_amount(amount)
+        .build(&wallet)
+        .await?
+        .call(wallet.get_provider()?)
+        .await?;
```
iqdecay, line 3798: I think this test is not "hard" enough: for instance I would like to see that providing a `gas_price` or `gas_limit` too low makes it fail. But maybe I'm being paranoid since they are essentially wrappers around other, well-tested functions.

iqdecay (changes requested):
Good work! This is a great first step for me to work on accepting arguments in `main` script functions. It would be better with some clarifiacations though :slightly_smiling_face:

packages/fuels-contract/src/script.rs
Salka1988: We can do it, but I don't think we need it.

packages/fuels-contract/src/script.rs
Salka1988: This was plan i forgot to delete!

packages/fuels-test-helpers/src/script.rs
Salka1988: It's always `let server` or `let srv`, so I think it would be bad to change the name.

packages/fuels-test-helpers/src/script.rs
Salka1988: Yup, ill change that!

packages/fuels/tests/harness.rs
Salka1988: It is already documented in wallet.rs. Is there a need to document in the tests.

packages/fuels/tests/harness.rs
Salka1988: This is a test to test if everything works as it should. I have already tested the situations you mention. They can be added if there is a demand.

packages/fuels-signers/src/wallet.rs
Salka1988: I don't think there's any need. I only changed the registry and the script remained the same.

iqdecay: Oh yes you're right, the documentation was "wrong" before, we didn't explain that we loaded inside 0x13.

iqdecay (approved):
LGTM!

hal3e (approved):
Good work! Thisi is much better than the first version I reviewed! :tada:

Issue #257, opened 2022-05-05
Remove `sway` wording


adlerjohn:
The SDK should operate independently of the high-level language. Any references to `sway` should be removed. If functionality is inextricably dependent on Sway specifically, then that functionality should be refactored out of this repo and into the Sway repo in a subsequent issue.

PR #589, opened 2022-09-22
refactor: remove sway wording

hal3e:
closes:https://github.com/FuelLabs/fuels-rs/issues/257

- AFAIK the rust SDK does not depend on any Sway functionality.
- I have tried to remove all the Sway mentions from the code and left the documentation intact. I need the teams opinion on the wording I used :)


packages/fuels-core/src/abi_encoder.rs
```diff
@@ -104,7 +104,7 @@ impl ABIEncoder {
 
         self.encode_discriminant(*discriminant);
 
-        // The sway compiler has an optimization for enums which have only Units as variants -- such
+        // The high-level language compiler has an optimization for enums which have only Units as variants -- such
```
MujkicA, line 107: Consistency

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -216,7 +216,7 @@ impl ABIDecoder {
         selected_variant: &ParamType,
         enum_width: usize,
     ) -> Result<Token, CodecError> {
-        // The sway compiler has an optimization where enums that only contain
+        // The FuelVM has an optimization where enums that only contain
```
MujkicA, line 219: Nit: be consistent about the wording here

MujkicA (approved):

Salka1988 (approved):
Looks good! You've got them all.

packages/fuels-core/src/abi_decoder.rs
hal3e: Do you guys prefer `FuelVM` or `high-level language` ?

packages/fuels-core/src/abi_encoder.rs
hal3e: Good catch! That was me changing stuff and not updating everywhere...

packages/fuels-core/src/abi_decoder.rs
segfault-magnet: What bothers me is that, as far as I understand it, the VM could care less about enums. 

It was the sway compiler that had this optimization. But how do you say that without saying sway :D 

High-level-language compiler has an optimization?

Or just be vague, my favourite, "Enums that contain only units are bla bla" don't say why :)

segfault-magnet (commented):
Soo, I have a good feeling the following is overkill, but just want to get it out there -- 

What do you think about a CI step which does a search of our docs and `.rs` files for sway wording? 

Or we will be careful to not introduce it back :D

hal3e:
> Soo, I have a good feeling the following is overkill, but just want to get it out there --
> 
> What do you think about a CI step which does a search of our docs and `.rs` files for sway wording?
> 
> Or we will be careful to not introduce it back :D

That is definitely an overkill :)

packages/fuels-core/src/abi_decoder.rs
digorithm: That's a good point. In this case, this optimization doesn't happen in the VM. 

I'm ok with what @segfault-magnet said, don't go too deep into why this optimization is done.

packages/fuels-core/src/code_gen.rs
```diff
@@ -1,5 +1,5 @@
 //! This module implements everything related to code generation/expansion
-//! from a fuel/sway ABI.
+//! from a fuel ABI.
```
digorithm, line 2:
```suggestion
//! from a FuelVM ABI.
```
For consistency!

tools/fuels-abi-cli/src/main.rs
```diff
@@ -5,7 +5,7 @@
 // TODO: currently disabled because it needs to be updated to the new JSON ABI.
 
 // #[derive(StructOpt, Debug)]
-// /// Sway/Fuel ABI coder.
+// /// Fuel ABI coder.
```
digorithm, line 8:
```suggestion
// /// FuelVM ABI coder.
```

digorithm (changes requested):
Very good! Just some tiny nits.

iqdecay (dismissed):

iqdecay (approved):

digorithm (approved):

PR #590, opened 2022-09-22
feat: create an `append_message_output` method

### LINKED ISSUE #579: Create an `append_message_output` method that is similar to `append_variable_output`
Similarly to the `append_variable_output` methods that exists today, we need an `append_message_output` methods that is similar but adds a "message output" instead of a "variable output", so that the SMO opcode can be used, via a `send_message()` function in Sway standard library which looks for a message output first.

iqdecay:
Closes #579 by creating an `append_message_output` method that works similarily
to `append_variable_output`.


iqdecay:
@mohammadfawaz I'm not sure I understood the context for this so I didn't include Sway tests as I would usually do, do you have some Sway code to show me where this API could interact?

nfurfaro:
@iqdecay 2 points:
1.) My[ draft PR](https://github.com/FuelLabs/sway/pull/2796) is just waiting for this PR to land and be released so I can test it correctly! 
Similar to how the `tro`(transfer to output) opcode requires that a variable output be appended in order to work, the `smo` opcode (send message output) requires a message output be appended to the call in order to work.

 2.) we should consider if there's any other way to approach testing, as we've already seen that there's a circular dep issue between the stdlib (the tests use the SDK) and the SDK (which has some tests which use sway (and the stdlib).
 It might be that we work on features like this one in tandem, and you can run your tests pointing at a stdlib branch other than `master` where the supporting changes have been made already. Once the SDK changes are released, the stdlib changes can merge to master, and then the SDK tests can be updated to use the `master` stdlib branch...
Open to ideas here.

packages/fuels-contract/src/contract.rs
```diff
@@ -465,13 +467,31 @@ where
...
+    /// Appends `num` `Output::Message`s to the transaction.
+    /// Note that this is a builder method, i.e. use it as a chain:
+    /// `my_contract_instance.my_method(...).add_message_outputs(num).call()`.
+    pub fn append_message_outputs(mut self, num: u64) -> Self {
+        let new_message_outputs: Vec<Output> = (0..num)
+            .map(|_| Output::Message {
+                recipient: Address::zeroed(),
+                amount: 0,
+            })
+            .collect();
```
segfault-magnet, line 486: Another way, if you like it:
```suggestion

        let new_message_outputs = vec![
            Output::Message {
                recipient: Address::zeroed(),
                amount: 0,
            };
            num as usize
        ];
```

or this:
```rust
        let new_message_outputs = std::iter::repeat(Output::Message {
            recipient: Address::zeroed(),
            amount: 0,
        })
        .take(num as usize)
        .collect();
```

packages/fuels/tests/test_projects/append_message_output/Forc.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+authors = ["lulwat"]
```
hal3e, line 2:
```suggestion
authors = ["Fuel Labs <contact@fuel.sh>"]
```

packages/fuels/tests/test_projects/append_message_output/src/main.sw
```diff
@@ -0,0 +1,14 @@
+contract;
```
hal3e, line 1: I could not find where we use this `contract`. Am I missing something ?

hal3e (changes requested):
Good work! I left a nit and a question..

packages/fuels/tests/test_projects/append_message_output/src/main.sw
iqdecay: No you're not missing anything, I'm going to remove it. e don't need to test Sway opcodes in the SDK (duh), I just got confused haha.

hal3e (approved):

Salka1988 (approved):

Issue #515, opened 2022-08-03
Collision between contract instance- and binding methods


MujkicA:
The original problem only applied to `new` (#277) and was resolved by introducing `[MyContract]Builder` which is used to create contract instances.
The addition #485 and #486 reintroduces the possibility of colliding method names in addition to making the `[MyContract]Builder` fix somewhat obsolete.

PR #592, opened 2022-09-26
refactor!: add methods() to contract abigen

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/515

- removed the `Builder` pattern from abigen
- added a new struct that holds the contract methods which is built by the `contract instance`

Now the workflow looks like this:
```rust
    abigen!(
        SimpleContract,
        "packages/fuels/tests/takes_ints_returns_bool-abi.json",
    );

    // `SimpleContract` is the name of the contract
    let contract_instance = SimpleContract::new(null_contract_id(), wallet);

    let call_handler = contract_instance.methods().takes_ints_returns_bool(42);
```
In the tests where we have multiple contract calls, I have saved the methods in a new variable and called the methods directly. For example:

```rust
    let contract_methods = SimpleContract::new(null_contract_id(), wallet).methods();

    let call_handler = contract_methods.takes_nested_struct(nested_struct_from_tokens);
```

### DIFF
--- a/docs/src/contracts/the-setup-contract-test-macro.md
+++ b/docs/src/contracts/the-setup-contract-test-macro.md
@@ -22,13 +22,13 @@ The input of the macro are the contract instance variable name, wallet variable 
 
 >**Note** The same contract can be deployed several times as the macro deploys the contracts with salt. You can also deploy different contracts to the same provider using a shared wallet.
 
-If you want to deploy contracts to the same provider you have to set the wallet name of the first macro to `wallet` and all the remaining wallet names to `None`. The first macro will create `wallet` and bring it into context and the other macros will use it instead of creating new ones. Let's see it in an example.
+If you want to deploy contracts to the same provider, you have to set the wallet name of the first macro to `wallet` and all the remaining wallet names to `None`. The first macro will create `wallet` and bring it into context, and the other macros will use it instead of creating new ones. Let's see it in an example.
 
 ```rust,ignore
 {{#include ../../../packages/fuels/tests/harness.rs:contract_setup_macro_multi}}
 ```
 
-In this example, three contracts are deploy on the same provider using the `wallet`. The second and third macro use the same contract but they have different IDs because of the deployment with salt. Both of them can call the first contract by using its ID.
+In this example, three contracts are deployed on the same provider using the `wallet`. The second and third macro use the same contract but have different IDs because of the deployment with salt. Both of them can call the first contract by using its ID.
 
 In addition, you can manually create the `wallet` variable and then use it inside the macro. This is useful if you want to create custom wallets or providers, but still want to use the macro to reduce boilerplate code. Below is an example of this approach.
 
--- a/docs/src/contracts/the-setup-contract-test-macro.md
+++ b/docs/src/contracts/the-setup-contract-test-macro.md
@@ -22,9 +22,16 @@ The input of the macro are the contract instance variable name, wallet variable 
 
 >**Note** The same contract can be deployed several times as the macro deploys the contracts with salt. You can also deploy different contracts to the same provider using a shared wallet.
 
-If you want to deploy contract to the same provider you have to set the wallet name of the first macro to `shared_wallet` and all the remaining wallet names to `None`. The first macro will create `shared_wallet` and bring it into context and the other macros will use it instead of creating new ones. Let's see it in an example.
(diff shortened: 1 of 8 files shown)


hal3e:
While I was updating the harness tests I figured that I could also update some test that I did not update when I made the `setup_test_contract!` macro. At that time, I did not update them because they used some custom wallets and/or providers. But then I remembered that if I named the custom wallet `shared_wallet` and declare it before the macro, I can still use the macro with the **_None_** value. But now I did not like the name `shared_wallet` anymore :smile: (Initially, I put `shared_wallet` as it would be shared between macros but that is not the only use case). So I decided to rename `shared_wallet` to just `wallet` and now we can have custom wallets or providers and still use the macro. Here is an example:

```rust
    let config = WalletsConfig::new(Some(2), Some(1), Some(DEFAULT_COIN_AMOUNT));

    let mut wallets = launch_custom_provider_and_get_wallets(config, None).await;
    let wallet = wallets.pop().unwrap();
    let wallet_2 = wallets.pop().unwrap();

    setup_contract_test!(
        contract_instance,
        None,
        "packages/fuels/tests/test_projects/contract_test"
    );
```
Let me know what you think about this :smile:

docs/src/contracts/the-setup-contract-test-macro.md
```diff
@@ -22,9 +22,16 @@ The input of the macro are the contract instance variable name, wallet variable
...
+If you want to deploy contracts to the same provider you have to set the wallet name of the first macro to `wallet` and all the remaining wallet names to `None`. The first macro will create `wallet` and bring it into context and the other macros will use it instead of creating new ones. Let's see it in an example.
 
 ```rust,ignore
 {{#include ../../../packages/fuels/tests/harness.rs:contract_setup_macro_multi}}
 ```
-In this example, three contracts are deploy on the same provider using the `shared_wallet`. The second and third macro use the same contract but they have different IDs because of the deployment with salt. Both of them can call the first contract by using its ID.
+
+In this example, three contracts are deploy on the same provider using the `wallet`. The second and third macro use the same contract but they have different IDs because of the deployment with salt. Both of them can call the first contract by using its ID.
```
Salka1988, line 31:
```suggestion
In this example, three contracts are deployed on the same provider using the `wallet`. The second and third macro use the same contract but they have different IDs because of the deployment with salt. Both of them can call the first contract by using its ID.
```

Salka1988 (changes requested):
Nice! :+1:

docs/src/contracts/the-setup-contract-test-macro.md
MujkicA, line 31:
```suggestion
In this example, three contracts are deployed on the same provider using the `wallet`. The second and third macro use the same contract but have different IDs because of the deployment with salt. Both of them can call the first contract by using its ID.
```

segfault-magnet (dismissed):
Love it.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -110,48 +110,50 @@ impl Abigen {
...
-                        pub fn _with_wallet(&self, mut wallet: WalletUnlocked) -> Result<Self, SDKError> {
+                        pub fn with_wallet(&self, mut wallet: WalletUnlocked) -> Result<Self, SDKError> {
                            let provider = self.wallet.get_provider()?;
                            wallet.set_provider(provider.clone());
 
                            Ok(Self { contract_id: self.contract_id.clone(), wallet: wallet })
                         }
+
+                        pub fn methods(&self) -> #methods_name {
+                            #methods_name { contract_id: self.contract_id.clone(), wallet: self.wallet.clone() }
```
MujkicA, line 144: Should this have a reference to the instance instead? What should the user expect to happen if they change the wallet of the instance with the  above `wallet()` method.

hal3e: They must call `methods()` again. AFAIK rust will not allow to have a reference to a dropped wallet. As the new one would replace it.

MujkicA: This creates a pitfall. It's not evident that if you change the wallet, the methods are affected too. 
Can we use a reference to the actual instance or add a warning to the docs?

hal3e: (I am writing this without trying it out).  `wallet()` and `with_wallet` both create  a new instance so I think the reference is again no go. IMO documentation should be enough.

hal3e: I implemented the second option.

digorithm (approved):
Super clean implementation; goodbye name conflicts! 🎉

Salka1988 (approved):

Issue #353, opened 2022-06-01
`Vec` and heap type support


adlerjohn:
# Description

Currently, the SDK only supports static-sized arrays in the format `[T; n]` where `T` is the type and `n` is the length. 

Soon, Sway will release `Vec`, a growable/mutable/dynamic vector. We need to support this on the SDK side.

# Blockers

- [x] https://github.com/FuelLabs/sway/pull/1118: Full `Vec` support in Sway
- [x] https://github.com/FuelLabs/sway/issues/1994: An output-related work that needs to be done in the compiler, i.e., converting the heap data into stack data and returning it in a `retd`
- [x] https://github.com/FuelLabs/sway/issues/1993: Additional type-related data in the JSON ABI needs to be outputted by the compiler (the `T` in `Vec<T>` needs to be resolved and present in the ABI)
- [x] https://github.com/FuelLabs/sway/issues/2386
- [x] https://github.com/FuelLabs/fuels-rs/pull/587
- [x] https://github.com/FuelLabs/sway/issues/2900
- [x] https://github.com/FuelLabs/fuels-rs/issues/600

# Design

`Vec` is implemented in Sway with a struct like this:

```Rust
struct RawVec<T> {
    ptr: u64,
    cap: u64,
}

pub struct Vec<T> {
    buf: RawVec<T>,
    len: u64,
}
```

Once a Sway user imports it into their contract, `Vec<T>` and `RawVec<T>` will be exposed in the JSON ABI as regular structs. This will be `abigen!`'d on the SDK side, turning these two Sway structs into equivalent Rust structs.

Passing data in and out using `Vec` will be the most challenging part. 

Passing in a `Vec<T>` to a contract call in Rust-land means we have to encode the length `len`, the pointer `ptr`, and the capacity `cap`. But where is the `ptr` pointing to? It will point to where the data lives _inside the `script_data`_, where we'll have to encode the actual content of the `Vec<T>`. This should suffice and, in Sway-land, a user should be able to access the contents of the `Vec<T>` param in their contract method. 

Taking the returned data, in Rust-land, from a contract call that returns a `Vec<T>` means [this](https://github.com/FuelLabs/sway/issues/1994), which should be simpler than the encoding part.

Once the SDK is unblocked by the compiler, the SDK needs to encode the vector data into `script_data` and point to it using the `Vec`'s `ptr`. 

## Update August 16th, 2022

[...]

PR #597, opened 2022-09-27
feat: vector input support

segfault-magnet:
closes: #353 

**Vectors in function outputs are currently not supported. Waiting on compiler support.**

Most notable changes:
1. Unresolved Bytes
Encoding something via the `ABIEncoder` will no longer give you raw bytes right away. You get something called `UnresolvedBytes`. Think of an unresolved symbol during linking :D 

If you give `UnresolvedBytes` the address at which the bytes are going to be placed, then it will reward you with the bytes themselves. This is so that we may resolve any pointers (currently only vectors) within the data.

<s>2. Reverse decoding
Returning a vector involves logging its elements. We need to know where those logs are in, what could be, a sea of other unrelated logs. We cannot require the vec element logs to be the first thing logged since that would mean whatever you do to calculate the vector must not log.

The only reasonable approach to me was to say that the last logs of the method must be the vector elements themselves.

So that means reverse decoding since you don't know from `ParamType`s how big the vector is going to be. So you kind of decode in reverse until everyone is happy :) ~

3. Decoding now requires the ReturnData as well as Logs to decode. 

To generate methods that actually return vectors, we need the sway contract methods themselves to return vectors.

But returning a vector from sway gives you very little to work with. Only 3 WORDs, the pointer, the capacity and the length of the vector you're returning. 

The rest of the information should be logged rather than returned. Until further notice that burden falls on the authors of the contracts. 

There is a caveat to logging in sway as well -- if you log a vector you're not going to be logging its elements. Just like when you return a vector -- you'll only get those 3 WORDS. 

This results in an edge case where if you have a Vec<Vec<..>> you need to jump through an extra hoop to have it decode properly. This is due to `log` not penetrating the Vec iron curtain, so if you log the parent vec you're going to just get it's ptr, cap and len and not those of its children elements.

That is why you need to log the inner vectors themselves **after** you log their elements. This is because we're decoding from the right (aka reverse) and we need to get the vec length info before we get the elements. 

And why we're decoding in reverse is explained somewhere above.

There is a sync shortly regarding vectors, hopefully, the compiler team can figure something out to either make this logging automatic or provide a recursive helper which can penetrate the veil of a vector :) This behavior is documented in the SDK book. </s>


docs/src/types/vectors.md
```diff
@@ -0,0 +1,183 @@
+# Vectors
+
+## Passing in vectors
+You can pass a Rust std::vec::Vec into your contract function transparently. The following code calls a sway function which accepts and then returns a `Vec<SomeStruct<u32>>`.
```
mohammadfawaz, line 4:
```suggestion
You can pass a Rust `std::vec::Vec` into your contract method transparently. The following code calls a Sway function which accepts and then returns a `Vec<SomeStruct<u32>>`.
```

docs/src/types/vectors.md
```diff
@@ -0,0 +1,183 @@
+# Vectors
+
+## Passing in vectors
```
mohammadfawaz, line 3: I think we're meant to add an empty line after titles. Otherwise, the Markdown linter will complain.

docs/src/types/vectors.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+### A simple case:
+```Rust
+contract;
+
+use std::logging::log;
+use std::vec::Vec;
+use std::option::Option;
+use std::assert::assert;
```
mohammadfawaz, line 27:
```suggestion
use std::logging::log;
```
The other are now in the prelude. Assuming you're on `v0.24.4` or after.

docs/src/types/vectors.md
```diff
@@ -0,0 +1,183 @@
...
+    while i < vec.len() {
+        let el = vec.get(i);
+
+        match el {
+            Option::Some(val) => {
+                log(val)
+            },
+            _ => {
+                assert(false);
+            }
```
mohammadfawaz, line 56:
```suggestion
            Option::Some(val) => log(val),
            _ => assert(false),
```

docs/src/types/vectors.md
```diff
@@ -0,0 +1,183 @@
...
+```
+
+You can use a vector just like you would use any other type -- i.e. a `[Vec<u32>; 2]` or a `SomeStruct<Vec<Bits256>>` etc.
+
+## Returning vectors
+There is a mandatory extra step to returning vectors -- you need to `log` every element before returning the vector.
+
+You must not mix any unrelated `log`s once you start logging vector elements. 
+
+These logs must be the last ones you make.
```
mohammadfawaz, line 17: If we're logging the elements, why are we also returning in it? Is there some magic happening to combine the logs and the return?

docs/src/types/vectors.md
```diff
@@ -0,0 +1,183 @@
...
+
+
+### A simple case:
+```Rust
+contract;
```
mohammadfawaz, line 22: Can we include actual code from the test directory instead of pasting code here?

packages/fuels/tests/harness.rs
```diff
@@ -3805,3 +3827,141 @@ async fn test_script_interface() -> Result<(), Error> {
...
+        StorageConfiguration::default(),
+    )
+    .await?;
+
+    let contract_instance = MyContractBuilder::new(contract_id.to_string(), wallet).build();
+
+    {
+        // vec of u32s
+        let arg = vec![0, 1, 2];
+        let result = contract_instance.u32_vec(arg.clone()).call().await?.value;
```
mohammadfawaz, line 3853: Hmm I see. So I wonder if the compiler should inject the logging automatically here.

docs/src/types/vectors.md
segfault-magnet: I wasn't aware our CI had a markdown linter. Added the newlines.

docs/src/types/vectors.md
segfault-magnet: 0.24.4 didn't want to compile.

I've just tried 0.24.5 and it works. Will update the CI version of forc and remove this.

mohammadfawaz: Ahhhh it is possible that `0.24.4` has `Vec` and `assert` only. `Option` needs `0.24.5`.

docs/src/types/vectors.md
segfault-magnet: The only way we know when somebody wants to return something from a function is if they actually return it.

When they do we will generate a function that has that return type -- i.e. a `Vec<...>`

But the return data isn't enough to populate the vector... so it is enriched with the logs.

docs/src/types/vectors.md
```diff
@@ -0,0 +1,177 @@
...
+
+fn log_vec<T>(vec: Vec<T>) {
+    let mut i = 0;
+    while i < vec.len() {
+        let el = vec.get(i);
+
+        match el {
+            Option::Some(val) => log(val),
+            _ => assert(false),
+        };
```
mohammadfawaz, line 53: Actually, now that I'm looking at these match statements, you should probably just use `unwrap()` here since you already know from the condition of the loop that `get()` will not return `None`.

docs/src/types/vectors.md
segfault-magnet: Sure. Wanted to originally but those were a bit convoluted with assertion checks for equality and what not.

I'll add some that will serve mostly as an example and link to them here.

packages/fuels/tests/harness.rs
segfault-magnet: If it could be automatic and result in a single log receipt, I'd be a happy man.

docs/src/types/vectors.md
segfault-magnet: Funny thing. `Option<something>` doesn't have an `unwrap` method, at least that is what the compiler (v0.24.4) tells me. Perhaps I'll have more luck in `0.24.5`, will try it now.

mohammadfawaz: It should! https://github.com/FuelLabs/sway/blob/44f6bcb50cec49275a30afa23c5a0d5fde86c585/sway-lib-std/src/option.sw#L47
If that's not working, then there's a bug somewhere 🤔

segfault-magnet: For this example:
```
contract;

abi MyContract {
    fn test_function();
}

    fn test_function() {
        let mut a_vec = ~Vec::new();
        a_vec.push(1);

        let el = a_vec.get(1).unwrap();
    }
}
```
This is the output on `0.24.4`:
```
❯ forc build
  Creating a new `Forc.lock` file. (Cause: lock file did not match manifest)
  Removing core
  Removing std git+https://github.com/fuellabs/sway?tag=v0.24.5#e695606d8884a18664f6231681333a784e623bc9
    Adding core
    Adding std git+https://github.com/fuellabs/sway?tag=v0.24.4#f7800e18979dd8e4c8d672b8e7aae1758c2edb88
   Created new lock file at /home/segfault_magnet/example/Forc.lock
  Compiled library "core".
  Compiled library "std".
error
  --> /home/segfault_magnet/example/src/main.sw:13:24
   |
11 |
12 |
13 | 	let el = a_vec.get(1).unwrap();
   |                       ^^^^^^ No method named "unwrap" found for type "Option<u64>".
14 |     }
15 | }
   |
____

  Aborting due to 1 error.
Error: Failed to compile example

✦ ~/example
```

But on 0.24.5 it works :partying_face: . So I'm going to be changing these to unwraps

segfault-magnet: Ahhh... 

0.24.5 will work as well if you `use std::option::Option`.

docs/src/types/vectors.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Passing in vectors
+
+You can pass a Rust `std::vec::Vec` into your contract method transparently. The following code calls a Sway function which accepts and then returns a `Vec<SomeStruct<u32>>`.
+
+```rust,ignore
+{{#include ../../../packages/fuels/tests/harness.rs:passing_in_vec}}
+```
+
+You can use a vector just like you would use any other type -- i.e. a `[Vec<u32>; 2]` or a `SomeStruct<Vec<Bits256>>` etc.
```
MujkicA, line 11:
```suggestion
You can use a vector just like you would use any other type -- e.g. a `[Vec<u32>; 2]` or a `SomeStruct<Vec<Bits256>>` etc.
```
Maybe more appropriate since it's an example.

docs/src/types/vectors.md
```diff
@@ -0,0 +1,171 @@
...
+        log_vec(parent_info_vec);
+
+        parent
+    }
+}
+```
+
+### Nested vectors
+
+There is one more step you must take if you're logging a vector nested immediately inside another vector -- i.e. `Vec<Vec<u32>>`
```
MujkicA, line 108:
```suggestion
There is one more step you must take if you're logging a vector nested immediately inside another vector -- e.g. `Vec<Vec<u32>>`
```

MujkicA (commented):
Code looks good, nice work :tada:

packages/fuels/tests/harness.rs
```diff
@@ -3805,3 +3827,141 @@ async fn test_script_interface() -> Result<(), Error> {
...
+
+    let contract_id = Contract::deploy(
+        "tests/test_projects/vectors/out/debug/vectors.bin",
+        &wallet,
+        TxParameters::default(),
+        StorageConfiguration::default(),
+    )
+    .await?;
+
+    let contract_instance = MyContractBuilder::new(contract_id.to_string(), wallet).build();
```
hal3e, line 3848: You will have to change if because of the `methods()` PR. might a s well use the `setup_contract_test!` macro.

hal3e (commented):
Nice work! This is my first review pass. I have to rest a bit and try to wrap my head around the `resolve` thing :smile:

docs/src/types/vectors.md
```diff
@@ -0,0 +1,171 @@
...
+
+Step 4.
+We've finished logging the elements of the parent vector. Since the type is `Vec<Vec<..>>` we have an additional step -- call `log` on each element of the vector:
+
+
+`log(inner_vec_1)`
+
+`log(inner_vec_2)`
+
+In the example this is done by the `log_vec` helper.
```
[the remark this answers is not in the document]
segfault-magnet: I would like to avoid saying we need to "log the parent vector" because we shouldn't do that. We should log the elements of its elements, and log the elements themselves but not the parent vector.

I'll integrate the rest of the suggestions accordingly.

segfault-magnet:
> Nice work! This is my first review pass. I have to rest a bit and try to wrap my head around the `resolve` thing smile
Ty.
I'm glad you said that. This means that I probably need to put some comments in.

hal3e (dismissed):
Thanks for adding the comments. It is much easier now to understand what is happening where. LGTM! Good work! :tada:

Salka1988 (approved):
As Hal3e said, thanks for the explanation! LGTM :+1:

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -141,6 +215,23 @@ impl ABIEncoder {
...
+        // len. This means that we must place the encoded vector elements
+        // somewhere else. Hence the use of Data::Dynamic which will, when
+        // resolved, leave behind in its place only a pointer to the actual
+        // data.
+        Ok(vec![
+            Data::Dynamic(encoded_data),
+            Self::encode_u64(cap),
+            Self::encode_u64(len),
+        ])
+    }
```
digorithm, line 234: image

Very nice abstraction you got going on here!

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -212,22 +213,17 @@ impl Abigen {
...
-            || split[1] == ADDRESS_NATIVE_TYPE
-            || split[1] == OPTION_NATIVE_TYPE
-            || split[1] == RESULT_NATIVE_TYPE
+    pub fn is_native_type(type_field: &str) -> anyhow::Result<bool> {
+        let name = extract_custom_type_name_from_abi_type_field(type_field)?;
+
+        // "RawVec" is part of the Vec structure. Not used in the SDK and thus
+        // not generated.
+        Ok(
+            ["ContractId", "Address", "Option", "Result", "Vec", "RawVec"]
```
digorithm, line 222: These should still be `const`s extracted away from this function, though. Easier to maintain/find/refer to. Plus, there's a new `"Identity"` one in master; you'll see it once you sync with it!

packages/fuels-core/src/code_gen/resolved_type.rs
```diff
@@ -24,6 +24,16 @@ impl ResolvedType {
...
+    // Used to prevent returning vectors until we get the compiler support for
+    // it.
+    #[must_use]
+    pub fn uses_vectors(&self) -> bool {
+        lazy_static! {
+            static ref RE: Regex = Regex::new(r"\bVec\b").unwrap();
+        }
+        RE.is_match(&self.type_name.to_string())
+            || self.generic_params.iter().any(ResolvedType::uses_vectors)
+    }
```
digorithm, line 36: image

packages/fuels/tests/test_projects/vectors/src/main.sw
```diff
@@ -0,0 +1,128 @@
...
+        inner_vec_2.push(inner_inner_vec_3);
+
+        let inner_inner_vec_4 = vec_from([9, 10, 11]);
+        inner_vec_2.push(inner_inner_vec_4);
+
+        expected.push(SomeStruct { a: inner_vec_2 });
+
+        assert(arg == expected);
+    }
+}
```
digorithm, line 128: Really nice testing stratetegy!

digorithm (changes requested):
Wonderful work. Loved the refactors and abstraction you introduced you did along the way. 

Left a micro nit. Also, sync with master when possible!

packages/fuels-core/src/code_gen/resolved_type.rs
segfault-magnet: True dat :D , but the '\b' is worth its weight in gold.

packages/fuels-core/src/code_gen/abigen.rs
segfault-magnet: Hopefully, I can sway you with the following :D 

> These should still be consts 

Godbolt shows both `let a = "something"` and `const a: &str = "something"` results in the same asm.


> extracted away from this function
They have meaning only in this function. I'd rather not place them somewhere else.

> Easier to maintain/find/refer to.
They are inside a 10line function. Only to be used here. The keywords are used only once and in such a way that their usage gives them meaning -- i.e. `if my type matches any of the following ones`. 

> there's a new "Identity" one in master; you'll see it once you sync with it!
I've added it just now :)

digorithm: > Godbolt shows both let a = "something" and const a: &str = "something" results in the same asm.

My suggestion is _really_ not related to the generated asm 😂. 100% related to readability. But this: 

> They have meaning only in this function. I'd rather not place them somewhere else.

Makes it ok for it to be inside this function (and, thus, not a `const`). The moment we start referencing this from other places, then we should extract it out. But yeah, not needed right now!

digorithm (approved):
:shipit:

PR #599, opened 2022-09-28
Fix transfer to contract

MujkicA:
I've noticed that our transfer-to-contract script is calculating the offset of the asset id incorrectly.
The corresponding test was passing because it was checking for the base asset id which is all 0s.

This PR corrects the script, and adapts the test to use a random asset id.

### DIFF
--- a/packages/fuels/tests/harness.rs
+++ b/packages/fuels/tests/harness.rs
@@ -3761,9 +3761,9 @@ async fn test_script_interface() -> Result<(), Error> {
     let script = vec![
         Opcode::gtf(0x10, 0x00, GTFArgs::ScriptData),
         Opcode::ADDI(0x11, 0x10, ContractId::LEN as u16),
-        Opcode::LW(0x11, 0x11, 0),
-        Opcode::ADDI(0x12, 0x11, WORD_SIZE as u16),
-        Opcode::TR(0x10, 0x11, 0x12),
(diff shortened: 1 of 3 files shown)


packages/fuels/tests/harness.rs
```diff
@@ -3761,9 +3761,9 @@ async fn test_script_interface() -> Result<(), Error> {
     let script = vec![
         Opcode::gtf(0x10, 0x00, GTFArgs::ScriptData),
         Opcode::ADDI(0x11, 0x10, ContractId::LEN as u16),
-        Opcode::LW(0x11, 0x11, 0),
-        Opcode::ADDI(0x12, 0x11, WORD_SIZE as u16),
-        Opcode::TR(0x10, 0x11, 0x12),
+        Opcode::LW(0x12, 0x11, 0),
+        Opcode::ADDI(0x13, 0x11, WORD_SIZE as u16),
+        Opcode::TR(0x10, 0x12, 0x13),
```
digorithm, line 3766: Oof, nice find!

digorithm (approved):

packages/fuels/tests/harness.rs
iqdecay: Oh come we didn't catch in the tests ?

MujkicA: Yup, the previous test was just weak. My bad.

PR #601, opened 2022-09-30
feat: add Identity as native type

hal3e:
Using `Identity` in two deployed contracts would result in naming collision as both contract would make their own `Identity` enum. 

This PR fixes that by treating `Identity` as a native type.

- Updated abigen to use native `Identity`
- Added test for different `Identity` use-cases


digorithm (approved):
Super clean, great job!

Issue #594, opened 2022-09-26
Integration tests (`harness.rs`) refactor/re-org


digorithm:
Currently, our main harness file is nearing 4 thousand lines. It’s getting hard to navigate and find specific tests. Not only that, but this file contains a mixed bag of *many* different tests. To improve this navigation and general maintainability, we have to refactor this file by breaking it into smaller, responsibility-driven files. 

The `harness.rs` inside `fuels/tests` should act like a regular `lib.rs` aggregating many other siblings `*.rs` files. Then, alongside `harness.rs`, we should have other test files named in an easier way to find specific tests. Grouping all harness tests by specific responsibility.

Examples and initial ideas:

1. From `compile_bindings_from_contract_file` to `compile_bindings_enum_input` ⇒ `bindings.rs`, a test group focused on testing the binding compilation process using simple types. 
2. From `create_struct_from_decoded_tokens` to `create_nested_struct_from_decoded_tokens` ⇒ `from_token.rs`; a test group focused on testing the `::from_token()` functionality.
3. Contract call behaviour-related tests, maybe call it `behavior.rs`: `test_reverting_transaction`, `multiple_read_calls`, `test_contract_calling_contract`, `test_gas_errors`, `test_call_param_gas_errors`
4.  `call_with_structs`, `call_with_empty_return`, `abigen_different_structs_same_arg_name`, `test_large_return_data` ⇒ `types.rs`; tests related to types.
5. Tests related to infrastructure; `test_provider_launch_and_connect`. Maybe call this one `infrastructure.rs`?
6. `test_script_interface` ⇒ `script.rs`, integration tests related to scripts.

These are just some initial ideas. The implementer should be able to check test by test in the `harness.rs` and group them by responsibility. **By the end of it, `harness.rs` should contain only includes from other test files, e.g.:**

```rust
mod script;
mod behavior;
mod types;
mod from_tokes; 
// And so on.
```

And the `fuels/` crate should look like this:

```
.
├── Cargo.toml
├── src
│   └── lib.rs
└── tests
    ├── harness.rs
    ├── types.rs
    ├── behavior.rs
    ├── infrastructure.rs
    ├── script.rs
    ├── ... and so on... 
    └── test_projects/
```

PR #607, opened 2022-10-04
refactor: change organisation of the integration tests

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/594

AFAIK there are two options for refactoring the `harness.rs`
The first option is similar to what @digorithm said. It looks like this:
```
.
├── Cargo.toml
├── src
│   └── lib.rs
└── tests
    ├── types.rs
    ├── behavior.rs
    ├── infrastructure.rs
    ├── script.rs
    ├── ... and so on... 
    └── test_projects/
```
In this case we do not need the `harness.rs` and the `mod`s. Because all the files will be separately run. So adding them to the harness would just repeat the test.

The second option (the one that is in this branch) is that we separate the files as proposed in https://doc.rust-lang.org/book/ch11-03-test-organization.html. In this case, the test are separated into modules and they have to be included in the `harness.rs` with the `mod` statement. I liked this approach because we could also put the test projects into the appropriate folders and get a nice separation for them as well. The `test_project` folder also got big and unorganized... The new structure would looks like this:
```
.
├── Cargo.toml
├── src
│   └── lib.rs
└── tests
    ├── behaviors
    │   ├── auth_testing_abi
    │   ├── ...
    │   └── mod.rs
    ├── bindings
    ├── contract_instance
    ├── examples
    ├── from_token
    ├── infrastructure
    ├── logs
    ├── predicates
    ├── scripts
    ├── types
    ├── storage
    └── harness.rs
```
Each group would have a folder with all the related `test_projects` and a `mod.rs` file with the tests.

I have extended the `build-test-projecs` script and it can now recursively traverse folders to find the `forc` projects. I have also made a small update on how the info is displayed by the script. Previously it would build all the projects and show the results at the end. Now it will print as it builds...

Regardless which option we choose at the end, I need some feedback on the _groups_. Are the names OK and did I put the right test in the right groups. 

For example, the test `contract_test` is used all over the place. Should we have a `common` group or is it ok if it stays in `behaviors`.  Also I was not sure where to put these: `fn test_amount_and_asset_forwarding()`, `fn test_get_gas_used()`, `fn test_large_return_data()`


scripts/build-test-projects/src/main.rs
```diff
@@ -84,6 +48,42 @@ fn main() {
...
+                (1, "[✓]", "succeeded")
+            };
+            println!("  {}: {} {}!", checkmark, child_path.display(), status);
+
+            summary.push(success);
+        }
+    }
+    summary
+}
+
```
segfault-magnet, line 86: A more radical approach :D Warning, you might not like it :D 
Discovering projects is its own phase:
```rust
fn discover_projects(path: &Path) -> Vec<PathBuf> {
    fs::read_dir(path)
        .expect("failed to walk directory")
        .filter_map(|entry| entry.ok())
        .map(|entry| entry.path())
        .filter(|path| path.is_dir())
        .flat_map(|path| {
            if dir_contains_forc_manifest(&path) {
                vec![path]
            } else {
                discover_projects(&path)
            }
        })
        .collect()
}
```

Building projects returns an iterator of `TestResult`:
```rust
fn build_recursively(path: &Path) -> impl Iterator<Item = TestResult> {
    discover_projects(path).into_iter().map(|path| {
        let output = std::process::Command::new("forc")
            .args(["build", "--generate-logged-types", "--path"])
            .arg(&path)
            .output()
            .expect("failed to run `forc build` for example project");

        let compilation_result = CompilationResult {
            path,
            stdout: String::from_utf8(output.stdout).expect("Forc output is not valid utf8"),
            stderr: String::from_utf8(output.stderr).expect("Forc output is not valid utf8"),
        };

        if output.status.success() {
            TestResult::Success(compilation_result)
        } else {
            TestResult::Failure(compilation_result)
        }
    })
}
```
with test results defined as:
```rust

struct CompilationResult {
    path: PathBuf,
    stdout: String,
    stderr: String,
}

enum TestResult {
    Success(CompilationResult),
    Failure(CompilationResult),
}
```

if you implement `Display` for the result:
```rust
impl Display for TestResult {
    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
        match self {
            TestResult::Success(CompilationResult { path, .. }) => {
                write!(f, "  {}: [✓] succeeded!", path.display())
            }
            TestResult::Failure(CompilationResult {
                path,
                stdout,
                stderr,
            }) => {
                write!(f, "{stdout}")?;
                write!(f, "{stderr}")?;
                write!(f, "  {}: [x] failed!", path.display())
            }
        }
    }
}
```

then you can lazily consume the iterator and lift the decision to print to stdout as high as it can go:
```rust
    let (succeeded, failed): (Vec<_>, Vec<_>) = build_recursively(&absolute_path)
        .into_iter()
        .inspect(|result| println!("{result}"))
        .partition(|result| matches! {result, TestResult::Success(_)});
```

MujkicA:
I'm concerned that we will be tempted to put future tests into a general folder like `behaviors` which leads us back to square one. 
Also, what would be the reasoning for e.g. having a separate folder for `storage` but not for `gas_estimation`.

I propose we agree to a soft rule like: If some functionality warrants a separate section in the docs, it should also have its own test folder. This should also help users navigate since we already encourage them to look into the tests for inspiration.

iqdecay (changes requested):
I agree with AhmedM, maybe you could add some documentation in the `README` about how tests are structured and what the suggestions are for adding new ones? I feel like having the organization just in the commit message really won't be enough for referencing in the future. Maybe even a doc section? I really like the new organization!

hal3e:
I have refactored the `build-test-projects` script based on @segfault-magnet's suggestions. While I was at it I dediced to add color and make it look more like the output from `cargo`. I hope you like it.
Here is how it looks now when everything is ok:
2022-10-10-162130_357x406_scrot

and when we have an error:
2022-10-10-162553_497x561_scrot

iqdecay (changes requested):
Super neat, I love the color and the lib! Still think you should add some docs though

hal3e:
I will :smile:  I have to reorg the tests also.

hal3e:
I have reorganized the test and now we have the following structure:
```rust
mod bindings; // generating Rust bindings
mod contracts; // contracts, contract calls and call behaviors
mod from_token; // generating instances from decoded tokens
mod logs; // parsing and decoding contract logs
mod predicates; // calling predicates and encoding predicate data
mod providers; // launching and querying providers
mod scripts; // building and calling scripts
mod storage; // storage initialization and forwarding
mod types; // encoding/decoding of native and custom types
mod wallets; // wallet creation and balance checks
```
I have also added a new section called `Contributing` and a sub-section called `Integration tests structure` where I have added some soft rules for creating new test modules.

Let me know if you have concrete suggestion with the test separation or where to put the info in the docs.

scripts/build-test-projects/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+
+    pub fn display_forc_info(&mut self) -> Result<(), std::io::Error> {
+        let output = std::process::Command::new("forc")
+            .args(["--version"])
+            .output()?;
+
+        let version =
+            String::from_utf8(output.stdout).expect("failed to parse forc --version output");
+
+        self.stdout.set_color(self.green.clone().set_bold(true))?;
```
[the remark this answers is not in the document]
hal3e: Good points. I thought about both when I decided to add color and came to the same conclusion. Catching write errors would IMO make the code more complex and checking if we are writing to a `tty` would require to import another crate.. I am ok to leave it like this. We can upgrade the script later if needed :smile:

scripts/build-test-projects/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+                self.stdout.set_color(&self.green)?;
+                writeln!(&mut self.stdout, "ok")?;
+            }
+            BuildResult::Failure(c) => {
+                write!(
+                    &mut self.stdout,
+                    "build {} ... ",
+                    c.get_display_path(abs_path).display()
+                )?;
+                self.stdout.set_color(&self.red)?;
```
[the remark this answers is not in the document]
hal3e: Good idea. I will add the functions. It should make the code clearer.

segfault-magnet (approved):

digorithm:
It's looking good, I really like the new org; Build is broken though, seems to be a small thing. Good to merge after it's fixed!

digorithm (approved):
Awesome work. Extra thumbs up for the new behavior around `build-test-projects`. It's much better now.

docs/src/contributing/tests-structure.md
```diff
@@ -0,0 +1,11 @@
+# Integration tests structure in `fuels-rs`
+
+The integration tests of `fuels-rs` cover almost all aspects of the SDK and have grown significantly as more functionality is added. To make the test more manageable, they were split into several `modules` together with the corresponding `Sway` projects. The modules are defined in the `packages/fuels/test/harness.rs` file and currently have the following structure:
```
iqdecay, line 3:
```suggestion
The integration tests of `fuels-rs` cover almost all aspects of the SDK and have grown significantly as more functionality were added. To make the test more manageable, they were split into several `modules` together with the corresponding `Sway` projects. The modules are defined in the `packages/fuels/test/harness.rs` file and currently have the following structure:
```

docs/src/contributing/tests-structure.md
```diff
@@ -0,0 +1,11 @@
...
+
+```rust,ignore
+{{#include ../../../packages/fuels/tests/harness.rs:test_modules}}
+```
+
+Even though adding new test modules is subjective, please consider these guidelines when adding new ones:
+ - Add a new module when creating a new section in the `Fuels Rust SDK` book - e.g. `Types`
+ - Add a new module if there are more than 3 test and more than 100 lines of code and they form a group of tests - e.g. `mod storage`
```
iqdecay, line 11:
```suggestion
Even though test organization is subjective, please consider these guidelines before adding a new category (`module`):
 - Add a new module when creating a new section in the `Fuels Rust SDK` book - e.g. `Types`
 - Add a new module if there are more than 3 test and more than 100 lines of code and they form a group of tests - e.g. `mod storage`
 
 Otherwise, we recommend putting the integration test inside the existing modules above. 
```

iqdecay (approved):
Nice! Left a nit on formulation :)

docs/src/contributing/tests-structure.md
segfault-magnet: as more functionality **was** added.

iqdecay (approved):
:ship: it

PR #613, opened 2022-10-05
Variable output estimation

### LINKED ISSUE #596: Variable outputs estimation
Exactly like https://github.com/FuelLabs/fuels-rs/issues/595. Replace `Contract input set` for `variable outputs`. Same goals, design, and blockers.

MujkicA:
Closes #596 

This PR implements a method that automatically sets the number of required output variables by simulating the call and checking if the tx reverted with a special return code. 
The current name is `auto_setup()`, but feel free to suggest alternatives.

I've also changed `simulate()`in `script.rs` to throw a `RevertTransactionError` if a revert occurs. Prior to this change, using `simulate()` on a contract call that fails would result in bug because we would attempt to decode a response from a failed call.

Options for the UI:

- (Currently implemented) Builder method that modifies and returns the `ContractCallHandler`. This gives the user more control but the await-call-await chain is a bit awkward.

```rust
let _ = contract_methods
            .some_method()
            .auto_setup(None)
            .await
            .call()
            .await?;
```

- Have a special call method that performs estimation and before execution. The downside to this is that the user wouldn't know if an error occurred during simulation or during the actual call.

```rust
let _ = contract_methods
            .some_method()
            .auto_setup_and_call()
            .await?;
```

Would also like to get an opinion on this:
Should the method throw an error if, after running all the attempts, the simulation fails?


digorithm:
Should be unblocked now by https://github.com/FuelLabs/sway/releases/tag/v0.25.0, right?

packages/fuels-core/src/constants.rs
```diff
@@ -15,3 +15,5 @@ pub const BASE_ASSET_ID: AssetId = AssetId::new([0u8; 32]);
 pub const DEFAULT_GAS_ESTIMATION_TOLERANCE: f64 = 0.2;
 pub const GAS_PRICE_FACTOR: u64 = 1_000_000_000;
 pub const MAX_GAS_PER_TX: u64 = 100_000_000;
+
+pub const FAILED_TRANSFER_TO_OUTPUT_SIGNAL: u64 = 0xffff_ffff_ffff_0001;
```
iqdecay, line 19: Can you add a comment explaining what this means? Is this some kind of error code from `fuel-core`?

packages/fuels-contract/src/contract.rs
```diff
@@ -512,16 +513,40 @@ where
         .await
     }
 
+    /// Simulates calls and attempts to set output variables based on the received error.
+    /// Forwards the reiceived error if it cannot be fixed.
+    pub async fn try_resolve(mut self, max_attempts: Option<u64>) -> Self {
```
iqdecay, line 518: I feel like the naming should be improved. `try_resolve` is too broad, and looks like a Rust method. Something like `auto_set_variable_outputs` would be better IMO, what do you think ?

iqdecay (changes requested):
This will be great for UX! But I feel that since we're probably going to encourage users to use this, we should maybe add a docs section about it? Since it will become kind  of a "good practice". 
Also not very sure about the naming.

packages/fuels-contract/src/contract.rs
MujkicA: `try_resolve` is just a placeholder but we will need something more general than `auto_set_variable_outputs ` since this method will also set message outputs and external contracts.

MujkicA:
> This will be great for UX! But I feel that since we're probably going to encourage users to use this, we should maybe add a docs section about it? Since it will become kind of a "good practice". Also not very sure about the naming.

I left out the docs because I figured it would be better to wait for message output and external contract estimation, rather than rewrite them 2 times.

packages/fuels-contract/src/contract.rs
```diff
@@ -358,6 +361,39 @@ pub struct ContractCall {
...
 
 impl ContractCall {
+    pub fn append_variable_outputs(&mut self, num: u64) {
+        let new_variable_outputs: Vec<Output> = (0..num)
+            .map(|_| Output::Variable {
+                amount: 0,
+                to: Address::zeroed(),
+                asset_id: AssetId::default(),
+            })
+            .collect();
```
hal3e, line 371:
```suggestion
        let new_variable_outputs = vec![
            Output::Variable {
                amount: 0,
                to: Address::zeroed(),
                asset_id: AssetId::default(),
            };
            num as usize
        ];
```

hal3e (commented):
LGTM.

packages/fuels-contract/src/contract.rs
```diff
@@ -512,16 +527,40 @@ where
         .await
     }
 
+    /// Simulates calls and attempts to set output variables based on the received error.
+    /// Forwards the reiceived error if it cannot be fixed.
+    pub async fn auto_setup(mut self, max_attempts: Option<u64>) -> Self {
```
digorithm, line 532: `auto_setup` isn't a good name. It doesn't tell me _what_ it is automatically being set up; I have to read the doc-level comment to know that. It gets extra confusing once the user has to reason about the "automatically set up of something" and the need for a maximum number of attempts, e.g., "why does it need to do many attempts just to set up?!".

It's better to give it a name that makes sense to have `max_attemps`. Probably with `"estimate"` in it. I understand we don't want to put `"output variables"` in it because this will likely be used for estimating the contract input set as well. So... I propose `set_estimated_tx_dependencies`:
1. `set` hints that it's mutating something;
2. `estimated` because it's an estimation (and it takes `max_attemps` attempts);
3. `tx_dependencies` because it's doing this operation on the dependencies needed for this transaction to work (output variables and, soon, the contract input set);
4. We're `set`ting the `estimated` `tx_dependencies` for the current transaction.

Open to suggestions/improvements on it, of course.

packages/fuels-contract/src/contract.rs
```diff
@@ -512,16 +527,40 @@ where
         .await
     }
 
+    /// Simulates calls and attempts to set output variables based on the received error.
+    /// Forwards the reiceived error if it cannot be fixed.
```
digorithm, line 531:
```suggestion
    /// Simulates calls and attempts to set output variables based on the received error.
    /// Forwards the received error if it cannot be fixed.
```

digorithm (changes requested):
Nice, it's looking really good from a functionality perspective; super neat to offer this API.

I don't love the new `.await` added to the chain, but I don't _think_ users will have to do this all the time. Time will tell, though; not a huge DevEx hit.

About this:

> Would also like to get an opinion on this:
> Should the method throw an error if, after running all the attempts, the simulation fails?

I believe so, yes,

packages/fuels-contract/src/contract.rs
```diff
@@ -624,6 +661,43 @@ impl MultiContractCallHandler {
...
+        let provider = self.wallet.get_provider()?;
+
+        script.simulate(provider).await?;
+
+        Ok(())
+    }
+
+    pub async fn set_estimated_tx_dependencies(
+        mut self,
+        max_attempts: Option<u64>,
```
mohammadfawaz, line 676: Are we going to reuse this function for estimating the number of Message Outputs as well? If so, will both outputs share a `max_attempts`?

packages/fuels-contract/src/contract.rs
segfault-magnet: It's a better name.

It's just the `set` part that rubs me wrong. We are setting it, sure, but we're also doing the estimating. 

Perhaps it's just that I'm not used to seeing a "setter" do such work. 

But that means we're to split the estimation and setting -- which causes issues of its own. 

Or use action-y names like `estimate_tx_deps` but that doesn't have the set in it. You'd have to see it takes a `&mut self`

digorithm: Yeah, I'm not a huge fan of `set`, but it does tell us that it's mutating; that's its sole purpose here.

But you raise a good point: `estimate_tx_deps` is better because no `set`, but if a dev see it takes a `&mut self`... then it's pretty obvious it's mutating. At least one would hope it's pretty obvious haha.

So I'd say let's go with `estimate_tx_deps`. Also, are we sure there's no ambiguity/confusion between `deps` and `dependencies`?

MujkicA: I was thinking the same, but I figured I prefer it to state that it is setting something.

packages/fuels-contract/src/contract.rs
MujkicA: Yes, the idea is to re-use it. I thought about having x attempts per estimation target, but I guess that can be revisited in future PR's. Since this is not documented yet, we can afford to change the interface until we find the right combination.

packages/fuels-contract/src/contract.rs
```diff
@@ -358,6 +361,40 @@ pub struct ContractCall {
...
+        match self.variable_outputs {
+            Some(ref mut outputs) => outputs.extend(new_variable_outputs),
+            None => self.variable_outputs = Some(new_variable_outputs),
+        }
+    }
+
+    /// Appends `num` `Output::Message`s to the transaction.
+    /// Note that this is a builder method, i.e. use it as a chain:
+    /// `my_contract_instance.my_method(...).add_message_outputs(num).call()`.
+    pub fn append_message_outputs(&mut self, num: u64) {
```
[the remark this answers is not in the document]
MujkicA: This the `ContractCall` method, not `CallHandler`. I think there is no need for this to be a builder, but making it `pub(crate)` would probably be better.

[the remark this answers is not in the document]
segfault-magnet: The comment reads: 'Note that this is a builder method, i.e. use it as a chain:
    /// `my_contract_instance.my_method(...).add_message_outputs(num).call()`.' 
I have a feeling we're not talking about the same thing :D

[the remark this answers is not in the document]
segfault-magnet: I'm confused. A comment above `append_message_outputs` which talks about how to use `add_message_outputs`.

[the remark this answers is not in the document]
MujkicA: I say we do away with all of them

packages/fuels-contract/src/contract.rs
```diff
@@ -466,18 +495,7 @@ where
     /// Note that this is a builder method, i.e. use it as a chain:
     /// `my_contract_instance.my_method(...).add_message_outputs(num).call()`.
     pub fn append_message_outputs(mut self, num: u64) -> Self {
-        let new_message_outputs = vec![
```
iqdecay, line 469: nice

packages/fuels-contract/src/contract.rs
```diff
@@ -446,18 +486,7 @@ where
     /// Note that this is a builder method, i.e. use it as a chain:
     /// `my_contract_instance.my_method(...).add_variable_outputs(num).call()`.
     pub fn append_variable_outputs(mut self, num: u64) -> Self {
-        let new_variable_outputs: Vec<Output> = (0..num)
```
iqdecay, line 449: nice

packages/fuels-contract/src/contract.rs
```diff
@@ -550,15 +597,15 @@ where
 #[must_use = "contract calls do nothing unless you `call` them"]
 /// Helper that handles bundling multiple calls into a single transaction
 pub struct MultiContractCallHandler {
-    pub contract_calls: Option<Vec<ContractCall>>,
+    pub contract_calls: Vec<ContractCall>,
```
iqdecay, line 600: makes sense

packages/fuels-contract/src/contract.rs
```diff
@@ -583,14 +627,11 @@ impl MultiContractCallHandler {
...
-        Script::from_contract_calls(
-            self.contract_calls
-                .as_ref()
-                .expect("No calls added. Have you used '.add_calls()'?"),
-            &self.tx_parameters,
-            &self.wallet,
-        )
-        .await
+        if self.contract_calls.is_empty() {
+            panic!("No calls added. Have you used '.add_calls()'?");
```
iqdecay, line 631: sweet

packages/fuels/tests/harness.rs
```diff
@@ -4180,3 +4180,119 @@ async fn test_vector() -> Result<(), Error> {
...
+        &wallets[0],
+        TxParameters::default(),
+        StorageConfiguration::default(),
+    )
+    .await?;
+
+    let contract_instance = MyContract::new(contract_id.to_string(), wallets[0].clone());
+    let contract_methods = contract_instance.methods();
+    let mint_asset_id = AssetId::from(*contract_id.hash());
+    let addresses: [Address; 3] = wallets
```
iqdecay, line 4205: Unrelated note: maybe we should get some methods on `Vec<Wallets>` that provide eg the vec of their addresses, the vec of their balances, etc? Wdyt?

packages/fuels-contract/src/contract.rs
```diff
@@ -514,14 +532,43 @@ where
...
     pub async fn simulate(self) -> Result<CallResponse<D>, Error> {
-        Self::call_or_simulate(self, true).await
+        Self::call_or_simulate(&self, true).await
+    }
+
+    /// Simulates the call and attempts to resolve missing tx dependencies.
+    /// Forwards the received error if it cannot be fixed.
+    pub async fn estimate_tx_dependencies(
+        mut self,
+        max_attempts: Option<u64>,
```
iqdecay, line 549: What is the logical reason for allowing the user to provide a maximum attempt? I feel like this argument would just force the user to pass in `None` 99.99% of the time. What do you think?

packages/fuels/tests/harness.rs
```diff
@@ -4180,3 +4180,119 @@ async fn test_vector() -> Result<(), Error> {
...
+            .await;
+
+        assert!(matches!(response, Err(Error::RevertTransactionError(..))));
+    }
+
+    {
+        // Should add 3 output variables automatically
+        let _ = contract_methods
+            .mint_to_addresses(amount, addresses)
+            .estimate_tx_dependencies(Some(3))
```
iqdecay, line 4237: Here, you should also test that with passing `None`, it also works.

packages/fuels/tests/harness.rs
```diff
@@ -4180,3 +4180,119 @@ async fn test_vector() -> Result<(), Error> {
...
+    let mut multi_call_handler = MultiContractCallHandler::new(wallets[0].clone());
+    (0..3).for_each(|_| {
+        let call_handler = contract_methods.mint_to_addresses(amount, addresses);
+        multi_call_handler.add_call(call_handler);
+    });
+
+    let _ = multi_call_handler
+        .estimate_tx_dependencies(None)
+        .await?
+        .call::<((), (), ())>()
```
iqdecay, line 4289: Is this the syntax the end-user would have to use in multi-call contexts? Can we make it better? I feel like it could be abstracted away, but maybe it involves something too complicated.

packages/fuels-core/src/constants.rs
```diff
@@ -15,3 +15,6 @@ pub const BASE_ASSET_ID: AssetId = AssetId::new([0u8; 32]);
 pub const DEFAULT_GAS_ESTIMATION_TOLERANCE: f64 = 0.2;
 pub const GAS_PRICE_FACTOR: u64 = 1_000_000_000;
 pub const MAX_GAS_PER_TX: u64 = 100_000_000;
+
+// Revert return value that indicates missing output variables
```
iqdecay, line 19: :smile:

iqdecay (changes requested):
Very nice! I left some questions, but I really like this :tada:

packages/fuels-contract/src/contract.rs
digorithm: The reason is to provide granular control of a latency-sensitive part of their code. Yes, a big part of the users won't care about this control; they can just pass in `None`. The users who care will be triggered if they cannot control the number of _network trips_ of a function call that could be in a latency-sensitive part of their code.

packages/fuels/tests/harness.rs
MujkicA: This is would be a prettier alternative: 

```rust
let (counter, array): (u64, [u64; 2]) = multi_call_handler.call().await?.value;
```
The user must, at some point, define what types they are expecting as return. We spent quite some time weighing different options when this was developed and concluded this was the friendliest.

packages/fuels-contract/src/contract.rs
iqdecay: Ah yes I overlooked the network part, very touchy subject indeed haha.

packages/fuels/tests/harness.rs
MujkicA: Maybe we could implement it as a wrapper for `Vec<Wallets>`. Good idea IMO.

packages/fuels/tests/contracts/token_ops/src/main.sw
```diff
@@ -16,7 +16,7 @@ impl TestFuelCoin for Contract {
     fn mint_coins(mint_amount: u64) {
         mint(mint_amount);
     }
-
+
```
segfault-magnet, line 19: 3-4 spaces left on row.
```suggestion

```

segfault-magnet (approved):

packages/fuels/tests/harness.rs
iqdecay: Nice, can you open an issue for this?

iqdecay (approved):
Looking great, :ship:  it!

packages/fuels/tests/harness.rs
MujkicA: I've added it as a discussion for now so that others can weigh in.

PR #620, opened 2022-10-10
feat: support main function arguments in scripts

### LINKED ISSUE #591: Support running scripts that take `main()` arguments
Scripts `main()` functions can now take arguments. However, currently, there's no way to run these scripts with the SDK and pass in arguments to the `main()` function. 

As an effort to improve the script DevEx in the SDK, we should support this.

It was introduced in PR https://github.com/FuelLabs/sway/pull/2592.

iqdecay:
This PR closes #591 by making it easy to pass arguments to the `main` function of a Sway script. 

## Work done

* Adds a `script_abigen` macro that generates the Rust binding with the script types.
* Completely reworks the way scripts and calls are handled. 
* This script abigen works similarily to the classic `abigen`, the workflow looks like this:

The workflow looks like this: 
```rust
    script_abigen!(
        MyScript,
        "packages/fuels/tests/scripts/script_with_arguments/out/debug/script_with_arguments-abi.json"
    );
    let bin_path =
        "../fuels/tests/scripts/script_with_arguments/out/debug/script_with_arguments.bin";
    let instance = MyScript::new(wallet, bin_path);

    let bim = Bimbam { val: 90 };
    let bam = SugarySnack {
        twix: 100,
        mars: 1000,
    };
    let result = instance.main(bim, bam).call().await?;
    let expected = Bimbam { val: 2190 };
    assert_eq!(result.value, expected);
```

## Limits
- No support for `Vectors`.
- No support of inputs/outputs or exetrnal contracts. 
- Lack of unit tests on various combinations of arguments.
- No logging of values.

The fix to these limitations will come in later PRs.


packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -54,6 +57,11 @@ impl Abigen {
         self
     }
 
+    pub fn script(mut self) -> Self {
```
MujkicA, line 60: We could remove this with the bellow alternative.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -72,6 +80,9 @@ impl Abigen {
     /// set of `TokenStream`. This generated Rust code is the brought into scope
     /// after it is called through a procedural macro (`abigen!()` in our case).
     pub fn expand(&self) -> Result<TokenStream, Error> {
+        if self.is_script {
```
MujkicA, line 83: Potentially cleaner alternative:

```rust
pub fn expand(&self, is_script: bool) -> Result<TokenStream, Error> {
        if is_script {
            return self.expand_script();
        }

        self.expand_contract()
    }
```

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,6 +65,44 @@ pub fn expand_function(
     })
 }
 
+/// Generate the function (as a `TokenStream`) used to encode into script data the arguments to the
```
MujkicA, line 68: Nit: The return value in the signature already conveys the TokenStream part.
```suggestion
Generate the function used to encode arguments to the`main` function of a script.
```

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,6 +65,44 @@ pub fn expand_function(
...
+/// `main` function of a deployed script.
```
MujkicA, line 69: Could be confusing since scripts aren't actually "deployed".
```suggestion
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -192,13 +219,53 @@ impl Abigen {
...
+                .abi
+                .functions
+                .iter()
+                .map(|function| expand_function(function, &self.types))
+                .collect::<Result<Vec<TokenStream>, Error>>()?,
+            true => self
+                .abi
+                .functions
+                .iter()
+                .filter(|function| function.name == "main")
```
MujkicA, line 265: Curious, is a script valid if doesn't include a "main" method? Or in other words, what should be the expected behaviour on our side if not main is fond here?

packages/fuels-test-helpers/src/script.rs
```diff
@@ -15,26 +15,35 @@ use fuels_types::errors::Error;
 /// without having to setup a node or contract bindings.
 pub async fn run_compiled_script(
```
MujkicA, line 16: At some point we should somehow consolidate this, `ScriptBuilder` and the `Script` abstraction into a consistent interface.

Just leaving this as note, not expecting any changes.

MujkicA (changes requested):

segfault-magnet (commented):
I'm wondering if we can achieve something more ergonomic:
```rust
some_macro!(MyScript, path_to_abi);

let instance = MyScript::new(provider);

let result = instance.methods().main(arg1, arg2).call().await?;
```

And have the generated `MyScript::methods` contains a `main` method which will, when called, convert the arguments into tokens and use the saved provider to run the script on a node.

Basically almost identical to what we do with contracts -- might lend itself to some great deduplication as well.

packages/fuels-core/src/code_gen/abigen.rs
iqdecay: Trying to `forc build` a script that doesn't contain a `main` method yields: 
`Script declaration contains no main function. Scripts require a main function.`. 
I think we should probably `panic`, maybe?

packages/fuels-core/src/code_gen/abigen.rs
iqdecay: No it's actually needed for other methods that check the value of the `is_script` boolean. Especially if we try to factorize things down further.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -182,6 +208,7 @@ impl Abigen {
                 #![allow(unused_imports)]
 
                 #includes
+                use fuels::core::abi_encoder::ABIEncoder;
```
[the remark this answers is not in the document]
iqdecay: Honestly I tried, you either end up with two bunch of quotes that look a lot alike (one being included in the other), or you "append"  the `contract` quotes to the basic `script` quotes, thus losing the somewhat alphabetized and clean thing we have. After having tried, I'm chosing to leave it as is, and maybe in the future we will reorganize the way we generate includes.

packages/fuels-core/src/code_gen/abigen.rs
iqdecay: So I actually return an `InvalidData` error when the only method of a script is different from `main` (inside the `generate_script_argument_encoding_function`.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -192,13 +219,53 @@ impl Abigen {
...
+            true => self
+                .abi
+                .functions
+                .iter()
+                .filter(|function| function.name == "main")
+                .map(|function| generate_script_argument_encoding_function(function, &self.types))
+                .collect::<Result<Vec<TokenStream>, Error>>()?,
+        };
 
         Ok(quote! { #( #tokenized_functions )* })
```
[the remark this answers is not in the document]
iqdecay: It does look much better, ty!

hal3e (commented):
Just to remind you to remove the `harness.rs`. Currently `github` shows 4,409 add lines.

iqdecay:
After discussion with @simonr0204 and @digorithm  I decided against doing QoL refactorings in this PR. The point is to get this PR to unblock Simon as quickly as possible, and the improvement in UX will be done in a later PR.

digorithm:
A short story of @iqdecay descending into madness
image

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -87,93 +99,143 @@ impl Abigen {
...
+                use fuels::types::enum_variants::EnumVariants;
+                use std::str::FromStr;
+                use std::collections::{HashSet, HashMap};
+            }
+        }
+    }
 
-                        #fetch_logs
+    pub fn expand_script(&self) -> Result<TokenStream, Error> {
+        let name = &self.contract_name;
```
digorithm, line 213: Setting the script name as `contract_name` could confuse new readers. If that's unavoidable, we should add a comment explaining why we're grabbing the `contract_name` here.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -195,12 +258,21 @@ impl Abigen {
...
+                .abi
+                .functions
+                .iter()
+                .map(|function| expand_function(function, &self.types))
+                .collect::<Result<Vec<TokenStream>, Error>>()?,
+            true => self
+                .abi
+                .functions
+                .iter()
+                .filter(|function| function.name == "main")
```
digorithm, line 272: I understand that `InvalidData` will be returned in the (very unlikely) case of the script not containing `main()`. Having said that, it might be _much_ easier to debug (both for us and for a Sway user completely unaware of the internals of the SDK) if we check, right here, whether we found `main` or not, and pop up a very explicit error if that's not the case.

digorithm (changes requested):
Very good work -- super excited to see the release of this feature!

Left a couple of nits.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -1,14 +1,15 @@
 use std::collections::{HashMap, HashSet};
 
 use crate::code_gen::bindings::ContractBindings;
+use crate::code_gen::functions_gen::generate_script_argument_encoding_function;
```
Br1ght0ne, line 4:
```suggestion
use crate::code_gen::functions_gen::generate_script_main_function;
```

Br1ght0ne (changes requested):
I think this slipped past. Everything else looks cool as heck :)

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,6 +65,50 @@ pub fn expand_function(
...
+    );
+
+    let name = safe_ident("main");
+
+    Ok(quote! {
+        #doc
+        pub async fn #name(&self #(,#arg_declarations)*) -> Result<Vec<Receipt>, SDKError> {
+            let provider = self.wallet.get_provider().expect("Provider not set up").clone();
+            let tokens = [#(#arg_names.into_token()),*];
+            let script_data = ABIEncoder::encode(&tokens)?.resolve(0);
```
segfault-magnet, line 102: Resolving 0 here will not work for vectors, the proper loading address needs to be determined.

I'm for either implementing vector support as a script argument or failing in compile-time if somebody should attempt to call the script gen on such a script until such support is added.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,6 +65,50 @@ pub fn expand_function(
...
+            Ok(run_script_binary(self.binary_filepath.as_str(),
```
segfault-magnet, line 103: Reusing the return-parsing logic from contracts should not be too difficult. Either Data or ReturnData needs to be filtered out and then passed on to the decoder with the correct output type.

iqdecay:
Edited the workflow and PR description, see above.

packages/fuels-contract/src/script.rs
```diff
@@ -403,6 +406,91 @@ impl Script {
...
+pub fn get_decoded_output(
+    output_param: ParamType,
+    receipts: &mut Vec<Receipt>,
+) -> Result<Token, Error> {
+    // Multiple returns are handled as one `Tuple` (which has its own `ParamType`)
+    let (encoded_value, index) = match output_param.get_return_location() {
+        ReturnLocation::ReturnData => {
+            match receipts.iter().find(|&receipt| {
+                matches!(receipt,
+                    Receipt::ReturnData { id: _, data, .. } if !data.is_empty())
```
MujkicA, line 420: This seems to have been copied before a recent bug fix that includes a check for the contract_id here.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -24,12 +25,14 @@ pub struct Abigen {
...
-    /// The contract name as an identifier.
-    contract_name: Ident,
+    /// The contract or script name as an identifier.
+    program_name: Ident,
 
     abi: ProgramABI,
 
     types: HashMap<usize, TypeDeclaration>,
+
+    is_script: bool,
```
MujkicA, line 35: I would split `functions()` into two methods, and avoid having unnecessary indirection. Noting it down on this PR before I forget.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,6 +65,63 @@ pub fn expand_function(
     })
 }
 
+/// Generate the function (as a `TokenStream`) used to encode into script data the arguments to the
+/// `main` function of a deployed script.
```
MujkicA, line 69:
```suggestion
/// `main` function of a script.
```

MujkicA (commented):
Looks good. We should de-duplicate ASAP to avoid more trouble with `get_decoded_output`

packages/fuels-contract/src/script.rs
iqdecay: There is no contract id here as we're executing a script, unless I misunderstood how the contract ids work

packages/fuels-core/src/code_gen/abigen.rs
iqdecay: I'm not sure what you mean

MujkicA: Instead of having the method `functions()` exhibit two behaviours depending on this field, we split it into two methods. `expand_contract` and `expand_script` would then call the corresponding method. 
Ideally, we remove this field entirely.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -96,15 +104,20 @@ pub fn generate_script_main_function(
...
-                                    Some(script_data)).await?
+            let arg_name_tokens = [#(#arg_names.into_token()),*];
+            let script_data = ABIEncoder::encode(&arg_name_tokens)?.resolve(0);
+            let mut receipts = run_script_binary(self.binary_filepath.as_str(),
+                                                    Some(TxParameters::default()),
+                                                    Some(provider),
+                                                    Some(script_data)
             )
+            .await?;
+            let output_param = #output_type::param_type();
```
segfault-magnet, line 117: If the `output_type` turns out to be a type with generic parameters then this is not the correct syntax for calling a method associated with such a type. Rather it would look something like this:
```rust
SomeType::<GenericA,GenericB...>::param_type();
```
This is handled already in the `single_param_type_call` fn.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -77,8 +77,16 @@ pub fn generate_script_main_function(
         ));
     }
 
-    let args = function_arguments(main_function_abi, types)?;
+    let output_type = resolve_fn_output_type(main_function_abi, types)?;
 
+    let args = function_arguments(main_function_abi, types)?;
+    if args.iter().filter(|c| c.field_type.uses_vectors()).count() != 0 {
+        return Err(Error::CompilationError(
+            "Script main function contains a vector in its argument types. This currently isn't \
```
segfault-magnet, line 85: Vectors as contract method inputs are supported. The reason they might not be for scripts is the `resolve()` business if I'm not mistaken. Haven't checked but we should make a task for it.

In any case, returning vectors isn't supported anywhere, so you should add a
`output_type.uses_vectors` guard somewhere.

segfault-magnet (commented):
Looking good. A few gotchas.

`get_decoded_output` should be deduped. There is an identical one used in `contract.rs`.

packages/fuels-core/src/code_gen/functions_gen.rs
iqdecay: The guard for outputs is already contained in `resolve_fn_output_types` :)
Created issue #655.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,18 +66,76 @@ pub fn expand_function(
...
+            let arg_name_tokens = [#(#arg_names.into_token()),*];
+            let script_data = ABIEncoder::encode(&arg_name_tokens)?.resolve(0);
+            let mut receipts = run_script_binary(self.binary_filepath.as_str(),
+                                                    Some(TxParameters::default()),
+                                                    Some(provider),
+                                                    Some(script_data)
+            )
+            .await?;
+            let output_tokens = get_decoded_output(#output_params, &mut receipts)?;
+            let output = #output_type::from_token(output_tokens)?;
```
segfault-magnet, line 121: You're going to have the same issue here as you did earlier with the param type call since from token is also an associated fn.

You can either do the same logic or do the following workaround and leave rust to figure out the type:
```rust
    return Tokenizable::from_token(output_token);
```


Also, rename `output_tokens` to `output_token` since there can only be one.

Also, since this is the second time we're noticing stuff related to generics, add a test with generic script fn args?

Make sure there isn't a third thing we've missed.

packages/fuels-contract/src/script.rs
MujkicA: Keep in mind for the refactor, this can potentially lead to the wrong value being decoded when we call multiple contracts via the script.

packages/fuels-test-helpers/src/script.rs
```diff
@@ -1,65 +1,21 @@
...
 
     #[tokio::test]
     async fn test_run_compiled_script() -> Result<(), Error> {
         // ANCHOR: run_compiled_script
         let path_to_bin = "../fuels/tests/logs/logging/out/debug/logging.bin";
-        let return_val = run_compiled_script(path_to_bin, TxParameters::default(), None).await?;
+        // TODO: use default provider
+        let (provider, _) = setup_test_provider(vec![], vec![], None).await;
+        // Provide `None` to all 3 other arguments so the function uses the default for each
+        let return_val = run_script_binary(path_to_bin, None, Some(provider), None).await?;
```
MujkicA, line 18: Might be cleaner to just move `run_script_binary` to be a provider method or expand on our `Script` type and provide an interface that is similar to how we handle contracts. Just some ideas for the refactoring.

MujkicA (approved):
Looks for now

iqdecay:
PR blocked until https://github.com/FuelLabs/fuels-rs/issues/664 is in! :)

docs/src/testing/scripts.md
```diff
@@ -1,7 +1,19 @@
 # Running scripts
 
-`run_compiled_script` is a helper function for testing simple Sway scripts and reducing boilerplate code related to setting up contracts and deployment. As the argument, it takes the path to the generated binary file (`.bin`).
+`run_compiled_script` is a helper function for testing simple Sway scripts and reducing boilerplate code related to setting up contracts and deployment. As arguments, it takes: 
+- The path to the generated binary file (`.bin`);
+- The transactions parameters as an Option;
```
mohammadfawaz, line 5:
```suggestion
- The transactions parameters as an `Option`;
```

docs/src/testing/scripts.md
```diff
@@ -1,7 +1,19 @@
...
+- The provider as an Option;
```
mohammadfawaz, line 6:
```suggestion
- The provider as an `Option`;
```

packages/fuels-abigen-macro/src/lib.rs
```diff
@@ -22,6 +22,19 @@ pub fn abigen(input: TokenStream) -> TokenStream {
         .into()
 }
 
+/// Abigen proc macro definition and helper functions/types for scripts
+#[proc_macro]
+pub fn script_abigen(input: TokenStream) -> TokenStream {
```
mohammadfawaz, line 27: Does `script_abigen` handle `loggedTypes`? That is, can I get parse logged values from the `Log` receipts and their types from the JSON ABI, and print them nicely?

packages/fuels/tests/scripts/script_generic_arguments/src/main.sw
```diff
@@ -0,0 +1,18 @@
...
+struct GenericBimbam<U> {
+    val: U,
+}
+
+struct GenericSnack<T,V> {
+    twix: GenericBimbam<T>,
+    mars: V,
+}
+
+fn main(bim: GenericBimbam<u8>, bam: GenericSnack<u16,u32>) -> GenericSnack<u64, u32> {
```
mohammadfawaz, line 12: It would be nice to see more complex types in the argument list. For example:
* `Vec`, `Option`, `Result`.
* Enum, array, string, tuple, `b256`, `bool`
I know all of these are tested for contracts, but we should make sure we don't regress for scripts as well.

mohammadfawaz (commented):
Really nice. Thanks @iqdecay. Just some small comments/questions.

packages/fuels-abigen-macro/src/lib.rs
iqdecay: I don't think I currently added support for this. This PR is a first effort into bringing scritp on par with contracts. Then I will PR a big "refactor" for making the script code clearer, then will add the missing features.

packages/fuels/tests/scripts/script_generic_arguments/src/main.sw
iqdecay: Yes I think it's important too, I will add these projects. `Vec` will not be supported yet because of the way it is encoded.

packages/fuels-abigen-macro/src/lib.rs
mohammadfawaz: Sounds good!

packages/fuels/tests/scripts/script_generic_arguments/src/main.sw
iqdecay: Some of these test types I will add in a later PR

iqdecay:
CI is finally passing I'm going to cry lol

iqdecay:
EDIT: it turns out `Tuples` are actually supported now for some unknown reason! Also PRed the refactorization in #681. The features will come later.

packages/fuels-contract/src/script.rs
```diff
@@ -405,6 +408,89 @@ impl Script {
...
+    if let Some(i) = index {
+        receipts.remove(i);
+    }
+    let decoded_value = ABIDecoder::decode_single(&output_param, &encoded_value)?;
+    Ok(decoded_value)
+}
+
+/// Run the script binary located at `binary_filepath` and return its resulting receipts,
+/// without having to setup a node or contract bindings.
+pub async fn run_script_binary(
```
MujkicA, line 453: Since there is no wallet involved, the tx is transmitted without any inputs to cover the fee. This will work as long as the tx parameters are set to default (gas cost = 0). However, if the user supplies their own parameters, this may fail .

iqdecay: Oh yeah I just realized it's super ill-designed… I think the refactor fixes this but not even sure lol. I'm going to test this.

docs/src/testing/scripts.md
```diff
@@ -1,7 +1,19 @@
 # Running scripts
 
-`run_compiled_script` is a helper function for testing simple Sway scripts and reducing boilerplate code related to setting up contracts and deployment. As the argument, it takes the path to the generated binary file (`.bin`).
+`run_compiled_script` is a helper function for testing simple Sway scripts and reducing boilerplate code related to setting up contracts and deployment. As arguments, it takes:
```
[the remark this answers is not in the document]
digorithm: Yes, let's avoid using "simple" (or "easy" or similar adjectives) throughout our docs.

packages/fuels-contract/src/script.rs
digorithm: Hmm... then I believe this should be supported in _this_ PR, not in a subsequent one; No proper scenario will use parameters like gas cost set to zero.

packages/fuels-contract/src/script.rs
```diff
@@ -405,6 +408,89 @@ impl Script {
...
+    binary_filepath: &str,
+    tx_params: Option<TxParameters>,
+    provider: Option<Provider>,
+    script_data: Option<Vec<u8>>,
+) -> Result<Vec<Receipt>, Error> {
+    let script_binary = std::fs::read(binary_filepath)?;
+    let provider = match provider {
+        None => {
+            // let server = FuelService::new_node(Config::local_node()).await.unwrap();
+            // Provider::connect(server.bound_address.to_string()).await?
```
digorithm, line 463: No need to keep this, then.
```suggestion
```

packages/fuels-contract/src/script.rs
iqdecay: Yes I'm going to refactor and test this inside this PR

packages/fuels-contract/src/execution_script.rs
```diff
@@ -0,0 +1,130 @@
...
+use crate::contract::ContractCall;
+use crate::contract_calls_utils::{
+    build_script_data_from_contract_calls, calculate_required_asset_amounts, get_data_offset,
+    get_instructions, get_transaction_inputs_outputs,
+};
+
+/// TransactionExecution provides methods to create and a call/simulate a transaction that carries
+/// out contract method calls or script calls
+#[derive(Debug)]
+pub struct TransactionExecution {
```
hal3e, line 25: I would change `TransactionExecution` to something else. IMO it is a bit confusing. Unfortunately, I do not have a good alternative name. Maybe some other reviewer will have something...

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -43,7 +43,7 @@ impl Abigen {
         Ok(Self {
             types: Abigen::get_types(&parsed_abi),
             abi: parsed_abi,
-            contract_name: ident(contract_name),
+            program_name: ident(contract_name),
```
hal3e, line 46: I would just use `name`.. No need to make it confusing with `program`
```suggestion
            name: ident(contract_name),
```

hal3e (dismissed):
Great work! I like how you implemented this! :tada:  I have left some change requests below

docs/src/getting-started/running-scripts.md
```diff
@@ -0,0 +1,15 @@
+# Running scripts
+
+You can run a script using its JSON-ABI and the path to its binary file. You can run the scripts with arguments. For this, you have to use the `script_abigen!` macro, which is not unlike the `abigen!` macro seen [previously](../contracts/the-abigen-macro.md).
```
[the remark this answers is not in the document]
Br1ght0ne: I think `not unlike == not much different from == similar`. Probably could use a simpler rewording though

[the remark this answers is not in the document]
iqdecay: will reword to `similar`

packages/fuels-contract/src/execution_script.rs
iqdecay: Yeah I'm not really a fan of this naming either…

Br1ght0ne:
(review is WIP)

I created #704 (which targets your branch), containing some docs updates I found unreasonably long for suggestions :)
Please don't hesitate to merge this PR at any time, as #704 will be retargeted to `master` after that.

packages/fuels-contract/src/script_calls.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use crate::execution_script::{CompiledScript, TransactionExecution};
+
+pub struct ScriptInterface {
+    pub script: TransactionExecution,
+    pub wallet: WalletUnlocked,
+}
+
+#[derive(Debug)]
+pub struct ScriptCallResponse<D> {
```
MujkicA, line 25: This seems to implement the same behaviour as `ContractCallResponse`. Keeping both is more mental strain for the user and cumbersome to maintain. 
How about generalising this to something like `VMResponse`

Another thing to keep in mind is that #690 moves log methods to the response. ATM I'm unsure of what implication logging has in the context of scripts.

packages/fuels-contract/src/script_calls.rs
```diff
@@ -0,0 +1,181 @@
...
+            value,
+            gas_used: Self::get_gas_used(&receipts),
+            receipts,
+        }
+    }
+}
+
+#[derive(Debug)]
+/// Contains all data relevant to a single script call
+pub struct ScriptCall {
```
MujkicA, line 55: In the case of contracts, we keep `ContractCall` and `ContractCallHandler` separate to make multi-calls easier to construct. In the case of scripts, we don't have that requirement, so we could combine them into a single struct. 

Wether that is a cleaner design, in the end, is for you to decide, but we should hide `ScriptCall` from the user at any rate to avoid confusion.

Ideally we then have `ScriptCallHandler` (or maybe `ScriptHandler`) as the only entry point to doing something with scripts in the SDK. It replaces the functionality of `run_compiled_script` and enables main arguments.

packages/fuels-contract/src/execution_script.rs
MujkicA: Maybe a trait would be more appropriate here. Something that implies that the transaction will be processed by the VM interpreter. Is `ExecutableTransaction` too vague?

MujkicA (commented):
Overall I think this is step in the right direction, but could use some more polish to unify the UX. I left some thoughts.

packages/fuels-contract/src/script_calls.rs
```diff
@@ -0,0 +1,187 @@
...
+    parameters::{CallParameters, TxParameters},
+    Tokenizable,
+};
+use fuels_signers::{provider::Provider, WalletUnlocked};
+
+use fuels_types::{errors::Error, param_types::ParamType};
+
+use crate::execution_script::{CompiledScript, TransactionExecution};
+
+pub struct ScriptInterface {
```
MujkicA, line 19: This seems to be unused

packages/fuels-contract/src/script_calls.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+#[derive(Debug)]
+/// Contains all data relevant to a single script call
+pub struct ScriptCall {
+    pub compiled_script: CompiledScript,
+    pub script_data: Vec<u8>,
+    pub inputs: Vec<Input>,
+    pub outputs: Vec<Output>,
+    pub call_parameters: CallParameters,
```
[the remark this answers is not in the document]
iqdecay: I left it there because I misunderstood what `CallParameters` are for and used in. Do you confirm that they would not be useful in a script call, even if the script calls some contracts?

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -197,17 +220,69 @@ impl Abigen {
         })
     }
 
-    pub fn functions(&self) -> Result<TokenStream, Error> {
+    /// Generates the includes necessary for the abigen.
+    /// TODO(iqdecay): change the necessary imports on scripts
```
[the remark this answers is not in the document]
iqdecay: It hasn't, I will take care of it.

packages/fuels-contract/src/contract.rs
```diff
@@ -404,51 +440,58 @@ impl ContractCall {
...
+                None => (vec![], None),
             }
-        };
-        if let Some(i) = index {
-            receipts.remove(i);
         }
-
-        let decoded_value = ABIDecoder::decode_single(&self.output_param, &encoded_value)?;
-        Ok(decoded_value)
+    };
```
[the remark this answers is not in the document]
iqdecay: Committed to the branch, thanks!

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -65,18 +66,66 @@ pub fn expand_function(
...
+
+    Ok(quote! {
+        #doc
+        pub fn #name(&self #(,#arg_declarations)*) -> ScriptCallHandler<#output_type> {
+            let arg_name_tokens = [#(#arg_names.into_token()),*];
+            let script_data = ABIEncoder::encode(&arg_name_tokens).expect("Cannot encode script
+            arguments").resolve(0);
+            let compiled_script = CompiledScript::from_binary_filepath(self.binary_filepath
+                .as_str()).expect("Could not read from binary filepath");
+            let provider = self.wallet.get_provider().expect("Provider not set up").clone();
```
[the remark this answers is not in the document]
iqdecay: Committed to the branch too, thanks!

packages/fuels-contract/src/script_calls.rs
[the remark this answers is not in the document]
MujkicA: In the case of contract calls, `CallParameters` are possible because we control what addresses the script data is loaded from since we build the calling script fusing fuel asm opcodes.
When we load a script directly from a binary, we have no control over the addresses the data is being loaded from.

I am sure of the above, but I wonder what would happen if we were to load a binary of a script that performs a call. Does forc embed the data into the binary?

packages/fuels-contract/src/script_calls.rs
MujkicA: It is possible. But if you have a script that does multi-calls, all the relevant information is contained by a single `ScriptCall`:
- script binary
- script data
- inputs
- outputs

`ContractCall` holds information specific to a single call e.g. selector, contract id, output variables. Note all of these would be contained in the above fields if we use 'ScriptCall'.

packages/fuels-contract/src/execution_script.rs
MujkicA: On second thought, `ExecutableTransaction` might not be ideal since fuel-vm implements a trait with the same name and we make se of it in our provider.

packages/fuels-contract/src/script_calls.rs
iqdecay: Got it, thanks for clarifying! I'm removing it, I had added it for extensibility.

packages/fuels-contract/src/script_calls.rs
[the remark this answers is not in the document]
iqdecay: This is outside my knowledge, @digorithm  @mohammadfawaz?

packages/fuels-contract/src/execution_script.rs
digorithm: Good call y'all. `TransactionExecution` had me very confused. 

`ExecutableSDKTransaction ` is a no for me.

I'm also not liking having `VM` all over the codebase now. It's not very descriptive; what VM? I'd suggest replacing `VM` with `Fuel`. In this case, `ExecutableFuelCall`. SDK users won't be very interested in lower-level concepts, such as virtual machines (that's for us, SDK developers, to take care of). 

The same applies to `VMCallResponse` I've been seeing in this PR. `FuelCallResponse` would be at a more appropriate abstraction level.

iqdecay: Going for `FuelCallResponse` and `ExecutableFuelCall` then, unless more objections!

packages/fuels-contract/src/call_response.rs
```diff
@@ -0,0 +1,36 @@
+use fuel_tx::Receipt;
+
+/// [`VMCallResponse`] is a struct that is returned by a call to the contract or script. Its value
```
digorithm, line 3:
```suggestion
/// [`FuelCallResponse `] is a struct that is returned by a call to the contract or script. Its value
```

packages/fuels-contract/src/execution_script.rs
iqdecay: Do you think `ExecutableFuelCall` should be a trait? If it's a trait it would only have one implementation AFAIU, so I'm not sure it's really the best idea. @digorithm @Br1ght0ne @MujkicA @hal3e

packages/fuels-contract/src/contract_calls_utils.rs
```diff
@@ -0,0 +1,718 @@
...
+/// Specifies offsets of [`Opcode::CALL`] parameters stored in the script
+/// data from which they can be loaded into registers
+pub(crate) struct CallOpcodeParamsOffset {
+    pub asset_id_offset: usize,
+    pub amount_offset: usize,
+    pub gas_forwarded_offset: usize,
+    pub call_data_offset: usize,
+}
+
+pub(crate) fn calculate_required_asset_amounts(calls: &[ContractCall]) -> Vec<(AssetId, u64)> {
```
digorithm, line 28: This needs a doc-level comment.

packages/fuels-contract/src/contract_calls_utils.rs
```diff
@@ -0,0 +1,718 @@
...
+    let amounts_per_asset_id = calls
+        .iter()
+        .map(|call| (call.call_parameters.asset_id, call.call_parameters.amount));
+    sum_up_amounts_for_each_asset_id(amounts_per_asset_id)
+}
+
+fn sum_up_amounts_for_each_asset_id(
```
digorithm, line 35: This needs a doc-level comment.

packages/fuels-contract/src/execution_script.rs
digorithm: Hmmm, I'm not 100% sold on the idea of this being a trait 🤔

packages/fuels-contract/src/script_calls.rs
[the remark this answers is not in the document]
digorithm: I'm also unsure here, will do some digging.

> I am sure of the above, but I wonder what would happen if we were to load the binary of a script that performs a call. Does forc embed the data into the binary?

@iqdecay can you test this?

packages/fuels-contract/src/script_calls.rs
```diff
@@ -0,0 +1,155 @@
...
+    async fn call_or_simulate(&self, simulate: bool) -> Result<FuelCallResponse<D>, Error> {
+        let mut tx = Transaction::script(
+            self.tx_parameters.gas_price,
+            self.tx_parameters.gas_limit,
+            self.tx_parameters.maturity,
+            self.script_call.compiled_script.script_binary.clone(),
+            self.script_call.script_data.clone(),
+            self.script_call.inputs.clone(),  // TODO(iqdecay)
+            self.script_call.outputs.clone(), // TODO(iqdecay)
+            vec![vec![0, 0].into()],          //TODO(iqdecay): witnesses
```
digorithm, line 120: Need more context on these `TODO`s here!

digorithm (changes requested):
Feels like we're getting close 😄 

This PR would've been easier to review if we had separated new functionality from refactoring. 

In most cases, a little refactoring makes sense for a certain change. But, in this case, the change is significant and complex, and adding complex refactoring work on top of it makes it tough to review.

Just something to keep in mind for the next time!

packages/fuels-contract/src/script_calls.rs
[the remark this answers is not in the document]
iqdecay: I will test this in a PR related to contract calls. Leaving the `call_parameters` field here for now.

packages/fuels-contract/src/execution_script.rs
iqdecay: Ok so I'm not making it a trait for now then.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -184,17 +205,89 @@ impl Abigen {
...
+            };
+            quote! {
+                use fuels::contract::logs::LogDecoder;
+                use fuels::core::types::*;
+                use fuels::core::{Tokenizable, Token, Parameterize, try_from_bytes};
+                use fuels::signers::WalletUnlocked;
+                use fuels::types::enum_variants::EnumVariants;
+                use fuels::types::errors::Error as SDKError;
+                use fuels::types::param_types::ParamType;
+                #specific_includes
```
MujkicA, line 258:
```suggestion
                    use fuels::types::{
                        ResolvedLog,
                        bech32::Bech32ContractId
                    };
                    use std::collections::{HashSet, HashMap};
                    use std::str::FromStr;
                }
            };
            quote! {
                use fuels::contract::logs::LogDecoder;
                use fuels::core::types::*;
                use fuels::core::{Tokenizable, Token, Parameterize, try_from_bytes};
                use fuels::signers::WalletUnlocked;
                use fuels::types::enum_variants::EnumVariants;
                use fuels::types::errors::Error as SDKError;
                use fuels::types::param_types::ParamType;
                use fuels::tx::{ContractId, Address, Receipt};
                #specific_includes
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -184,17 +205,89 @@ impl Abigen {
...
+                        ContractCallHandler,
+                        get_decoded_output
+                    };
+                    use fuels::core::abi_decoder::ABIDecoder;
+                    use fuels::core::code_gen::{
+                        function_selector::resolve_fn_selector,
+                        get_logs_hashmap,
+                    };
+                    use fuels::core::{EnumSelector, StringToken, Identity};
+                    use fuels::tx::{ContractId, Address, Receipt};
```
MujkicA, line 241:
```suggestion
```

packages/fuels-core/src/code_gen/abigen.rs
MujkicA, line 258:
```suggestion
                use fuels::tx::{ContractId, Address, Receipt};
                #specific_includes
```

MujkicA (commented):
Scripts with `ContractId`, `Address` as main argument will fail because they need this import.
 
[Related](https://github.com/FuelLabs/sway-applications/pull/304)

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -71,18 +75,77 @@ pub fn expand_function(
...
+    Ok(quote! {
+        #doc
+        pub fn #name(&self #(,#arg_declarations)*) -> ScriptCallHandler<#output_type> {
+            let arg_name_tokens = [#(#arg_names.into_token()),*];
+            let script_data = ABIEncoder::encode(&arg_name_tokens).expect("Cannot encode script
+            arguments").resolve(0);
+            let script_binary = std::fs::read(self.binary_filepath.as_str())
+                                        .expect("Could not read from binary filepath");
+            let provider = self.wallet.get_provider().expect("Provider not set up").clone();
+            // TODO(iqdecay): handle log decoding in scripts
```
digorithm, line 123: Well, goes without saying that log decoding is a high priority, and we should address it asap! Issue created here: https://github.com/FuelLabs/fuels-rs/issues/713.

digorithm (approved):
Alright, I know I said this before, but again: this was a bit rough to review. Lots of big refactoring moves that add a lot to the review process, on top of the essential complexity of the _actual_ changes, which was also great (understandably).

Took me a long time to review it all, multiple times, but in large PRs, there are always risks of something wrong slipping into the codebase. That's why smaller PRs are a lot better for the overall health of the codebase. 

Other than this issue, it looks good to me. Excited to release this! 

Important/urgent follow-up works: `Vec` support and log decoding in scripts.

PR #621, opened 2022-10-12
Some minor cosmetic updates to the SDK Sway tests

mohammadfawaz:
This started because I noticed that `__log` is used in the `fuels-rs` book instead of `std::logging::log`, so I wanted to update that in the tests. Users should really not be using intrinsics directly (intrinsics always start with `__`) which is why we provide wrappers around them in the standard library.

Then I noticed that a whole bunch of things could be cosmetically improved such as:
* Removing unnecessary imports due to the standard library prelude.
* Running the new formatter.


digorithm (approved):
Ah, very nice -- hadn't realized how outdated some of the SDK's Sway code was. Thanks for doing this!

PR #622, opened 2022-10-12
feat!: Enforce Bech32 type-safety on contract instance creation

### LINKED ISSUE #618: Make creating a contract instance more ergonomic
Users have to jump through hoops to create a contract instance. We work with bech32 encoded ids, while users often have a hex id from a previously deployed contract.

```rust
// some contract id the user received from fuel-core or forc
let id = "0x76cdb8a504479867b9157aae6c3fcfa03d99fa815fc7afb707adc44a5e9b2e95";
let id = ContractId::from_str(id).unwrap();
let bech32_id = Bech32ContractId::from(id);

let contract_instance = MyContract::new(
        bech32_id.to_string(),
        wallet.clone(),
);
```

We could make this more ergonomic by having the `FromStr` implementation detect whether the provided string is hex or bech32.

MujkicA:
Closes #618

~~Constructing a `Bech32Address` or `Bech32ContractId` from a string now first tries to decode it as bech32. If that fails for reasons other than an invalid checksum, the string is treated as a regular hex string.~~

Creating an abstraction over bech32 or hex strings causes us to loose the benefit of error detection that bech32 brings.  
I've changed `MyContract:new()` to accept a `Bech32ContractId` instead of a string. This way we avoid the back and forth conversion that happens when deploying and calling new. 
The docs have been updated to include examples on how to create an instance from both hex and bech32 format strings.


digorithm (commented):
Very nice, major QoL improvement. Hopefully, no weird errors will come from it! 

Could you also add an example for the book where you instantiate the same contract instance using both a bech32 and its hex equivalent?

packages/fuels-types/src/bech32.rs
```diff
@@ -114,32 +137,108 @@ mod test {
...
+                let bech32_contract_id = &Bech32ContractId::from_str(hex_encodings[i]).unwrap();
+                assert_eq!(*bech32_contract_id.hash(), pubkey_hashes[i]);
+            }
+        }
+
+        {
+            for i in 0..3 {
+                let bech32_contract_id = &Bech32Address::from_str(bech32m_encodings[i]).unwrap();
+                assert_eq!(*bech32_contract_id.hash(), pubkey_hashes[i]);
+            }
```
hal3e, line 215:
```suggestion
            for (b32m_e, pbkh) in bech32m_encodings.iter().zip(pubkey_hashes) {
                let bech32_contract_id = &Bech32ContractId::from_str(b32m_e).unwrap();
                assert_eq!(*bech32_contract_id.hash(), pbkh);
            }

            for (hex_e, pbkh) in hex_encodings.iter().zip(pubkey_hashes) {
                let bech32_contract_id = &Bech32ContractId::from_str(hex_e).unwrap();
                assert_eq!(*bech32_contract_id.hash(), pbkh);
            }
        }

        {
            for (b32m_e, pbkh) in bech32m_encodings.iter().zip(pubkey_hashes) {
                let bech32_contract_id = &Bech32Address::from_str(b32m_e).unwrap();
                assert_eq!(*bech32_contract_id.hash(), pbkh);
            }
```

packages/fuels-types/src/bech32.rs
```diff
@@ -114,32 +137,108 @@ mod test {
...
 
-        let expected_contract_id = ContractId::new(pubkey_hash);
+            for i in 0..3 {
+                let bech32_contract_id = &Bech32Address::from_str(hex_encodings[i]).unwrap();
+                assert_eq!(*bech32_contract_id.hash(), pubkey_hashes[i]);
+            }
```
hal3e, line 220:
```suggestion
            for (hex_e, pbkh) in hex_encodings.iter().zip(pubkey_hashes) {
                let bech32_contract_id = &Bech32Address::from_str(hex_e).unwrap();
                assert_eq!(*bech32_contract_id.hash(), pbkh);
            }
```

packages/fuels-types/src/errors.rs
```diff
@@ -27,6 +27,8 @@ impl From<Utf8Error> for CodecError {
 
 #[derive(Error, Debug)]
 pub enum Error {
+    #[error("General error: {0}")]
+    GeneralError(String),
```
hal3e, line 31: I do not like the name `GeneralError` so much. I understand why you would give it as this error is constructed from any `str`.
The best solution would be if we could update the error type in `fuel-types`. Until then maybe we call it `FromStrError` ?

hal3e (commented):
LGTM. Left some suggestions.

packages/fuels-types/src/errors.rs
MujkicA: I agree that `GeneralError` is not ideal. I wanted to give this some more thought, but it got lost in the process. On the other hand, I think going with `FromStrError` adds to the proliferation of one-off error types.

examples/types/src/lib.rs
```diff
@@ -86,6 +86,12 @@ mod tests {
             Bech32Address::from_str(string).expect("failed to create Bech32 address from string");
         assert_eq!([0u8; 32], *bech32_address.hash());
 
+        // From a hex string.
+        let string = "0x0000000000000000000000000000000000000000000000000000000000000000";
```
iqdecay, line 90: this is technically not a string but a `&str`, I'm not sure the comment is good here.  Same remark with the ones above, and maybe we could implement an "auto" conversion from `String` -> `bech32_address`.

packages/fuels-types/src/bech32.rs
```diff
@@ -114,32 +137,108 @@ mod test {
...
+    fn test_from_invalid_bech32_string() {
+        {
+            let expected: Error = bech32::Error::InvalidChecksum.into();
+            let invalid_bech32_with_hex_chars =
+                "fadeaffe1dfed7c25adadadd715e3f309ecf07dca4e3a6d9de9cdadcddac50d7f";
+
+            {
+                let result = &Bech32ContractId::from_str(invalid_bech32_with_hex_chars)
+                    .expect_err("should error");
+                assert_eq!(result.to_string(), expected.to_string());
```
iqdecay, line 234: Here I think you should check for the error, even though IIUC, it's going to be `GeneralError` which is very broad.

packages/fuels-types/src/bech32.rs
```diff
@@ -114,32 +137,108 @@ mod test {
...
+            }
+
+            {
+                let result = &Bech32Address::from_str(invalid_bech32_with_hex_chars)
+                    .expect_err("should error");
+                assert_eq!(result.to_string(), expected.to_string());
```
iqdecay, line 240: Same

packages/fuels-types/src/bech32.rs
iqdecay:
```suggestion
            for (hex_encoding, pubkey_hash) in hex_encodings.iter().zip(pubkey_hashes) {
                let bech32_contract_id = &Bech32Address::from_str(hex_encoding).unwrap();
                assert_eq!(*bech32_contract_id.hash(), pubkey_hash);
            }
```
I love verbose lol

iqdecay (changes requested):
Looking good, some nits :)

digorithm (approved):
Awesome work; type-safety ftw :-)

Br1ght0ne (approved):
LGTM, love to see changes like this!

hal3e (approved):

PR #623, opened 2022-10-13
fix: Ignore lock files and out/ directory

digorithm:


iqdecay (changes requested):
I think you're still adding the targets? But I'm not sure

digorithm:
> I think you're still adding the targets? But I'm not sure

Nope, this is the current `git status` output:
image

digorithm:
> > I think you're still adding the targets? But I'm not sure
> 
> Nope, this is the current `git status` output: image

But I'm going to add to the ignore file just in case my fs is doing some funny business haha.

hal3e (approved):

iqdecay (approved):

PR #625, opened 2022-10-13
feat: concurrently build test projects

hal3e:
This PR is based on @segfault-magnet's work from the `e2e` repository. 

We updated the script to run the builds concurrently. You can change the number of concurrent futures with an environment variable like so:

```shell
 NUM_CONC_BUILDS=12 cargo run --bin build-test-projects
```
the default value is currently set to  1. Which means that there is no concurrency. Once `forc` can build multiple projects concurrently we will update the default.


hal3e:
I set the default num of current builds to 1. Which means that there is no concurrency. Once `forc` can build multiple projects concurrently we will update the default.

segfault-magnet (approved):
so clean you can eat off it.

digorithm (approved):
Very nice, LGTM!

PR #628, opened 2022-10-15
Contract input set estimation

Salka1988:
Add `set_contracts_automatic` which automatically adds missing `contract_id`s.

blocked by: 
https://github.com/FuelLabs/fuel-core/pull/702

In addition, we will need a new release of `fuel-core` that supports `transparent_receipt::Receipt'` as the return value of the `dry_run'` function.

~~Important:
I noticed that introducing the contract_id in Panic and all the necessary logic in fuel-core drastically increases the gas_fee, which can be seen from the tests.~~


Salka1988:
Blocked by `fuel-core`

packages/fuels/tests/contracts.rs
```diff
@@ -645,3 +645,70 @@ async fn contract_call_futures_implement_send() -> Result<(), Error> {
...
+    let bits = *foo_contract_id.hash();
+    let contract_methods = foo_caller_contract_instance.methods();
+
+    let mut multi_call_handler = MultiContractCallHandler::new(wallet.clone());
+    multi_call_handler.tx_params(Default::default());
+
+    (0..3).for_each(|_| {
+        let call_handler = contract_methods.call_foo_contract(Bits256(bits), true);
+        multi_call_handler.add_call(call_handler);
+    });
```
hal3e, line 705: I would add three different `call_handlers` with different contract calls just in case. Maybe even mix in one that doesn't use a contract as an input.

hal3e (commented):
LGTM. Good work :tada:

packages/fuels/tests/contracts.rs
Salka1988: I would have to add two more `sway` files that would do the same. Maybe I shouldn't do that, but in the multicall test I added a call that doesn't need a contract id

packages/fuels-contract/src/contract.rs
```diff
@@ -438,6 +443,12 @@ impl ContractCall {
         let decoded_value = ABIDecoder::decode_single(&self.output_param, &encoded_value)?;
         Ok(decoded_value)
     }
+
+    fn check_if_contract_not_in_inputs(receipts: &[Receipt]) -> Option<&Receipt> {
```
MujkicA, line 447: Nit: for me `check_` usually implies that the method is returning a boolean. Maybe something like `find_contract_not_in_inputs` is more descriptive.

packages/fuels-contract/src/contract.rs
```diff
@@ -550,14 +581,23 @@ where
...
+            let result = self.simulate_without_decode().await;
 
             match result {
                 Err(Error::RevertTransactionError(_, receipts))
                     if ContractCall::is_missing_output_variables(&receipts) =>
                 {
                     self = self.append_variable_outputs(1);
                 }
+
+                Err(Error::RevertTransactionError(_, receipts)) => {
```
MujkicA, line 593: At a glance, given this branching structure, if an `Error::RevertTransactionError` occurs that is neither due to missing variables or a missing contract, the loop will continue. Instead, the error should be propagated.

packages/fuels-contract/src/contract.rs
```diff
@@ -694,6 +734,17 @@ impl MultiContractCallHandler {
                         .take(1)
                         .for_each(|call| call.append_variable_outputs(1));
                 }
+
+                Err(Error::RevertTransactionError(_, receipts)) => {
+                    if let Some(receipt) = ContractCall::check_if_contract_not_in_inputs(&receipts)
```
MujkicA, line 739: Same as above

packages/fuels-contract/src/script.rs
```diff
@@ -409,8 +410,9 @@ impl Script {
         if receipts
             .iter()
             .any(|r|
-                matches!(r, Receipt::ScriptResult { result, .. } if *result != ScriptExecutionResult::Success)
-        ) {
+                matches!(r, Receipt::ScriptResult { result, .. } if *result != ScriptExecutionResult::Success) |
+                    matches!(r, Receipt::Panic { reason, .. } if *reason.reason() == PanicReason::ContractNotInInputs )
```
MujkicA, line 414: I'm pretty sure the execution result should != `ScriptExecutionResult::Success` if there is missing contracts. No need for an additional check.

packages/fuels/tests/contracts/foo_caller_contract/src/main.sw
```diff
@@ -18,4 +19,8 @@ impl FooCaller for Contract {
 
         !response
     }
+
+    fn return_bool(value: bool) -> bool {
```
MujkicA, line 23: Nit: use a different contract entirely to "add call that does not need ContractId". 
`foo_caller_contract` is very specific in what it does, having this method here seems out of place.

packages/fuels/tests/contracts.rs
```diff
@@ -645,3 +645,76 @@ async fn contract_call_futures_implement_send() -> Result<(), Error> {
...
+        foo_caller_contract_instance,
+        None,
+        "packages/fuels/tests/contracts/foo_caller_contract"
+    );
+
+    let bits = *foo_contract_id.hash();
+    let res = foo_caller_contract_instance
+        .methods()
+        .call_foo_contract(Bits256(bits), true)
+        .estimate_tx_dependencies(None)
```
MujkicA, line 671: Nit: confirm the call fails first before trying with `estimate_tx_dependencies`. Similar to how it is done in `test_output_variable_estimation`

packages/fuels/tests/contracts.rs
```diff
@@ -645,3 +645,95 @@ async fn contract_call_futures_implement_send() -> Result<(), Error> {
...
+    setup_contract_test!(
+        foo_caller_contract_instance,
+        None,
+        "packages/fuels/tests/contracts/foo_caller_contract"
+    );
+
+    let bits = *foo_contract_id.hash();
+
+    {
+        // Should fail due to lack of output variables
```
MujkicA, line 670:
```suggestion
        // Should fail due to missing external contracts
```

MujkicA (approved):

packages/fuels-contract/src/contract.rs
```diff
@@ -550,14 +581,24 @@ where
...
                 {
                     self = self.append_variable_outputs(1);
                 }
+
+                Err(Error::RevertTransactionError(_, ref receipts)) => {
+                    if let Some(receipt) = ContractCall::find_contract_not_in_inputs(receipts) {
+                        let contract_id = Bech32ContractId::from(*receipt.contract_id().unwrap());
+                        self = self.append_contract(contract_id);
+                    } else {
+                        return Err(result.err().unwrap());
```
digorithm, line 598: This error here will be _very_ cryptic when returned to the user. 

The user will be trying to `estimate_tx_dependencies`, but if it _is_ an `Error::RevertTransactionError` and for some weird reason that's out of our control, `ContractCall::find_contract_not_in_inputs` returns `None`, the `Err` here will be _very_ hard to understand.

So we should return this `Err` with a string saying something like `"Couldn't estimate tx dependencies because we couldn't find the missing contract input"`.

packages/fuels-contract/src/contract.rs
```diff
@@ -694,6 +735,19 @@ impl MultiContractCallHandler {
...
+
+                Err(Error::RevertTransactionError(_, ref receipts)) => {
+                    if let Some(receipt) = ContractCall::find_contract_not_in_inputs(receipts) {
+                        let contract_id = Bech32ContractId::from(*receipt.contract_id().unwrap());
+                        self.contract_calls
+                            .iter_mut()
+                            .take(1)
+                            .for_each(|call| call.append_external_contracts(contract_id.clone()));
+                    } else {
+                        return Err(result.err().unwrap());
```
digorithm, line 747: Same comment about the returned error.

packages/fuels-contract/src/contract.rs
```diff
@@ -694,6 +735,19 @@ impl MultiContractCallHandler {
                         .take(1)
                         .for_each(|call| call.append_variable_outputs(1));
                 }
+
+                Err(Error::RevertTransactionError(_, ref receipts)) => {
+                    if let Some(receipt) = ContractCall::find_contract_not_in_inputs(receipts) {
+                        let contract_id = Bech32ContractId::from(*receipt.contract_id().unwrap());
+                        self.contract_calls
+                            .iter_mut()
+                            .take(1)
```
digorithm, line 744: Wait, what's the purpose of this `take(1)` in this context?

digorithm (changes requested):
Very good work! Left one suggestion to improve error handling.

packages/fuels-contract/src/contract.rs
Salka1988: There are more `ContractCalls` in `pub contract_calls: Vec<ContractCall`. In this way, I put the `contract_id` only in the `ContractCall` where it should be, that is, in the one that calls the missing `contract_id`.

hal3e (approved):

packages/fuels-contract/src/contract.rs
MujkicA: If `ContractCall::find_contract_not_in_inputs` returns `None` it should mean that the `Error::RevertTransactionError` was due to reasons other than missing output variables or contract inputs. We should forward the original error to the user in that case since it cannot be fixed by estimating dependencies anyways.

Salka1988: This was the initial idea.

digorithm (approved):
LGTM -- great work @Salka1988!

PR #636, opened 2022-10-17
refactor(fuels-core): add tests for resolved_type

### LINKED ISSUE #635: Write unit tests for `resolved_type`
From @segfault-magnet in #568:

> `resolved_type` should be unit tested to ensure we're resolving all the different types available (see `ParamTypes`).
> The `From<&ResolvedType> for TokenStream` should be tested as well.

Br1ght0ne:
Close #635.


segfault-magnet (commented):
Love the dedupe effort :) 

I'd bake in some generics when testing structs and enums. Just so that we cover it in any capacity.

And vector tests are still pending. If you have any pesky issues, feel free to reach out :)

segfault-magnet (commented):
Unfortunately, vec ABI changed in forc v0.28.0+ and we've adapted to it here:
https://github.com/FuelLabs/fuels-rs/pull/651

Since #651 is quite urgent I have a feeling it's going to be merged before this is. If that turns out to be the case, then the ABI for vecs would need to be updated to how forc now generates it (with the raw ptr).

segfault-magnet (approved):

Issue #605, opened 2022-10-04
Increase `Provider` on-chain information coverage


digorithm:
Currently, our `Provider` lacks more APIs to access on-chain information. These are essentially wrappers on `fuel-core`'s API, but since the SDK is usually the layer users will be working with, `Provider` needs to provide access to more on-chain information. The extra information we need to provide:

- [x] Blocks (wrapper on `fuel-core`'s `block` API)
- [x] Coins (wrapper on `fuel-core`'s `coin` and `coins` API)
- [x] Contract and contract balance (wrapper on `fuel-core`'s `contract` and `contract_balance` API)

digorithm:
This seems to be done, closing it.

Issue #350, opened 2022-05-29
test: add `Provider` balance API unit tests 


vnepveu:
They are currently only tested throught the `harness`, which isn't what unit tests aresupposed to do.

PR #637, opened 2022-10-17
feat: add new provider api and tests

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/605 and https://github.com/FuelLabs/fuels-rs/issues/350

I have updated provider with the new APIs and added respective unit tests.


iqdecay (changes requested):
Looking really good

Salka1988 (approved):
:+1:

digorithm (dismissed):
👍

Br1ght0ne (dismissed):
👍

packages/fuels-signers/src/provider.rs
```diff
@@ -321,6 +326,14 @@ impl Provider {
         Ok(balances)
     }
 
+    pub async fn get_contract(
```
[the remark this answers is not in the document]
hal3e: I added a comment but moved the code a bit so it is not reflected here.

segfault-magnet (dismissed):

digorithm:
This is _really_ cool @segfault-magnet, I love it. But, I believe we should implement and experiment with this on the events querying issue you're working on; it's a lot more likely the amount logs will be huge, and users will likely want to query it in near real-time, which justifies using streams and having this wonderful interface you came up with. 

For the case of the current `Provider` APIs, I think we should keep it simple, _for now_, and avoid premature optimization. I'd say if the client's API takes in a `PaginationRequest`, the SDK wrapper API should also take it as well. cc @hal3e

packages/fuels-signers/src/provider.rs
```diff
@@ -284,7 +289,12 @@ impl Provider {
                      owner: _,
                      amount,
                      asset_id,
-                 }| (asset_id.to_string(), amount.try_into().unwrap()),
+                 }| {
+                    (
+                        AssetId::from_str(&asset_id.to_string()).unwrap(),
```
MujkicA, line 294: Is there a reason why this isn't just `asset_it`

packages/fuels-signers/src/provider.rs
```diff
@@ -432,4 +491,352 @@ impl Provider {
...
+    }
+
+    #[tokio::test]
+    async fn test_coin_api() -> Result<(), Error> {
+        let (_, (coins, _), provider) = setup_provider_api_test().await;
+
+        let (coin_id, _) = &coins[0];
+        let hex_coin_id = format!("{:#x}", coin_id);
+
+        let expected_coin = provider
```
MujkicA, line 541: Shouldn't "expected" be the hardcoded value as opposed to the value retrieved from the tested method?

packages/fuels-signers/src/provider.rs
hal3e: Yes. `AssetId` here is `fuels_types::AssetId` and `asset_id` is `fuel_gql_client::client::schema::AssetId`. There is no direct conversion.

packages/fuels-signers/src/provider.rs
hal3e: Yes, you are right :smile:  I will update the tests.

hal3e:
In the last commit, I have added a POC for the new provider API. The code is a bit more complicated than before but the UX is IMO much better. The idea was to give the user the possibility to do the pagination but still make the new API clean and easy to use. 
As this is a POC I "converted" only two functions to get your opinion. If  you like it I will convert the rest. Let's compare the old and new `get_coins` API. Previously we would return all coins associated to an address and there was no way to change it..
```rust
let coins = provider.get_coins(wallet.address(), asset_id).await?;
```
with the new API the call looks like this
```rust
let num_results = 2;
let coins = provider
    .get_coins_new(wallet.address(), asset_id, num_results)
    .call()
    .await?
    .results;
```
we use a builder pattern to create a `ProviderPaginationCaller` (I desperately need a better name :smile: ) and then we make and await the acctual call. We can also set the `cursor`, if we have one from a previous call, and we can also set the direction. Here is an example.
```rust
let paginated_result = provider
    .get_coins_new(wallet.address(), asset_id, 4)
    .call()
    .await?;
dbg!(paginated_result.results.len());

let paginated_result = provider
    .get_coins_new(wallet.address(), asset_id, 2)
    .with_cursor(paginated_result.cursor)
    .with_direction(PageDirection::Backward)
    .call()
    .await?;
dbg!(paginated_result.results.len());
```
Now about the code. If you look at the `ProviderPaginationCaller` struct you will see some complex types. The tricky part was to make the struct accept generic types and save the async function so that we can call it later with the `call()` method. As we do not know the size of the returned `Futures` we had to use `dyn Future` trait and use a `Box`. If anybody could make this code cleaner I would highly appreciate it :smile: . Here are the crazy types:
```rust
type BoxFnFuture<'a, T, U> = Box<dyn Fn(PaginationRequest<T>) -> BoxFutureResult<'a, U> + 'a>;
type BoxFutureResult<'a, U> = Box<dyn Future<Output = Result<U, ProviderError>> + 'a + Unpin>;

pub struct ProviderPaginationCaller<'a, T, U>
where
    T: Debug,
{
    /// The cursor returned from a previous query to indicate an offset
    pub cursor: Option<T>,
    /// The number of results to take
    pub results: usize,
    /// The direction of the query (e.g. asc, desc order).
    pub direction: PageDirection,
    // The function to call
    pub function: BoxFnFuture<'a, T, U>,
}
```

digorithm:
Talking strictly related to the UX of this new API (and not related to the actual implementation), I believe the API has a nicer UX this way. 

An important context to remember is how this would look like _without_ @hal3e's proposal. We would need to _always_ create a `PaginationRequest`; even if you're feeling lazy and don't want to paginate at all, which is most use cases. So it would look like this:

```Rust
let pagination_request = PaginationRequest { // I have to import this
                        cursor: None, // Always passing None until I'm actually paginating
                        results: 100,
                        direction: PageDirection::Forward, // And also import this one;
                    };
                    
let coins = provider
    .get_coins_new(wallet.address(), asset_id, pagination_request) // Always need to pass a pagination request.
    .await?
    .results;
```

So, as we can see, we gain something by not having to `.call()`, but then we _always_ have to create a `PaginationRequest` (or pass `None` if we were to change to accept an `Option<PaginationRequest>`).

With @hal3e's proposal, we drop the mandatory `PaginationRequest`, and whenever we need to paginate, we chain the methods `with_cursor()` and `with_direction()`. The downside is that we add a `.call()` to the workflow. 

IMO, the `.call()` isn't too bad as it aligns with contract calls; all contract calls do have a `.call()` (or `.simulate()`). Kinda sounds like whenever you're calling something on-chain, you append the `.call()` to make it explicit, which I think is kinda cool :). 

Let me know your thoughts @FuelLabs/sdk-rust.

hal3e:
@MujkicA and @leviathanbeak were able to reproduce the inconsistent behavior when getting paginated results from the client. @leviathanbeak will investigate it from the client side.

Issue #631, opened 2022-10-17
fix: cannot `ctrl-c` running the tests


iqdecay:
This bug was introduced in #607. 
Hitting `ctrl-c` currently does not kill the test-running process.

PR #639, opened 2022-10-17
fix: exiting tests

hal3e:
should close: https://github.com/FuelLabs/fuels-rs/issues/631

This is a possible solution for the ctrl-c exit problem with the integration tests. The tests are now separated into files and are not included as modules. Please try it and if we like it I will update the docs.


iqdecay (approved):
I like this, and tested it locally an can now `Ctrl-C` the tests as I please! Great :)

EDIT: approval on condition that the `TODO` added about the structure is done hehe

digorithm (dismissed):
Nice, I pretty much prefer the `name.rs` pattern over `name/mod.rs` :) 

Note: I believe we should address the `TODO` before merging this in!

iqdecay (approved):
:ship:  it

PR #643, opened 2022-10-19
feat: create `Bits256` from hex string

digorithm:
Small ergonomics improvement. 

Enables the user to do `let bits256 = Bits256::from_hex_str(hex_str)?;` to get a `Bits256` from their hex strings instead of having to do some more manual conversions.

### DIFF
--- a/packages/fuels-core/src/types/bits_256.rs
+++ b/packages/fuels-core/src/types/bits_256.rs
@@ -48,6 +48,7 @@ impl Tokenizable for Bits256 {
     }
 }
 
+#[cfg(test)]
 mod tests {
     use super::*;
     use crate::Tokenizable;
--- a/packages/fuels-core/src/types/bits_256.rs
+++ b/packages/fuels-core/src/types/bits_256.rs
@@ -82,16 +82,22 @@ mod tests {
     }
 
     #[test]
-    fn from_hex_str() {
+    fn from_hex_str() -> Result<(), Error> {
         // ANCHOR: from_hex_str
         let hex_str = "0101010101010101010101010101010101010101010101010101010101010101";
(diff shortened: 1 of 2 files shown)


hal3e (dismissed):
Great QoL improvement! :tada:  I left some minor suggestions.

hal3e (dismissed):

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -8,6 +8,22 @@ use fuels_types::param_types::ParamType;
...
+    /// Accepts both `0x` prefixed and non-prefixed hex strings.
+    pub fn from_hex_str(hex: &str) -> Result<Self, Error> {
+        let hex = if let Some(stripped_hex) = hex.strip_prefix("0x") {
+            stripped_hex
+        } else {
+            hex
+        };
+
+        let mut bytes = [0u8; 32];
+        hex::decode_to_slice(hex, &mut bytes as &mut [u8]).expect("Decoding failed");
```
segfault-magnet, line 22:
```suggestion
        hex::decode_to_slice(hex, &mut bytes as &mut [u8]).map_err(|e| {
            Error::InvalidData(format!("Could not convert hex str '{hex}' to Bits256! {e}"))
        })?;
 ```
Currently, we always return Result::Ok in this fn.

If somebody passes in a `&str` which, when decoded, is longer than the available buffer size of 32B, this fn will panic. 

Maybe we should convert that to an Error?

digorithm: Yeah, I tested for that (locally while developing). I figured since this is such a low-stakes helper it wouldn't be a problem. But it costs virtually nothing to make it nicer and not panic (you even wrote it already 😂 ) so let's do it!

segfault-magnet (approved):

Issue #638, opened 2022-10-17
ci: discover why CI didn't detect bad documentation path


iqdecay:
See #632.

PR #647, opened 2022-10-25
fix(check-docs): check whole file includes

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/638

Previously, the `regex` used to match `include` statements would skip includes that did not have an `anchor`. I have updated the regex and now the anchor is optional. Previously the program would stop if the `path` in the include statement is not correct, now we partition the result and show the path errors the same way as the other errors.


digorithm (approved):
Sounds about right to me; :shipit:

Issue #600, opened 2022-09-29
Support `raw_ptr` in `abigen!`


mohammadfawaz:
When https://github.com/FuelLabs/sway/issues/2551 is fully implemented, `raw_ptr` will appear in the JSON ABI (both inputs and ouptput). `raw_ptr` is basically just a `u64` under the hood. The compiler will make sure that `raw_ptr` does not show up anywhere other than in `Vec`.

PR #651, opened 2022-10-26
fix: Support `untyped raw ptr`

segfault-magnet:
fixes: #600 

Since the `untyped raw ptr` is not directly used in the SDK we'll skip its generation and avoid delving deeper into `Vec` when encountered in type resolving logic such as `resolve_type` or `ParamType::try_from_type_declaration`.


mohammadfawaz:
Thank you for this @segfault-magnet! Is `Vec` working again in the ABI with this change?

segfault-magnet:
> Thank you for this @segfault-magnet! Is `Vec` working again in the ABI with this change?

vector e2e tests are passing. They are relatively extensive so I'm pretty sure we're in the clear.

Vector returns are still not supported, but I guess that goes without saying.

mohammadfawaz:
> 

Excellent, looking forward to the next release then 😄  Thank you for addressing this so quickly!

packages/fuels/tests/types/generics/src/main.sw
```diff
@@ -25,7 +25,7 @@ enum EnumWGeneric<N> {
 
 struct MegaExample<T, U> {
     a: ([U; 2], T),
-    b: ([EnumWGeneric<StructWTupleGeneric<StructWArrayGeneric<PassTheGenericOn<T>>>>; 1], u32),
+    b: Vec<([EnumWGeneric<StructWTupleGeneric<StructWArrayGeneric<PassTheGenericOn<T>>>>; 1], u32)>,
```
digorithm, line 28: 😆

digorithm (approved):
Awesome work; left a tiny nit but feel free to ignore/push back.

packages/fuels/tests/types/generics/src/main.sw
segfault-magnet: A PITA to maintain, but I do feel safe seeing it pass.

digorithm (approved):
:shipit:

Issue #650, opened 2022-10-25
refactor: InputMessage support


Salka1988:
The `resources_to_spend` function has replaced the `coins_to_spend` function and it returns both `Coins` and `Messages` to spend. We should adapt the existing code to use the new function `resources_to_spend`.

PR #652, opened 2022-10-26
refactor: Input message support

Salka1988:
closes https://github.com/FuelLabs/fuels-rs/issues/650


leviathanbeak:
since no one reviewed it yet, could you reset your commits so that "unnecessary" data is not part of the commit history

segfault-magnet (dismissed):

digorithm (dismissed):
Nice and clean; LGTM.

packages/fuels-signers/src/provider.rs
```diff
@@ -232,6 +232,35 @@ impl Provider {
         Ok(coins)
     }
 
+    /// Get some spendable messages for address `from` that add up at least to
+    /// amount `amount`. The returned messages are actual messages that can be spent. The number
```
MujkicA, line 236: IIUC, we cannot guarantee that the retrieved messages will add up to amount because we request `Resource`s and filter for messages later.

Are we sure that we want to go forward with this approach?

Salka1988: We will talk about this at the meeting.

hal3e (approved):

PR #659, opened 2022-10-29
docs: implication that returning vectors is a thing.

segfault-magnet:
Part of the changed sentence isn't correct because it was written when returning vectors was possible.


digorithm (approved):
Nice catch!

PR #663, opened 2022-11-01
Actualized the `tx_pool::Config` with updated values

xgreenx:
`utxo_validation` and `chain_config` of `Config` are not the same as fields in the `tx_pool::Config`.

### DIFF
--- a/examples/predicates/src/lib.rs
+++ b/examples/predicates/src/lib.rs
@@ -38,15 +38,10 @@ mod tests {
             })
             .collect::<Vec<_>>();
 
-        let (provider, _) = setup_test_provider(
-            all_coins,
-            vec![],
-            Some(Config {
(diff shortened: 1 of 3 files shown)


hal3e (commented):
Would it make sense to update the structure of  `local_node()` so that we have the same fields as with fuel-core-lib ? This way we could avoid the `#[cfg(feature = "fuel-core-lib")]`

xgreenx:
Closed in favor of https://github.com/FuelLabs/fuel-core/pull/743/commits/6c4e07e4859b0db15c369e5dd3418f6370626aea

`fuel-core` will override `TxPoolConfig` to use correct `ChainConfig` and `utxo_validation`. In the future we will rework configs to avoid this kind of issue.

PR #668, opened 2022-11-02
feat: support evm address

### LINKED ISSUE #661: Add support for the `EvmAddress` type
Sway has an `EvmAddress` type which is very similar to the `Address` type.
It would be nice to be able to use the type as contract call inputs and outputs.

***Note**: An EVM address is only 20 bytes, so the first 12 bytes of the inner `value` are set to zero
```
/// The Address type, a struct wrappper around the inner `value`.
pub struct EvmAddress {
    value: b256,
}
```

MujkicA:
Closes #661 

The `EvmAddress` is implemented equivalent to the Sway type, as a wrapper over `Bits256`. And just like in Sway, the first 12 bytes are zeroed out when it is instantiated from `Bits256`.
Encoding/decoding this type is treated as a struct with a single `Bits256` field.


MujkicA:
Needs forc update

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -50,19 +50,80 @@ impl Tokenizable for Bits256 {
...
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+// ANCHOR: evm_address
+pub struct EvmAddress {
+    value: Bits256,
+}
+// ANCHOR_END: evm_address
+
+impl EvmAddress {
+    fn clear_12_bytes(bytes: [u8; 32]) -> [u8; 32] {
```
segfault-magnet, line 61: Since it's private might save space and just make it a fn instead of an associated fn.

Goes without saying that it doesn't make any difference functionally.

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -50,19 +50,80 @@ impl Tokenizable for Bits256 {
...
+    fn from_token(t: Token) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
+        if let Token::Struct(tokens) = t {
+            let first_token = tokens.into_iter().next();
+            if let Some(Token::B256(data)) = first_token {
+                Ok(EvmAddress::from(Bits256(data)))
+            } else {
+                Err(Error::InstantiationError(format!(
```
segfault-magnet, line 96: What I've found is that these 'expected X got Y' are not that great to run into without a backtrace.

When I remember I try to say something like: "EvmAddress could not be constructed from the given token. Reason: Expected Struct(B256) got: {token:?}"

packages/fuels/tests/types/evm_address/src/main.sw
```diff
@@ -0,0 +1,25 @@
...
+    fn evm_address_as_input(evm_addr: EvmAddress) -> bool;
+    fn evm_address_from_literal() -> EvmAddress;
+    fn evm_address_from_argument(raw_address: b256) -> EvmAddress;
+}
+
+impl EvmTest for Contract {
+    fn evm_address_as_input(evm_addr: EvmAddress) -> bool {
+        let evm_addr2 = EvmAddress::from(0x1616060606060606060606060606060606060606060606060606060606060606);
+
+        evm_addr == evm_addr2
```
segfault-magnet, line 15: Good idea. Better to return the comparison result rather than revert due to assert usage in sway. It helps narrow down the issue by separating it from a ton of other reasons why a revert might happen.

We should probably do it this way in the future, maybe even update all existing sway fn-s that use assertions to validate input args in some future general refactoring.

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -50,19 +50,80 @@ impl Tokenizable for Bits256 {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+// ANCHOR: evm_address
+pub struct EvmAddress {
+    value: Bits256,
```
iqdecay, line 56: I think it should be documented "in-code" that the first twelve bytes are zeroed out

packages/fuels-core/src/types/bits_256.rs
iqdecay: I like that it's an associated function since it's a very specific "utility", it would be weird to see it implemented outside this context.

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -50,19 +50,80 @@ impl Tokenizable for Bits256 {
...
+        let mut bytes = bytes;
+        bytes[..12].copy_from_slice(&[0u8; 12]);
+
+        bytes
+    }
+}
+
+impl From<Bits256> for EvmAddress {
+    fn from(b256: Bits256) -> Self {
+        let value = Bits256(Self::clear_12_bytes(b256.0));
```
iqdecay, line 71: A reference to why clear the 12 bytes would be warranted here

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -102,4 +163,37 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_into_token_evm_addr() {
+        let data = [0u8; 32];
+        let evm_address = EvmAddress::from(Bits256(data));
+
+        let token = evm_address.into_token();
+
+        assert_eq!(token, Token::Struct(vec![Token::B256(data)]));
```
iqdecay, line 197: I think it would be better t o check that the data is valid on non-0 vectors, here you're not making sure that it works as it should with the conversion I think

packages/fuels/tests/bindings.rs
```diff
@@ -443,6 +443,74 @@ async fn compile_bindings_b256_input() {
...
+
+    let mut hasher = Sha256::new();
+    hasher.update("test string".as_bytes());
+
+    // ANCHOR: evm_address_arg
+    let b256 = Bits256(hasher.finalize().into());
+    let arg = EvmAddress::from(b256);
+
+    let call_handler = contract_instance.methods().takes_evm_address(arg);
+    // ANCHOR_END: evm_address_arg
```
iqdecay, line 499: the presence of anchors makes me think maybe you forgot to add docs ?

packages/fuels/tests/types/evm_address/src/main.sw
iqdecay: Oh yeah that's smart

iqdecay (changes requested):
Looking good, some questions

packages/fuels-core/src/types/bits_256.rs
MujkicA: How would you go about this?

packages/fuels-core/src/types/bits_256.rs
```diff
@@ -68,6 +68,7 @@ impl EvmAddress {
 
 impl From<Bits256> for EvmAddress {
     fn from(b256: Bits256) -> Self {
+        // An evm address is only 20 bytes
```
iqdecay, line 71: Is there a spec refering to the fact that we decided that the bytes to clear where the leftmost ones?

docs/src/types/evm_address.md
```diff
@@ -0,0 +1,15 @@
+# EvmAddress
+
+In the Rust SDK, evm addresses can be represented with the 'EvmAddress' type. Its definition matches with the Sway standard library type with the same name and will be converted accordingly when interacting with contracts:
```
iqdecay, line 3:
```suggestion
In the Rust SDK, Ethereum Virtual Machine (EVM) addresses can be represented with the 'EvmAddress' type. Its definition matches with the Sway standard library type with the same name and will be converted accordingly when interacting with contracts:
```

docs/src/types/evm_address.md
```diff
@@ -0,0 +1,15 @@
...
+{{#include ../../../packages/fuels-core/src/types/bits_256.rs:evm_address}}
+```
+
+Here's an example:
+
+```rust,ignore
+{{#include ../../../packages/fuels/tests/bindings.rs:evm_address_arg}}
+```
+
+> **Note:** when creating an `EvmAddress` from `Bits256`, the first 12 bytes will be cleared because an evm address is only 20 bytes long.
```
iqdecay, line 15: Explain the choice of clearing the leftmost ones. If it is a deliberate choice (ie not based on spec), then this choice should be stated clearly ("We chose the convention of clearing the leftmost bytes").

packages/fuels-core/src/types/bits_256.rs
iqdecay: So it would be comment duplication but a comment along the line of the one you wrote on the `clear_bytes` function would be good enough imo. It's just so people don't get confused.

iqdecay (dismissed):
I think we should make it clearer if we're going with the leftmost clearing just by convention or specs. Otherwise great!

hal3e (approved):
LGTM. Good work! Left just a tiny comment.

docs/src/types/evm_address.md
MujkicA: The choice was made based on the Sway implementation of the same type. I find this a bit tricky to comment since we had a task to specifically remove all mentions of sway.

digorithm (approved):
Very nice!

PR #672, opened 2022-11-04
fix: remove predicates config flag

digorithm:
###### Description of changes

Seems like `fuel-core` doesn't take the `--predicates` flag anymore. So we're removing it from the SDK as well.


nfurfaro (approved):

hal3e (commented):
LGTM

PR #676, opened 2022-11-05
fuel-core v0.14 tai64 timestamp support

Voxelot:
###### Description of changes

Update fuels-rs to the latest fuel-core v0.14. This release was necessary to unblock the bridging team as fuel-core was incorrectly using datetime strings instead of tai64 when representing time. No breaking changes to the VM or forc like fuel-core v0.13, so this is a very lightweight update.


digorithm (approved):
Nice -- :shipit:

PR #685, opened 2022-11-07
feat: add format and clean to `test-projects`

hal3e:
This PR updates the `build-test-projects` scripts. Now we are able to format, check formatting and clean sway projects.

What was done:
- rename `build-test-projects` to `test-projects` - as we are not only building, changing the name made sense
- added `clap` to make a nice CLI with sub-commands and flags
- refactor the script so that we can concurrently run different commands (currently `forc` and `forc-fmt`)
- added formatting check for sway projects to CI
- updated docs
- used the tool to format all test projects - that is why we have changes in the test projects in this PR

How to use the new CLI:
- `test-projects build` - will build all the projects
- `test-projects clean` - will clean the output of all projects
- `test-projects format` - will format all projects
- `test-projects format --check` - will check the formatting but will not change any files

I have also added optional OPTIONS:
- `-n, --num-concurrent` - we can set the number of concurrent projects to run. Currently the default is 1.
- `--bin-path` - we can manually set the path to `forc` and `forc-fmt` (useful if we need to use specific versions)
- `--projects-path` - we can specify the test projects path. You can now build only parts of the test projects or use any other directory where you have sway projects.

All the info above is available in the `help` output of the CLI:
```shell
Build or format all projects under the main test suite with `forc`/`forc-fmt`

Usage: test-projects [OPTIONS] <COMMAND>

Commands:
  clean   Cleans `forc` build output
  build   Builds all test projects with `forc`
  format  Formats all test projects with `forc-fmt`
  help    Print this message or the help of the given subcommand(s)

Options:
  -n, --num-concurrent <NUM>  Number of concurrent projects [default: 1]
      --bin-path <DIR>        Specify where to find `forc` and `forc-fmt`
      --projects-path <DIR>   Specify test projects path
  -h, --help                  Print help information
  -V, --version               Print version information
```

> **Note** We use `cargo run` to run the script. So for example to check the formatting you would do:
```shell
cargo run --bin test-projects -- format --check
```


hal3e:
The format check is working in the CI this is how it looks like:
ci

digorithm (approved):
Looking really good. The new CLI is great!

Issue #679, opened 2022-11-06
Return the readable reason of a revert


digorithm:
# Problem

Consider the following setup: 

```rust
contract;

abi MyContract {
    fn test_function() -> bool;
}

impl MyContract for Contract {
    fn test_function() -> bool {
        require(false, "something is not right");
        true
    }
}
```

And the test file:

```rust
#[tokio::test]
async fn can_get_contract_id() {
    let (instance, _id) = get_contract_instance().await;
    let response = instance.methods().test_function().call().await;
    dbg!(&response);
}
```

Currently, the top-level error will be:

```plaintext
Err(
    RevertTransactionError(
        "Revert(18446744073709486080)"
```

Which tells nothing to the user. If the user manually digs deeper, one of the receipts will be:

```plaintext
LogData {
                id: 0x3d041ddc77907490df673f72cf579763bd7a77ab8e31cf20738f97da7ad225f9,
                ra: 0,
                rb: 0,
                ptr: 11696,
                len: 24,
                digest: 0xd5f441dadfdac1d8f13db35714929fec7c5c9c4cf125cb05d27ff4dcbeb4ba3a,
                data: [
                    115,
                    111,
                    109,
                    101,
                    116,
                    104,
                    105,
                    110,
                    103,
                    32,
                    105,
                    115,
                    32,
                    110,
                    111,
                    116,
                    32,
                    114,
                    105,
                    103,
                    104,
                    116,
                    116,
                    116,
                ],
                pc: 11684,
                is: 11624,
            }
```

That `data` piece is the string `"something is not right"` used in the revert. This should be the top-level error.

# Solution 

We have to find a way to parse that `data` into a string and return it if the contract call reverted. Currently, we have this in `send_transaction`: https://github.com/FuelLabs/fuels-rs/blob/88eb85313f8e5f7fb60c14c6220b9adbf304b845/packages/fuels-signers/src/provider.rs#L81. 

```Rust
 let (status, receipts) = self.submit_with_feedback(&tx.clone().into()).await?;

  if let TransactionStatus::Failure { reason, .. } = status {
      Err(Error::RevertTransactionError(reason, receipts))
  } else {
      Ok(receipts)
  }
```

We have to detect it was a revert. Once that's done, we might have to extend `RevertTransactionError` to include the string representing the error. 

Of course, the actual solution design is up for discussion. Our only goal is to return that string in `LogData.data` after a revert so that users don't have to dig deep to find it.

[...]

PR #690, opened 2022-11-09
feat!: add parsed log to revert error

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/679

Had to redo the PR as the previous implementation did not work for multi-calls from different contracts :/. 
@segfault-magnet helped and updated the `ParamTypes` for `struct` and `enum` and added names for the types and for the fields/variants.
This made it possible to use the `ParamType` and decoded tokens directly to make logs. We do not need to generate all the types in the abigen aymore. 

Now the workflow is:
1. Generate the mappings between `(contract_id, log_id)` and `ParamType` in the abigen (we need the `contract_id`  because we can have multiple contracts when using multi-call)
2. Propagate this mapping to `ContractCallHandler`/`MultiCallHandler` - we use it here for the revert log mapping
3. Propagate to `CallResponse` used for general logging.

- The mapping is propagated through the new struct `LogDecoder`.
- A new trait `DecodeLog` that is implemented for `ParamType` was added. This trait is used to get a decoded log from a `ParamType` and the corresponding data.

I have also:
- added integration tests
- added unit tests
- updated how the logging is done - We can use the `CallResponse` directly
- updated the docs
- changed how the name of the contract instance mod is generated (now it is snake case: `contract_instance_mod` and the struct is class case `ContractInstance`


packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -173,6 +177,25 @@ impl Abigen {
                     impl #methods_name {
                         #contract_functions
                     }
+
```
segfault-magnet, line 180:
```suggestion
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -173,6 +177,25 @@ impl Abigen {
...
+                        fn logs_with_type<D: Tokenizable + Parameterize>(receipts: &[Receipt]) -> Result<Vec<D>, SDKError> {
+                            Self::logs_with_type::<D>(receipts)
+                        }
+                    }
+
+                    // Implemented manually so that we can use Debug on types that use depend on the Logging trait
+                    impl fmt::Debug for #name {
+                        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+                            write!(f, "{}", #name_str)
```
segfault-magnet, line 196: Any reason this differs from the derived version of Debug? I.e. why not `derive(Debug)` it?

Salka1988 (commented):
Nice! LGTM :tada:

packages/fuels-core/src/code_gen/abigen.rs
hal3e: I thought we can not use the macro inside of our procedural macro. I was wrong :D Thanks!

docs/src/calling-contracts/logs.md
```diff
@@ -8,16 +8,16 @@ Consider the following contract method:
 {{#include ../../../packages/fuels/tests/logs/logged_types/src/main.sw:produce_logs}}
 ```
 
-You can access the logged values in Rust by calling `logs_with_type::<T>` from a contract instance, where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
+You can access the logged values in Rust by calling `logs_with_type::<T>` from a call response, where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
```
MujkicA, line 11:
```suggestion
You can access the logged values in Rust by calling `logs_with_type::<T>` from a `CallResponse,` where `T` is the type of the logged variables you want to retrieve. The result will be a `Vec<T>`:
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -25,7 +26,7 @@ pub struct Abigen {
     no_std: bool,
 
     /// The contract name as an identifier.
```
MujkicA, line 28:
```suggestion
```

packages/fuels/tests/contracts.rs
```diff
@@ -642,6 +642,105 @@ async fn contract_call_futures_implement_send() -> Result<(), Error> {
...
+    );
+
+    let contract_methods = contract_instance.methods();
+    {
+        let error = contract_methods
+            .require_primitive()
+            .call()
+            .await
+            .expect_err("Should return a revert error");
+        dbg!(&error); //TODO: remove after the PR review
```
MujkicA, line 670: Just a reminder

packages/fuels/tests/contracts.rs
```diff
@@ -642,6 +642,105 @@ async fn contract_call_futures_implement_send() -> Result<(), Error> {
...
+async fn test_multi_call_require() -> Result<(), Error> {
+    setup_contract_test!(
+        contract_instance,
+        wallet,
+        "packages/fuels/tests/contracts/require"
+    );
+
+    let contract_methods = contract_instance.methods();
+
+    // The output of the error depends on the order of the contract
```
MujkicA, line 707: Should this also be mentioned in the book?

packages/fuels-contract/src/contract.rs
```diff
@@ -633,23 +668,36 @@ impl MultiContractCallHandler {
...
+        err
+    }
+
     /// Call contract methods on the node, in a state-modifying manner.
-    pub async fn call<D: Tokenizable + Debug>(&self) -> Result<CallResponse<D>, Error> {
-        Self::call_or_simulate(self, false).await
+    pub async fn call<D: Tokenizable + Debug>(&self) -> Result<CallResponse<D, C>, Error> {
+        Self::call_or_simulate(self, false)
+            .await
+            .map_err(Self::revert_error_map)
```
MujkicA, line 685: Can we use ContractCallHandler::revert_error_map() instead here?

packages/fuels-core/src/code_gen/abigen.rs
MujkicA: Since it is stored as a string now

packages/fuels/tests/contracts.rs
hal3e: I had a look. I can not insert it easily anywhere in the docs. Maybe I make a new subsection for using `require` in sway ?

packages/fuels-contract/src/contract.rs
```diff
@@ -634,22 +669,25 @@ impl MultiContractCallHandler {
     }
 
     /// Call contract methods on the node, in a state-modifying manner.
-    pub async fn call<D: Tokenizable + Debug>(&self) -> Result<CallResponse<D>, Error> {
-        Self::call_or_simulate(self, false).await
+    pub async fn call<D: Tokenizable + Debug>(&self) -> Result<CallResponse<D, C>, Error> {
+        Self::call_or_simulate(self, false)
+            .await
+            .map_err(ContractCallHandler::<D, C>::revert_error_map)
```
segfault-magnet, line 675: I'd extract `revert_error_map` into a private free function. No need to mention the `ContractCallHandler` inside the multi-call handler.

Especially no need to know `D` since you only need `C` to map the error.

so a `map_err(decode_revert_err::<C>)` or however else it is written to please the compiler.

packages/fuels/tests/contracts.rs
MujkicA: On second thought, it may be unnecessary. It is intuitive that the first failed require will be the one to pop up since it stops execution anyways.

MujkicA (dismissed):
LGTM, just remove the dbg!

segfault-magnet (approved):

hal3e:
After trying to merge the last commit from master, I figured out that the new method will not work if we use the `MultiCallHandler` with methods from different calls. The approach that I used in this PR requires that the generic type `C` to be the same for all `ContractCallHandler`s. @segfault-magnet and I figured out that we could avoid the whole abigen generation of the logged types if we add some info to the `ParamType` for struct and enum. The new approach is almost done. I will mark the PR as draft until I have it cleaned up.

packages/fuels-contract/src/contract.rs
```diff
@@ -526,7 +621,6 @@ where
     /// your method returns `bool`, it will be a bool, works also for structs thanks to the
     /// `abigen!()`). The other field of CallResponse, `receipts`, contains the receipts of the
     /// transaction.
-    #[tracing::instrument]
```
MujkicA, line 529: If this is not needed, can we remove `tracing` from the the Cargo file?

hal3e: Good catch! I will have a look.

packages/fuels/tests/logs.rs
```diff
@@ -184,18 +182,268 @@ async fn test_fetch_logs() -> Result<(), Error> {
...
+        multi_call_handler
+            .add_call(call_handler_1)
+            .add_call(call_handler_2);
+
+        let error = multi_call_handler
+            .call::<((), ())>()
+            .await
+            .expect_err("Should return a revert error");
+
+        dbg!(&error);
```
MujkicA, line 445: Should probably be removed with the rest of the dbg statements

MujkicA (approved):
Nice work, the tests look good 👍

packages/fuels-contract/src/contract.rs
hal3e: This feature is not working since we changed the test structure :/ https://github.com/FuelLabs/fuels-rs/pull/607
Since we have logging, (and now also logging for require ) I think we can remove it completely. If a user wants to print the receipts he can do it manually. What do you guys think ?

packages/fuels-core/src/tokenizer.rs
```diff
@@ -55,7 +53,14 @@ impl Tokenizer {
...
                 fields: struct_params,
                 ..
-            } => Ok(Self::tokenize_struct(trimmed_value, struct_params)?),
+            } => Ok(Self::tokenize_struct(
+                trimmed_value,
+                &struct_params
+                    .iter()
+                    .map(|(_, param_type)| param_type)
+                    .cloned()
+                    .collect::<Vec<_>>(),
```
iqdecay, line 62: The fact that this `iter().map().clone().collect()` construction appears so many times makes me think there should be an associated function or something

packages/fuels/tests/logs.rs
```diff
@@ -184,18 +182,262 @@ async fn test_fetch_logs() -> Result<(), Error> {
...
+
+        let error = multi_call_handler
+            .call::<((), ())>()
+            .await
+            .expect_err("Should return a revert error");
+
+        assert_is_revert_containing_msg("StructDeeplyNestedGeneric", error);
+    }
+    Ok(())
+}
```
iqdecay, line 387: this is great

iqdecay (changes requested):
This is looking great, I love the new API! Left some remarks regarding code duplication

packages/fuels-contract/src/contract.rs
```diff
@@ -1,41 +1,111 @@
...
+        });
+
+        ids_with_data
+            .map(|((c_id, id), data)| {
+                let param_type = self
+                    .logs_map
+                    .get(&(c_id, id))
+                    .ok_or_else(|| Error::InvalidData("Failed to find log id".into()))?;
+
+                param_type.decode_log(&data)
```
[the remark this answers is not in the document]
hal3e: I will updated the `id` variable to be `c_id`  but the other variables are basically in the sway specs and have those names..

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -320,10 +312,16 @@ fn skip(slice: &[u8], num_bytes: usize) -> Result<&[u8], CodecError> {
...
 mod tests {
     use super::*;
+    use std::iter::{repeat, zip};
+
     use fuels_types::enum_variants::EnumVariants;
     use fuels_types::errors::Error;
     use std::vec;
 
+    fn unused_field_names(types: Vec<ParamType>) -> Vec<(String, ParamType)> {
+        zip(repeat("unused".to_string()), types).collect()
```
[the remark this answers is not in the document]
hal3e: @segfault-magnet's magic :tada:

packages/fuels-core/src/abi_encoder.rs
```diff
@@ -242,11 +222,16 @@ mod tests {
     use fuels_types::{errors::Error, param_types::ParamType};
     use itertools::chain;
     use sha2::{Digest, Sha256};
+    use std::iter::{repeat, zip};
     use std::slice;
 
     const VEC_METADATA_SIZE: usize = 3 * WORD_SIZE;
     const DISCRIMINANT_SIZE: usize = WORD_SIZE;
 
+    fn unused_field_names(types: Vec<ParamType>) -> Vec<(String, ParamType)> {
```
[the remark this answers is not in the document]
hal3e: Makes sense as it is used on several places for testing...

packages/fuels-core/src/lib.rs
```diff
@@ -162,9 +161,98 @@ pub fn pad_string(s: &str) -> Vec<u8> {
...
+        (ParamType::String(..), Token::String(str_token)) => {
+            format!("SizedAsciiString {{ data: \"{}\" }}", str_token.data)
+        }
+        (ParamType::Array(inner_type, _), Token::Array(tokens)) => {
+            let elements = inner_types_log(tokens, inner_type, ", ")?;
+            format!("[{elements}]")
+        }
+        (ParamType::Vector(inner_type), Token::Vector(tokens)) => {
+            let elements = inner_types_log(tokens, inner_type, ", ")?;
+            format!("[{elements}]")
```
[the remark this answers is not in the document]
hal3e: I used the `std::Debug` as reference and it writes both the same way.

packages/fuels-core/src/tokenizer.rs
hal3e: I do not know of another way. We cannot use `into_iter` because it would move the values, so we have to clone it at some point..

iqdecay: I was not talking about the construction itself (I tried to improve it via `unzip` but couldn't haha), but the fact that maybe we should have a function associated to the `struct_params` that does this -> it's basically an unzip. If I'm not clear I can try to come up with a code example.

packages/fuels-contract/src/contract.rs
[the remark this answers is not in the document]
iqdecay: Ok so it's really close to the metal, no comment needed then

packages/fuels-core/src/tokenizer.rs
hal3e: Oh ok i  got you now. I thought you were looking for an iterator function that would do these steps :smile: I will add it somewhere.

packages/fuels-contract/src/contract.rs
```diff
@@ -1,41 +1,111 @@
...
+    fs,
+    marker::PhantomData,
+    panic,
+    path::Path,
+    str::FromStr,
+};
 
 pub const DEFAULT_TX_DEP_ESTIMATION_ATTEMPTS: u64 = 10;
 
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
digorithm: This is minor, but I'd put it in a new file called `logs.rs` under `src/contract/`.

packages/fuels-contract/src/contract.rs
```diff
@@ -75,12 +157,12 @@ impl<D> CallResponse<D> {
             .expect("could not retrieve gas used from ScriptResult")
     }
 
-    pub fn new(value: D, receipts: Vec<Receipt>) -> Self {
-        Self {
-            value,
-            gas_used: Self::get_gas_used(&receipts),
-            receipts,
-        }
+    pub fn get_logs(&self) -> Result<Vec<String>, Error> {
```
digorithm, line 160: Since this and `get_logs_with_type<T>` are very user-facing, we should have doc-level comments here.

packages/fuels-contract/src/contract.rs
```diff
@@ -552,14 +647,18 @@ where
 
     /// Call a contract's method on the node, in a state-modifying manner.
     pub async fn call(self) -> Result<CallResponse<D>, Error> {
-        Self::call_or_simulate(&self, false).await
+        Self::call_or_simulate(&self, false)
+            .await
+            .map_err(|err| decode_revert_error(err, &self.log_decoder))
```
digorithm, line 652: 😍 this is wonderful!

packages/fuels-core/src/lib.rs
```diff
@@ -162,9 +161,106 @@ pub fn pad_string(s: &str) -> Vec<u8> {
...
 
+pub fn unzip_param_types(param_types: &[(String, ParamType)]) -> Vec<ParamType> {
+    param_types
+        .iter()
+        .map(|(_, param_type)| param_type)
+        .cloned()
+        .collect()
+}
+
+pub trait DecodeLog {
```
digorithm, line 172: Nit: usually, trait names are adjectives rather than verbs. So `DecodeLog` would be `DecodableLog`. Should we stick to this convention?

packages/fuels-core/src/lib.rs
```diff
@@ -162,9 +161,106 @@ pub fn pad_string(s: &str) -> Vec<u8> {
...
+        }
+        (ParamType::Tuple(types), Token::Tuple(tokens)) => {
+            let elements = zip(types, tokens)
+                .map(|(ptype, token)| paramtype_decode_log(ptype, token))
+                .collect::<Result<Vec<_>, _>>()?
+                .join(", ");
+
+            format!("({elements})")
+        }
+        _ => return Err(Error::InvalidData("alaj nevalja".to_string())),
```
digorithm, line 251: `"alaj nevalja"` say what now? 😆

digorithm (changes requested):
A few nits. But, overall, _outstanding_ work! I love the new abstractions you've introduced. Surprisingly, it got cleaner than before (log feature-wise).

I still need to do another, deeper pass, but wanted to give you something actionable in the meantime.

packages/fuels-core/src/lib.rs
hal3e: haha good catch.

digorithm (approved):
This is an amazing, long-waited SDK feature. Excellent work @hal3e!

PR #692, opened 2022-11-09
get_asset_inputs_for_amount uses resources

### LINKED ISSUE #691: `get_asset_inputs_for_amount` should pull for `Coin`s and `Message`s
# Problem

It's been noticed that a user could receive fewer coins/messages than the requested amount. This is happening because the SDK filters only coins or messages. The problem with this approach is that users may get an insufficient balance error. Since users request a specific target amount to spend, any filtering of the `resourcesToSpend` response may cause users to come up short of the amount they expected, and their transactions will fail.

# Solution

The function below

```Rust
pub async fn get_asset_inputs_for_amount(
        &self,
        asset_id: AssetId,
        amount: u64,
        witness_index: u8,
    ) -> Result<Vec<Input>, Error> {
        let spendable = self.get_spendable_coins(asset_id, amount).await?;
        let mut inputs = vec![];
        for coin in spendable {
            let input_coin = Input::coin_signed(
                UtxoId::from(coin.utxo_id),
                coin.owner.into(),
                coin.amount.0,
                asset_id,
                TxPointer::default(),
                witness_index,
                0,
            );
            inputs.push(input_coin);
        }
        Ok(inputs)
    }
```

Has to be updated not only to pull from `get_spendable_coins`, but from the client's `resources_to_spend`, which will return both `Message`s and `Coin`s.

MujkicA:
Closes #691

Preparing asset inputs now works with `Resource` instead of `Coin`. I didn't pay too much attention to cleanliness to push it quickly.

Docs update WIP.

### DIFF
--- a/examples/providers/src/lib.rs
+++ b/examples/providers/src/lib.rs
@@ -63,12 +63,12 @@ mod tests {
         assert_eq!(coins.len(), 1);
         // ANCHOR_END: get_coins
 
-        // ANCHOR: get_spendable_coins
-        let spendable_coins = provider
-            .get_spendable_coins(wallet.address(), BASE_ASSET_ID, 1)
+        // ANCHOR: get_spendable_resources
+        let spendable_resources = provider
+            .get_spendable_resources(wallet.address(), BASE_ASSET_ID, 1)
             .await?;
-        assert_eq!(spendable_coins.len(), 1);
-        // ANCHOR_END: get_spendable_coins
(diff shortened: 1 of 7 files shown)


digorithm (dismissed):
Excellent work @MujkicA! Thanks for addressing this _so_ quickly and precisely!

digorithm (approved):
Me approving my own changes like
image

Issue #688, opened 2022-11-07
Create a `B512` type to match `std::b512::B512`


nfurfaro:
This is implemented in Sway as:
```
pub struct B512 {
    bytes: [b256; 2],
}
```

I'm currently working around this by making contract functions accept 2 `b256` values and constructing the `B512` in sway.

arboleya:
@luizstacio @digorithm Should I consider this type for the new [ABI `typegen`](https://github.com/FuelLabs/fuels-ts/pull/570)?

Initially, I followed only the primitives listed here:
 - https://fuellabs.github.io/sway/v0.30.1/basics/built_in_types.html#primitive-types

The only mention of `B512` is under _**signature recovery**_:
 - https://fuellabs.github.io/sway/v0.30.1/blockchain-development/hashing_and_cryptography.html?highlight=B512#signature-recovery

nfurfaro:
`B512` is a stdlib type, like `Address`, `EvmAddress`, `Option` & `Result`.
It seems that sooner or later, for every type added to `std` there will be a request to use that type in the SDK.
Perhaps moving forward, as new types are proposed for `std` we can coordinate with the SDK teams to roll out support more in parallel?

arboleya:
Interesting. I wasn't paying attention to this.

Are these `stdlib` types documented somewhere?

I found only [this](https://fuellabs.github.io/sway/v0.30.1/introduction/standard_library.html) that points to [this](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/prelude.sw), but there's no mention to `B512`.

Anyway, I believe `Rust` and `Typescript` SDK differ in some aspects.

Perhaps @luizstacio or @digorithm can shine a light on this matter.

digorithm:
> Are these stdlib types documented somewhere?

@arboleya some are documented [here](https://fuellabs.github.io/sway/v0.30.1/basics/blockchain_types.html).

In terms of actual implementation, it differs between the Rust and TS SDKs, in some aspects. For instance, `Option` and `Result` types _are not_ generated in the Rust SDK because Rust natively has `Option` and `Result`, so the SDK basically "maps" the `Option`/`Result` from Sway to `Option`/`Result` from Rust. The TS SDK does something similar, but with different mappings/strategies (e.g. there's no `Result<T>` in Typescript).

Regarding this:

> Should I consider this type for the new https://github.com/FuelLabs/fuels-ts/pull/570?

[...]

PR #694, opened 2022-11-10
feat: add B512 support

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/688

What was done:
- add the `B512` type and excluded it from the `Abigen`
- add `B512::from((Bits253, Bits256))` to construct a `B512` ( this is how it is done in the std lib)
- add test and docs
- change name of `bits_256.rs` to `bits.rs` - we have there `Bits256`, `EvmAddress` and `B512`


packages/fuels-core/src/types/bits.rs
```diff
@@ -90,16 +144,22 @@ impl Tokenizable for EvmAddress {
...
-        if let Token::Struct(tokens) = &t {
-            let first_token = tokens.iter().next();
+        if let Token::Struct(tokens) = t {
+            let first_token = tokens.into_iter().next();
             if let Some(Token::B256(data)) = first_token {
-                return Ok(EvmAddress::from(Bits256(*data)));
+                Ok(EvmAddress::from(Bits256(data)))
+            } else {
+                Err(Error::InstantiationError(format!(
+                    "Expected `b256`, got {:?}",
```
segfault-magnet, line 153: Technically we could be more strict in checking the tokens against our expectations. We should be given a struct with only one token inside of it. Here we're willing to accept multiple tokens and just consume the first one.

If we ever got multiple tokens here it would be a good sign that something is broken.

Also, when informing that the first token was not of the right type, we don't mention that there happened to be others. Information that might be useful since these would probably be the tokens of some other struct. 

Maybe there is a way to get the full error message and take ownership of the first element?

docs/src/types/B512.md
```diff
@@ -0,0 +1,13 @@
+# B512
+
+In the Rust SDK, the `B512` definition matches the Sway standard library type with the same name and will be converted accordingly when interacting with contracts:
```
iqdecay, line 3: Didn't we say we would avoid saying `Sway` in the docs? I never fully understood why, @digorithm it feels like saying `Fuel` here wouldn't make sense either. Wdyt?

packages/fuels-core/src/types/bits.rs
iqdecay: Agreed that we should check that length is exactly one

packages/fuels-core/src/types/bits.rs
```diff
@@ -50,6 +50,60 @@ impl Tokenizable for Bits256 {
...
+    }
+}
+
+impl Tokenizable for B512 {
+    fn from_token(t: Token) -> Result<Self, Error>
+    where
+        Self: Sized,
+    {
+        if let Token::Struct(tokens) = t {
+            let first_token = tokens.into_iter().next();
```
iqdecay, line 84: Check that length is exactly one

iqdecay (changes requested):
Looking good, I think we can just be a little more strict on conversion checks!

hal3e:
I have updated all manual `Tokenizable` implementations for our structs. Now we make sure that they receive the right token and the right number of tokens. By chance it turned out that all need only one token e.g. have one field.

iqdecay (approved):
LGTM! :ship: it

digorithm (approved):
Great work on the documentation; very clean! :shipit:

PR #699, opened 2022-11-15
Use non-tagged version of forc in CI 

Salka1988:


iqdecay (changes requested):
I understand the code and think it looks godd! I'm just not sure how we would use it in reality. Can you provide more explanation?

Salka1988:
> I understand the code and think it looks godd! I'm just not sure how we would use it in reality. Can you provide more explanation?

If I understood correctly, we need the option to use the `forc` version that has not yet been released and is located somewhere on a branch or under a revision number. If the assumption is correct, then we will do it like this:

PR #707, opened 2022-11-22
feat!: convert client types to fuels types

### LINKED ISSUE #711: Introduce an anti-corruption layer between the SDK and common fuel-core types
Using fuel-core types in our UI has many consequences, like the user having to:
- dig trough multiple wrappers to get the underlying data, e.g. `header.transactions_root.0 .0` to get `Bytes32`
- use `Address` instead of `Bech32Address`
It also makes our code and the UX vulnerable to changes made to these types.

MujkicA:
Closes #711 

The goal of this PR is to introduce an anti-corruption layer between the SDK and fuel-core and to provide a more consistent interface to our users.
It adds the Rust SDK equivalent of types we usually query from the client:
`Message`, 
`Coin`, 
`TransactionResponse`, 
`Block` etc. 

The new types have been applied to `provider` methods as a return type and to `fuels-test-helpers` methods like `setup_single_asset_coins`, `setup_single_message`. 
Previously, setup methods would use different types, from those returned by provider methods even though they describe the same data, e.g. `fuel_gql_client::...:: Coin` vs. 'fuel_core_interface::...:: Coin'.
This also lets us hide `Address` (in some cases) in favor of `Bech32Addres`, ridding us of `.into()` calls and thereby reducing confusion for the user.


packages/fuels-types/src/transaction_response.rs
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+impl From<ClientTransactionStatus> for TransactionStatus {
+    fn from(client_status: ClientTransactionStatus) -> Self {
+        match client_status {
+            ClientTransactionStatus::Submitted { .. } => TransactionStatus::Submitted(),
+            ClientTransactionStatus::Success { .. } => TransactionStatus::Success(),
+            ClientTransactionStatus::Failure { .. } => TransactionStatus::Failure(),
+            ClientTransactionStatus::SqueezedOut { .. } => TransactionStatus::SqueezedOut(),
```
iqdecay, line 64: aren't we losing precious information here?

iqdecay (approved):
Looks good to me! Just one question

packages/fuels-types/src/transaction_response.rs
MujkicA: Initially, I was thinking we could introduce whatever data we have there on a "per need basis" in favour of having "block_id" and "time" be part of `TransactionResponse` as optional fields.

iqdecay (approved):
:ship:  it

iqdecay (approved):

hal3e (approved):
LGTM!

PR #715, opened 2022-11-30
fix: pay tx fees with message

### LINKED ISSUE #714: Missing change output when paying tx fees with an InputMessage
When the paying wallet uses a single InputMessage to cover transaction fees, the remaining amount is burned because of a missing change Output.

MujkicA:
Closes #714 

When didn't update `wallet.add_fee_coins()` properly after the client endpoint for getting coins changed to return resources (messages and coins).
I've fixed the behaviour and included a test.


digorithm (approved):
Awesome job handling this fix so quickly!

Issue #713, opened 2022-11-30
Handle log decoding in `Script`s


digorithm:
Once https://github.com/FuelLabs/fuels-rs/pull/620/ is merged, we'll be able to support arguments in `Script`s' `main()` functions. But the current implementation doesn't support decoding of the logs yet. This issue is to track such functionality.

PR #716, opened 2022-11-30
feat: add logs for scripts

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/713

- Script calls can use the same logging interface as contract calls
- Added log tests for scripts
- Added require log tests for scripts
- Update documentation
- Renamed the `logging` and `logged_types` sway projects for clarity
- Moved `decode_revert_error()` to `logs.rs` -> open for suggestions where to put it...


docs/src/scripts/scripts.md
```diff
@@ -0,0 +1 @@
+# Running scripts
```
iqdecay, line 1: Obsolete file

packages/fuels-contract/src/contract.rs
```diff
@@ -481,18 +483,6 @@ pub fn get_decoded_output(
     Ok(decoded_value)
 }
 
-// Decode the logged type from the receipt of a `RevertTransactionError` if available
-fn decode_revert_error(err: Error, log_decoder: &LogDecoder) -> Error {
```
iqdecay, line 485: Moving this function makes complete sense.

packages/fuels-contract/src/logs.rs
```diff
@@ -85,3 +85,15 @@ impl LogDecoder {
             .extend(log_decoder.logs_map.clone().into_iter());
     }
 }
+
+// Decode the logged type from the receipt of a `RevertTransactionError` if available
```
iqdecay, line 89:
```suggestion
/// Decodes the logged type from the receipt of a `RevertTransactionError` if available
```

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -172,11 +174,17 @@ impl Abigen {
...
                 }
 
                 impl #name {
                     pub fn new(wallet: WalletUnlocked, binary_filepath: &str) -> Self {
-                        Self {wallet: wallet, binary_filepath: binary_filepath.to_string()}
+                        Self {
+                            wallet: wallet,
+                            binary_filepath: binary_filepath.to_string(),
+                            logs_map: get_logs_hashmap(&[#(#log_id_param_type_pairs),*],
+                                                       &Bech32ContractId::from(ContractId::zeroed()))
```
iqdecay, line 186: This makes me think we should accept an `Option<Bech32ContractID>` for the `get_logs_hashmap` function: it makes no sense to pass in a null contract_id, it should be hidden away in the function itself. Wdyt?
Also this way we don't have to have the `Bech32ContractID` include statement in the general case, which makes more sense.

packages/fuels/tests/logs.rs
```diff
@@ -439,5 +441,187 @@ async fn test_multi_call_require_log_multi_contract() -> Result<(), Error> {
...
+    );
+
+    let wallet = launch_provider_and_get_wallet().await;
+    let bin_path = "../fuels/tests/logs/script_logs/out/debug/script_logs.bin";
+    let instance = log_script::new(wallet.clone(), bin_path);
+
+    let response = instance.main().call().await?;
+
+    let logs = response.get_logs()?;
+    let _log_u64 = response.get_logs_with_type::<u64>()?;
```
iqdecay, line 463:
```suggestion
    let log_u64 = response.get_logs_with_type::<u64>()?;
```

packages/fuels/tests/logs.rs
```diff
@@ -439,5 +441,187 @@ async fn test_multi_call_require_log_multi_contract() -> Result<(), Error> {
 
         assert_is_revert_containing_msg("StructDeeplyNestedGeneric", error);
     }
+
+    Ok(())
+}
+
+#[tokio::test]
```
iqdecay, line 448:
```suggestion
#[tokio::test]
#![allow(unused_variables)]
```

packages/fuels/tests/logs.rs
```diff
@@ -439,5 +441,187 @@ async fn test_multi_call_require_log_multi_contract() -> Result<(), Error> {
...
+        let error = instance
+            .main(MatchEnum::RequireCustomGeneric())
+            .call()
+            .await
+            .expect_err("Should return a revert error");
+
+        assert_is_revert_containing_msg("StructDeeplyNestedGeneric", error);
+    }
+
     Ok(())
```
iqdecay, line 626: Those tests are really good

iqdecay (changes requested):
Very nice, super clean! Left some comments

packages/fuels/tests/logs.rs
hal3e: I am not using this field in this test.. but I left it there for the book example. This way I can show both and do not have to make a new anchor.. If we do not like this I will add the second anchor ...

packages/fuels/tests/logs.rs
hal3e: I have to give credit to @MujkicA. Those are his test but modified for scripts :smile:

iqdecay (approved):
Nice! LGTM :)

packages/fuels-contract/src/script_calls.rs
```diff
@@ -131,7 +133,9 @@ where
 
     /// Call a script on the node, in a state-modifying manner.
     pub async fn call(self) -> Result<FuelCallResponse<D>, Error> {
-        Self::call_or_simulate(&self, false).await
+        Self::call_or_simulate(&self, false)
+            .await
+            .map_err(|err| decode_revert_error(err, &self.log_decoder))
```
digorithm, line 138: I love the reusability here!

digorithm (approved):
NIce! The tests/docs are awesome.

Issue #655, opened 2022-10-27
feat: support `Vec` in script inputs


iqdecay:
In PR #620, (limited) support for scripts was added, but without the support for `Vec` as argument types. The scripts should be able to run with `Vec` types as inputs.

hal3e:
https://github.com/FuelLabs/fuels-rs/issues/712

digorithm:
Done.

PR #718, opened 2022-12-02
feat: add vector support for scripts

hal3e:
closes https://github.com/FuelLabs/fuels-rs/issues/655

Scripts now support vectors as arguments. 

This PR is the joint work of @iqdecay, @segfault-magnet  and me. I decided to open the PR ASAP as this feature is highly requested by users.

This PR is based on https://github.com/FuelLabs/fuels-rs/pull/716 as we needed logs for debugging. I recommend we merge the logs PR first and than this one.


packages/fuels/tests/scripts/script_vectors/src/eq_impls.sw
```diff
@@ -0,0 +1,199 @@
...
+
+        let mut i = 0;
+        while i < self.len() {
+            if self.get(i).unwrap() != other.get(i).unwrap() {
+                return false;
+            }
+            i += 1;
+        }
+        true
+    }
```
iqdecay, line 198: All of this is great, very clean!

packages/fuels/tests/scripts/script_vectors/src/main.sw
```diff
@@ -0,0 +1,114 @@
...
+        let inner_inner_vec_4 = vec_from([9, 10, 11]);
+        inner_vec_2.push(inner_inner_vec_4);
+
+        exp_vec_in_a_vec_in_a_struct_in_a_vec.push(SomeStruct { a: inner_vec_2 });
+
+        require(vec_in_a_vec_in_a_struct_in_a_vec == exp_vec_in_a_vec_in_a_struct_in_a_vec, "vec_in_a_vec_in_a_struct_in_a_vec err");
+    }
+
+    true
+}
```
iqdecay, line 114: splendid

packages/fuels-contract/src/contract_calls_utils.rs
```diff
@@ -256,19 +256,22 @@ fn convert_to_signed_resources(spendable_resources: Vec<Resource>) -> Vec<Input>
         .collect()
 }
 
+/// Get the base offset for a script
+pub fn get_base_script_offset() -> usize {
+    ConsensusParameters::DEFAULT.tx_offset() + fuel_tx::Script::script_offset_static()
```
iqdecay, line 261: So I think we should rather query the consensus parameters here rather than using a default.

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -115,13 +108,13 @@ pub fn generate_script_main_function(
         #doc
         pub fn #name(&self #(,#arg_declarations)*) -> ScriptCallHandler<#output_type> {
             let arg_name_tokens = [#(#arg_names.into_token()),*];
-            let script_data = ABIEncoder::encode(&arg_name_tokens).expect("Cannot encode script
-            arguments").resolve(0);
             let script_binary = std::fs::read(self.binary_filepath.as_str())
                                         .expect("Could not read from binary filepath");
+            let script_offset = get_base_script_offset() + script_binary.len();
```
iqdecay, line 113: Ok that's the catch. How did you realize we were missing a base script offset?

iqdecay (changes requested):
Splendid work, gg for finding the missing base offset we struggled so much on lol. Few nits.

packages/fuels-contract/src/contract_calls_utils.rs
hal3e: This is just for the offset. We do not use the real values.

packages/fuels-core/src/code_gen/functions_gen.rs
hal3e: When we were looking at the various offsets that we found manually I made a mistake and divided them by 8. I wanted to make  bytes but the offset was already bytes. Then I subtracted the script binary length and got weird results.. When I did it without the division I stared getting constant offsets and later found that it was the base offset above ^

hal3e:
I  marked this PR as draft until we merge https://github.com/FuelLabs/fuels-rs/pull/716

packages/fuels-core/src/code_gen/functions_gen.rs
hal3e: I initially wanted to turn the offset to bytes because I thought that the offset we got was way to big.. later @segfault-magnet told me that contract calls also have around a 10kB offset, we checked it and then it all clicked :D

packages/fuels-contract/src/contract_calls_utils.rs
iqdecay: Oh you mean the `tx_offset` value does not change with the consensus parameters? If so I think it's useful to clarify it.

hal3e: You are right it does change. But it only depends on `max_inputs` that is why we did not have problems until now :/

hal3e: This was a great catch. If the `max_inputs` parameter is changes our contract calls break as the offset is not ok. I will refactor the code a bit to take the consensus_parameter from the node into account.

hal3e:
@iqdecay found a bug regarding how we calculate the script offset. Until now we were creating it form default of the `ConsesnsusParameters`. If the `max_inputs` field was set to something else than the default our scripts could not be called ass the offset are wrong. I updated the code to take into consideration the actual consensus_parameters and added a test for contract_calls

hal3e:
I also had to change the `ScriptCallHandler` to accept `UnresolvedByets` instead of the actual script data. This was done so that I can querry the provider and get the current consensus_parameters.

iqdecay (changes requested):
Very nice, just a nit on naming

iqdecay (dismissed):

packages/fuels/tests/scripts.rs
```diff
@@ -162,6 +162,59 @@ async fn main_function_tuple_types() -> Result<(), Error> {
...
+    let tuple_in_vec = vec![(0, 0), (1, 1)];
+    let vec_in_tuple = (vec![0, 1, 2], vec![0, 1, 2]);
+    let vec_in_a_vec_in_a_struct_in_a_vec = vec![
+        SomeStruct {
+            a: vec![vec![0, 1, 2], vec![3, 4, 5]],
+        },
+        SomeStruct {
+            a: vec![vec![6, 7, 8], vec![9, 10, 11]],
+        },
+    ];
```
digorithm, line 197: 😂

digorithm (approved):
Amazing work! 🌟

PR #726, opened 2022-12-07
feat!: add predicates abigen

hal3e:
This PR is is based on @mohammadfawaz [sway branch](https://github.com/FuelLabs/sway/pull/3523) which supports main function arguments for predicates.

We do not have to use the `input_predicate_data(0)` data anymore instead we pass the data directly into the main function. This is an example how a new predicate looks like:

```rust
predicate;

struct TestStruct {
    value: u64,
}

struct AnotherTestStruct {
    value: u64,
    number: u64,
}

fn main(test_struct: TestStruct, test_struct2: AnotherTestStruct) -> bool {
    test_struct.value == (test_struct2.value + test_struct2.number)
}
```
Previously, we could not use`TestStruct` or `AnotherTestStruct` directly in our test because it was not generated. 
With the new abigen we generate all the types and we also generate an encoding function `encode_data` which has the same inputs as the predicate's main function...

This is how the new abigen is used:
```rust
    predicate_abigen!(
        MyPredicate,
        "packages/fuels/tests/predicates/predicate_structs/out/debug/predicate_structs-abi.json"
    );

    let predicate =
        MyPredicate::load_from("tests/predicates/predicate_structs/out/debug/predicate_structs.bin")?;
```
Send funds to a predicate:

```rust
    predicate
        .receive(&sender, amount, asset_id, None)
        .await?;
```

Spend funds from a predicate. First encode the data and then spend to a wallet:
```rust
    predicate
        .encode_data(
            TestStruct { value: 192 },
            AnotherTestStruct {
                value: 64,
                number: 128,
            },
        )
        .spend(&receiver, predicate_balance, asset_id, None)
        .await?;
```
What was done:
- Extend the `Abigen` to handle predicates and add the predicate abigen macro
  -  Added `AbigenType` enum to specify if it is a `contract`, `script` or `predicate`
- Update the wallet's `spend_predicate` function to calculate the appropriate data offset for dynamic data
  - This means that we can use vectors as inputs in predicates
  - I moved the offset calculating function to `fuels_core::offsets` as it was used in `fuels-signers` and `fuels-contract`
- Added test for:
  - Coin and Message predicates
  - Primitive and custom type inputs
  - Generics and Vectors
- Update the docs

Breaking changes:
- Predicate instantiation
- Sending funds to and receiving funds from predicates

thanks @mohammadfawaz for the help from the compiler side!


MujkicA:
Looks good IMO

mohammadfawaz:
I'm assuming you need a new release of `forc` for this to get merged?

hal3e:
> I'm assuming you need a new release of `forc` with my PR for this to get merged?

Yes. Currently the CI is using your branch.

packages/fuels-core/src/code_gen/abigen.rs
```diff
@@ -217,52 +241,159 @@ impl Abigen {
...
+                            .transfer(
+                                self.address(),
+                                amount,
+                                asset_id,
+                                tx_parameters
+                            )
+                            .await
+                    }
+
+                    pub async fn spend_to_wallet(&self, wallet: &WalletUnlocked, amount:u64, asset_id: AssetId, tx_parameters: Option<TxParameters>) -> Result<Vec<Receipt>, SDKError> {
```
digorithm, line 302: The same suggestion about shortening the function name here. `spend_to_wallet(&self, wallet: &WalletUnlocked)` mentions "wallet" three times. 

```Rust
fn spend(&self, to: &WalletUnlocked) [...]
```

The above would achieve the same semantics but a lot more concisely!

digorithm (changes requested):
I did a first pass focusing on the docs and interfaces. I thought I should publish the review without fully finishing it so you can work through the suggestions while I do another deeper pass!

By the way, it's looking _beautiful_!

packages/fuels/tests/predicates.rs
```diff
@@ -1,404 +1,556 @@
...
+    predicate
+        .encode_data(
+            TestStruct { value: 192 },
+            AnotherTestStruct {
+                value: 64,
+                number: 128,
+            },
+        )
+        .spend(&receiver, predicate_balance, asset_id, None)
         .await?;
```
digorithm, line 303: 😍

digorithm (commented):
This is so good! 

Consider this an approval, but since we have to wait for `forc` to be released so we can pin a released version, this is just a comment to prevent an accidental merge.

packages/fuels/tests/predicates/predicate_address/src/main.sw
```diff
@@ -1,10 +1,7 @@
 predicate;
 
-use std::inputs::input_predicate_data;
+fn main(input: Address) -> bool {
```
mohammadfawaz, line 3: Do you mind adding one last test where `main` has a single argument that is a copy type (such as `u64`). Sorry if you've added that already and I missed it.

packages/fuels-signers/src/wallet.rs
```diff
@@ -679,15 +682,27 @@ impl WalletUnlocked {
...
-        let predicate_data = data.unwrap_or_default();
+        // Iterate through the spendable resources and calculate the appropriate offsets
+        // for the coin or message predicates
+        let mut offset = base_predicate_offset(&predicate.consensus_parameters().await?);
         let inputs = spendable_predicate_resources
             .into_iter()
             .map(|resource| match resource {
                 Resource::Coin(coin) => {
-                    self.create_coin_predicate(coin, asset_id, code.clone(), predicate_data.clone())
+                    offset += coin_predicate_data_offset(code.len());
```
segfault-magnet, line 692: Thinking if this shouldn't be unit tested. Quite a lot happening. 

To e2e test this you'd have to have multiple coins and messages tied to the predicate and have the predicate accept vectors as data and then spend the resources from the predicate.

Is there such a test? I know vectors are tested but wasn't sure how many inputs were used -- whether the offset is ever anything other than `base_predicate_offset`?

segfault-magnet (commented):
Excellent work. Especially with the bindings and the predicate data offset.

.github/workflows/ci.yml
```diff
@@ -16,10 +16,10 @@ env:
   CARGO_TERM_COLOR: always
   DASEL_VERSION: https://github.com/TomWright/dasel/releases/download/v1.24.3/dasel_linux_amd64
   RUSTFLAGS: "-D warnings"
-  FUEL_CORE_VERSION: 0.15.0
+  FUEL_CORE_VERSION: 0.15.1
   RUST_VERSION: 1.64.0
   FORC_VERSION: 0.31.1
```
mohammadfawaz, line 21: Ready to be bumped to `0.32.1`

packages/fuels-signers/src/wallet.rs
hal3e: We have exactly that test. And all the other test also include multiple messages and coins. I thought about moving this logic somewhere else but then got stuck and could not decide where to put it so I left it there..

The first offset is always `base_predicate_offset` (of course if it is not changed in the VM) then we have a list of inputs and we have to iterate through them and add the right offset to point to the predicate data part of the specific input

packages/fuels/tests/predicates/predicate_address/src/main.sw
hal3e: We do not have a test like that (there is one with two unsigned ints). I will add it :)

Salka1988 (commented):
Awsome! :medal_sports:

packages/fuels-core/src/code_gen/functions_gen.rs
```diff
@@ -126,6 +116,38 @@ pub fn generate_script_main_function(
     })
 }
 
+pub fn generate_predicate_encode_function(
```
digorithm, line 119: Does this function need to be`pub`?

digorithm (dismissed):
LGTM -- wonderful work. These tests are 10/10.

packages/fuels-core/src/code_gen/functions_gen.rs
hal3e: Good catch. It needs to be `pub(crate)` only.

digorithm (approved):
:shipit:

Issue #727, opened 2022-12-07
Required assets from multiple calls not correctly collated


segfault-magnet:
If we do a multi-call so that:
* call 1 requires 100 of asset_1
* call 2 requires 100 of asset_2
* call_3 requires 100 of asset_1

our current logic won't collate that properly, instead of requesting 200 of asset_1 and 100 of asset_2, it will require 100 of asset_1, then 100 of asset_2, then 100 of asset_1. 

Besides doing an unnecessary network call, requesting spendable resources for asset_1 multiple times might be suboptimal with regard to the algorithm used for choosing which resources to use.

PR #728, opened 2022-12-07
bug: asset ids not collated properly.

segfault-magnet:
closes: #727 

Just came across this gem while perusing the code.

group_by requires that the elements are sorted by the key you're grouping on.


iqdecay (approved):
Woah nice catch!

hal3e (approved):

PR #734, opened 2022-12-08
feat(contracts): add tx_id to FuelCallResponse

### LINKED ISSUE #733: Ergonomic way to fetch the transaction ID(s) from a contract call
Suppose the following workflow:

```Rust
            .methods()
            .initialize_counter(42)
            .call()
            .await?;
```

Currently, I believe there's no easy way to get the transaction ID of the transaction executed under `initialize_counter(42)`. We should provide that under the CallResponse that's returned upon calling a contract method.

Br1ght0ne:
###### Description of changes

Close #733.

- for scripts, `tx_id` is `Transaction::Script.id()`
- for contract calls, `tx_id` is `ExecutableFuelCall.tx.id()`

### Checklist


digorithm (approved):
Short and neat. gg!

Just gotta fix the tests, but after that feel free to merge!

packages/fuels/tests/scripts.rs
```diff
@@ -15,7 +15,7 @@ async fn test_transaction_script_workflow() -> Result<(), Error> {
     let provider = wallet.get_provider()?;
     let receipts = execution_script.execute(provider).await?;
 
-    let response = call_handler.get_response(receipts)?;
+    let response = call_handler.get_response(receipts, execution_script.tx.id())?;
```
MujkicA, line 18: Feels a bit off to pass the tx id here only so that it can be included in the response struct. Suddenly the users needs to know about the internals of 2 layers.
We could call  `get_executable_call()` and extract the id from within `get_response` but this is unreliable since we have no guarantee that the user won't edit the transaction before executing. Maybe caching the underlying transaction and giving the user a mut ref would solve this.

digorithm: Ha! I actually had the same nit but brushed it off. 

Now that someone else has also thought of that... that's enough confirmation to me that we should do the way @MujkicA suggested.

`get_executable_call()` is callable from within `get_response`, so... why not? Less work for the user.

packages/fuels-contract/src/contract.rs
```diff
@@ -859,7 +864,7 @@ impl MultiContractCallHandler {
     }
 
     /// Create a [`FuelCallResponse`] from call receipts
-    pub fn get_response<D: Tokenizable + Debug>(
+    pub async fn get_response<D: Tokenizable + Debug>(
```
iqdecay, line 867: it feels weird to me that now this function is an `async`. Isn't there a way to get the `tx_id` without having to do an async call? I feel like it's an extra network and I'm wondering if it's necessary to do it.

iqdecay (changes requested):
Just one question regarding async.

packages/fuels-contract/src/contract.rs
digorithm: @Br1ght0ne Thoughts on this? It seems like it's the only unresolved conversation here. We should ship this soon! 😁

Br1ght0ne: ddf32b3 instead gets `tx_id` from `call/simulate` calls, building on the assumption that you definitely should have `tx_id` if you own `receipts` (to be passed to `get_response`).

packages/fuels/tests/scripts.rs
Br1ght0ne: I do confer with @iqdecay that hitting the network in `get_response` feels unnecessary, but passing `tx_id` here is still wonky. I'm exploring memoizing/caching the transaction after execution and providing it through `get_response` in an impossible-to-misuse manner.

iqdecay (changes requested):
So I like that we don't need to call the network to get the `tx_id`, but I do agree with the comment about changing the `get_response` API. 
First of all this is afaiu a breaking change that we should consider carefully.
Second, I do feel like it is wonky. 
I'd like to see what a caching solution would like I think :smile: 

Can't we include an extra `Receipts` type that just gives us the txid? But it would feel bad to "lie" about it being a tx receipt.

digorithm:
Hmm, seems like we're bound to modify the API, regardless of the options. Having to append `.await` is a change to its API, anyway. 

Are we deciding between `let response = call_handler.get_response(receipts).await?;` and `let response = call_handler.get_response(receipts, executable_call.tx.id())?;`? 

If so, the first option doesn't require the user to know more internal details and manually fetch the tx id, which I believe to be better than the second option.

iqdecay:
Agreed but I feel like Br1ght0ne is hinting that he maybe has a solution that isn't async **and** doesn't need the user to know the tx_id. 
My question was: is it too urgent for us to wait on the implementation which seems tricky? Or can we wait a bit.

hal3e (approved):

MujkicA (dismissed):

iqdecay (changes requested):
LGTM, but please add documentation and tests

packages/fuels-programs/src/contract.rs
```diff
@@ -420,6 +420,8 @@ impl ContractCall {
 pub struct ContractCallHandler<T: Account, D> {
     pub contract_call: ContractCall,
     pub tx_parameters: TxParameters,
+    /// Initially `None`, this gets set after the transaction is submitted
```
iqdecay, line 423:
```suggestion
    // Initially `None`, gets set to the right tx id after the transaction is submitted
```

packages/fuels-programs/src/contract.rs
```diff
@@ -754,6 +759,8 @@ pub struct MultiContractCallHandler<T: Account> {
     pub contract_calls: Vec<ContractCall>,
     pub log_decoder: LogDecoder,
     pub tx_parameters: TxParameters,
+    /// Initially `None`, this gets set after the transaction is submitted
```
iqdecay, line 762:
```suggestion
    // Initially `None`, gets set to the right tx id after the transaction is submitted
```

packages/fuels-programs/src/script_calls.rs
```diff
@@ -58,6 +58,8 @@ impl ScriptCall {
 pub struct ScriptCallHandler<T: Account, D> {
     pub script_call: ScriptCall,
     pub tx_parameters: TxParameters,
+    /// Initially `None`, this gets set after the transaction is submitted
```
iqdecay, line 61:
```suggestion
    // Initially `None`, gets set to the right tx id after the transaction is submitted
```

iqdecay (dismissed):
nice

iqdecay (approved):
Lgtm

hal3e (approved):

PR #735, opened 2022-12-08
add `mint transaction` support

### LINKED ISSUE #702: Add support for the new mint transaction
The client's `Transaction` now has a `Mint` type, used to mint new coins. The SDK should have a helper interface to help users mint coins. Where this interface will live is up for discussion and is part of resolving this issue.

Salka1988:
closes #702 

add `mint transaction` support in

### DIFF
--- a/packages/fuels-signers/src/provider.rs
+++ b/packages/fuels-signers/src/provider.rs
@@ -19,6 +19,7 @@ use fuel_gql_client::{
     interpreter::ExecutableTransaction,
 };
 use fuels_core::constants::{DEFAULT_GAS_ESTIMATION_TOLERANCE, MAX_GAS_PER_TX};
+use fuels_core::tx::{Checkable, Output, TxPointer};
 use fuels_types::{
     bech32::{Bech32Address, Bech32ContractId},
     block::Block,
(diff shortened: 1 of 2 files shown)


Salka1988:
Is there or will there be some kind of endpoint on the `fuel-client` through which I will be able to do `mint transactions`
@digorithm @Voxelot

digorithm:
> Is there or will there be some kind of endpoint on the `fuel-client` through which I will be able to do `mint transactions` @digorithm @Voxelot

I'm not sure I follow your question. Why would we need an _endpoint_ specifically for that? Doesn't sending a `Transaction` of type `Mint` already work for sending a transaction of type mint?

packages/fuels-signers/src/provider.rs
```diff
@@ -141,6 +142,24 @@ impl Provider {
         }
     }
 
+    pub async fn mint_transaction(
```
digorithm, line 145: `mint_transaction` sounds like a verb. i.e., we're minting a transaction. Which is wrong. 

I'd suggest renaming it to `mint` only. Plus, add some doc-level comments here.

packages/fuels/tests/contracts.rs
```diff
@@ -770,3 +771,26 @@ async fn test_contract_call_with_non_default_max_input() -> Result<(), Error> {
...
+        1,
+        AssetId::from([3u8; 32]),
+    )];
+
+    let a = provider.mint_transaction(output_coins).await?;
+
+    dbg!(&a);
+
+    dbg!(wallet.get_balances().await?);
+
```
digorithm, line 794: Remove `dbg!` statements.

digorithm (changes requested):
A couple of nits.

Also, it seems like mint transactions aren't supported by the client. Just noticed that: https://github.com/FuelLabs/fuel-core/blob/b50f0a9ec8a924ad6cf2e3cb12e4699d70bbb9e1/fuel-txpool/src/txpool.rs#L76. 

@Voxelot can you confirm that that's the case?

Salka1988:
We’re not prioritizing this now and it needs to be better specified.

PR #737, opened 2022-12-09
fix `setup_test_client` parameters

### LINKED ISSUE #730: `setup_test_client` uses default parameters when provided with `ChainConfig` and `ConsensusParameters`
When using `setup_test_client` by passing both `ChainConfig` and `ConsensusParameters`, it will ignore the provided  `ConsensusParameters` and replace them with a default value.

We should also consider improving the UX around setting multiple parameters from different config structs like `block_gas_limit` and `max_gas_per_tx` and provide docs.

Salka1988:
closes #730

Now when using `setup_test_client` by passing both `ChainConfig` and `ConsensusParameters`, the provided parameters won't be ignored


Salka1988:
```rust
pub async fn setup_test_client(
    coins: Vec<Coin>,
    messages: Vec<Message>,
    node_config: Option<Config>,
    chain_config: Option<ChainConfig>,
)
```
regarding the second part of this PR, I can suggest that we create a builder pattern struct and put all the parameters we need in it. :man_shrugging:

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -165,17 +165,25 @@ pub async fn setup_test_client(
...
 
+    chain_conf = if consensus_parameters_config.is_some() {
+        ChainConfig {
+            transaction_parameters: consensus_parameters_config.unwrap(),
+            ..chain_conf
+        }
+    } else {
+        chain_conf
+    };
+
```
segfault-magnet, line 186:
```rust
    let chain_conf = chain_config.unwrap_or_else(|| ChainConfig {
        transaction_parameters: consensus_parameters_config.unwrap_or_default(),
        initial_state: Some(StateConfig {
            coins: Some(coin_configs),
            contracts: None,
            messages: Some(message_configs),
            ..StateConfig::default()
        }),
        ..ChainConfig::local_testnet()
    });
```

Salka1988: > ```rust
>     let chain_conf = chain_config.unwrap_or_else(|| ChainConfig {
>         transaction_parameters: consensus_parameters_config.unwrap_or_default(),
>         initial_state: Some(StateConfig {
>             coins: Some(coin_configs),
>             contracts: None,
>             messages: Some(message_configs),
>             ..StateConfig::default()
>         }),
>         ..ChainConfig::local_testnet()
>     });
> ```

It can't be like this. This was exactly what was causing the bug.

segfault-magnet: Good catch!

digorithm:
> ```rust
> pub async fn setup_test_client(
>     coins: Vec<Coin>,
>     messages: Vec<Message>,
>     node_config: Option<Config>,
>     chain_config: Option<ChainConfig>,
>     consensus_parameters_config: Option<ConsensusParameters>,
> )
> ```
> 
> regarding the second part of this PR, I can suggest that we create a builder pattern struct and put all the parameters we need in it. 🤷‍♂️ @digorithm
> 
> or we could exclude `consensus_parameters` because we already have them in `ChainConfig` and user could do something like this:
> 
> ```rust
> let chain_config = ChainConfig {
>     transaction_parameters: ConsensusParameters::default(),
>     ..ChainConfig::local_testnet()
> };
> ```

That makes sense to me @Salka1988.

PR #738, opened 2022-12-09
feat: Add `messagesTypes` to ProgramABI

mohammadfawaz:
This is needed for https://github.com/FuelLabs/sway/issues/3409. Basically, we want to be able to decode the data sent via `smo`, just like logs, so messages need to store information in the JSON ABI.

Doesn't look like this impacts anything at the moment as everything compiles fine.

Spec change: https://github.com/FuelLabs/fuel-specs/pull/444
Compiler change: https://github.com/FuelLabs/sway/pull/3564 that needs this (currently using a local copy of `fuels-rs`)


digorithm (approved):
LGTM!

PR #743, opened 2022-12-14
feat: support `raw_slice` returns from scripts and contracts

### LINKED ISSUE #703: Handle `raw_slice` in the SDK
`raw_slice` is a new type that was introduced in the compiler to allow returning dynamically-sized data from scripts. The idea is that someone could write
```rust
fn main() -> raw_slice {
    // Convert to a vector
    let vec: Vec<u64> = Vec::new();

    vec.push(1);
    vec.push(2);
    vec.push(3);

    // Return it
    vec.as_raw_slice()
}
```
and they would get a `ReturnData` receipt that contains a byte representation of `[1, 2, 3]`.

We recently discussed an option to hide this from users who are testing contracts as follows:

The user writes a contract method that returns `Vec<u64>`:
```rust
fn foo() -> Vec<u64>;
```
From the JSON ABI, it would be clear that this is a vector.

When the SDK creates a script that calls `foo()`, it writes:
```rust
fn main() -> raw_slice {
    ...
    caller.foo().as_raw_slice()
}
```
Note that returning from the contract to the script itself does not need a `raw_slice` since the pointer from the contract is valid in the script, but **not** valid outside the script, hence the need for a `raw_slice`.

With above, users who are testing their contracts never need to think about raw slices, but they would only ever be allowed to return a single vector at a time, which is an annoying but reasonable limitation for the time being.

If we like this direction, I can create a similar issue in `fuels-ts`.

iqdecay:
This PR closes #703 by adding support for `raw untyped slice` in scripts. 
It adds the possibility of returning a vector from a script, as basically an array of `u64`.


packages/fuels-core/src/types/raw_slice.rs
```diff
@@ -0,0 +1,73 @@
+use crate::{Parameterize, Token, Tokenizable};
+use fuels_types::errors::Error;
+use fuels_types::param_types::ParamType;
+use std::fmt::{Debug, Display, Formatter};
+
+// To be used when interacting with contracts which have strings in their ABI.
```
segfault-magnet, line 6: I'm not sure the explanation about strings and their length is warranted here.

I'd stop at: 
`RawSlice` is a mapping of the contract type `raw_slice` -- currently the only way of returning dynamically sized data from a contract.

packages/fuels-core/src/types/raw_slice.rs
```diff
@@ -0,0 +1,73 @@
...
+            ))),
+        }
+    }
+
+    fn into_token(self) -> Token {
+        Token::RawSlice(self.data)
+    }
+}
+
+impl TryFrom<Vec<u64>> for RawSlice {
```
segfault-magnet, line 44: If you change the `new` to return `Self` then this can be an implementation of `From` instead of `TryFrom`.

packages/fuels-core/src/types/raw_slice.rs
```diff
@@ -0,0 +1,73 @@
...
+use fuels_types::errors::Error;
+use fuels_types::param_types::ParamType;
+use std::fmt::{Debug, Display, Formatter};
+
+// To be used when interacting with contracts which have strings in their ABI.
+// The length of a string is part of its type -- i.e. str[2] is a
+// different type from str[3]. The FuelVM strings only support ascii characters.
+#[derive(Debug, PartialEq, Clone, Eq)]
+pub struct RawSlice {
+    data: Vec<u64>,
```
segfault-magnet, line 11: I'm not sure if we should encapsulate the data. We don't need to guard it from anything like we needed to for the sized ascii type.

Maybe doing a simple wrapper around the data as we did for `Bits256` and `B512`. Especially because we only care about the data inside and not the wrapper itself, which is a necessary evil for the `Parameterize` and `Tokenizeable` mechanisms to work.

Otherwise, we'd need to implement a bunch of traits to make the usage of this type painless. Like `PartialEq`, `AsRef`, and others.

packages/fuels/tests/scripts/script_raw_slice/.gitignore
```diff
@@ -0,0 +1,2 @@
+out
```
segfault-magnet, line 1: We should probably see what to do about the `.gitignores` that seem to have started being generated by forc.

Currently only this forc project and one regarding predicate data commit a `.gitignore`. We can probably lift that out somewhere on the level of the whole e2e dir.

packages/fuels/tests/scripts/script_raw_slice/src/main.sw
```diff
@@ -0,0 +1,16 @@
+script;
+
+fn main(length: u64) -> raw_slice {
+    let mut vec = Vec::new();
+    vec.push(42);
```
segfault-magnet, line 5: I'm not sure if we gain much by randomizing the elements here, except in complexity.

Perhaps a single monotonically-increasing vector with the last element being `length`-1:

1 -> 0
2 -> 0,1
...
n -> 0, 1, ..., n-1

It's not all zeroes so we're covered on that front, and I doubt something will create monotonically increasing junk data.

packages/fuels/tests/scripts.rs
```diff
@@ -290,3 +290,36 @@ async fn test_script_call_with_non_default_max_input() -> Result<(), Error> {
...
+    let num_coins = 1;
+    let amount = 1000;
+    let config = WalletsConfig::new(Some(num_wallets), Some(num_coins), Some(amount));
+
+    let mut wallets = launch_custom_provider_and_get_wallets(config, None, None).await;
+    let wallet = wallets.pop().unwrap();
+    let bin_path = "../fuels/tests/scripts/script_raw_slice/out/debug/script_raw_slice.bin";
+    let instance = bimbam_script::new(wallet.clone(), bin_path);
+
+    let mut expected = vec![42, 1337, 1987];
```
segfault-magnet, line 310: If we go for the monotonically increasing vector then this might become:
```rust
    for length in 0..3 {
        let response = instance.main(length).call().await?.value;
        assert_eq!(response, (0..length).collect::<Vec<_>>());
    }
```

segfault-magnet (commented):
Haven't seen all the discussions around this, is this supported only by scripts, or can contract methods return `raw_slice` as well?

If so we should probably cover that with tests as well.

packages/fuels-core/src/types/raw_slice.rs
iqdecay: This is a leftover from a copy-paste haha

packages/fuels/tests/scripts/script_raw_slice/.gitignore
iqdecay: I added the `.gitignore` pattern for test projects in the root gitignore, the same way we ignore build artifacts.

MujkicA (commented):
LGTM
I would wait for @segfault-magnet to give it one more pass before merging.

digorithm (dismissed):
Looking really nice and clean. Let's wait for @segfault-magnet's last pass since he pushed a bunch of comments on here.

segfault-magnet (approved):
Contracts can return raw_slices too, if you see the benefit you can add a test for that. 

From a functional point of view, it should be tested. But if we loosen our blackboxing a little then we know that the same logic is going to handle the type resolving and data decoding for the raw slice regardless of the program type.

In any case, LGTM!

digorithm (dismissed):

iqdecay:
Added a test for contracts, I didn't think it worked too lol

packages/fuels-core/src/traits/tokenizable.rs
```diff
@@ -557,6 +577,27 @@ mod tests {
...
     }
 
+    #[test]
+    fn test_from_token_raw_slice() -> Result<(), Error> {
+        let data = vec![42; 11];
+        let token = Token::RawSlice(data.clone());
+
+        let slice = RawSlice::from_token(token)?;
+
+        assert_eq!(slice.0, data);
```
hal3e, line 587: You implemented `PartialEq` for `Vec`
```suggestion
        assert_eq!(slice, data);
```

hal3e (dismissed):
LGTM. Good work! Left a nit...

digorithm (approved):
Great work, very clean implementation!

Issue #742, opened 2022-12-14
Methods without `#[payable]` annotation cannot accept coins


anton-trunov:
Once https://github.com/FuelLabs/sway/pull/3450 is merged the SDK also needs to forbid accepting coins in non-payable functions. The current plan is to introduce function attributes into the JSON ABI schema to mark payable functions as such. I'm happy to take any comments/suggestions on this.

Spec issue: https://github.com/FuelLabs/fuel-specs/issues/446
Compiler change: https://github.com/FuelLabs/sway/pull/3450

digorithm:
@anton-trunov could you open the same issue on the TS SDK repo as well (https://github.com/FuelLabs/fuels-ts)?

anton-trunov:
@digorithm Done! https://github.com/FuelLabs/fuels-ts/issues/675

PR #744, opened 2022-12-15
feat: Add `attributes` to `ABIFunction`

anton-trunov:
This is needed for https://github.com/FuelLabs/sway/pull/3450 and for #742 eventually.

Spec issue: https://github.com/FuelLabs/fuel-specs/issues/446
Compiler change: https://github.com/FuelLabs/sway/pull/3450

**Correction**: the explanation below is only true at the parser level, the distinction between `foo` and `foo()` disappears later in the pipeline, so I changed the type of `arguments` field to `Vec<String>`.

The `Option<Vec<String>>` in

```
pub struct Attribute {
     pub name: String,
     pub arguments: Option<Vec<String>>,
 }
```

is needed because in Sway a parameterless attribute `foo` and an attribute with empty parameter list `foo()` are not the same.

### DIFF
--- a/examples/rust_bindings/src/abi.json
+++ b/examples/rust_bindings/src/abi.json
@@ -21,8 +21,7 @@
         "name": "",
         "type": 0,
         "typeArguments": null
-      },
-      "attributes": []
+      }
     },
     {
       "inputs": [
@@ -37,8 +36,7 @@
         "name": "",
         "type": 0,
         "typeArguments": null
-      },
-      "attributes": []
+      }
     }
   ]
 }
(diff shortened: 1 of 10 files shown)


segfault-magnet:
This seems to be a breaking change. The CI is complaining: 

> SerdeJson(Error("missing field `attributes`", line: 43, column: 5))


I haven't checked more closely but it seems that the way we're adding this is breaking backward compatibility -- i.e. we won't be able to release until the compiler starts providing the attribute field. Do you have an ETA this? 

If it's too long we should probably look into guiding serde to default-populate the field if it is not present in the json file and still remain backward compatible.

anton-trunov:
@segfault-magnet I made the function attributes optional -- it turns out to be too hard to introduce a breaking change like an obligatory field because of close coupling of the Sway and Fuels-rs repos.

iqdecay (approved):
lgtm

PR #748, opened 2022-12-15
Automatic fee payment with modified inputs

### LINKED ISSUE #729: Scripts/Calls with custom tx inputs can get overwritten
When the user sets custom inputs on a script, there is a chance that they will be overwritten if they contain the base asset.

Due to how `wallet.add_fee_coins()` modifies the input set to include enough assets to cover tx fees, the manually added inputs may end up being replaced.

Salka1988:
closes #729

bugs fixed: 

description https://github.com/FuelLabs/fuels-rs/issues/729#issue-1482093180

~set `tx` in `execution_script.rs` to `pub(crate)` and add `extend_transaction_inputs` function which prohibits adding `inputs` with already existing utxo_id or message_id to `tx.inputs`~


packages/fuels-contract/src/execution_script.rs
```diff
@@ -107,10 +111,59 @@ impl ExecutableFuelCall {
...
+        let (tx_utxo_ids, tx_message_ids): (HashSet<Bytes32>, HashSet<MessageId>) =
+            self.tx.inputs().iter().partition_map(|input| match input {
+                Input::CoinSigned { utxo_id, .. } => Either::Left(utxo_id.tx_id()),
+                Input::CoinPredicate { utxo_id, .. } => Either::Left(utxo_id.tx_id()),
+                Input::Contract { utxo_id, .. } => Either::Left(utxo_id.tx_id()),
+                Input::MessageSigned { message_id, .. } => Either::Right(message_id),
+                Input::MessagePredicate { message_id, .. } => Either::Right(message_id),
+            });
+        (tx_utxo_ids, tx_message_ids)
+    }
```
segfault-magnet, line 150:
```suggestion
    fn extract_inputs_from_tx(&mut self) -> (HashSet<Bytes32>, HashSet<MessageId>) {
        self.tx.inputs().iter().partition_map(|input| match input {
            Input::CoinSigned { utxo_id, .. } => Either::Left(utxo_id.tx_id()),
            Input::CoinPredicate { utxo_id, .. } => Either::Left(utxo_id.tx_id()),
            Input::Contract { utxo_id, .. } => Either::Left(utxo_id.tx_id()),
            Input::MessageSigned { message_id, .. } => Either::Right(message_id),
            Input::MessagePredicate { message_id, .. } => Either::Right(message_id),
        })
    }
```

packages/fuels-contract/src/execution_script.rs
```diff
@@ -107,10 +111,59 @@ impl ExecutableFuelCall {
...
+        tx_message_ids: &HashSet<MessageId>,
+    ) -> Option<Input> {
+        inputs.to_owned().iter().cloned().find(|input| match input {
+            Input::CoinSigned { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
+            Input::CoinPredicate { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
+            Input::Contract { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
+            Input::MessageSigned { message_id, .. } => tx_message_ids.contains(message_id),
+            Input::MessagePredicate { message_id, .. } => tx_message_ids.contains(message_id),
+        })
+    }
```
segfault-magnet, line 164:
```suggestion
    fn check_inputs_for_duplicates(
        inputs: &[Input],
        tx_utxo_ids: &HashSet<Bytes32>,
        tx_message_ids: &HashSet<MessageId>,
    ) -> Option<Input> {
        inputs
            .iter()
            .find(|input| match input {
                Input::CoinSigned { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
                Input::CoinPredicate { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
                Input::Contract { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
                Input::MessageSigned { message_id, .. } => tx_message_ids.contains(message_id),
                Input::MessagePredicate { message_id, .. } => tx_message_ids.contains(message_id),
            })
            .cloned()
    }
```
No need to clone the whole collection twice, find the element first and then clone only that element.

Or better still, return a `Option<&Input>` because you're only calling `utxo_id(&self)`  on the returned value. So you can do that without any copying/cloning:
```rust
    fn check_inputs_for_duplicates<'a>(
        inputs: &'a [Input],
        tx_utxo_ids: &HashSet<Bytes32>,
        tx_message_ids: &HashSet<MessageId>,
    ) -> Option<&'a Input> {
        inputs.iter().find(|input| match input {
            Input::CoinSigned { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
            Input::CoinPredicate { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
            Input::Contract { utxo_id, .. } => tx_utxo_ids.contains(utxo_id.tx_id()),
            Input::MessageSigned { message_id, .. } => tx_message_ids.contains(message_id),
            Input::MessagePredicate { message_id, .. } => tx_message_ids.contains(message_id),
        })
    }
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -516,25 +516,21 @@ impl WalletUnlocked {
         // If the tx doesn't consume any UTXOs, attempting to repeat it will lead to an
         // error due to non unique tx ids (e.g. repeated contract call with configured gas cost of 0).
         // Here we enforce a minimum amount on the base asset to avoid this
+        let is_consuming_utxos = tx.inputs().iter().any(|input| match input {
+            Input::CoinSigned { .. }
+            | Input::CoinPredicate { .. }
+            | Input::MessageSigned { .. }
+            | Input::MessagePredicate { .. } => true,
+            Input::Contract { .. } => false,
+        });
```
segfault-magnet, line 525:
```suggestion
        let is_consuming_utxos = tx
            .inputs()
            .iter()
            .any(|input| !matches!(input, Input::Contract { .. }));
```

segfault-magnet (commented):
You need to add this file to the `Summary.md` in order for it to be visible to users.

Salka1988:
> You need to add this file (`modify-contract-call-tx.md`) to the `Summary.md` in order for it to be visible to users.

Done

packages/fuels-contract/src/execution_script.rs
```diff
@@ -114,10 +120,181 @@ impl ExecutableFuelCall {
...
     }
+
+    pub fn set_inputs(
+        &mut self,
+        inputs: Vec<Input>,
+    ) -> std::result::Result<&mut ExecutableFuelCall, Error> {
+        let mut hash_set_utxo_id = HashSet::new();
+        let mut hash_set_messages = HashSet::new();
+
+        for input in &inputs {
```
segfault-magnet, line 137: Seems to be a few places where duplicates are being handled, you can do something like this to clean it up:
```rust
    fn extract_duplicate_ids(inputs: &[Input]) -> Vec<String> {
        inputs
            .iter()
            .filter_map(|input| {
                input
                    .utxo_id()
                    .map(Either::Left)
                    .or_else(|| input.message_id().map(Either::Right))
            })
            .duplicates()
            .map(|id| match id {
                Either::Left(id) => format!("{id:?}"),
                Either::Right(id) => format!("{id:?}"),
            })
            .collect()
    }
    
    fn validate_no_duplicate_ids(inputs: &[Input]) -> Result<(), Error> {
        let duplicate_ids = Self::extract_duplicate_ids(inputs);

        if duplicate_ids.is_empty() {
            Ok(())
        } else {
            Err(Error::InvalidData(format!(
                "Provided inputs are not unique! Duplicated ids: {duplicate_ids:?}"
            )))
        }
    }
```

then you can call `validate_no_duplicate_ids` wherever you need it, for example:
```rust
    pub fn extend_inputs(&mut self, inputs: Vec<Input>) -> Result<&mut ExecutableFuelCall, Error> {
        Self::validate_no_duplicate_ids(&inputs)?;

        self.tx.inputs_mut().extend(inputs);

        Ok(self)
    }
```
or
```rust
    pub fn set_inputs(
        &mut self,
        inputs: Vec<Input>,
    ) -> std::result::Result<&mut ExecutableFuelCall, Error> {
        Self::validate_no_duplicate_ids(&inputs)?;

        *self.tx.inputs_mut() = inputs;
        Ok(self)
    }
```

segfault-magnet (commented):
I'll do another pass regarding tests, need to check whether something is being tested.

packages/fuels-contract/src/execution_script.rs
Salka1988: This looks much cleaner. I'll take this but I'll have to change a few things to make it work. Thanx!

packages/fuels-contract/src/execution_script.rs
```diff
@@ -114,10 +119,136 @@ impl ExecutableFuelCall {
...
+
+    pub fn get_inputs(&mut self) -> &Vec<Input> {
+        self.tx.inputs()
+    }
+
+    pub fn extend_inputs(&mut self, inputs: Vec<Input>) -> Result<&mut ExecutableFuelCall, Error> {
+        Self::validate_no_duplicate_ids(chain!(self.tx.inputs(), &inputs))?;
+        self.tx.inputs_mut().extend(inputs);
+        Ok(self)
+    }
```
segfault-magnet, line 177: We might have a problem here.

When we request `get_asset_inputs_for_amount` from the node, we'll get resources of the node's choosing following a randomized algorithm to prevent dust accumulation. 

During the formation of our tx we ask for resources to cover the gas fee and cover whatever assets the contract calls required. The node is free to return whichever resources it deems appropriate to cover our request.

Later on, after we've formed the tx, somebody might want to add some more inputs for whatever reason. They're probably going to do that using `extend_inputs`. And they're probably going to get those inputs by calling `get_asset_inputs_for_amount`. 

There is nothing (AFAIK) preventing the possibility that the node is going to return resources that we've already 'reserved' by either covering the transaction costs or because a contract call demanded them. The node can't know we're planning to do anything with the resources it had suggested just a few moments before while we were building the tx.

So a user might find this snippet to work most of the time:

```rust
let mut a_call = contract.methods().some_method().get_executable_call();
let some_more_inputs = wallet.get_asset_inputs_for_amount(...);
a_call.extend_inputs(some_more_inputs);
```
but once in a while, the node will return a resource we've already placed in our tx inputs and then this code will error out due to the duplicated ids.

How often this happens would depend on the client's algorithm for choosing resources.

It would be unfortunate for the user to have his application error-out even though he had the resources to cover his request, but the node choose an already 'reserved' resource to avoid dust.

We could either:
1.  leave the user to his own devices and have him request all his resources and filter out the ones we've already chosen for his tx, or repeat `get_asset_inputs_for_amount` until the node changes its mind, or, 
2. develop a way of extending inputs such that we request resources from the node only once for a given asset id.
3. both of the above, let the user use the error-prone one if he knows what he's doing

The #1 way is more flexible because the user would choose *exactly* which resource he wants to extend as inputs. I'm out of my depth here but that could be used to make decisions in the contract? Heard about predicates making decisions based on inputs.

The #2 way is more friendly, just have the user extend the inputs by noting his wish of `(asset_id, amount)` and we'll find the coins for that wish and for our other needs at the same time, avoiding duplicates. 

So something like:
```rust
call.add_input_resources(asset_id, amount);
```

and right before executing the script, we would do one pass per asset id to actually find all the needed resources.


Dunno. Been doing my own thing for a while now, might be out of the loop. @FuelLabs/sdk-rust opinions?

Salka1988: You're right. I will take this into account and try to implement something

Salka1988:
I've reverted this branch back to b0c2e80 where there is a fix for #729. While solving the problem we came across another one which is described here https://github.com/FuelLabs/fuels-rs/pull/748#discussion_r1055594621. As @digorithm said, it would be better to create a new issue/PR for this new bug and solve it separately, which I will do.

docs/src/cookbook/modify-contract-call-tx.md
```diff
@@ -0,0 +1,34 @@
+# Modifying the contract call transaction
+
+The Rust SDK lets you modify the underlying transaction of a contract call. An example use case for this would be to create a transaction that atomically executes a contract call and transfers assets between addresses. 
+
+First we need to include some traits to access the inputs/outputs of the transaction:
```
MujkicA, line 5:
```suggestion
First, we need to include some traits to access the inputs/outputs of the transaction:
```

docs/src/cookbook/modify-contract-call-tx.md
```diff
@@ -0,0 +1,34 @@
...
+
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:modify_call_inputs_include}}
+```
+
+Now let's setup 2 wallets that hold a base asset coin, and a coin of some other asset.
```
MujkicA, line 11:
```suggestion
Now let's set up two wallets that hold a base asset coin and a coin of another asset.
```

docs/src/cookbook/modify-contract-call-tx.md
```diff
@@ -0,0 +1,34 @@
...
+
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:modify_call_inputs_setup}}
+```
+
+In the next step, we deploy our test contract, create an instance of the contract, get the call handler and
```
MujkicA, line 17:
```suggestion
In the next step, we deploy our test contract, create an instance of the contract and get the call handler. Finally, we get the underlying call using `get_executable_call()`. 
```

docs/src/cookbook/modify-contract-call-tx.md
```diff
@@ -0,0 +1,34 @@
...
+finally, the underlying call by using `get_executable_call()`.
```
MujkicA, line 18:
```suggestion 
```

docs/src/cookbook/modify-contract-call-tx.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+In the next step, we deploy our test contract, create an instance of the contract, get the call handler and
+finally, the underlying call by using `get_executable_call()`. 
+
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:modify_call_inputs_instance}}
+```
+
+Since the executable call acts as a wrapper fot the actual transaction, we can use it to modify fields like the inputs, outputs, maturity etc. Here, we setup inputs coming from `wallet_1` to be transferred to `wallet_2`.
```
MujkicA, line 24:
```suggestion
Since the executable call acts as a wrapper for the actual transaction, we can use it to modify fields like the inputs, outputs, maturity, etc. Here, we set up inputs from `wallet_1` to be transferred to `wallet_2`.
```

docs/src/cookbook/modify-contract-call-tx.md
```diff
@@ -0,0 +1,34 @@
...
+
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:modify_call_inputs_execute}}
+```
+
+Finally, we verify the result of the contract call and the transfer.
```
MujkicA, line 30:
```suggestion
Lastly, we verify the result of the contract call and the transfer.
```

MujkicA (commented):
Some suggestions for the docs

packages/fuels-signers/src/wallet.rs
```diff
@@ -519,25 +519,87 @@ impl WalletUnlocked {
...
+    }
+
+    /// Add base asset inputs to the transaction to cover the estimated fee.
+    /// The original base asset amount cannot be calculated reliably from
+    /// the existing transaction inputs because the selected resources may exceed
+    /// the required amount to avoid dust. Therefore we require it as an argument.
+    ///
+    /// Requires contract inputs to be at the start of the transactions inputs vec
+    /// so that their indexes are retained
+    pub async fn add_fee_resources1<Tx: Chargeable + field::Inputs + field::Outputs>(
```
MujkicA, line 559: This needs to be removed

Salka1988: Done!

MujkicA (approved):
LGTM

digorithm (approved):
Great work!

PR #749, opened 2022-12-16
release: Bump versions to `0.32.3`

anton-trunov:
Let's release the function attributes JSON ABI change into the wild!


iqdecay (approved):

mohammadfawaz:
Shouldn't we be bumping to 0.33.0 instead to avoid the same issues that we faced a few days ago with messagesTypes? Changes to the JSON ABI types breaks the compiler.

PR #750, opened 2022-12-16
release: Bump versions to `0.33.0`

anton-trunov:
The functions attributes feature for JSON ABI breaks compatibility with 0.32 for the Sway compiler because of the changes in `fuels-types`, so the `fuels-rs` version should be corrected from 0.32.3 to 0.33.0.

This changes is following this comment https://github.com/FuelLabs/fuels-rs/pull/749#issuecomment-1354594423 by @mohammadfawaz.


digorithm (approved):
Please, let's be careful with the release versions; there was also [another PR](https://github.com/FuelLabs/fuels-rs/pull/737) that had a breaking change. `0.32.3` shouldn't have existed 😃

cc @iqdecay since you approved the release PR.

PR #761, opened 2023-01-03
fix: fix transfer script

### LINKED ISSUE #630: Script in `build_transfer_tx` should be empty
It was mentioned by @adlerjohn, in our internal conversations that an empty script should still be a valid script. The transfer script being built [here](https://github.com/FuelLabs/fuels-rs/blob/4f154c77f83ca62c9b7be4bb49b2422836ac9a99/packages/fuels-signers/src/wallet.rs#L281) isn't empty; it contains only a `ret(1)`. The rationale behind it being `ret(1)` _seems_ to be lost. Removing the line linked and effectively passing `script: vec![]` causes some tests to fail. 

We should investigate why that is the case and try and make it work with an empty script, if possible. For more motivation behind it and/or more explanation, reach out to @adlerjohn.

iqdecay:
This closes #630.

### DIFF
--- a/docs/src/SUMMARY.md
+++ b/docs/src/SUMMARY.md
@@ -39,8 +39,6 @@
   - [Predicates](./getting-started/predicates.md)
     - [Predicate data](./predicates/predicate-data.md)
     - [Signatures example](./predicates/send-spend-predicate.md)
-  - [Scripts](./scripts/index.md)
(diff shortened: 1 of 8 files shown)


hal3e (approved):
LGTM :tada:

digorithm (approved):
:shipit:

PR #770, opened 2023-01-06
Methods without #[payable] cannot accept assets

### LINKED ISSUE #742: Methods without `#[payable]` annotation cannot accept coins

MujkicA:
Closes #742

`.call_params()` now returns a `Result` which either resolves to the `ContractCallHandler` instance or an `InvalidCallParameters` error. The latter happens when the user tries to set call parameters with an amount > 0 on a contract method that isn't payable.

Initially, I wanted to avoid implementing it like this. One previous attempt was to have a generic param <T> on `ContractCall` that would define the `call_params` field as a payable or non-payable version of the current `CallParameters` struct. But this meant that for `MultiContractCallHandler` handler to work, we would need to define a trait for the `CallParameters`. The resulting interface would provide some kind of setters/getters for both the payable and non-payable params version. If non-payable call params have setters for the amount, they have to be fallible, which is just the current solution but more bloat.

Open to suggestions.


mohammadfawaz:
`forc 0.33.0` was out yesterday if you'd like to update the version in this PR.

packages/fuels-types/src/lib.rs
```diff
@@ -52,6 +52,16 @@ pub struct ABIFunction {
     pub attributes: Option<Vec<Attribute>>,
 }
 
+impl ABIFunction {
+    pub fn is_payable(&self) -> bool {
+        if let Some(attributes) = &self.attributes {
+            return attributes.iter().any(|attr| attr.name == "payable");
+        }
+
+        false
```
segfault-magnet, line 61: another way of writing it. I've no strong preference either way. I just wanted to point out the ability to iterate through an `Option` and flatten the contained `Vec`.
```suggestion
        self.attributes
            .iter()
            .flatten()
            .any(|attr| attr.name == "payable")
```

segfault-magnet (dismissed):
Clean, simple. 

The first weekend I find myself not doing anything, I think I'll try and see what a type-safe alternative looks like. 

My gut tells me it's going to be ugly because it's going to make the call handlers generic -- but there is a small chance that a clean solution might be found.

segfault-magnet (dismissed):

hal3e (dismissed):

digorithm:
CI is broken and branch is out of sync!

MujkicA:
> CI is broken and branch is out of sync!

The forc bugfix has recently been merged, we still need a new release for our test contracts to build properly.

https://github.com/FuelLabs/sway/issues/3741

Salka1988:
> > CI is broken and branch is out of sync!
> 
> The forc bugfix has recently been merged, we still need a new release for our test contracts to build properly.
> 
> [FuelLabs/sway#3741](https://github.com/FuelLabs/sway/issues/3741)

Maybe you could use FORC PATCH BRANCH: "" FORC PATCH REVISION: ""  from ci.yml for testing until they issue a new release?

Salka1988 (dismissed):

MujkicA:
#790 needs to be merged first.

digorithm:
Can we sync this with master? 👀 https://github.com/FuelLabs/fuels-rs/pull/790 is closed, so this should be unblocked.

PR #772, opened 2023-01-08
refactor: move `ContractCall`s into their own module

iqdecay:
- This PR is preparatory work in order to implement an `estimate_tx_dependencies` that works for both scripts and contracts.


iqdecay:
Closing as it is going to be a nightmare before the big PRs come in.

PR #777, opened 2023-01-10
fix: do not modify receipts

hal3e:
We should not delete receipts when decoding the output from a contract call.


digorithm (approved):
Nice, thanks for the quick fix!

PR #778, opened 2023-01-11
feat: add custom assets to contract call

### LINKED ISSUE #758: Develop a way to expand `inputs` and avoid duplicates
When we request `get_asset_inputs_for_amount` from the node, we'll get resources of the node's choosing following a randomized algorithm to prevent dust accumulation.

During the formation of our tx we ask for resources to cover the gas fee and cover whatever assets the contract calls required. The node is free to return whichever resources it deems appropriate to cover our request.

Later on, after we've formed the tx, somebody might want to add some more inputs for whatever reason. They're probably going to do that using extend_inputs. And they're probably going to get those inputs by calling `get_asset_inputs_for_amount`.

There is nothing (AFAIK) preventing the possibility that the node is going to return resources that we've already 'reserved' by either covering the transaction costs or because a contract call demanded them. The node can't know we're planning to do anything with the resources it had suggested just a few moments before while we were building the tx.

So a user might find this snippet to work most of the time:

```rust
let mut a_call = contract.methods().some_method().get_executable_call();
let some_more_inputs = wallet.get_asset_inputs_for_amount(...);
a_call.extend_inputs(some_more_inputs);
```

but once in a while, the node will return a resource we've already placed in our tx inputs and then this code will error out due to the duplicated ids. This is problematic.

We could either:

1- leave the user to his own devices and have him request all his resources and filter out the ones we've already chosen for his tx, or repeat get_asset_inputs_for_amount until the node changes its mind, or,
2- develop a way of extending inputs such that we request resources from the node only once for a given asset id.
    both of the above, let the user use the error-prone one if he knows what he's doing

The `#1` way is more flexible because the user would choose exactly which resource he wants to extend as inputs. I'm out of my depth here but that could be used to make decisions in the contract? Heard about predicates making decisions based on inputs.

The  `#2` way is more friendly, just have the user extend the inputs by noting his wish of (asset_id, amount) and we'll find the coins for that wish and for our other needs at the same time, avoiding duplicates.
So something like:
```rust
call.add_input_resources(asset_id, amount);
```
and right before executing the script, we would do one pass per asset id to actually find all the needed resources.

### LINKED ISSUE #803: Enable users to attach custom asset transfers to contract calls
As part of solving #758 we want to enable custom asset transfers within the same transaction as a contract call. 
This should be possible without having the user modify the tx inputs directly.

Salka1988:
Add new feature to the `ContractCall` struct by introducing a new field called custom_assets. Additionally, a new method called `add_custom_asset` is added to the struct, which allows users to add a custom asset to the contract call by providing the asset's asset_id, amount, and an optional address for the recipient of the generated outputs. 

closes #803


docs/src/SUMMARY.md
```diff
@@ -28,6 +28,7 @@
     - [Connecting wallets](./calling-contracts/calls-with-different-wallets.md)
     - [Transaction parameters](./calling-contracts/tx-params.md)
     - [Call parameters](./calling-contracts/call-params.md)
+    - [Custom asset transfers](./calling-contracts/custom-asset-transfer.md)
```
iqdecay, line 31:
```suggestion
    - [Custom assets transfer](./calling-contracts/custom-assets-transfer.md)
```

docs/src/calling-contracts/custom-asset-transfer.md
```diff
@@ -0,0 +1,7 @@
+# Custom asset transfers
```
iqdecay, line 1:
```suggestion
# Custom assets transfer
```

iqdecay: Filename should be `custom-assets-transfer` IMO

packages/fuels-contract/src/contract.rs
```diff
@@ -346,6 +348,7 @@ pub struct ContractCall {
     pub message_outputs: Option<Vec<Output>>,
     pub external_contracts: Vec<Bech32ContractId>,
     pub output_param: ParamType,
+    pub custom_assets: HashMap<(AssetId, Option<Bech32Address>), u64>,
```
iqdecay, line 351: Why is it an `Option<Bech32Address>` and not just a `Bech32Address`?

packages/fuels-contract/src/contract.rs
```diff
@@ -433,6 +436,12 @@ impl ContractCall {
             |r| matches!(r, Receipt::Panic { reason, .. } if *reason.reason() == PanicReason::ContractNotInInputs ),
         )
     }
+
+    pub fn add_custom_asset(&mut self, asset_id: AssetId, amount: u64, to: Option<Bech32Address>) {
+        let key = (asset_id, to);
+        let sum = self.custom_assets.get(&key).unwrap_or(&0) + amount;
```
iqdecay, line 442: nice

packages/fuels-contract/src/contract.rs
iqdecay: I think it should be `Bech32Address` with the `None` option being replaced by the user's address.

iqdecay (changes requested):
If the interface is much much better than the previous attempt, I'm not really convinced this solves the issue #758. I feel like a complete solution would also change the way `add_tx_fees` works, because we want to make sure we are not reserving the same `Resource` twice.

docs/src/calling-contracts/custom-asset-transfer.md
MujkicA: How about custom asset transfer since it is a single transfer of a single asset

examples/contracts/src/lib.rs
```diff
@@ -713,4 +713,39 @@ mod tests {
...
+
+        let contract_id = Contract::deploy(
+            "../../packages/fuels/tests/contracts/contract_test/out/debug/contract_test.bin",
+            &wallet,
+            TxParameters::default(),
+            StorageConfiguration::default(),
+        )
+        .await?;
+
+        let contract_instance = MyContract::new(contract_id, wallet);
```
segfault-magnet, line 737:
```suggestion
        setup_contract_test!(
            Wallets("wallet"),
            Abigen(
                name = "MyContract",
                abi = "packages/fuels/tests/contracts/contract_test"
            ),
            Deploy(
                name = "contract_instance",
                contract = "MyContract",
                wallet = "wallet"
            )
        );

        let other_wallet = WalletUnlocked::new_random(None);
```

packages/fuels-programs/src/call_utils.rs
```diff
@@ -178,13 +192,36 @@ pub(crate) fn get_transaction_inputs_outputs(
...
         extract_message_outputs(calls)
     )
     .collect();
     (inputs, outputs)
 }
 
+fn generate_custom_outputs(calls: &[ContractCall]) -> Vec<Output> {
+    calls
+        .iter()
+        .flat_map(|call| call.custom_assets.iter().collect::<Vec<_>>())
```
segfault-magnet, line 206:
```suggestion
        .flat_map(|call| &call.custom_assets)
```
You can sum references to amounts later on, no need to copy them all into another collection.

packages/fuels/tests/contracts.rs
```diff
@@ -944,6 +944,75 @@ async fn test_contract_call_with_non_default_max_input() -> Result<(), Error> {
...
+
+    let balance_asset_1 = wallet_1.get_asset_balance(&asset_id_1).await?;
+    let balance_asset_2 = wallet_1.get_asset_balance(&asset_id_2).await?;
+    assert_eq!(balance_asset_1, initial_amount - amount_1);
+    assert_eq!(balance_asset_2, initial_amount - amount_2);
+
+    let balance_asset_1 = wallet_1.get_asset_balance(&asset_id_1).await?;
+    let balance_asset_2 = wallet_1.get_asset_balance(&asset_id_2).await?;
+    assert_eq!(balance_asset_1, initial_amount - amount_1);
+    assert_eq!(balance_asset_2, initial_amount - amount_2);
```
segfault-magnet, line 1011:
```suggestion
    let balance_asset_1 = wallet_2.get_asset_balance(&asset_id_1).await?;
    let balance_asset_2 = wallet_2.get_asset_balance(&asset_id_2).await?;
    assert_eq!(balance_asset_1, initial_amount + amount_1);
    assert_eq!(balance_asset_2, initial_amount + amount_2);
```
You might have forgotten to change the second block to validate that the second wallet got the assets. 

Otherwise the two blocks are identical.

packages/fuels-programs/src/contract.rs
```diff
@@ -447,6 +449,12 @@ impl ContractCall {
             |r| matches!(r, Receipt::Panic { reason, .. } if *reason.reason() == PanicReason::ContractNotInInputs ),
         )
     }
+
+    pub fn add_custom_asset(&mut self, asset_id: AssetId, amount: u64, to: Option<Bech32Address>) {
+        let key = (asset_id, to);
+        let sum = self.custom_assets.get(&key).unwrap_or(&0) + amount;
+        self.custom_assets.insert(key, sum);
```
segfault-magnet, line 456:
```suggestion
        *self.custom_assets.entry((asset_id, to)).or_default() += amount;
```

packages/fuels-contract/src/contract.rs
MujkicA: I figured it would be neat to also have the possibility to just add any custom amount without sending it to any address. If anything is supposed to consume that asset, it can, otherwise it gets returned to the user via Output:change.
But I guess it isn't obvious enough what happens with it, is it getting burned or returned as change?

packages/fuels/tests/contracts.rs
```diff
@@ -1006,10 +1006,68 @@ async fn test_add_custom_assets() -> Result<(), Error> {
...
+    for length in 0..=10 {
+        let response = contract_methods.return_raw_slice(length).call().await?;
+        assert_eq!(response.value, (0..length).collect::<Vec<_>>());
+    }
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_deploy_error_messages() -> Result<(), Error> {
```
iqdecay, line 1046: I think this test doesn't belong here.

iqdecay (changes requested):
I still think this implementation does not solve issue #758 because we will still have the issue that the `add_fees` method can potentially attribut resources that were "reserved" by the user or vice-versa.

packages/fuels/tests/contracts.rs
MujkicA: Git did something weird there, this test is not part of this PR

MujkicA:
> I still think this implementation does not solve issue https://github.com/FuelLabs/fuels-rs/issues/758 because we will still have the issue that the add_fees method can potentially attribut resources that were "reserved" by the user or vice-versa.

True. The plan is to prohibit the user from directly altering inputs of contract call on the tx level in a future PR. This is just there to replace the functionality.

iqdecay:
>True. The plan is to prohibit the user from directly altering inputs of contract call on the tx level in a future PR. This is just there to replace the functionality.

Ok then LGTM! Just edit the PR description so it reflects that it does not actually solve #758

packages/fuels/tests/contracts.rs
iqdecay: Yeah I figured

iqdecay (dismissed):
LGTM, please edit PR description

iqdecay (approved):
LGTM with the new PR description 😉

Issue #736, opened 2022-12-09
fuels-rs types reorganization


hal3e:
Based on user and team members feedback we have decided to reorganizes parts of `fuels-rs`. Specifically we will focus on the following action items: 
- Re-exporting types in a single place — users will only care about this single place to pull types.
- Keep trait impl inside fuels-core, move specific types to `fuels-types`
- We’re moving traits impl from `lib.rs` to new file name TBD
- Check if a symbol can be imported through only one path
- A package can only export symbols it owns
- Define the role of fuels-contract and rename it to what ever is appropriate
- Delete tools/fuels-abi-cli from the repo (It should probably live somewhere else in Fuels)

PR #779, opened 2023-01-13
refactor!: reorganize types

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/736

These are the task defined in the issue:

- [x] Re-exporting types in a single place — users will only care about this single place to pull types.
  - I have tried to export all of our types through the `fuels` package and use those imports in our tests

- [x] Keep trait impl inside fuels-core, move specific types to `fuels-types`
  - I have moved the types from `fuel-core` to `fuels_types::core`. Now all of our types are defined in `fuels-types`
  
- [x] We’re moving traits impl from `lib.rs` to new file name TBD
  - I have added a new folder `traits` and added three new files: `tokenizable.rs`, `parameterize.rs` and `decodable_log.rs`. The three new files hold the trati implementations for all types (that need them). 
  
- [x] Check if a symbol can be imported through only one path
  - AFAIK our types can only be imported from a single place now.

- [x] A package can only export symbols it owns

- [x] Define the role of fuels-contract and rename it to what ever is appropriate
  - I have given the name `fuels-programs`.
  
- [x] Delete tools/fuels-abi-cli from the repo (It should probably live somewhere else in Fuels)

The PR is big but it is basically just moving things around.

Breaking changes:

- changed paths from `fuels_core::types` to `fuels_types::core`
- traits path also changed to `fuels_core::traits::*`


packages/fuels-types/src/core.rs
```diff
@@ -0,0 +1,119 @@
...
+
+use fuel_types::bytes::padded_len;
+use strum_macros::EnumString;
+
+use std::fmt;
+
+pub mod bits;
+pub mod byte;
+pub mod native;
+pub mod sized_ascii_string;
```
segfault-magnet, line 11: I wouldn't make this public. 
These types are split into files, not as some grand organization plan but rather to limit file size to something manageable. 

This doesn't need to be seen outside, especially since you're exporting everything a few lines below.

segfault-magnet (commented):
You seem to be a fan of nested `use` statements. We can't support this via the CI ([due to this](https://github.com/rust-lang/rustfmt/issues/4991))

But since things are already mixed, perhaps you can give some structure to the imports. You can run a one-time custom rustfmt on the code to achieve the changes you like. Just change the rustfmt.toml adding these lines:
```
unstable_features = true
group_imports = "StdExternalCrate"
imports_granularity = "Crate"
```
then run `cargo +nightly fmt --all` on `fuels-rs`.

segfault-magnet (dismissed):
Went through the PR with @hal3e, LGTM!

packages/fuels-core/src/utils.rs
```diff
@@ -29,3 +30,14 @@ pub fn ident(name: &str) -> Ident {
 pub fn safe_ident(name: &str) -> Ident {
     syn::parse_str::<SynIdent>(name).unwrap_or_else(|_| ident(&format!("{}_", name)))
 }
+
+pub fn get_logs_hashmap(
```
iqdecay, line 34: I'm not convinced by the name here

packages/fuels-programs/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "fuels-contract"
+name = "fuels-programs"
```
iqdecay, line 2: :smiley:

iqdecay (approved):
Almost perfect! Leaving as approved because I'm nitpicking :smile:.
Congrats on this, seriously.

docs/src/calling-contracts/tx-params.md
```diff
@@ -6,7 +6,7 @@ Transaction parameters are:
 2. Gas limit;
 3. Maturity.
 
-You can configure these parameters by creating an instance of [`TxParameters`](https://github.com/FuelLabs/fuels-rs/blob/adf81bd451d7637ce0976363bd7784408430031a/packages/fuels-contract/src/parameters.rs#L7) and passing it to a chain method called `tx_params`:
+You can configure these parameters by creating an instance of [`TxParameters`](https://github.com/FuelLabs/fuels-rs/blob/adf81bd451d7637ce0976363bd7784408430031a/packages/fuels-programs/src/parameters.rs#L7) and passing it to a chain method called `tx_params`:
```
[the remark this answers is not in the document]
hal3e: I updated the docs and added an `ANCHOR` to not rely on external links.

packages/fuels-core/src/utils.rs
hal3e: changed to `log_type_lookup`

MujkicA: Maybe something that is more descriptive of what the hashmap holds. 
`get_contract_id_log_id_mapping`?

MujkicA (dismissed):

iqdecay (dismissed):
LGTM! :ship: it

Salka1988 (dismissed):

digorithm (changes requested):
Looking really good, a step in the right direction, and a massive devex improvement!

One thing, though: The CI workflow file (https://github.com/FuelLabs/fuels-rs/blob/master/.github/workflows/ci.yml) needs to be updated accordingly; otherwise, the publish action will break, as it's verifying the tags for publishing for `fuels-contract` and not the new `fuels-program`.

iqdecay (approved):
let's gooo

Issue #752, opened 2022-12-20
Solve the circular dependency issue between the SDK, compiler, and tooling


digorithm:
# Problem

Currently, there's a circular dependency between the SDK, compiler, and other tooling projects. This is essentially due to the SDK using `forc` in the CI builds to build Sway projects for its tests _and_ the compiler's harness tests using the SDK for testing.

# Solution

After many discussions (that happened IRL in Mexico recently, [notes here](https://www.notion.so/fuellabs/Circular-Dependencies-97494685807d49e58ad882022a41cd05)), we've settled with the following approach: 

- [x] Move the types in `packages/fuels-types/src/lib.rs` to the following repository: https://github.com/FuelLabs/fuels-types; 
  - https://github.com/FuelLabs/fuels-types/issues/1
- [x] Setup a CI environment (identical to `fuels-rs`'s CI) to publish this new package, `fuel-abi-types`, to crates.io. This new `fuel-abi-types` that contains the `ProgramABI` type will be shared between the SDK and the compiler;
  - https://github.com/FuelLabs/fuels-types/issues/2
- [x] Once `fuel-abi-types` has been published, remove the `ProgramABI`-related types from `packages/fuels-types/src/lib.rs` from `fuels-rs` and pull them from the new package, `fuel-abi-types`;
  - https://github.com/FuelLabs/fuels-rs/issues/753
- [ ] Instead of relying on `forc`, the binary, to build the Sway projects in the `fuels-rs` repo, we'll call directly `build()` from `sway`'s `forc-pkg` package. Effectively removing the dependency on `forc` itself;
  - https://github.com/FuelLabs/fuels-rs/pull/780 
- [x] Remove the dependency of `sway-core` and `forc-pkg` on `fuels-rs` and replace it with a dependency on the shared repo `fuel-abi-types`. This may _just_ work given that the dependency is already on a package named `fuel-abi-types` (the one from `fuels-rs` currently). 
  - https://github.com/FuelLabs/sway/issues/3638

Then, we'll achieve the proposed DAG (Directed Acyclic Graph):
- JSON ABI types (`fuel-abi-types`) → SDK
- JSON ABI types (`fuel-abi-types`) → `sway-core`
- `sway-core` → `forc-pkg`
- `forc-pkg` → SDK
- SDK + `sway-core` → `sdk-harness` tests

mohammadfawaz:
I updated the description to include a compiler task that may end up being a no-op.

[...]

PR #780, opened 2023-01-13
remove forc 

Salka1988:
add forc-pkg and functions

CI now utilizes the test-projects-workspace script to construct tests. This script employs the forc-pkg build function to construct all tests within a single call, as opposed to individually building each test. The Forc.toml file located within the tests directory contains the [workspace] configuration, listing all tests as members. While this method may result in slightly slower build times compared to the previous test-projects method, it eliminates the need for the forc library. However, there is still room for optimization to speed up the build process.


Salka1988:
Do we need forc-fmt in CI?

iqdecay:
I'm not sure I understand what issue this PR solves, can you maybe please reformulate? I feel like we're still dependent on `forc` here, but maybe that's not the point?

digorithm:
> I'm not sure I understand what issue this PR solves, can you maybe please reformulate? I feel like we're still dependent on `forc` here, but maybe that's not the point?

The point is better described here https://github.com/FuelLabs/fuels-rs/issues/752 and in the Notion document linked in it; an artifact that came out of a meeting with @mohammadfawaz and other folks from the compiler and client team that happened in the company retreat.

.github/workflows/ci.yml
```diff
@@ -41,31 +41,29 @@ jobs:
...
           command: run
-          args: --bin test-projects -- build
+          args: --bin test-projects-workspace -- build
 
-      - name: Check format of Sway test projects
-        uses: actions-rs/cargo@v1
-        with:
-          command: run
-          args: --bin test-projects -- format --check
+#      - name: Check format of Sway test projects todo: do we need this?
```
hal3e, line 62: I would definitely keeps this in the CI. This will ensure that all our our sway projects are formatted on the main branch.

Cargo.toml
```diff
@@ -25,4 +25,5 @@ members = [
     "scripts/test-projects",
     "scripts/check-docs",
     "tools/fuels-abi-cli",
+    "scripts/test-projects-workspace"
```
hal3e, line 28: I would not change the script name. `test-projects` seems ok even though we use the `workspace` way of building our projects.

scripts/test-projects-workspace/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+forc-pkg = "0.32.2"
+forc-util = "0.32.2"
+#forc-fmt = "0.32.2" todo do we need this
+walkdir = "2.3.2"
+glob = "0.3.1"
+
+termcolor = "1.1"
+tokio = { version = "1.21.2", features = ["process", "macros", "rt", "rt-multi-thread"] }
+futures-util ="0.3.24"
+clap = { version ="4.0.15", features =["cargo", "color", "derive"] }
```
hal3e, line 19: I would keep the CLI interface we had `test_projects {build, format etc.}` If we agrree that we do not need that anymore than you can remove `clap` as a dependency.

scripts/test-projects-workspace/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[dependencies]
+forc-pkg = "0.32.2"
+forc-util = "0.32.2"
+#forc-fmt = "0.32.2" todo do we need this
+walkdir = "2.3.2"
+glob = "0.3.1"
+
+termcolor = "1.1"
+tokio = { version = "1.21.2", features = ["process", "macros", "rt", "rt-multi-thread"] }
+futures-util ="0.3.24"
```
hal3e, line 18: AFAIK we do not need `tokio` and `fucures-util` as you only run one build now.

hal3e (changes requested):

.github/workflows/ci.yml
iqdecay: Agreed that we need to keep the Sway projects properly formatted

Cargo.toml
iqdecay: Agreed

packages/fuels/tests/Forc.toml
```diff
@@ -0,0 +1,71 @@
+[workspace]
```
iqdecay, line 1: I think we should have a check to ensure all tests directories are listed here. Maybe open an issue for this?

packages/fuels/tests/contracts/auth_testing_abi/src/main.sw
```diff
@@ -1,6 +1,6 @@
 library auth_testing_abi;
 
 abi AuthTesting {
-    fn is_caller_external() -> bool;
+      fn is_caller_external() -> bool;
```
iqdecay, line 4:
```suggestion
     fn is_caller_external() -> bool;
```

.github/workflows/ci.yml
```diff
@@ -41,31 +41,29 @@ jobs:
...
+#        uses: actions-rs/cargo@v1
+#        with:
+#          command: run
+#          args: --bin test-projects -- format --check
```
iqdecay, line 66:
```suggestion
      - name: Check format of Sway test projects
        uses: actions-rs/cargo@v1
        with:
          command: run
          args: --bin test-projects-workspace -- format --check
```

If I'm not mistaken this is how the formatting would look like without the `forc` binary?

iqdecay (changes requested):
Nice! Sorry I didn't understand earlier, got confused by the difference between the `forc` crate and the `forc` binary. Asked for some changes :)

.github/workflows/ci.yml
Salka1988: We'll keep it :)

Cargo.toml
Salka1988: This name is temporary

scripts/test-projects-workspace/Cargo.toml
Salka1988: I added it because I was testing fmt. In the end, we will see if it will be needed or not. Perhaps none of these changes will be necessary.

packages/fuels/tests/Forc.toml
Salka1988: Maybe. If we don't decide that we will leave it as it was, because we don't use force as a dependency

Salka1988:
We are closing this PR due to the details listed here https://github.com/FuelLabs/fuels-rs/issues/752#issuecomment-1386110240

PR #781, opened 2023-01-17
refactor: move ContractCalls into their own module

iqdecay:
This PR is preparatory work in order to implement an `estimate_tx_dependencies` that works for both scripts and contracts.
This PR is an adapted version of #772, now that the previous big PRs are in. 

- Rename the `call_utils` module into `contract_call_utils` because it contains a majority of functions that are used on `ContractCall`s.
- Move utils that are used with both `ScriptCall`s and `ContractCall`s inside the new `call_utils` module. 
- Move `ContractCall` and `ContractCallHandler` types into their own `contract_call` module
- Use a new folder structure for `fuels-programs`:
```bash
src
├── calls
│   ├── call_response.rs
│   ├── call.rs
│   ├── call_utils.rs
│   ├── contract_call.rs
│   ├── contract_call_utils.rs
│   └── script_call.rs
├── calls.rs
├── contract.rs
├── execution_script.rs
├── lib.rs
└── logs.rs
```


hal3e (commented):
I had a first pass through the changes. IMO the structure of `fuels-programs` is confusing. We have:
- call_response.rs
- call_utils.rs
- contract_call.rs
- contract_call_utils.rs
- contract.rs
- execution_script.rs
- lib.rs
- logs.rs
- script_call.rs
 
 @FuelLabs/sdk-rust what do you think we add some structure with folders. Maybe separate with `contract` and `script` folders. Or a `call` folder for the call related files...

iqdecay:
> I had a first pass through the changes. IMO the structure of `fuels-programs` is confusing. We have:
> 
>     * call_response.rs
> 
>     * call_utils.rs
> 
>     * contract_call.rs
> 
>     * contract_call_utils.rs
> 
>     * contract.rs
> 
>     * execution_script.rs
> 
>     * lib.rs
> 
>     * logs.rs
> 
>     * script_call.rs
> 
> 
> @FuelLabs/sdk-rust what do you think we add some structure with folders. Maybe separate with `contract` and `script` folders. Or a `call` folder for the call related files...

I like the idea of making a distinction between `contracts` and `scripts`. Maybe I went a little overboard with the separation into files.

segfault-magnet (commented):
graph

Not related to this PR per-se:
I hacked together a tool to generate graphs from use statements -- we have a few weird things going on in our dependency graph. 

For example:
* There is a dependency cycle between `contract_call` and `execution_script` where contract call needs `ExecutableFuelCall` from `execution_script` and `execution_script` needs `ContractCall` from `contract_call`.
* A similar cycle appears between `contract_call_utils` and `contract_call`
* `contract_call_utils` are mostly used by `execution_script` (a thing to be used by both scripts and contracts alike) although by the name one would think that `contract_call` would be the one using them the most
* We go from a contract to an executable script to mentioning contracts again, I have a feeling that overall our abstraction isn't being reduced the deeper into the graph we go.

We should refactor this code in the future to at least remove the cyclic dependencies. 

Related to this PR:
Agree with what Halil said about folders.

iqdecay:
Tried an improved structure, that solves the first cyclic dependency.

packages/fuels-programs/src/calls.rs
```diff
@@ -0,0 +1,5 @@
+pub mod call_response;
+pub mod call_utils;
+pub mod contract_call;
+pub mod contract_call_utils;
+pub mod script_call;
```
segfault-magnet, line 5:
```suggestion
pub mod call_response;
mod call_utils;
pub mod contract_call;
mod contract_call_utils;
pub mod script_call;
```

packages/fuels-programs/src/calls/call_utils.rs
```diff
@@ -0,0 +1,143 @@
...
+use fuels_types::bech32::Bech32ContractId;
+use fuels_types::core::Token;
+use fuels_types::errors::Error;
+use fuels_types::param_types::{ParamType, ReturnLocation};
+use fuels_types::{bech32::Bech32Address, resource::Resource};
+use itertools::Itertools;
+
+/// Based on the receipts returned by the call, the contract ID (which is null in the case of a
+/// script), and the output param, decode the values and return them.
+pub fn get_decoded_output(
```
segfault-magnet, line 16:
```suggestion
pub(crate) fn get_decoded_output(
```

packages/fuels-programs/src/calls/contract_call.rs
```diff
@@ -0,0 +1,611 @@
...
+};
+use crate::{
+    calls::call_response::FuelCallResponse,
+    calls::call_utils::get_decoded_output,
+    execution_script::ExecutableFuelCall,
+    logs::{decode_revert_error, LogDecoder},
+};
+
+/// How many times to attempt to resolve missing tx dependencies.
+pub const DEFAULT_TX_DEP_ESTIMATION_ATTEMPTS: u64 = 10;
```
segfault-magnet, line 35: A duplicate can be found in `fuels_programs::contract`. Should probably be deleted since it isn't used.

packages/fuels-programs/src/calls/contract_call_utils.rs
```diff
@@ -222,87 +210,17 @@ fn generate_custom_outputs(calls: &[ContractCall]) -> Vec<Output> {
...
-        .iter()
-        .map(|resource| match resource {
-            Resource::Coin(coin) => coin.asset_id,
-            Resource::Message(_) => BASE_ASSET_ID,
-        })
-        .collect()
-}
-
 fn extract_variable_outputs(calls: &[ContractCall]) -> Vec<Output> {
     calls
```
segfault-magnet, line 214:
```rust
    calls
        .iter()
        .flat_map(|call| &call.variable_outputs)
        .cloned()
        .collect()
```
Perhaps this is better because the original would clone whole vectors and then discard them moving the elements into the final collected vector.

This way you're iterating through each element and cloning the element inside placing it in its final resting place in the collected vector.

packages/fuels-programs/src/calls/contract_call_utils.rs
```diff
@@ -222,87 +210,17 @@ fn generate_custom_outputs(calls: &[ContractCall]) -> Vec<Output> {
...
-        .map(|(idx, contract_id)| {
-            Input::contract(
-                UtxoId::new(Bytes32::zeroed(), idx as u8),
-                Bytes32::zeroed(),
-                Bytes32::zeroed(),
-                TxPointer::default(),
-                contract_id,
-            )
-        })
+        .flat_map(|call| call.message_outputs.clone())
```
segfault-magnet, line 223: The same suggestion here to extract clone into `cloned()`

packages/fuels-programs/src/calls/script_call.rs
```diff
@@ -81,7 +92,8 @@ where
             script_binary,
             encoded_args,
             inputs: vec![],
-            outputs: vec![],
+            message_outputs: vec![],
+            variable_outputs: vec![],
```
segfault-magnet, line 96: Is there some benefit from keeping message outputs and variable outputs separate in both `ContractCall` and `ScriptCall`? 

We couldn't find a reason why they couldn't just be one collection of `Outputs`. 

AFAIK the only restriction is to have contract outputs first because that is the way we've populated the contract input indexes. The rest can go in any order.

If that is true then these outputs don't need to be separated and we can simplify the code while we're here.

packages/fuels-types/src/lib.rs
```diff
@@ -8,7 +8,7 @@ pub mod block;
 pub mod chain_info;
 pub mod coin;
 pub mod constants;
-mod core;
+pub mod core;
```
segfault-magnet, line 11: Everything from core is publicly exported here to provide a uniform type path access in order to simplify abigen.

There is no need for this mod to be made public.

segfault-magnet (commented):
Joint review with @hal3e.

packages/fuels-programs/src/calls/script_call.rs
```diff
@@ -107,8 +119,12 @@ where
         self
     }
 
-    pub fn with_outputs(mut self, outputs: Vec<Output>) -> Self {
-        self.script_call = self.script_call.with_outputs(outputs);
+    pub fn with_message_outputs(mut self, message_outputs: Vec<Output>) -> Self {
```
MujkicA, line 122: We agreed to remove this in #795, just a reminder to be careful not to re-introduce it with this PR

packages/fuels-programs/src/calls/script_call.rs
iqdecay: That was my initial reaction too. I tried building with both of them merged but it gets really messy downstream when you are actually building the transaction, because the code to extract only one type of message is ugly. 
Also it was already the case with contracts so I just imitated.
But I can try again and we can see how it looks.

iqdecay: It actually also breaks the `estimate_tx_dependency` functions for a reason that's not immediate to diagnose. I suggest we put this into a separate issue because it would be beyond the scope of this PR.

iqdecay: Nvm it works fine, I thought we had to separate the two for generating outputs but you're rigth, as long as contract outputs are here, we don't need to differentiate them.

MujkicA:
I made several changes to the same files in #801, most notably I removed `execution_script.rs` completely. I believe it resolves some of the dependency issues.
Overall, I am unsure of the benefits of the new structure introduced here.
Could you help me understand what about these changes is essential to enable `estimate_tx_dependencies`  for both scripts and contracts.

iqdecay:
I believe the "only once" implementation of the shared behaviors between `ContractCall` and `ScriptCall` is an important improvement for code structure and quality. The idea of this PR was to be preparatory work for this implementation, and ultimately to be able to have `estimate_tx_dependencies` be a generic function. 
I'm going to update my PR to take into account #801 which seems like a great improvement.

MujkicA:
> I believe the "only once" implementation of the shared behaviors between `ContractCall` and `ScriptCall` is an important improvement for code structure and quality. The idea of this PR was to be preparatory work for this implementation, and ultimately to be able to have `estimate_tx_dependencies` be a generic function. I'm going to update my PR to take into account #801 which seems like a great improvement.

Lmk if I can be of help

iqdecay:
Closing in favor of an upcoming PR.

Issue #785, opened 2023-01-18
`Derive`ing parts of `abigen!` generated code


segfault-magnet:
`Abigen` generates types and their implementations of `Parameterize` and `Tokenizable`. For all other types, we have to do that manually. 

This manual labor can be eliminated by writing a `derive`-compatible procedural macro so that we might generate the traits as follows:
```rust
#[derive(Parameterize, Tokenizable)]
struct SomeStruct{
 a: u8
 }
```

This not only reduces the boilerplate implementations for other types, but it also simplifies `Abigen` since we would no longer generate the implementations there but rather just slap a `#![derive(Parameterize, Tokenizable)]` onto the generated types.

This has consequences in tests where we would have less code to match against an expectation. Test separation is also improved since we can now separate testing the trait implementation expansion vs testing everything else.

The tests would become simpler as well since we would no longer have to do the long-winded setup with `TypeApplications` and `TypeDeclarations` while testing the trait implementations.

PR #790, opened 2023-01-19
refactor!: move abigen to fuels-macros crate

hal3e:
This PR is related to https://github.com/FuelLabs/fuels-rs/issues/785

The new `derive` macros will be implemented in the `fuels-macros` package, formerly `fuels-abigen-macro`.

To use the  `derive` macros in `fuels-types` we need to import `fuels-macros` there. However, this will create a dependency cycle:

`fuels-types` <- `fuels-core` <- `fuels-macros` <- `fuels-types` ...

@segfault-magnet and I agreed that the natural place to break the cycle is `fuels-core` <- `fuels-macros`. This dependency is mainly due to the `Abigen` code being in `fuels-core`. We do not see a need to have the code generation in `fuels-core` so we have moved it to the `fuels-macros`, where we have the rest of the code generation.

Breaking change:
- `Abigen` is not public anymore - if there should come a time when `Abigen` needs to be public we can discuss alternatives like creating a new crate.

<s>P.S. GitHub bugged out on the line numbers... we did not delete 1500 lines :smile: </s>
The deleted lines come from `functions_gen.rs` which should have been deleted by the big PR but made it through. It wasn't part of the compilation, just hanging there.

Also, the tokenizer is now dead, since it was only used by the cli tool, and we still haven't decided on what to do with it.


packages/fuels-types/src/param_types.rs
```diff
@@ -310,7 +319,8 @@ fn try_vector(the_type: &Type) -> Result<Option<ParamType>, Error> {
 }
 
 fn try_enum(the_type: &Type) -> Result<Option<ParamType>, Error> {
-    let result = if has_enum_format(&the_type.type_field) {
```
iqdecay, line 313: Why weren't all those methods kept in the `fuel-abi-types` repo? Those utils are useful and avoid having to define helpers in all downstream crates.

packages/fuels-types/src/param_types.rs
```diff
@@ -292,8 +290,19 @@ fn try_struct(the_type: &Type) -> Result<Option<ParamType>, Error> {
     Ok(result)
 }
 
+fn has_struct_format(field: &str) -> bool {
+    field.starts_with("struct ")
+}
```
iqdecay, line 295: Why isn't it in the `fuels-abi-types` utils?

packages/fuels-types/src/utils.rs
```diff
@@ -1,69 +0,0 @@
-use lazy_static::lazy_static;
```
iqdecay, line 1: So IIUC this is leftover from the `fuel-abi-type` repo migration?

iqdecay (dismissed):
Lots of cleanliness improvements, I love it! Left some questions.

packages/fuels-types/src/utils.rs
hal3e: That was moved to the `fuel-abi-type` repo recently.

packages/fuels-types/src/param_types.rs
segfault-magnet: `fuel-abi-types` *

We only put what we absolutely had to in the `fuel-abi-types` . Technically we no longer have sole ownership over that project so changes we might want to do will have to be checked against all the users of the library.

Also, changes to `fuel-abi-types`, breaking or not, have to go through a release to be made available to `fuels-rs`, which causes delays.

So:
1. If we put it there, we have to be careful while changing it.
2. Nobody is using it except us, no need to share.
3. Anything put there takes time to modify.

Also, we'll probably talk to the compiler team about saving the fn selectors in the json abi. If we manage to get that through, then this will be deleted.

iqdecay: OOOH saving the fn selectors in the JSON ABI sounds super smart lol. 
I get it, but I feel like any library that would like to work over the `fuel-abi-types` would need the helpers function we use? For example someone writing another SDK? Don't we want to make this available? 
But I agree this would be a larger discussion.

iqdecay (approved):
LGTM!

digorithm (changes requested):
Excellent work with the error handling consistency here (moving from `Result<T, E>` to `Result<T>`). 

One thing, though, CI file must handle the new `fuels_macros` package; otherwise, crate publishing won't work!

digorithm (approved):
:shipit:

digorithm (approved):

PR #794, opened 2023-01-22
fix: add both chain_config and coins to node config

### LINKED ISSUE #793: `launch_custom_provider_and_get_wallets` ignores coin configuration if `ChainConfig` is `Some`
# Problem
When calling `launch_custom_provider_and_get_wallets`, if the user specifies a `ChainConfig` _without_ an explicit initial state and then passes the coin configuration as part of the `WalletsConfig`, these coins won't be added to the generated wallet.

See the example below. 

```Rust
async fn init_wallets() -> Vec<WalletUnlocked> {
    use fuel_chain_config::ChainConfig;
    use fuel_tx::ConsensusParameters;

    // Here, we define a chain config specifying only the max gas per transaction and block gas limit, nothing else.
    let chain_config = ChainConfig {
        transaction_parameters: ConsensusParameters::DEFAULT.with_max_gas_per_tx(10_000_000_000),
        block_gas_limit: 10_000_000_000,
        .. ChainConfig::default()
    };

    // But we also want 4 wallets with 3 coins of `2_000_000_000` each. 
    launch_custom_provider_and_get_wallets(
        WalletsConfig::new(
            Some(4),             /* Single wallet */
            Some(3),             /* Single coin (UTXO) */
            Some(2_000_000_000), /* Amount per coin */
        ),
        None,
        Some(chain_config),
    )
    .await
}
```

These wallets won't be properly generated because the coins are ignored if `ChainConfig` is `Some`, see the code below:

https://github.com/FuelLabs/fuels-rs/blob/ef71ed6879d3ec5d38630b53acfd28b28ecaa1e6/packages/fuels-test-helpers/src/lib.rs#L155-L184

What needs to be done is, if the user specifies a `ChainConfig` _and_ a configuration for the coins in their test wallets, _both_ things need to happen.

Salka1988:
This PR fixes a bug in the launch custom provider and_get_wallets function where the coin_configs and messages were being omitted from the fuel node config file. The fix involves adding the coin_configs and messages to the initial_state of the chain_config variable, as seen in the updated code:

```rust
    let mut chain_config = chain_config.unwrap_or_else(|| ChainConfig::local_testnet());

    chain_config.initial_state = Some(StateConfig {
        coins: Some(coin_configs),
        contracts: None,
        messages: Some(messages),
        height: None,
    });

    if let Some(transaction_parameters) = consensus_parameters_config {
        chain_config.transaction_parameters = transaction_parameters;
    }
```

closes #793

### DIFF
--- a/packages/fuels-test-helpers/src/lib.rs
+++ b/packages/fuels-test-helpers/src/lib.rs
@@ -12,7 +12,7 @@ use fuel_chain_config::StateConfig;
 #[cfg(feature = "fuel-core-lib")]
 pub use fuel_core::service::Config;
 #[cfg(feature = "fuel-core-lib")]
-use fuel_core::service::{DbType, FuelService};
+use fuel_core::service::FuelService;
 use fuel_gql_client::client::FuelClient;
 use fuel_tx::{Bytes32, ConsensusParameters, UtxoId};
 use fuels_core::constants::BASE_ASSET_ID;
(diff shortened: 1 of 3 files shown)


iqdecay (approved):
LGTM! :ship: it

digorithm (commented):
Late to the party, but thanks for addressing this so quickly!

PR #795, opened 2023-01-23
refactor: implement shared behaviors between program calls

iqdecay:
This PR uses a macro to implement shared behaviors between `ScriptCall`s and `ContractCall`s.
This is dependent on #781.


packages/fuels-programs/src/calls/call.rs
```diff
@@ -0,0 +1,100 @@
...
+    ($target:ty) => {
+        impl ProgramCall for $target {
+            fn with_external_contracts(self, external_contracts: &[Bech32ContractId]) -> Self {
+                Self {
+                    external_contracts: external_contracts.to_vec(),
+                    ..self
+                }
+            }
+
+            fn with_call_parameters(self, call_parameters: CallParameters) -> Self {
```
MujkicA, line 48: `CallParameters` don't make much sense in the context of ScriptCalls. There is no way to convert them to the right script data so that they are forwarded to the `CALL` instruction.

iqdecay: Thanks! I always get confused around `CallParameters`.

packages/fuels-programs/src/calls/call.rs
```diff
@@ -0,0 +1,100 @@
...
+                Self {
+                    call_parameters,
+                    ..self
+                }
+            }
+
+            fn with_variable_outputs(self, variable_outputs: &[Output]) -> Self {
```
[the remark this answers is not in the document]
iqdecay: You're right, using `append_variable_output` should be enough. It's a YAGNI.

packages/fuels-programs/src/calls/call.rs
```diff
@@ -0,0 +1,76 @@
...
+use crate::{calls::contract_call::ContractCall, calls::script_call::ScriptCall};
+
+// Trait implemented by contract instances so that
+// they can be passed to the `set_contracts` method
+pub trait SettableContract {
+    fn id(&self) -> Bech32ContractId;
+    fn log_decoder(&self) -> LogDecoder;
+}
+
+pub(crate) trait ProgramCall {
```
segfault-magnet, line 18: I love to see deduplication any day -- but :D 

I'm not seeing that we're using this new trait `ProgramCall` anywhere. Besides not having a use for it rn, it seems forced/unfortunately named. A ProgramCall which you cannot call -- the methods inside revolve mostly around transaction inputs/outputs. 

The dedupe effort can still be used without the trait by just calling the macro to produce the code. You're still in the same crate and you can have multiple impl blocks for a struct:
```rust
struct SomeStruct;
impl SomeStruct { 
    fn foo() {
        
    }
}
impl SomeStruct {
    fn goo() {
        
    }
}
```

also, you can then leave the macro here and call it below the original impl blocks in both call handlers, to keep the impls close.

packages/fuels-programs/src/calls/contract_call_utils.rs
```diff
@@ -589,57 +590,6 @@ mod test {
         assert_eq!(expected_inputs, inputs_as_signed_coins);
     }
 
-    #[test]
-    fn variable_outputs_appended_to_outputs() {
```
segfault-magnet, line 593: Are these two tests no longer needed? 

Couldn't find them elsewhere.

segfault-magnet (commented):
Reviewed alongside @hal3e.

packages/fuels-programs/src/calls/call.rs
```diff
@@ -0,0 +1,76 @@
...
+    fn with_external_contracts(self, external_contracts: &[Bech32ContractId]) -> Self;
```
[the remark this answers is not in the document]
iqdecay: Didn't know that, thanks!

packages/fuels-programs/src/calls/call.rs
iqdecay: That makes a lot of sense, thanks a lot. The trait lead me to having a macro and now I can remove the trait haha.

digorithm (commented):
Two notes:

1. It sounds like the PR's title and description are outdated; There's no `ProgramCall` trait introduced. There's no trait at all.

2. I'm not necessarily against this approach at all, and I recognize the initial problem this PR solves: a set of behavior were being implemented _twice_ by two structs -- not good. However, I've been having a hard time justifying using a _macro_ for this. Macros are inherently harder to maintain and read. Sometimes, only macros can properly solve a problem. Is this the case here? It seems we want a set of behaviors shared among two structs. Would a good old composition [0] strategy solve this problem without introducing macros? It feels like we're trying to force the usage of macros here. Again, not against this approach, but from an outsider's perspective (which I have right now, as I didn't implement this), it looks like a simpler approach could work.

_[0]: a base struct `Base` that implements the set of behaviors would be a component of `ScriptCall` and `ContractCall`, etc._

iqdecay:
I agree with your point about the macro not being necessary. My design process was: trying to use compositon, realizing it would be hard, implementing a trait, implementing the macro, removing the trait. Now I'll try composition again, I didn't think of making it a component of `ContractCall` directly, I was trying to compose inside of `ContractCallHandler`.

PR #796, opened 2023-01-25
refactor: fuels-types Error

hal3e:
Because of the refactoring we have done in the SDK, some error types defined in `fuels_types::Error` are not needed anymore. In this PR I have tried to remove the unnecessary error variants and I have introducing an `error!` macro (thanks @segfault-magnet for the macro implementation)

More specifically, I have:
- removed `anyhow` completely from the SDK.
  - We had a `From` implementation that turned any `anyhow` error to `ParseTokenStreamError` which was wrong.
- removed the `CodecError`
- removed Error variants:
  - `ParseIntError`
  - `ParseBoolError`
  - `ParseHexError`
  - `ParseTokenStreamError`
  - `CompilationError`
- added `Result` type alias `Result<T> = std::result::Result<T, Error>` 

Added the `error!` macro. This macro makes it easier to write the error manually. Prevously we had to do something like this:
```rust
Err(Error::InstantiationError(format!(
       "Expected `bool`, got {:?}",
       other
))),
```
now we can write:
```rust
Err(error!(InstantiationError,
       "Expected `bool`, got {:?}",
       other
)),
```
In addition, we can pass a string literal to the macro i.e. we do not have to use `.into()` or `to_string()` to turn it into a `String`.
This macro can only be used with errors that have a `String` field. Those are the errors that we manually make.
IMO the macro makes the error instantiation cleaner. If the team does not like it we can remove it :smile:


segfault-magnet (commented):
I like it. 

If we're gonna make me the muse, might as well introduce the Result type alias so that we don't write `Result<T, Error>` all the time :D 

Follow the footsteps of countless libs before us

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -127,30 +116,31 @@ impl ABIDecoder {
...
         }
         let u64_length = bytes.len() / u64_size;
         let (tokens, bytes_read) = Self::decode_multiple(&vec![ParamType::U64; u64_length], bytes)?;
         let elements = tokens
             .into_iter()
             .map(u64::from_token)
-            .collect::<Result<Vec<u64>, Error>>()
-            .map_err(|e| CodecError::InvalidData(e.to_string()))?;
+            .collect::<Result<Vec<u64>>>()
+            .map_err(|e| Error::InvalidData(e.to_string()))?;
```
iqdecay, line 135: Use the brand new `error!` macro here?

packages/fuels/tests/types.rs
```diff
@@ -1,3 +1,4 @@
+use std::result::Result as StdResult;
```
iqdecay, line 1: I don't understand why we don't use just the `Result` everywhere?

packages/fuels/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub mod prelude {
         tx::Salt,
         types::{
             bech32::{Bech32Address, Bech32ContractId},
-            errors::Error,
+            errors::{Error, Result},
```
iqdecay, line 78: This is suprising to me. Are we defining our own `Result` type?

iqdecay (changes requested):
So I really like the `error!` macro and general cleanup, but I don't think the new `Result` type is a good idea. We're changing the semantics of what a `Result` is, and I think it would be confusing for downstream projects.

packages/fuels/src/lib.rs
digorithm: @iqdecay Are you talking about this https://github.com/FuelLabs/fuels-rs/pull/796/files#diff-f9b98bdf9d483a3e5875ff6df383880efd6759d37e2c24cbab51d56fdc8c59f2R32?

If so, this is a fairly common and encouraged Rust pattern: https://doc.rust-lang.org/rust-by-example/error/result/result_alias.html.

MujkicA (dismissed):
LGTM

packages/fuels/src/lib.rs
iqdecay: Oh I didn't know that! Then if it's recommended it will not confuse people 😊

iqdecay (approved):
LGTM!

packages/fuels-test-helpers/src/utils.rs
```diff
@@ -128,8 +135,8 @@ mod tests {
         }
 
         #[tokio::test]
-        async fn returns_value_on_success() -> anyhow::Result<()> {
-            let successfully_generates_value = || async { Ok(12345u64) as anyhow::Result<u64> };
+        async fn returns_value_on_success() -> Result<()> {
+            let successfully_generates_value = || async { Ok(12345u64) as Result<u64> };
 
             let value = retry(
```
segfault-magnet, line 141: We might as well remove the `retry` helper. 

Made sense when we didn't have the client API for tweaking the block height, now I'm not sure anybody uses it.

The only place it is 'used' is `confirm_blocks_created` which probably nobody uses and probably shouldn't use since we now have a reliable API to tweak the block height.

iqdecay: Agreed

digorithm (approved):
Much better! Thanks for the clean-up 😁

digorithm (approved):

PR #800, opened 2023-01-25
Methods without #[payable] cannot accept assets

MujkicA:
Closes https://github.com/FuelLabs/fuels-rs/issues/742

.call_params() now returns a Result which either resolves to the ContractCallHandler instance or an InvalidCallParameters error. The latter happens when the user tries to set call parameters with an amount > 0 on a contract method that isn't payable.

### DIFF
--- a/packages/fuels-programs/src/contract.rs
+++ b/packages/fuels-programs/src/contract.rs
@@ -110,6 +110,7 @@ impl Contract {
         signature: Selector,
         args: &[Token],
         log_decoder: LogDecoder,
+        is_payable: bool,
     ) -> Result<ContractCallHandler<D>, Error> {
         let encoded_selector = signature;
 
@@ -129,6 +130,7 @@ impl Contract {
             message_outputs: None,
             external_contracts: vec![],
             output_param: D::param_type(),
+            is_payable,
         };
 
         Ok(ContractCallHandler {
@@ -360,6 +362,7 @@ pub struct ContractCall {
     pub message_outputs: Option<Vec<Output>>,
     pub external_contracts: Vec<Bech32ContractId>,
     pub output_param: ParamType,
+    pub is_payable: bool,
 }
 
 impl ContractCall {
@@ -560,6 +563,10 @@ where
         self
     }
 
+    pub fn is_payable(&self) -> bool {
+        self.contract_call.is_payable
+    }
+
     /// Sets the transaction parameters for a given transaction.
     /// Note that this is a builder method, i.e. use it as a chain:
 
@@ -579,9 +586,12 @@ where
     /// let params = CallParameters { amount: 1, asset_id: BASE_ASSET_ID };
     /// my_contract_instance.my_method(...).call_params(params).call()
     /// ```
-    pub fn call_params(mut self, params: CallParameters) -> Self {
+    pub fn call_params(mut self, params: CallParameters) -> Result<Self, Error> {
+        if !self.is_payable() && params.amount > 0 {
+            return Err(Error::InvalidCallParameters());
+        }
         self.contract_call.call_parameters = params;
-        self
+        Ok(self)
     }
 
     /// Appends `num` [`Output::Variable`]s to the transaction.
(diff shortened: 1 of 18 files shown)


MujkicA:
Same changes as https://github.com/FuelLabs/fuels-rs/pull/770 minus the merge conflicts.

hal3e (approved):
Nice and clean. LGTM

Salka1988 (dismissed):
:+1:

packages/fuels-types/src/errors.rs
```diff
@@ -56,6 +56,8 @@ pub enum Error {
     ValidationError(#[from] CheckError),
     #[error("Revert transaction error: {}, receipts: {:?}", .0, .1)]
     RevertTransactionError(String, Vec<Receipt>),
+    #[error("Tried to forward assets to a contract method that is not payable.")]
+    InvalidCallParameters(),
```
iqdecay, line 60: `InvalidCallParameters` would indicate a validation issue for the call parameters, ie that the provided call parameters don't conform to the spec. This is not the case here. I think a better name could be `ForwardAssetToNonPayable` or something.

iqdecay (changes requested):
Left a naming nit, this is really good! Please indicate that the changes are breaking and how to fix those breaking changes in the PR description (useful for release notes).

packages/fuels-types/src/errors.rs
segfault-magnet: Or `AssetsForwardedToNonPayableMethod`. 

Since the asset amount forwarding happens in the `CallParameters` they are kinda invalid. But, like @iqdecay , I'd make it more specific until we have at least one more reason why `CallParameters` might fail.

Br1ght0ne (approved):

packages/fuels-types/src/errors.rs
```diff
@@ -28,7 +28,7 @@ pub enum Error {
     #[error("Revert transaction error: {}, receipts: {:?}", .0, .1)]
     RevertTransactionError(String, Vec<Receipt>),
     #[error("Tried to forward assets to a contract method that is not payable.")]
-    InvalidCallParameters(),
+    AssetsForwardedToNonPayableMethod(),
```
iqdecay, line 31: Nice

iqdecay (dismissed):
Perfect, :ship: it with a nice squash commit message

packages/fuels-macros/src/abigen/code_gen/abi_types.rs
```diff
@@ -118,7 +115,10 @@ impl FullABIFunction {
             abi_function.name.clone(),
             inputs,
             FullTypeApplication::from_counterpart(&abi_function.output, types),
-            abi_function.attributes.clone(),
+            abi_function
+                .attributes
+                .as_ref()
+                .map_or(vec![], Clone::clone),
```
segfault-magnet, line 121: It's getting crowded, I'd extract the attributes into a separate variable:
```rust
        let attributes = abi_function.attributes.iter().flatten().cloned().collect();
        FullABIFunction::new(
            abi_function.name.clone(),
            inputs,
            FullTypeApplication::from_counterpart(&abi_function.output, types),
            attributes,
        )
```

Also the flatten->cloned->collect is another approach. I'm fine with both.

packages/fuels-programs/src/contract.rs
```diff
@@ -615,9 +617,9 @@ where
     /// let params = CallParameters { amount: 1, asset_id: BASE_ASSET_ID };
     /// my_contract_instance.my_method(...).call_params(params).call()
     /// ```
-    pub fn call_params(mut self, params: CallParameters) -> Result<Self, Error> {
+    pub fn call_params(mut self, params: CallParameters) -> Result<Self> {
         if !self.is_payable() && params.amount > 0 {
-            return Err(Error::InvalidCallParameters());
+            return Err(Error::AssetsForwardedToNonPayableMethod());
```
segfault-magnet, line 622: I'd declare this error as `Error::AssetsForwardedToNonPayableMethod` to avoid having to append parentheses whenever you're creating it.

When you add the parentheses the variant type is no longer Unit, but an empty tuple.

Also you then have to add the parentheses when matching as well: 
```rust
        let some_error = Error::AssetsForwardedToNonPayableMethod();
        match some_error {
            Error::AssetsForwardedToNonPayableMethod() => {}
            _ => {}
        }
```

hal3e (approved):
LGTM

digorithm (approved):
Awesome work -- excited to get this out 😍

PR #801, opened 2023-01-25
Improve devx for extracting data from Script

### LINKED ISSUE #765: Improve devx for extracting data from `fuel_tx::transaction::types::Script`
Getting the innards of the `Script` type is a bit of a hassle.

Its fields are `pub(crate)` so accessing them directly isn't possible. 

There are getters for the data contained within but they are visible only if you import the appropriate traits. So to get the script data out of a `Script` our users would need to do something like this:
```rust
    use ::fuels::core::tx::field::ScriptData;
    let tx: ::fuels::core::tx::Script = foo_caller_contract_instance
        .methods()
        .call_foo_contract(Bits256(bits), true)
        .get_executable_call()
        .await
        .unwrap()
        .tx;
    tx.script_data();
```

ATTOW we have the `Script` type in the following public APIs:
* [field](https://github.com/FuelLabs/fuels-rs/blob/1f72d26189a83306bcbaa357cd6b24c9cf5efd1f/packages/fuels-contract/src/execution_script.rs#L26) of the `ExecutableFuelCall`
* `build_transfer_tx`[ returns it.](https://github.com/FuelLabs/fuels-rs/blob/1f72d26189a83306bcbaa357cd6b24c9cf5efd1f/packages/fuels-signers/src/wallet.rs#L268)
* `build_contract_transfer_tx` also [returns it](https://github.com/FuelLabs/fuels-rs/blob/1f72d26189a83306bcbaa357cd6b24c9cf5efd1f/packages/fuels-signers/src/wallet.rs#L283).
* `build_message_to_output_tx` also[ returns it](https://github.com/FuelLabs/fuels-rs/blob/1f72d26189a83306bcbaa357cd6b24c9cf5efd1f/packages/fuels-signers/src/wallet.rs#L330).

Three possible approaches to improving devx are:
1. Create our own Script type, and make it convertible back-n-forth with the original as need be. Ours will have all of its fields public.
**Pros:** 
  Solves the problem throughout our codebase, don't need to think about it anymore.
**Cons:**
  Takes a bit of work to implement. 
2. Add delegating getters to `ExecutableFuelCall` whose implementation will make use of the appropriate traits in order to access the data from `Script`. 
**Pros:**
  Quick
**Cons:**
  If `Script`s start appearing elsewhere we might be right back on square one.
3. Add all the necessary traits to the fuels prelude.
**Pros:**
  Quickest
**Cons:**
  Not sure. Maybe bloat? I'd rather extract the tumor rather than appease it.

MujkicA:
Closes #765, #807

This PR introduces `ScriptTransaction` and `CreateTransaction`  which act as a wrappers for the script and create tx type. They both implement `Transaction` which guarantees that the type can be transformed to a `fuel-tx::Transaction` and be consumed by the client.

The wrappers include getters and builder-like setters to provide a better devx when creating custom scripts or reading the underlying tx from a contract call.

Building a tx:
```rust
let mut tx = ScriptTransaction::new(&inputs, &outputs, params)
                .with_script(binary)
                .with_script_data(data);
                
// or to extend inputs/outputs
tx.inputs_mut().extend(other_inputs);
tx.outputs_mut().extend(other_outputs);
```

Reading tx fields from a contract/script call:
```rust
let mut tx = call_handler.build_tx().await?;
let data = tx.script_data();
let id = tx.id();
```

Some other side-effects of this change:

- removed `ExecutionScript` since this abstraction doesn't provide any functionality at this point.
- constants and params were moved from fuels-core to fuels-types to avoid circular dep
- moved tx validity check to `provider.send_transaction()`


hal3e (approved):

hal3e (changes requested):
Sorry. I did approve but still have to review the PR.

MujkicA:
> Not sure what to do. Not a fan of having the ScriptTransaction stop short of the provider.
> 
> I also tried to see why we couldn't integrate it into the provider. Then the mess started.
> 
> It seemed weird to me that `FuelClient` had these super simple data types as arguments for its methods, but then our provider imposed a bunch of weird foreign bounds on its args. Stuff like: `ExecutableTransaction`, `Chargable` etc.
> 
> Then I realized that is because we're using various helpers from other fuel crates, and these traits originate from them (e.g. `ExecutableTransaction` comes from `fuel-vm`).
> 
> `ExecutableTransaction` is a particularly nasty one to implement if we want a wrapper. It needs like 200 lines just to delegate everything to `Script`.
> 
> Then I got to thinking about whether we really need all these bounds. `Chargable`, `Executable`, `IntoChecked`... etc. Turns out we only need a subset in order to use the helpers and our bounds are too wide making the helper boilerplated.
> 
> I'm for one of two options:
> 
> 1. Bite the bullet, write wrappers and have them be used everywhere in the SDK. This means the provider as well -- figure out the minimum of what we really need to use the fuel helpers and implement only that for the wrappers.
> 2. Add the necessary imports in the prelude and ignore this issue for now.

I dislike option 1. The cost of maintenance might not be worth it. If we end up needing more and more helpers, this doesn't scale well.
Option 2 seems good enough, unless there is some downside to it I'm not seeing.

MujkicA:
> > Not sure what to do. Not a fan of having the ScriptTransaction stop short of the provider.
> > I also tried to see why we couldn't integrate it into the provider. Then the mess started.
> > It seemed weird to me that `FuelClient` had these super simple data types as arguments for its methods, but then our provider imposed a bunch of weird foreign bounds on its args. Stuff like: `ExecutableTransaction`, `Chargable` etc.
> > Then I realized that is because we're using various helpers from other fuel crates, and these traits originate from them (e.g. `ExecutableTransaction` comes from `fuel-vm`).
> > `ExecutableTransaction` is a particularly nasty one to implement if we want a wrapper. It needs like 200 lines just to delegate everything to `Script`.
> > Then I got to thinking about whether we really need all these bounds. `Chargable`, `Executable`, `IntoChecked`... etc. Turns out we only need a subset in order to use the helpers and our bounds are too wide making the helper boilerplated.
> > I'm for one of two options:
> > 
> > 1. Bite the bullet, write wrappers and have them be used everywhere in the SDK. This means the provider as well -- figure out the minimum of what we really need to use the fuel helpers and implement only that for the wrappers.
> > 2. Add the necessary imports in the prelude and ignore this issue for now.
> 
> I dislike option 1. The cost of maintenance might not be worth it. If we end up needing more and more helpers, this doesn't scale well. Option 2 seems good enough, unless there is some downside to it I'm not seeing.

I ended up finding a way to propagate the wrapper trough our API without re-implementing anything too heavy from `fuel-tx`.  I will move the implementation of  #807 to this branch and re-open the PR.

MujkicA:
Updated the PR description

packages/fuels-types/src/transaction.rs
```diff
@@ -0,0 +1,314 @@
...
+            params.maturity,
+            script,
+            script_data,
+            inputs.to_vec(),
+            outputs.to_vec(),
+            vec![],
+        )
+        .into()
+    }
+}
```
iqdecay, line 314: This is great, but I think the usage of a macro warrants at least 1 test to show that the implementation behaves the way we expect it to. This way, if we modify it, we can ensure we're not breaking anything.

iqdecay (changes requested):
This is great stuff! Missing a test IMO, and as usual with me some naming nits.

packages/fuels-programs/src/call_utils.rs
```diff
@@ -334,15 +395,46 @@ fn extract_unique_contract_ids(calls: &[ContractCall]) -> HashSet<ContractId> {
...
+    provider: &Provider,
+    tx: &T,
+) -> Result<Vec<Receipt>> {
+    let receipts = provider.dry_run(tx).await?;
+    validate_script_succedded(&receipts)?;
+
+    Ok(receipts)
+}
+
+fn validate_script_succedded(receipts: &[Receipt]) -> Result<()> {
```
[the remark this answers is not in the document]
digorithm: Since this will commonly be used with the `?` operator, `has_script_succeeded)?;` sounds the most natural to me.

packages/fuels-types/src/transaction.rs
digorithm: @iqdecay Are you talking about the `impl_tx_wrapper` macro? If so, in this case, if the macro implementation fails, it fails to comply with the implementation of the trait `Transaction`, failing right away at compile-time.

iqdecay: I was thinking about the inside behavior, the same way we have unit tests for regular method implementation. Or are the methods here too simple?

hal3e (approved):
This is awesome! Great job!

iqdecay (approved):
LGTM

digorithm (dismissed):
Awesome work, @MujkicA!

Salka1988 (approved):
:+1:

segfault-magnet (commented):
Reviewed alongside @hal3e 
The suggestions made into a PR, if you like them:
https://github.com/FuelLabs/fuels-rs/pull/847/files

digorithm (approved):

PR #804, opened 2023-01-26
feat!: `derive`able `Parameterize` and `Tokenizable`

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/785 - co-developed with @segfault-magnet 

If all your `fields`/`variants` of your custom `struct`/`enum` implement `Parameterizable` and `Tokenizable` you can derive the two traits using the new macros. Here is an example: 

```rust
#[derive(Parameterize, Tokenizable)]
struct MyStruct2<T: Tokenizable + Parameterize> {
    a: T,
}

#[derive(Parameterize, Tokenizable)]
pub struct MyStruct<T: Tokenizable + Parameterize> {
    field1: u64,
    field3: Address,
    field2: Option<MyStruct2<T>>,
}
```

If you do not have `fuels` as a dependency you will have to specify the full path to the `fuels_types` crate. Here is an example of how we used the derive macro inside `fuels_types`:
```rust
#[derive(Debug, PartialEq, Eq, Copy, Clone, Parameterize, Tokenizable)]
#[FuelsTypesPath("crate")]
pub struct B512 {
    pub bytes: [Bits256; 2],
}
```

Breaking changes:
- Moved `Parameterize` and `Tokenizable` traits from `fuels_core::traits` to `fuels_types::traits`
- Unit `enum` variants no longer have parentheses i.e.  before `SomeEnum::A()` now `SomeEnum::A`
- `EvmAddress` - field `value` is not public anymore. We derived both `Parametrize` and `Tokenizable` using the derive macro. The method `from_token` will not clear the first  12 bytes anymore. The reasoning is that: Tokens coming from Sway will always have the first 12 bytes cleared. Tokens from a manually created `EvmAddress` will always have the first twelve bytes cleared.


packages/fuels-macros/src/abigen/code_gen/custom_types.rs
```diff
@@ -0,0 +1,741 @@
+use std::collections::HashSet;
```
segfault-magnet, line 1: This file was moved, but GitHub didn't pick up on that. 

You'll see it later on in red.

MujkicA (dismissed):

Salka1988 (dismissed):

packages/fuels-macros/src/lib.rs
```diff
@@ -76,3 +81,30 @@ pub fn setup_contract_test(input: TokenStream) -> TokenStream {
...
+}
+
+#[proc_macro_derive(TryFrom)]
+pub fn try_from(stream: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(stream as DeriveInput);
+
+    generate_try_from_impl(input)
+        .unwrap_or_else(|e| e.to_compile_error())
+        .into()
+}
```
digorithm, line 110: This whole thing is a work of art. Excellent work breaking down the layers in a very readable/maintainable way!

packages/fuels-macros/src/parse_utils.rs
```diff
@@ -80,3 +81,118 @@ where
...
+            panic!("This cannot happen in valid Rust code. Fields::Unit only appears in enums")
+        }
+    }?;
+
+    let (names, types) = named_fields
+        .into_iter()
+        .map(|field| {
+            let name = field
+                .ident
+                .expect("FieldsNamed to only contain named fields!.");
```
digorithm, line 147:
```suggestion
                .expect("FieldsNamed to only contain named fields.");
```

Sorry, I wish my eyes couldn't catch such small typos 😂

packages/fuels-macros/src/parse_utils.rs
```diff
@@ -80,3 +81,118 @@ where
...
+                        "Must have exactly one element",
+                    ));
+                }
+                fields_unnamed.unnamed.into_iter().next()
+            }
+            Fields::Unit => None,
+            Fields::Named(named_fields) => {
+                return Err(Error::new_spanned(
+                    named_fields,
+                    "Struct-like enum variants are not supported!",
```
digorithm, line 181:
```suggestion
                    "Struct-like enum variants are not supported.",
```

Incredibly minor: let's try and consistently stick with period (`.`) instead of exclamation marks (`!`) for error messages 😁

digorithm (changes requested):
I love this so much 😍 

Great work @hal3e @segfault-magnet 

Requesting changes just for the tiny typos. Feel free to merge after!

packages/fuels-macros/src/lib.rs
hal3e: @segfault-magnet's refactors after the initial implementation are always amazing!

digorithm (approved):

PR #805, opened 2023-01-26
Reorganize navigation, add testing basics page

sarahschwartz:
- Add README FAQ instructions for how to run the docs locally
- Hide blank debugging pages from nav
- Reorganize navigation so the "Getting Started" section is not bloated
- Add a page for testing basics


README.md
```diff
@@ -36,3 +36,22 @@ See [the `fuels-rs` book](https://fuellabs.github.io/fuels-rs/latest/)
 
 In order to make the SDK for Fuel feel familiar with those coming from the [ethers.js](https://github.com/ethers-io/ethers.js) ecosystem, this project opted for an `s` at the end. The `fuels-*` family of SDKs is inspired by The Ethers Project.
 
+### How can I run the docs locally?
+
+Install mdbook by running:
```
hal3e, line 41:
```suggestion
Install `mdbook` by running:
```

docs/src/testing/basics.md
```diff
@@ -0,0 +1,80 @@
+# Testing Basics
+
+If you're new to Rust, you'll want to review these important tools to help you build tests.
+
+## The `assert!` macro
+
+You can use the `assert!` macro to enforce certain outcomes in your test. This macro throws an error and fails the test if the expression inside evaluates to `false`.
```
hal3e, line 7: `assert!()` will not throw an error but invoke `panic!()`.
```suggestion
You can use the `assert!` macro to enforce certain outcomes in your test. This macro invokes `panic!()` and fails the test if the expression inside evaluates to `false`.
```

docs/src/testing/basics.md
```diff
@@ -0,0 +1,80 @@
...
+
+```rust, ignore
+assert!(value == 5);
+```
+
+## The `assert_eq!` macro
+
+The `assert_eq!` macro works a lot like the `assert` macro, however instead it accepts two values, and throws and error if those values are not equal.
```
hal3e, line 15: same as above

docs/src/testing/basics.md
```diff
@@ -0,0 +1,80 @@
...
+
+```rust, ignore
+assert_eq!(balance, 100);
+```
+
+## The `assert_ne!` macro
+
+The `assert_ne!` macro works just like the `assert_eq!` macro, but it will throw an error if the two values are equal.
```
hal3e, line 23: same as above

docs/src/SUMMARY.md
```diff
@@ -5,67 +5,68 @@
...
-    - [B512](./types/B512.md)
-    - [EvmAddress](./types/evm_address.md)
-    - [Vectors](./types/vectors.md)
-- [API](./getting-started/api.md)
-- [Debugging](./debugging/debugging.md)
-  - [Debugging abigen errors](./debugging/abigen.md)
-  - [Debugging contract call errors](./debugging/contract-calls.md)
-  - [Debugging network errors](./debugging/network.md)
-  - [The Function selector](./debugging/function-selector.md)
+- [Connecting to a Fuel node](./getting-started/connecting.md)
```
hal3e, line 8: Awesome! Now that I see it, I can not understand how nobody noticed it before :smile:

hal3e (commented):
Thanks for the contribution! I left some nits

docs/src/testing/basics.md
segfault-magnet: I'm not sure how using `assert` _enforces_ outcomes, rather it _asserts_ (states or declares) that something should be true. Otherwise, the test will panic.

docs/src/testing/basics.md
```diff
@@ -0,0 +1,80 @@
...
+
+```rust, ignore
+assert_ne!(address, 0);
+```
+
+## The `println!` macro
+
+You can use the `println!` macro to print values to the console.
```
segfault-magnet, line 31: Technically the output goes to `stdout` which may or may not be a tty. 

But I'm not sure if being technically accurate serves any purpose here besides guarding against bureaucrats :D.

docs/src/testing/basics.md
```diff
@@ -0,0 +1,80 @@
...
+Using `{}` will print the value, and using `{:?}` will print the value plus its type.
+
+Using `{:?}` will also allow you to print values that do not have the `Display` trait implemented but do have the `Debug` trait. Alternatively you can use the `dbg!` macro to print these types of variables.
+
+```rust, ignore
+println!("WALLET 1 PROVIDER {:?}", wallet_1.get_provider().unwrap());
+dbg!("WALLET 1 PROVIDER {}", wallet_1.get_provider().unwrap());
+```
+
+To print more complex types that don't have it already, you can implement your own formatted display method with the `fmt` library from the Rust standard library.
```
segfault-magnet, line 47: `fmt` isn't a library in its own right. It's a _module_ inside the rust std lib.

docs/src/testing/basics.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Run Commands
+
+You can run your tests to see if they pass or fail with
+
+```
+cargo test
+```
+
+If you want to see anything printed to the console from your tests, use the `nocapture` flag.
```
segfault-magnet, line 76: Technically the output is only hidden if the test succeeds. 

If it failed then the output will be visible.

segfault-magnet (commented):
image

README.md
```diff
@@ -36,3 +36,22 @@ See [the `fuels-rs` book](https://fuellabs.github.io/fuels-rs/latest/)
...
+```shell
+cargo install mdbook
+```
+
+Next, navigate to the `docs` folder and run the command below to start a local server.
+
+```shell
+mdbook serve
+```
+
```
MujkicA, line 52: Maybe I'm paranoid, but for me this doesn't open a browser tab automatically and I fear there is a group of users that could get confused.

```suggestion

To view the docs, navigate to the localhost address outputted by `mdbook`.

```

MujkicA:
LGTM!
I would wait for @digorithm to give it one final look before merging.

digorithm (approved):
Thank you so much @sarahschwartz!

Issue #799, opened 2023-01-25
Feature request: Add support for the new `assert_eq`


nfurfaro:
I'm adding a new function in `std::assert` called `assert_eq` which will male use of a new error code found [here](https://github.com/FuelLabs/sway/blob/e7a74665615123cd3276a1e9905a4fc7fe1d4040/sway-lib-std/src/error_signals.sw#L18).

The SDK support needed will consist of the ability to:

- recognize the revert error (the same way the SDK handles the error for a failed `require`
- log the values of the 2 previous `log` or `logd` receipts in a presentable way, perhaps following the rus example of "lhs != rhs"

hal3e:
Is there a `sway` PR that we can track here ?

nfurfaro:
@hal3e there will be shortly; I will drop a link here as soon as I open it. Just working on the tests.

hal3e:
Blocked until https://github.com/FuelLabs/sway/issues/3946 is fixed.

PR #811, opened 2023-01-31
feat: add support for assert_eq logs

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/799

The SDK now supports decoding logs from an `assert_eq` revert. In Sway, `assert_eq(a,b)` will log both `a` and `b` and then revert with a specific `revert_id`. Previously, we only had logs from `require` reverts but now we have to check the `revert_id` and see what kind of revert it is. 

I have:
- Turned `RevertTransactionError` to a struct that now also hold the specific `revert_id`. 
- Refactored error generation in `ExecutableFuelCall` when using `simulate`. Now we will get a reason if the simulation fails.
- Refactored the log decoding code and separated it into two cases: `require` and `assert_eq`.
- Added `LogId` tuple struct
- Moved `log_type_lookup` to `fuels_programs::logs` from `fuels_core::utils`
- Refactored `TransactionStatus::Failure` handling in `provider.rs` so that we can get the specific `revert_id`
- Added `assert_eq` tests for both `contract` and `script`

>***Note:*** the PR is blocked by: https://github.com/FuelLabs/sway/issues/3946 and a new `forc` release. However, I have manually created logs with the appropriate revert so that we can work on the SDK side while they fix the bug. Hopefully when the bug is fixed,  I will just have to change the sway code.


packages/fuels/tests/contracts/assert_eq/Forc.toml
```diff
@@ -0,0 +1,8 @@
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "assert_eq"
+
+[dependencies]
+std = { path = "../../../../../../sway/sway-lib-std" }
```
segfault-magnet, line 8: This might be left over from debugging. Or are you waiting on a release? Not sure if I saw the CI pass :D

hal3e: Yes :) We need a new release

packages/fuels-programs/src/logs.rs
```diff
@@ -87,14 +69,75 @@ impl LogDecoder {
     }
 }
 
+trait FilterMapLogReceipts {
+    type Output: Iterator<Item = ((Bech32ContractId, u64), Vec<u8>)>;
+    fn filter_map_log_receipts(self) -> Self::Output;
+}
+
+impl<'a, I: Iterator<Item = &'a Receipt>> FilterMapLogReceipts for I {
+    type Output = FilterMap<Self, fn(&Receipt) -> Option<((Bech32ContractId, u64), Vec<u8>)>>;
```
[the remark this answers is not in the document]
hal3e: I do not know if that would help us that much... We have `((Bech32ContractId,u64), Vec<u8>)` only here when we filter the receipts to get the `log_id` and `data`.  How about we make `(Bech32ContractId,u64)` a tuple struct - `LogId(Bech32ContractId,u64)` and we use that instead.. IMO this and the your renaming proposal would make things clearer.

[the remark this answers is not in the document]
segfault-magnet: A fair compromise. Whatever gets us towards naming stuff :D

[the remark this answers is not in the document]
hal3e: btw we have `DecodableLog` trait in fuels_core for `ParamType` :D

packages/fuels-signers/src/provider.rs
```diff
@@ -136,12 +137,33 @@ impl Provider {
...
+            program_state,
+            ..
+        } = status
+        {
+            let revert_id = program_state
+                .and_then(|state| match state {
+                    ProgramState::Revert(revert_id) => Some(revert_id),
+                    _ => None,
+                })
+                .unwrap_or(0);
```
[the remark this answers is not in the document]
hal3e: The test do not have a case of `TransactionStatus::Failure` without `revert_id` so let's use `expect` for now.

packages/fuels-programs/src/logs.rs
```diff
@@ -87,14 +69,75 @@ impl LogDecoder {
     }
 }
 
+trait FilterMapLogReceipts {
+    type Output: Iterator<Item = ((Bech32ContractId, u64), Vec<u8>)>;
+    fn filter_map_log_receipts(self) -> Self::Output;
```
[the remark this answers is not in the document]
hal3e: As mentioned above, we  use `DecodableLog` elsewhere. We could use the same name here but I would avoid it because of possible confusion.

How about something like this:
```rust
trait ExtractLogIdData {
    type Output: Iterator<Item = (LogId, Vec<u8>)>;
    fn extract_log_id_with_data(self) -> Self::Output;
}
```

packages/fuels-programs/src/logs.rs
```diff
@@ -16,22 +15,26 @@ use fuels_types::{
...
 }
 
 impl LogDecoder {
     /// Get all decoded logs from the given receipts as `String`
     pub fn get_logs(&self, receipts: &[Receipt]) -> Result<Vec<String>> {
         receipts
             .iter()
-            .filter_map_log_receipts()
-            .filter_map(|((c_id, log_id), data)| {
+            .extract_log_id_add_data()
```
segfault-magnet, line 34: log id **and** data 

:D :D

segfault-magnet (commented):
A typo, probably missed it cuz the CI is failing anyway.

hal3e:
I have added an implementation of `assert_eq` directly to the code so that I can make the CI pass. This is a temporary solution until the bug in sway is fixed.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -123,8 +123,7 @@ impl<'de> DeserializeAs<'de, Word> for HexNumber {
         match bytes.len() {
             len if len > WORD_SIZE => {
                 return Err(D::Error::custom(format!(
-                    "value cant exceed {} bytes",
-                    WORD_SIZE
+                    "value cant exceed {WORD_SIZE} bytes",
```
segfault-magnet, line 126:
```suggestion
                    "value can't exceed {WORD_SIZE} bytes",
```

packages/fuels/tests/scripts/script_assert_eq/src/main.sw
```diff
@@ -0,0 +1,59 @@
...
+    T: Eq
+{
+    let FAILED_ASSERT_EQ_SIGNAL = 0xffff_ffff_ffff_0003;
+    if (v1 != v2) {
+        log(v1);
+        log(v2);
+        revert(FAILED_ASSERT_EQ_SIGNAL);
+    }
+}
+// link: https://github.com/FuelLabs/sway/issues/3946
```
digorithm, line 16: Can you open a `fuels-rs` issue and link this sway issue, so we know when to tackle this TODO?

digorithm (approved):
Very neat!

packages/fuels/tests/scripts/script_assert_eq/src/main.sw
hal3e: I wanted to wait for the compiler team to fix the bug, but it seems that it could take some time. I propose we merge this with a new issue to remove the added code when the bug is fixed.

hal3e:
New issue is here: https://github.com/FuelLabs/fuels-rs/issues/829

segfault-magnet (approved):

PR #815, opened 2023-02-03
feat!: enable payments with predicates

### LINKED ISSUE #809: Allow users to pay for transaction fees using predicates
> **Note**
> Note for sprint and milestone planning: this is a stretch goal and might not be done by the beta-4 deadline.

Users have to use their `Wallet` to pay for a transaction. However, there are many use cases where a user wants to pass a `Predicate`, instead of a `Wallet`, to pay for a transaction. The actual UTXOs used to pay for the transaction must be unlocked in the predicate passed. 

For example: 

```Rust
let contract_instance = MyContract::new(contract_id, wallet);

    .methods()
    .await?;
```

Here, we link a `Wallet` to the contract instance. Underneath, it will grab this wallet's funds to pay for the transaction. 

With this new feature, one could pass in a `Predicate` instead of a `Wallet`, and, underneath, the predicate would be evaluated, and, if true, the UTXOs unlocked would be used to pay for the tx fees.

This is based by a feedback we got from an external developer, here's what was said about this feature request: 

> When I want to spend utxos from a predicate, there should be a wallet that offers the same kind of features as a wallet for a simple public key address. Currently, you need to initialize the wallet with a key, and that will also be the address that pays for the fees. If I have a predicate, I generally want to pay for the fees with the coins available in the utxos on the predicate address. A public-key based address should not be required to manage a predicate wallet.

Salka1988:
The PR aims to enable predicates to be used as wallets, allowing payments and transactions to be made with them. The first step taken in this direction was to modify the `predicate_bindings` function so that it returns the `Predicate` data type, which will be further used as we use the wallet.

```rust
MyPredicateHandler::load_from(
"../../packages/fuels/tests/predicates/predicate_signatures/out/debug/predicate_signatures.bin",
).set_data(signatures).get_predicate();
```
This code loads the predicate .bin together with the signature for unlocking the predicate and returns the `Predicate` data type, which can be used as a wallet for making transactions, etc.

In addition, I added a new public trait called `Account`, which includes a group of functions common to both wallets and predicates. Any function that receives `T: Account` as a parameter can use these common functions.

```rust
pub trait Account: std::fmt::Debug + Send + Sync { ... }
```

Predicates that have dynamic data as input require special handling because the underlying pointer needs to be adjusted with the right offset for the actual data. To resolve the pointers of a given predicate input, we need to know all the previous tx inputs. The problem here is that the transaction can be modified at any point which can potentially invalidate the offsets.

`fuels_types::Input` is introduced to circumvent this. For predicates, it stores the relevant resource and the predicate data as `UnresolvedBytes`.

Some of `ScriptTransaction`s methods have been moved to the new `ScriptTransactionBuilder`. The idea is that the builder works on our input types and still allows for modifications. Once `build()` is called, the predicate data is resolved for all predicate inputs and the resulting `ScriptTransaction` doesn't allow for further changes to the inputs.

Some caveats:
- We currently don't have a native `Output` type. Should we implement as a thin wrapper for the sake of consistency?
- `UnresolvedBytes` has been moved to fuels_types to avoid cyclic deps. This also means that `Data` moved with it. Since 
   `Data` is used for parts of the abigen, we had to make it public. Is there a better alternative?
- To deduplicate significant amounts of code, `get_asset_inputs_for_amount` accepts `witness_index` as a parameter. For predicates we just ignore it in the implementation but overall it doesn't make sense in the context of predicates.
 
closes #809


packages/fuels/Forc.lock
```diff
@@ -0,0 +1,396 @@
+[[package]]
```
iqdecay, line 1: Remove this file

segfault-magnet (commented):
Spent the day reviewing alongside @hal3e, @MujkicA , @Salka1988 

Pushed changes on what we agreed upon, wrote comments for everything else.

Not fully done, 10 more files tomorrow.

examples/predicates/src/lib.rs
```diff
@@ -89,10 +70,32 @@ mod tests {
         let signatures = [signature1, signature2, signature3];
         // ANCHOR_END: predicate_signatures
 
+        // ANCHOR: predicate_load
+        abigen!(Predicate(name="MyPredicate", abi="packages/fuels/tests/predicates/predicate_signatures/out/debug/predicate_signatures-abi.json"));
+
+        let mut predicate: Predicate = MyPredicate::load_from(
+            "../../packages/fuels/tests/predicates/predicate_signatures/out/debug/predicate_signatures.bin",
+        )?.set_data(signatures).get_predicate();
```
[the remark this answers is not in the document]
digorithm: > Mainly because the MyPredicate ... .get_predicate() looks a bit weird, getting a predicate from a predicate.

I thought the same but didn't come up with a nice solution like you, @segfault-magnet. This makes a lot of sense to me, and it's worth delaying this a bit more to get to a more friendly experience. 

If we're all in agreement (including you @Salka1988 ), I'd say let's do it!

examples/contracts/src/lib.rs
```diff
@@ -639,6 +644,48 @@ mod tests {
...
+        let contract_methods_wallet =
+            MyContract::new(contract_id.clone(), wallet.clone()).methods();
+
+        let predicate = Predicate {
+            address: Bech32Address::default(),
+            code: vec![],
+            data: UnresolvedBytes::default(),
+            provider: Some(wallet.provider()?.clone()),
+        };
+        let contract_methods_predicate = MyContract::new(contract_id, predicate).methods();
```
digorithm, line 675: This is so neat; loving seeing it! 🔥

packages/fuels/tests/predicates.rs
```diff
@@ -589,3 +562,558 @@ async fn spend_predicate_coins_messages_generics() -> Result<()> {
...
+        .set_asset_id(AssetId::default());
+
+    let response = contract_methods
+        .get_msg_amount() // Our contract method.
+        .tx_params(tx_params) // Chain the tx params setting method.
+        .call_params(call_params)? // Chain the call params setting method.
+        .call() // Perform the contract call.
+        .await?;
+
+    // Todo ContractCall add,   asset transfer append_custom_asset
```
segfault-magnet, line 1058: Seems to be a note to do a test of some sort?

segfault-magnet (commented):
Reviewed alongside @Salka1988 @MujkicA @hal3e 

Pushed changes we agreed upon. One pending comment remains.

Also @Salka1988 is working on the Predicate bindings. @MujkicA is refactoring errors and introducing a locked-account abstraction to support the locked/unlocked wallets idea.

examples/contracts/src/lib.rs
Salka1988: :fire:

segfault-magnet (commented):
Some nits. Waiting for @MujkicA's error refactoring and locked account trait.

packages/fuels-signers/src/predicate.rs
```diff
@@ -0,0 +1,131 @@
...
+    pub data: UnresolvedBytes,
+    pub provider: Option<Provider>,
+}
+
+impl Predicate {
+    pub fn address(&self) -> &Bech32Address {
+        &self.address
+    }
+
+    pub fn set_predicate(&mut self, provider: Provider) -> &mut Self {
```
hal3e, line 29:
```suggestion
    pub fn set_provider(&mut self, provider: Provider) -> &mut Self {
```

Salka1988: I took care of this not only here but also in other locations.

segfault-magnet: @hal3e  :laughing: so much talk about predicates went to my head :D

MujkicA (commented):
After another pass I made some small changes to docs and tests.

segfault-magnet (commented):
Reviewed alongside @hal3e .

hal3e:
just a reminder to update the PR title (this will later become the commit message). Maybe use something like:
`feat: enable payments with predicates` or something more broad as you can use predicates the same way as wallets...

examples/contracts/src/lib.rs
```diff
@@ -639,6 +643,36 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
```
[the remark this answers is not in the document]
MujkicA: Moved to `fuels/tests/contracts.rs`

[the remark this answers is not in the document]
digorithm: You'd be 100% right, but I don't think this is an implementation-specific edge case! 

It's not an entry-level example for someone that's just checking out the SDK, that's for sure. But it's a nice showcase of a use case that isn't as rare as one would think (having different calls use different accounts for fees).

docs/src/predicates/send-spend-predicate.md
```diff
@@ -20,26 +18,20 @@ Next, let's add some coins, start a provider and connect it with the wallets.
 {{#include ../../../examples/predicates/src/lib.rs:predicate_coins}}
 ```
 
-Now we can use the predicate abigen, which will create a predicate instance for us.
+Now we can use the predicate abigen, which will create a predicate encoder instance for us. To spend the funds that are now locked in the predicate, we have to provide two out of three signatures whose public keys match the ones we defined in the predicate. In this example, the signatures are generated from an array of zeros.
```
digorithm, line 21: Shortening it to improve a bit of clarity.

```suggestion
Now we can use the predicate abigen to create a predicate encoder instance for us. To spend the funds now locked in the predicate, we must provide two out of three signatures whose public keys match the ones we defined in the predicate. In this example, the signatures are generated from an array of zeros.
```

docs/src/predicates/send-spend-predicate.md
```diff
@@ -20,26 +18,20 @@ Next, let's add some coins, start a provider and connect it with the wallets.
...
 ```
 
-To spend the funds that are now locked in the predicate, we have to provide two out of three signatures whose public keys match the ones we defined in the predicate. In this example, the signatures are generated from an array of zeros.
-
-```rust,ignore
-{{#include ../../../examples/predicates/src/lib.rs:predicate_signatures}}
-```
-
-After generating the signatures, we can use the predicate's `encode_data` and `spend` functions to spend the funds. If the provided data is correct the `receiver` wallet will get the funds, and we will verify that the amount is indeed correct.
+We can use the `transfer` method from the [Account](../getting-started/account.md) trait to transfer the assets. If the provided predicate data is correct the `receiver` wallet will get the funds, and we will verify that the amount is indeed correct.
```
digorithm, line 33:
```suggestion
We can use the `transfer` method from the [Account](../getting-started/account.md) trait to transfer the assets. If the predicate data is correct, the `receiver` wallet will get the funds, and we will verify that the amount is correct.
```

docs/src/providers/querying.md
```diff
@@ -1,6 +1,6 @@
 # Interacting with the blockchain
 
-Once you have set up a provider, you're ready to interact with the Fuel blockchain. Here are a few examples of what you can do with a provider; for a more in-depth overview of the API, check the [official provider API documentation](https://docs.rs/fuels/latest/fuels/signers/provider/struct.Provider.html).
+Once you have set up a provider, you're ready to interact with the Fuel blockchain. Here are a few examples of what you can do with a provider; for a more in-depth overview of the API, check the [official provider API documentation](https://docs.rs/fuels/latest/fuels/accounts/provider/struct.Provider.html).
```
digorithm, line 3:
```suggestion
Once you set up a provider, you can interact with the Fuel blockchain. Here are a few examples of what you can do with a provider; for a more in-depth overview of the API, check the [official provider API documentation](https://docs.rs/fuels/latest/fuels/accounts/provider/struct.Provider.html).
```

docs/src/getting-started/predicates.md
```diff
@@ -1,15 +1,36 @@
 # Predicates
 
-Predicates, in Sway, are programs that return a Boolean value, and they do not have any side effects (they are pure).
+Predicates, in Sway, are programs that return a Boolean value, and they do not have any side effects (they are pure). Assets can be owned by a predicate address. The predicate address is generated from the compiled byte code and is the same as the `P2SH` address used in Bitcoin. Users can seamlessly send assets to the predicate address as they do for any other address. To spend the predicate funds, the user has to provide the original `byte code` of the predicate together with the `predicate data`. The `predicate data` will be used when executing the `byte code`, and if the predicate is validated successfully, the funds can be transferred.
```
digorithm, line 3: Since we're touching this, rephrasing some parts from passive to active voice for clarity.

```suggestion
Predicates, in Sway, are programs that return a Boolean value and do not have any side effects (they are pure). A predicate address can own assets. The predicate address is generated from the compiled byte code and is the same as the `P2SH` address used in Bitcoin. Users can seamlessly send assets to the predicate address as they do for any other address. To spend the predicate funds, the user has to provide the original `byte code` of the predicate together with the `predicate data`. The `predicate data` will be used when executing the `byte code`, and the funds can be transferred if the predicate is validated successfully.
```

docs/src/getting-started/account.md
```diff
@@ -0,0 +1,39 @@
...
+- [Predicate](../getting-started/predicates.md)
+
+## Transferring assets
+
+An account implements the following methods for transferring assets:
+- `transfer`
+- `force_transfer_to_contract`
+- `withdraw_to_base_layer`
+
+The following examples are provided for a `Wallet` account. A `Predicate` account would work similarly, but requires you to set its predicate data before attempting to spend resources owned by it.
```
digorithm, line 19:
```suggestion
The following examples are provided for a `Wallet` account. A `Predicate` account would work similarly but requires you to set its predicate data before attempting to spend resources owned by it.
```

docs/src/getting-started/account.md
```diff
@@ -0,0 +1,39 @@
...
+{{#include ../../../examples/wallets/src/lib.rs:wallet_contract_transfer}}
+```
+
+For transferring assets to the base layer chain, you can use `wallet.withdraw_to_base_layer`.
+
+```rust,ignore
+{{#include ../../../examples/wallets/src/lib.rs:wallet_withdraw_to_base}}
+```
+
+The above example creates an `Address` from a string and converts it to a `Bech32Address`. Next, it calls `wallet.withdraw_to_base_layer` by providing the address, the amount to be transferred, and the transaction parameters. Lastly, to verify that the transfer succeeded, the relevant message proof is retrieved with `provider.get_message_proof,` and the amount and the recipient is verified.
```
digorithm, line 39:
```suggestion
The above example creates an `Address` from a string and converts it to a `Bech32Address`. Next, it calls `wallet.withdraw_to_base_layer` by providing the address, the amount to be transferred, and the transaction parameters. Lastly, to verify that the transfer succeeded, the relevant message proof is retrieved with `provider.get_message_proof,` and the amount and the recipient are verified.
```

digorithm (changes requested):
Some doc-related suggestions!

segfault-magnet (commented):
Reviewed with @hal3e 

Only have one additional suggestion:

in `fuels-test-helpers` you enable the mod `signers` if we have `accounts`. Maybe rename that to `accounts` as well?

The contents are reexported anyways and the mod itself is private -- so this isn't a publicly noticeable thing but still better to expunge all mention of the previous crate name.

hal3e (approved):
Amazing work! Congrats!

digorithm (approved):
Phewww, this was a long review. I did the best I could, but there could be some issues that are hard to catch due to the size of this PR. 

That said, this is a beautiful piece of work 😭 ❤️ I'm so excited to see this released.

Great job, everybody (especially @Salka1988 for leading this); this was a massive team effort.

PR #818, opened 2023-02-03
Pre-publish verification

Voxelot:
###### Description of changes

Adds a step to dry run the publish crates action to help identify potential issues before publishing
Also removes usage of the deprecated rust-action
Simplifies tag version validation


digorithm (dismissed):
TIL about `dtolnay/rust-toolchain`. Very neat -- thanks!

Oh, the publishing check seems to be failing.

Voxelot:
Looks like the action doesn't support dry-run's outside of the normal publish flow (yet).

However, it's still good to have this check run only during regular publishing to help mitigate any failures mid-publish, potentially requiring a new version bump.

digorithm (approved):

PR #819, opened 2023-02-03
feat: Reintroduce `Abigen` publicly in a new crate `fuels-code-gen`

segfault-magnet:
Moving `Abigen` from `fuels-core` into `fuels-macro` and making it private broke the indexer which relied on `Abigen` due to issues of using `abigen!` inside other macros.

`Abigen` could not have been exported from `fuels-macro` since it is a `proc-macro` crate type that cannot export anything except proc macros.

The only solution was to create a new crate.


digorithm (dismissed):
Short and sweet; :shipit:!

digorithm (approved):

Issue #822, opened 2023-02-06
Generated wasm bindings not wasm compatible


segfault-magnet:
The indexer team pointed out that the wasm-generated bindings will not compile for the wasm target. An example was given:
```rust
        #[derive(
            Clone,
            Debug,
            Eq,
            PartialEq,
            :: fuels :: macros :: Parameterize,
            :: fuels :: macros :: Tokenizable,
            :: fuels :: macros :: TryFrom,
        )]
        pub struct Greeting {
            pub id: u64,
            pub greeting: ::fuels::types::SizedAsciiString<32usize>,
            pub person: self::Person,
        }
```
The above is a generated struct that references the `fuels` package. This prohibits compiling to the wasm target due to `fuels` having dependencies (e.g. `mio`) that aren't wasm compatible.

When doing wasm abigen, everything should point to the separate crates themselves (e.g. instead of fuels::types, do fuels_types, etc.).

There is also the problem of one of the crates (`fuels_types`) not being wasm compatible as of late, but this will be tracked in another issue.

PR #825, opened 2023-02-06
fix wasm compatibility when `no_std` is used in Abigen

segfault-magnet:
fixes: #822

My thanks to @Br1ght0ne for his help around wasm. Please check if this makes sense :D 

To be considered a hotfix in order to unblock the indexer team. I will restructure/clean it asap. 

Even though you can use parts of the rust stdlib in wasm, I've still decided to add `#![no_std]` to the wasm test. This at least catches any usage of `std::` in the expanded code. If we ever find ourselves using subsets of the std that are compatible with wasm but not available in `core` or `alloc` then we'll figure out how to write a better test.

The `fuels` dependency has been removed from the wasm test, so the passing of the test proves we're not referencing `fuels` anymore. 

The derive macros of `Tokenizable`, `Parameterize` and `TryInto` had to receive two new attributes: `FuelsCorePath` and `NoStd`. 

The first one is so that we don't use `fuels` when expanding the derive macro in wasm builds.

The second one communicates that the derive macro should generate wasm-compatible code. It might also be done using conditional compilation but then we would need to agree on the condition, i.e. do we add a feature flag `wasm`, do we use the built-in one `no_std`, etc. 

From now on, to ease the pain later, we should first see if something is available in `core` before referencing it from std when modifying Abigen. But more on that after the refactor.


Salka1988 (commented):
:+1:

digorithm (approved):
I appreciate the cleverness of the fix, but can we track the clean-up / long-term fix in a separate issue? There are more maintainable solutions than this and this is a bit fragile (regexes!).

segfault-magnet:
> I appreciate the cleverness of the fix, but can we track the clean-up / long-term fix in a separate issue? There are more maintainable solutions than this and this is a bit fragile (regexes!).

Working on a maintainable solution atm, will create a task to track it. Merging.

Issue #812, opened 2023-01-31
Add support for remaining error codes in std::error_signals


nfurfaro:
Along the lines of #811, we should show the user the reason for a revert whenever possible.
This [file](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/error_signals.sw) contains all the error codes that we should eventually display in some way to the user.
In addition, [this PR](https://github.com/FuelLabs/sway/pull/3951) adds a new error code for a failing `assert`.

PR #831, opened 2023-02-08
feat: support all revert signals

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/812

This PR adds appropriate error messages to all `revert` transactions based on the IDs in  [error_signals.sw](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/error_signals.sw)


digorithm (approved):
This is such a nice dev experience improved -- good stuff, man!

PR #833, opened 2023-02-10
Add latest block time and spendable resources with exclusion

### LINKED ISSUE #832: Extend provider API with retrieval for latest block time and spendable resources with exclusion
The provider API lacks a convenient method to retrieve the latest block time. 

We could further improve our API by providing a method to retrieve a set of spendable resources given a list of resources to be excluded. This will most likely be relevant to users that create custom scripts by specifying their own inputs.

MujkicA:
Closes #832

This PR changes `provider.get_spendable_resources` to accept a filter which let's you define the target asset, amount, owner and also specify excluded resource ids.
It also adds `provider.get_inputs_for_filter` which works similarly to `wallet.get_inputs_for_amount` but with the advantage of being able to specify excluded ids.

It also adds a convenience method to the provider to retrieve the latest block time.

```rust
let filter = ResourceFilter {
            from: wallet.address().clone(),
            amount: coin_amount_1,
            excluded_utxos: vec![coin_2_utxo_id],
            excluded_message_ids: vec![message_id],
            ..Default::default()
        };
let resources = provider.get_spendable_resources(filter).await.unwrap();
```

While implementing the above, I made a bug which wasn't caught by our tests.
Because of that, I also added a test for `wallet.transfer()` for transferring asset ids other than the base asset.


packages/fuels-programs/src/script_calls.rs
```diff
@@ -179,6 +179,7 @@ where
             outputs,
             vec![],
         );
+
```
MujkicA, line 182:
```suggestion
```

packages/fuels/tests/providers.rs
```diff
@@ -639,3 +670,61 @@ async fn test_parse_block_time() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_get_spendable_with_exclusion() -> Result<()> {
```
segfault-magnet, line 675: Tried to make the test a bit more readable:
```rust
#[tokio::test]
async fn test_get_spendable_with_exclusion() -> Result<()> {
    let coin_amount_1 = 1000;
    let coin_amount_2 = 500;

    let mut wallet = WalletUnlocked::new_random(None);
    let address = wallet.address();

    let coins = [coin_amount_1, coin_amount_2]
        .into_iter()
        .flat_map(|amount| setup_single_asset_coins(address, BASE_ASSET_ID, 1, amount))
        .collect::<Vec<_>>();

    let message_amount = 200;
    let message = given_a_message(address.clone(), message_amount);

    let coin_1_utxo_id = coins[0].utxo_id;
    let coin_2_utxo_id = coins[1].utxo_id;

    let message_id = message.message_id();

    let (provider, _) = setup_test_provider(coins, vec![message], None, None).await;

    wallet.set_provider(provider);

    let requested_amount = coin_amount_1 + coin_amount_2 + message_amount;
    {
        let resources = wallet
            .get_spendable_resources(BASE_ASSET_ID, requested_amount)
            .await
            .unwrap();
        assert_eq!(resources.len(), 3);
    }

    {
        let excluded = (vec![coin_2_utxo_id], vec![message_id]);
        let resources = wallet
            .get_spendable_resources_with_exclusion(BASE_ASSET_ID, coin_amount_1, excluded)
            .await
            .unwrap();

        match resources.as_slice() {
            [Resource::Coin(coin)] => {
                assert_eq!(coin.utxo_id, coin_1_utxo_id);
            }
            _ => {
                panic!("This shouldn't happen!")
            }
        }
    }

    Ok(())
}

fn given_a_message(address: Bech32Address, message_amount: u64) -> Message {
    setup_single_message(Bech32Address::default(), address, message_amount, 0, vec![])
}

```

segfault-magnet (commented):
reviewed alongside @hal3e

hal3e (commented):
Reviewed with @segfault-magnet

packages/fuels/tests/providers.rs
```diff
@@ -639,3 +668,70 @@ async fn test_parse_block_time() -> Result<()> {
...
+
+fn given_a_message(address: Bech32Address, message_amount: u64) -> Message {
+    setup_single_message(
+        &Bech32Address::default(),
+        &address,
+        message_amount,
+        0,
+        vec![],
+    )
+}
```
iqdecay, line 737: gud tests

iqdecay (changes requested):
Left some questions, looking good!

packages/fuels-signers/src/provider.rs
```diff
@@ -60,6 +63,92 @@ impl From<TimeParameters> for FuelTimeParameters {
     }
 }
 
+pub(crate) struct ResourceQueries {
+    utxos: Vec<String>,
+    messages: Vec<String>,
```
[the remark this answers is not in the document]
MujkicA: `ResourceQueries` is hidden from the user and exists only to "encapsulate" the create of the exclusion and spend query. The reason why I don't do it on the fly inside of iterators is because going from `UtxoId` -> `&str` requires you to own the `String` at some point. If it were inside an iterator it's ownership is only temp.

Since this abstraction is used only internally to make our code cleaner, do you think we still need a comment?

packages/fuels-signers/src/provider.rs
```diff
@@ -60,6 +63,92 @@ impl From<TimeParameters> for FuelTimeParameters {
     }
 }
 
+pub(crate) struct ResourceQueries {
```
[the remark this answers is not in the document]
MujkicA: There is two different query items: exclusion and spend query. See above.

packages/fuels-signers/src/provider.rs
```diff
@@ -60,6 +63,92 @@ impl From<TimeParameters> for FuelTimeParameters {
...
+    pub excluded_message_ids: Vec<MessageId>,
+}
+// ANCHOR_END: resource_filter
+
+impl ResourceFilter {
+    pub fn owner(&self) -> String {
+        self.from.hash().to_string()
+    }
+
+    pub(crate) fn resource_queries(&self) -> ResourceQueries {
```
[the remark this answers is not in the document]
MujkicA: Having the methods on `ResourceFilter ` would require us to keep the fields of `ResourceQueries` inside it. 
This downgrades the devx since a user can't initialise a filter conveniently like this:

```rust
let filter = ResourceFilter {
   amount,
   asset_id,
   ..Default::default()
}
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -605,7 +614,7 @@ impl WalletUnlocked {
         tx_parameters: TxParameters,
     ) -> Result<(String, Vec<Receipt>)> {
         let inputs = self
-            .get_asset_inputs_for_amount(asset_id, amount, 0)
+            .get_asset_inputs_for_amount(BASE_ASSET_ID, amount, 0)
```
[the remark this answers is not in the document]
MujkicA: Good catch. It's unfortunate that we don't have a test that would catch this. I'll add one.

packages/fuels-signers/src/lib.rs
```diff
@@ -147,30 +147,46 @@ mod tests {
...
         // Setup two sets of coins, one for each wallet, each containing 1 coin with 1 amount.
         let mut wallet_1 = WalletUnlocked::new_random(None);
         let mut wallet_2 = WalletUnlocked::new_random(None).lock();
 
-        let amount = 1000000;
-        let mut coins_1 = setup_single_asset_coins(wallet_1.address(), BASE_ASSET_ID, 1, amount);
-        let coins_2 = setup_single_asset_coins(wallet_2.address(), BASE_ASSET_ID, 1, amount);
-
-        coins_1.extend(coins_2);
+        let coins = setup_single_asset_coins(wallet_1.address(), BASE_ASSET_ID, 1, amount);
```
hal3e, line 174: Now, that we do not add coins to the second wallet the comments on lines `170` and `176` are not valid anymore.

packages/fuels-signers/src/lib.rs
```diff
@@ -268,15 +275,54 @@ mod tests {
...
         let resulting_amount = wallet_1_final_coins.first().unwrap();
-        assert_eq!(resulting_amount.amount(), 3);
+        assert_eq!(resulting_amount.amount(), AMOUNT - SEND_AMOUNT);
 
         let wallet_2_final_coins = wallet_2.get_coins(BASE_ASSET_ID).await?;
-        assert_eq!(wallet_2_final_coins.len(), 2);
+        assert_eq!(wallet_2_final_coins.len(), 1);
 
         // Check that wallet 2's amount is 7:
         // 5 initial + 2 that was sent to it.
```
hal3e, line 284: These comments are not valid anymore. Maybe delete all the comments as it is clear from the code what we are doing...

packages/fuels-signers/src/provider.rs
```diff
@@ -301,6 +387,51 @@ impl Provider {
         Ok(res)
     }
 
+    /// Returns a vector consisting of `Input::Coin`s and `Input::Message`s for the given
+    /// `ResourceFilter`. The `witness_index` is the position of the witness (signature)
+    /// in the transaction's list of witnesses. In the validation process, the node will
+    /// use the witness at this index to validate the coins returned by this method.
+    pub async fn get_asset_inputs_for_filter(
```
hal3e, line 394:
```suggestion
    pub async fn get_filtered_inputs(
```

packages/fuels-signers/src/wallet.rs
```diff
@@ -560,9 +543,13 @@ impl WalletUnlocked {
...
-        let spendable_predicate_resources = predicate
-            .get_spendable_resources(predicate_address, asset_id, amount)
-            .await?;
+
+        let filter = ResourceFilter {
+            from: predicate_address.clone(),
+            amount,
+            ..Default::default()
+        };
+        let spendable_predicate_resources = predicate.get_spendable_resources(filter).await?;
```
hal3e, line 552: Did we mean to use `provider` here ?
```suggestion
        let spendable_predicate_resources = provider.get_spendable_resources(filter).await?;
```

hal3e (commented):
Review with @segfault-magnet. Looks good. Left some nits

packages/fuels-signers/src/provider.rs
MujkicA: I prefer it to be analogous to `get_asset_inputs_for_amount`. The `asset` part is not irrelevant since the returned inputs are always resources. 
If you are still unhappy with it, how about `get_filtered_asset_inputs`

segfault-magnet: I'm fine with `get_filtered_asset_inputs` or even `get_asset_inputs(a_filter_here)` if that name is not taken already.

@hal3e ?

hal3e: I am for `get_asset_inputs(the_filter)` or `get_filtered_asset_inputs`

segfault-magnet (approved):

packages/fuels-signers/src/lib.rs
```diff
@@ -202,81 +217,108 @@ mod tests {
...
+        let _receipts = wallet_1
+            .transfer(
+                wallet_2.address(),
+                SEND_AMOUNT,
+                BASE_ASSET_ID,
+                TxParameters::default(),
+            )
+            .await?;
+
+        let wallet_1_final_coins = wallet_1.get_spendable_resources(BASE_ASSET_ID, 1).await?;
```
digorithm, line 272: I don't have an answer to this yet, and I'm not even sure this could be a source of confusion, but one thing came to mind when reading `wallet_1.get_spendable_resources(BASE_ASSET_ID, 1)`: two methods with the same name (`get_spendable_resources`) that run the same operation (filtering) on the same data (resources). One takes a neat `ResourceFilter`, the other... two separate args. It would be neat if they had the same signature, but I know `ResourceFilter` takes the address, and `get_spendable_resources` bound to the wallet doesn't take an address. How could we unify these two APIs?

MujkicA: I hoped to make the changes less annoying by retaining parts of the old API. But the point is valid; name and behavior are the same, which could be confusing. 
How about extracting the address from the filter struct?

digorithm: Yeah, I think that would be the best solution. Let's do it!

segfault-magnet: It would solve the signature issue, but I feel kinda queasy about it. 

That would make the original `get_spendable_resources` on the Provider look like this:
```rust
    pub async fn get_spendable_resources(
        &self,
        address: Bech32Address,
        filter: ResourceFilter,
    ) -> ProviderResult<Vec<Resource>> {
```
But the `address` is just as much part of the filtering as `filter` is. I'm ok with additional arguments alongside `filter` but only if they aren't directly part of the filtering. Like if we had `retry_count` for retrying the query or something like that.

I'm for either leaving it as is -- a shortcut to get wallet-owned spendable resources OR going full bananas with it.

We no longer have `ResourceFilters` but a `ResourceQuery` which you can run.

So searching for resources might look like this:
```rust
let resource_query = provider.resource_query().from(from).amount(100); 
resource_query.run();
```

Then you can use the same thing to populate the `from` field from the wallet:
```rust
let query = wallet.get_spendable_resources();
query.amount(100);
let result = query.run().await;
```
where `wallet` will populate `from` before returning it from `get_spendable_resources`.

The query object inside would have to hold the provider either by value or reference, whatever we see is more ergonomic.

digorithm: > But the address is just as much part of the filtering as filter is.

Yeah, I've been trying to convince myself that this would be okay... but, nah, you're right, feels clunky. The address is a filtering predicate. 

> I'm for either leaving it as is -- a shortcut to get wallet-owned spendable resources 

Yup, I'd say let's keep as is and call the one bound to the wallet a shortcut.

Issue #835, opened 2023-02-13
Remove the revert error signals from the SDK and use them from `fuel-abi-types`


hal3e:
The signals were added to the`fuel-abi-types` in https://github.com/FuelLabs/fuel-abi-types/pull/11
I need a new release of `fuel-abi-types`

hal3e:
https://github.com/FuelLabs/fuel-abi-types/pull/12 new release.

PR #836, opened 2023-02-13
refactor: use revert error codes from `fuels-abi-types`

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/835

The SDK now uses the revert error codes from `fuels-abi-types`


MujkicA (approved):
LGTM

Issue #837, opened 2023-02-15
`fuels-types` are not wasm compatible


segfault-magnet:
The indexer team is having issues due to `fuels-types` not being wasm compatible. 

The cause is the dependency on `fuel-core-client` which, after a while, depends on `mio` which isn't wasm compatible. 

The solution is to turn off parts of the `fuels-types` that aren't wasm compatible, hiding it behind a flag (that will be part of the crate's defaults).

PR #839, opened 2023-02-15
fix!: wasm compatibility for `fuels-types`

segfault-magnet:
closes: #837 
closes: #854

A big thank you to @hal3e for help with the CI. 

The wasm offending functionality is now hidden behind the feature `std` in `fuels-types` and in `fuels-core` as well. 

### Breakage:
@digorithm 
It is part of the crate's default features, so most users shouldn't feel the change -- especially since there weren't other other features that would have made somebody do `default-features = false` in their `Cargo.toml`. 

But if they did for whatever reason, then they might see parts of `fuels-types` disappear since the `std` feature would not be enabled. So that makes it a theoretically breaking change.

Also `fuels` had two feature flags: `fuel-core` and `fuel-core-lib`. `fuel-core` is removed since it was implicitly added because the optional dependency `fuel-core` had not been referred to by prepending `dep:` to it.

`fuel-core` used to enable the fuel-core-lib only partially and was not the right flag to use.

### About the implementation
Cargo features should be additive by design. This means that you cannot remove dependencies by adding features. 

This further means that wasm support is generally implemented by *hiding* wasm-offending code behind a feature (such as `std`). 

In our case, if we want to be WASM compatible, we must not enable the `std` feature on any `fuels-*` crates our project might depend on.

Feature unification makes this a bit difficult. If there is even one dependency that enabled `std` on some `fuels-*` crate, all other crates that are part of that compilation will also have the `std` feature enabled for that crate.

Because of this, I've made it default for our workspace dependencies to start off without defaults (and thus without the `std` flag). 

If any of our crates need it, they will have to enable the feature explicitly.


hal3e (dismissed):
LGTM! Thanks for sorting this out!

hal3e (approved):
LGTM!

Salka1988 (approved):
👍

digorithm (approved):
Aaaah, very nice work. This makes a lot of sense.

PR #840, opened 2023-02-15
fix: Fuel-Core breaking changes

segfault-magnet:
Fixes the fuel-core issue. 

A bad CLI default introduced in `fuel-core` broke the 'with-binary' build of the SDK. 

The fact that spawning fuel-core as a library now implements RAII, where dropping the handle kills the node broke the 'with-fuel-core-as-library' build.


packages/fuels-test-helpers/src/node.rs
```diff
@@ -43,6 +44,7 @@ impl Config {
             manual_blocks_enabled: false,
             vm_backtrace: false,
             silent: true,
+            poa_interval_millis: 2500,
```
digorithm, line 47: Note that this will be temporary while we wait for the bad default value in the `fuel-core` CLI to be fixed. Once fixed, we'll bring back the `Instant` config.

packages/fuels/tests/providers.rs
```diff
@@ -194,6 +194,7 @@ async fn test_input_message_pays_fee() -> Result<()> {
 }
 
 #[tokio::test]
+#[ignore]
```
digorithm, line 197: Ignore these now because you can't use `manual_blocks_enabled` while using PoA. Once we bring it back to `Instant`, we'll re-enable those.

digorithm (approved):

packages/fuels-test-helpers/src/node.rs
iqdecay: Add a //TODO or open an issue?

iqdecay (approved):

Issue #588, opened 2022-09-21
Configuration-time constants


digorithm:
Sway currently supports constants declared in `Forc.toml`, e.g.: 

```toml
[project]
authors = ["Fuel Labs <contact@fuel.sh>"]
entry = "main.sw"
license = "Apache-2.0"
name = "oracle"

[constants]
# Change to an Identity once https://github.com/FuelLabs/sway/issues/2647 is fixed
owner = { type = "b256", value = "0x09c0b2d1a486c439a87bcba6b46a7a1a23f3897cc83a94521a96da5c23bc58db" }

[dependencies]
```

The SDK must offer support for this feature so that users can, in their Rust code, set the value of a constant and deploy the contract without the need to recompile the code.

To achieve this, the compiler will output a new piece of data (e.g., a new file similar to the JSON ABI) that will tell us the type of the constant, name (? tbd), and its _location in the compiled bytecode_. With this information, the SDK should inject that constant's value into the compiled bytecode before deploying it. 

Note that there will be more discussions on the specifics of how this is going to be achieved. The Sway compiler currently blocks this work until then.

mohammadfawaz:
https://github.com/FuelLabs/sway/pull/3536 (now in review) introduces configuration-time constants as described in https://github.com/FuelLabs/sway-rfcs/pull/19 including the JSON ABI generation.

mohammadfawaz:
@digorithm I've removed the "Blocked" status as the most recently `forc` release contains this feature.

PR #844, opened 2023-02-16
feat: support configurable constants

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/588

With this PR, the SDK supports setting the configurable constants of contracts and scripts. This is an example how configurable constants are used in sway:

```rust
contract;

configurable {
    U8: u8 = 8u8,
    BOOL: bool = true,
    ARRAY: [u32; 3] = [253u32, 254u32, 255u32],
    STR_4: str[4] = "fuel",
}

abi TestContract {
    fn return_configurables() -> (u8, bool, [u32; 3], str[4]);
}

impl TestContract for Contract {
    fn return_configurables() -> (u8, bool, [u32; 3], str[4]) {
        (U8, BOOL, ARRAY, STR_4)
    }
}
```
We can set them in the SDK like this:
```rust
    abigen!(Contract(
        name = "MyContract",
        abi = "packages/fuels/tests/contracts/configurables/out/debug/configurables-abi.json"
    ));

    let new_str: SizedAsciiString<4> = "FUEL".try_into()?;
    let new_u8 = 32;

    let configurables = MyContractConfigurables::new()
        .set_U8(new_u8)
        .set_STR_4(new_str);

    let contract_id = Contract::deploy_with_parameters(
        &wallet,
        TxParameters::default(),
        StorageConfiguration::with_storage_path(Some(
            "tests/contracts/configurables/out/debug/configurables-storage_slots.json".to_string(),
        )),
        configurables.into(),
        Salt::default(),
    )
    .await?;

    let contract_instance = MyContract::new(contract_id, wallet.clone());
```
Every configurable constant in sway will get a `set` method in the new `MyContractConfigurables` struct. This struct will use the right offsets and encoded data to update the binary before deployment.

Scripts have the same configuration struct but as they are not deployed the script instance has a new method called `with_configurables`. Below is an example how you use it:

```rust
    abigen!(Script(name="MyScript", abi="packages/fuels/tests/scripts/script_configurables/out/debug/script_configurables-abi.json"));

    let bin_path = "../fuels/tests/scripts/script_configurables/out/debug/script_configurables.bin";
    let instance = MyScript::new(wallet, bin_path);

    let new_str: SizedAsciiString<4> = "FUEL".try_into()?;
    let new_u8 = 32;

    let configurables = MyScriptConfigurables::new()
        .set_U8(new_u8)
        .set_STR_4(new_str.clone());

    let response = instance
        .with_configurables(configurables.into())
        .main()
        .call()
        .await?;
```

I have also added integration tests and updated the documentation.

<s>Note: I have temporarily disabled `forc fmt --check` as `configurables` are not yet supported.</s>


mohammadfawaz:
[This PR ](https://github.com/FuelLabs/sway/pull/4105)enables formatter support

mohammadfawaz:
The fact that this is working without any dram is fascinating to me.. Well done!

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
```diff
@@ -0,0 +1,118 @@
...
+    program_bindings::{
+        abi_types::{FullConfigurable, FullTypeDeclaration},
+        resolved_type::{resolve_type, ResolvedType},
+    },
+    utils::safe_ident,
+};
+
+#[derive(Debug)]
+pub(crate) struct ResolvedConfigurable {
+    pub name: Ident,
```
segfault-magnet, line 17: You can, if you like, use `Component` here as it also represents a field/variant in a struct/enum. It will resolve the type and snake-case the name.

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
```diff
@@ -0,0 +1,118 @@
...
+        .map(|c| ResolvedConfigurable::new(c, shared_types))
+        .collect::<Result<Vec<_>>>()?;
+
+    let config_methods = generate_configurable_impl(name, &resolved_configurables);
+    let from_impl = generate_from_impl(name);
+
+    Ok(quote! {
+        #[derive(Clone, Debug, Default)]
+        pub struct #name {
+            pub offsets_with_data: ::std::vec::Vec<(u64, ::std::vec::Vec<u8>)>
```
segfault-magnet, line 51: Does this have to be public?

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        #config_methods
+        #from_impl
+    })
+}
+
+fn generate_configurable_impl(
+    name: &Ident,
```
segfault-magnet, line 60: I keep forgetting what `name` refers to. Maybe call it `configurable_struct_name` or something like that?

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+fn generate_encoder_code(ttype: &ResolvedType) -> TokenStream {
+    quote! {
+        ::fuels::core::abi_encoder::ABIEncoder::encode(&[
+                <#ttype as ::fuels::types::traits::Tokenizable>::into_token(value)
+            ])
+            .expect("Cannot encode configurable data")
+            .resolve(0)
```
segfault-magnet, line 104: If we don't support vectors atm (evident by the `resolve(0)` maybe we can put out a compile-time error via the `ResolvedType` which has a helper to determine whether the resolved type contains vectors. 

If a configurable is a vector or contains one, then we fail the abigen? 

Wdyt?

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+fn generate_from_impl(name: &Ident) -> TokenStream {
+    quote! {
+        impl From<#name> for ::fuels::programs::Configurables {
+            fn from(config: #name) -> Self {
+                ::fuels::programs::Configurables {
+                    offsets_with_data: config.offsets_with_data.clone()
```
segfault-magnet, line 113: Don't need to clone since From<#name> will take ownership of the configuration structure.

packages/fuels-programs/src/lib.rs
```diff
@@ -3,3 +3,19 @@ pub mod call_utils;
 pub mod contract;
 pub mod logs;
 pub mod script_calls;
+
+#[derive(Debug, Default)]
+pub struct Configurables {
+    pub offsets_with_data: Vec<(u64, Vec<u8>)>,
+}
+
+pub fn replace_configurables(configurables: Configurables, binary: &mut Vec<u8>) {
```
segfault-magnet, line 12: I'd make this a method of `Configurables` since it makes so much use of it.

segfault-magnet: Maybe to use slices here?
```rust
pub fn replace_configurables_2(configurables: Configurables, binary: &mut [u8]) {
    configurables
        .offsets_with_data
        .iter()
        .for_each(|(offset, data)| {
            let offset = *offset as usize;

            binary[offset..offset+data.len()].copy_from_slice(data)
        });
}
```

packages/fuels/tests/configurables.rs
```diff
@@ -0,0 +1,170 @@
+use fuels::prelude::*;
+use fuels_types::SizedAsciiString;
+
+#[tokio::test]
+async fn contract_uses_default_configurables() -> Result<()> {
+    abigen!(Contract(
```
segfault-magnet, line 6: If you want, you can use setup contract test macro here.

packages/fuels/tests/contracts/configurables/src/main.sw
```diff
@@ -0,0 +1,30 @@
...
+    STRUCT: StructWithGeneric<u8> = StructWithGeneric{field_1: 8u8, field_2: 16},
+    ENUM: EnumWithGeneric<bool> = EnumWithGeneric::VariantOne(true),
+}
+
+abi TestContract {
+    fn return_configurables() -> (u8, bool, [u32; 3], str[4], StructWithGeneric<u8>, EnumWithGeneric<bool>);
+}
+
+impl TestContract for Contract {
+    fn return_configurables(    ) -> (u8, bool, [u32; 3], str[4], StructWithGeneric<u8>, EnumWithGeneric<bool>) {
```
segfault-magnet, line 27: The formatter kinda went bananas here.

packages/fuels-programs/src/contract.rs
```diff
@@ -186,6 +187,7 @@ impl Contract {
         wallet: &WalletUnlocked,
         params: TxParameters,
         storage_configuration: StorageConfiguration,
+        configurables: Configurables,
```
segfault-magnet, line 190: Deploying is starting to have way too many parameters. 

We should probably group them inside a `DeployConfig` with sensible defaults and a builder-like interface so that we may nuke the two deploy methods in favor of one with this structure.

Wdyt?

segfault-magnet (commented):
Superb work!

Br1ght0ne (approved):
Super clean code layout, nice! LGTM.

hal3e:
> Question:
> 
>     1. Is the name `MyContractConfigurables` auto-generated from the name of the contract, i.e. `MyContract` in this case?
> 
>     2. What if we do this instead?
> 
> 
> ```rust
> let configurables = MyContractConfigurables::new()
>         .set("U8", new_u8) // Runtime error if `"U8"` is not found
>         .set("STR_4", new_str.clone()); // Runtime error if `"STR_4"` is not found
> ```
> 
> I worry that `set_u8` is ambiguous if we have two separate configurables named `U8` and `u8`. Lower case constant names are discourage via a warning but not disallowed. Also, I find the change from upper case to lower case in the method name a bit unsettling (though I understand why it needs to be done).
1. Yes it is.
2. I think we cannot use your suggestion. AFAIK rust does not allow function overloading and here we would have `set(&str, SOME_TYPE)` multiple times. However, I also wanted to use uppercase function names lilke: `set_STR_4` but it was suggested that I avoid this because it is not idiomatic rust. It will work though.

So now we have to decided what to do:
- allow uppercase function names like `set_STR_4`, `setStr4`, etc.
- figure out some other way how to distinguish between lower and upper case constants

mohammadfawaz:
> AFAIK rust does not allow function overloading and here we would have set(&str, SOME_TYPE) multiple times

Can we not make `set` generic though?

Just brainstorming some more ideas here, but what if the struct `MyContractConfigurables` had all the configurables listed, with the same names as their Sway counterpart and then the code would look like:
```rust
let configurables = MyContractConfigurables::new()
configurables.U8 = ..;
configurables.BOOL = ..;
...
```
I suppose the issue with that is that we now have to overwrite all the configurables or rely on a zero default, which may not match the initializers chosen in the Sway program. BUT, we _can_ get initializers from the bytecode because we know where to look 🤔 

Either way, I just wanted to put my thoughts about this out there. You're the experts on the feasibility of all of this and the best API for users 🙂

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
hal3e: We decided to not modify the string so I will leave the current implementation

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
hal3e: Actually no. Good catch!

packages/fuels/tests/configurables.rs
hal3e: I would rather keep the same structure because I show that if I do not modify anything the contract will use the default values. I mean, the setup macro would do the same but still I want it to be visible.

packages/fuels-programs/src/contract.rs
hal3e: I like the idea. What do you think we make a new issue for this ?

hal3e:
> > AFAIK rust does not allow function overloading and here we would have set(&str, SOME_TYPE) multiple times
> 
> Can we not make `set` generic though?
> 
> Just brainstorming some more ideas here, but what if the struct `MyContractConfigurables` had all the configurables listed, with the same names as their Sway counterpart and then the code would look like:
> 
> ```rust
> let configurables = MyContractConfigurables::new()
> configurables.U8 = ..; // This is also not idomatic, but I think it's okay. Configurables are a special concept that Rust does not have.
> configurables.BOOL = ..;
> ...
> ```
> 
> I suppose the issue with that is that we now have to overwrite all the configurables or rely on a zero default, which may not match the initializers chosen in the Sway program. BUT, we _can_ get initializers from the bytecode because we know where to look thinking
> 
> Either way, I just wanted to put my thoughts about this out there. You're the experts on the feasibility of all of this and the best API for users slightly_smiling_face

I started with the idea to have on struct that holds all configurables as fields but as you said I can not just any defaults and would have to read them from the binary. As abigen is completely separate from the deploying (and thus the binary) I would have to set the binary somewhere and read the defaults which is just extra steps to get the same results as the approach with methods.

I think a generic `set`should work (I did not try it). Now we have to decide if we like more: `set_u8(10u8)` or `set("U8", 10u8)`. 

P.S `set("U8", 10u8)` is more work in the SDK as I will have to compare the param_types to make sure we did not get a wrong type because there is nothing stopping me to do `set("U8", MyStruct)` if I do not do that. The method approach has the right type in the argument so that is not needed. @FuelLabs/sdk-rust thoughts ?

packages/fuels-programs/src/contract.rs
segfault-magnet: Sure. Maybe not to put all the current deploy fn arguments in one struct, maybe only to group the params and various configurations (the oft-defaultable part) and let the others remain as arguments. 

In any case a new issue is a good idea, the implementer can decide on the best course of action.

segfault-magnet:
> I think a generic `set`should work (I did not try it). Now we have to decide if we like more: `set_U8(10u8)` or `set("U8", 10u8)`.

I'm for compile-time validation through type-safety with each constant getting its own setter.

segfault-magnet (approved):

segfault-magnet (approved):

segfault-magnet (approved):

packages/fuels-programs/src/contract.rs
digorithm: Ha! I _just_ thought the same thing; it's getting way too complicated. But yeah, a separate issue/PR for this!

digorithm (approved):
Super clean implementation!

packages/fuels-programs/src/contract.rs
hal3e: https://github.com/FuelLabs/fuels-rs/issues/855

PR #846, opened 2023-02-16
Add message output estimation

### LINKED ISSUE #845: Add output message tx dependency estimation
Currently we support automatic setup of external contracts and output variables. The same can be implemented for output messages.

MujkicA:
Closes #845

Extends `estimate_tx_dependancies` to also add output messages if they're missing.


packages/fuels/tests/contracts.rs
```diff
@@ -703,11 +741,15 @@ async fn test_output_variable_estimation_multicall() -> Result<()> {
     let amount = 1000;
 
     let mut multi_call_handler = MultiContractCallHandler::new(wallets[0].clone());
-    (0..3).for_each(|_| {
+    (0..2).for_each(|_| {
```
digorithm, line 744: Why this change here? 👀

digorithm (approved):
LGTM -- just left a few clarification questions!

packages/fuels/tests/contracts.rs
MujkicA: Actually, there is no need for it

packages/fuels-programs/src/contract.rs
```diff
@@ -886,6 +898,15 @@ impl MultiContractCallHandler {
                         .for_each(|call| call.append_variable_outputs(1));
                 }
 
+                Err(Error::RevertTransactionError { receipts, .. })
+                    if ContractCall::is_missing_message_output(&receipts) =>
+                {
+                    self.contract_calls
+                        .iter_mut()
+                        .take(1)
```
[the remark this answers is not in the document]
MujkicA: I've simplified this a little and added a comment. 
The gist is that for multi calls, it doesn't matter to which call we append the output. They all get merged into a single tx.

packages/fuels-programs/src/contract.rs
```diff
@@ -858,10 +870,15 @@ impl MultiContractCallHandler {
                 Err(Error::RevertTransactionError { receipts, .. })
                     if ContractCall::is_missing_output_variables(&receipts) =>
                 {
-                    self.contract_calls
-                        .iter_mut()
-                        .take(1)
-                        .for_each(|call| call.append_variable_outputs(1));
+                    // It doesn't matter where the missing output is added,
+                    // all calls will be merged to a single script tx
+                    self.contract_calls[0].append_variable_outputs(1);
```
segfault-magnet, line 875: What makes you sure that `self.contract_calls.len() > 0` will always be true at this point?

If there is a reason can you add it to a comment? 

Otherwise you can do: 
```rust
                    if let Some(first_contract_call) = self.contract_calls.first_mut() {
                        first_contract_call.append_message_outputs(1);
                    }
```

packages/fuels-programs/src/contract.rs
```diff
@@ -858,10 +870,15 @@ impl MultiContractCallHandler {
...
+                }
+
+                Err(Error::RevertTransactionError { receipts, .. })
+                    if ContractCall::is_missing_message_output(&receipts) =>
+                {
+                    self.contract_calls[0].append_message_outputs(1);
                 }
 
                 Err(Error::RevertTransactionError { ref receipts, .. }) => {
```
segfault-magnet, line 884: A simpler way to write this:
```rust
                    if let Some(first_contract_call) = self.contract_calls.first_mut() {
                        call.append_external_contracts(contract_id.clone());
                    }
```

segfault-magnet (commented):
Reviewed alongside @hal3e. We've put our bigger impacting suggestions in a separate branch, you can compare the differences [here](https://github.com/FuelLabs/fuels-rs/compare/mujkica/add_message_output_estimation...mujkica/add_message_output_estimation_pr_suggestions).

Reasons for the suggestions:
1. If a revert happened not because of a missing variable or message output then the code assumes that it must have been a missing contract input. This isn't true.
2. The message inside `expect_err` in the abovementioned case will never be propagated. The result in question will always be an `Err` at that place in the code so `expect_err` will never panic. If the original intent was for the message to be propagated then our changes offer one way of achieving that.
3. The current code assumes there can only be one revert reason (since only one match branch can be selected). Although this is probably true atm, it costs nothing to write it in a manner that can support multiple reverts.

We haven't discussed the MultiCallHandler's estimations. Perhaps they might benefit from the same suggestions.

packages/fuels-programs/src/contract.rs
MujkicA: Using simulate or call requires that the underlying tx is built via `build_tx()` which fails if there are no contract calls.

packages/fuels-programs/src/contract.rs
MujkicA: This seems like the same thing as mentioned above

packages/fuels-programs/src/contract.rs
iqdecay: This trick should be documented

MujkicA: I've applied the changes to both single and multi calls. I did change the naming a little and opted of `iter_mut().take(1)` instead of the above.

Salka1988: Papi's got tricks up his sleeves.

iqdecay (approved):
LGTM

Salka1988 (approved):
:+1:

PR #848, opened 2023-02-16
feat: support returning `Vec<>` types from contracts

### LINKED ISSUE #746: Transform `Vec` return from a contract into a `raw_slice` return from the calling scripts
Say we have the following contract:
```rust
contract;

abi MyContract {
    fn returns_vec() -> Vec<u64>;
}

    fn returns_vec() -> Vec<u64> {
        let mut v:Vec<u64> = Vec::new();
        v.push(1);
        v.push(2);
        v
    }
}
```
The method `returns_vec()` returns a `Vec<u64>` and the JSON ABI contains that information. If the SDK were to try to decode returned data, it will just be looking at a pointer that means nothing outside of the VM memory. Instead, a calling script might do the following:
```rust
fn main() -> raw_slice {
    let caller = abi(MyContract, <contract_id>);
    caller.returns_vec().as_raw_slice()
}
```
I'm proposing that the SDK does this automatically as follows:
1. The SDK knows that the return type _from the contract_ is a `Vec` and knows it element type.
2. The `CALL` opcode will give us back a base pointer (say in `$r1`) and a length (say in `$r2`)  (basically information contained in the struct `Vec`). Instead of returning those values directly and decoding them, we need an extra step that somehow crafts a `retd $r1 $r2` instruction. This instruction will generate a `ReturnData` receipt with all of the actual vector data is included. This is basically what `as_raw_slice()` is doing under the hood, but in a more convoluted manner. 

If we do the above, then the user will seamlessly be able to return vectors from contracts. That still won't work for multiple `Vec` returns (e.g. tuple of vectors), but that's okay for now.

iqdecay:
This PR closes #746 by allowing the user to return a vector containing any type
from a smart contract.  
It relies on one main change:
* The script that uses the `CALL` opcode on the contract now needs the output
  type of the contract call to work properly. Indeed, to know how many bytes
  the vector takes in VM memory and return all these bytes, you need to know
  the encoding width of the inner type of the vector.


packages/fuels/tests/types/vector_output/src/main.sw
```diff
@@ -0,0 +1,133 @@
...
+    fn tuple_in_vec() -> Vec<(u64, u32)>;
+    fn u16_in_vec(len: u16) -> Vec<u16>;
+    fn u32_in_vec(len: u32) -> Vec<u32>;
+    fn u64_in_vec(len: u64) -> Vec<u64>;
+    fn u64_in_vec(len: u64) -> Vec<u64>;
+    fn u8_in_vec(len: u8) -> Vec<u8>;
+}
+
+impl VectorsOutputContract for Contract {
+    fn array_in_vec() -> Vec<[u64; 4]> {
```
mohammadfawaz, line 29: Ouuuf looks so good.

packages/fuels/tests/types.rs
```diff
@@ -1677,3 +1678,103 @@ async fn test_b512() -> Result<()> {
...
+
+    {
+        let expected: Vec<[u64; 4]> = vec![
+            ([1, 1, 1, 1]),
+            ([2, 2, 2, 2]),
+            ([3, 3, 3, 3]),
+            ([4, 4, 4, 4]),
+        ];
+        let response = contract_methods.array_in_vec().call().await?.value;
+        assert_eq!(response, expected);
```
digorithm, line 1733: This is working so beautifully! 😍

packages/fuels-programs/src/contract.rs
```diff
@@ -471,13 +471,53 @@ pub fn get_decoded_output(
...
+                    return_data_receipt
+                ));
+            };
+            Some(
+                return_data_receipt
+                    .data()
+                    .expect("ReturnData should have data")
+                    .to_vec(),
+            )
+        }
```
segfault-magnet, line 520: A functional alternative
Iterate in tuples of two. If you find a pair that satisfies the given conditions, return the data from the second one:
```rust
            let vector_data = receipts
                .iter()
                .tuple_windows()
                .find_map(|(current_receipt, next_receipt)| extract_vec_data(current_receipt, next_receipt, contract_id))
                .cloned()
                .expect("Could not extract vector data");
```
```rust
fn extract_vec_data<'a>(
    current_receipt: &Receipt,
    next_receipt: &'a Receipt,
    contract_id: ContractId,
) -> Option<&'a Vec<u8>> {
    match (current_receipt, next_receipt) {
        (
            Receipt::ReturnData {
                id: first_id,
                data: first_data,
                ..
            },
            Receipt::ReturnData {
                id: second_id,
                data: vec_data,
                ..
            },
        ) if *first_id == contract_id
            && first_data.is_empty()
            && *second_id == ContractId::zeroed() =>
        {
            Some(vec_data)
        }
        _ => None,
    }
}

```

packages/fuels/tests/types.rs
```diff
@@ -1677,3 +1678,103 @@ async fn test_b512() -> Result<()> {
...
+    let response = contract_methods.u8_in_vec(10).call().await?;
+    assert_eq!(response.value, (0..10).collect::<Vec<_>>());
+    let response = contract_methods.u16_in_vec(11).call().await?;
+    assert_eq!(response.value, (0..11).collect::<Vec<_>>());
+    let response = contract_methods.u32_in_vec(12).call().await?;
+    assert_eq!(response.value, (0..12).collect::<Vec<_>>());
+    let response = contract_methods.u64_in_vec(13).call().await?;
+    assert_eq!(response.value, (0..13).collect::<Vec<_>>());
+    let response = contract_methods.bool_in_vec().call().await?;
+    assert_eq!(response.value, [true, false, true, false].to_vec());
```
segfault-magnet, line 1706:
```suggestion
    let contract_methods = contract_instance.methods();

    let response = contract_methods.u8_in_vec(10).call().await?;
    assert_eq!(response.value, (0..10).collect::<Vec<_>>());

    let response = contract_methods.u16_in_vec(11).call().await?;
    assert_eq!(response.value, (0..11).collect::<Vec<_>>());

    let response = contract_methods.u32_in_vec(12).call().await?;
    assert_eq!(response.value, (0..12).collect::<Vec<_>>());

    let response = contract_methods.u64_in_vec(13).call().await?;
    assert_eq!(response.value, (0..13).collect::<Vec<_>>());

    let response = contract_methods.bool_in_vec().call().await?;
    assert_eq!(response.value, [true, false, true, false].to_vec());

```

packages/fuels/tests/types.rs
```diff
@@ -1677,3 +1678,103 @@ async fn test_b512() -> Result<()> {
...
+    );
+    let contract_methods = contract_instance.methods();
+
+    {
+        let expected: Vec<[u64; 4]> = vec![
+            ([1, 1, 1, 1]),
+            ([2, 2, 2, 2]),
+            ([3, 3, 3, 3]),
+            ([4, 4, 4, 4]),
+        ];
```
segfault-magnet, line 1731: Parentheses are not needed.
```suggestion
        let expected: Vec<[u64; 4]> = vec![[1, 1, 1, 1], [2, 2, 2, 2], [3, 3, 3, 3], [4, 4, 4, 4]];
```

packages/fuels/tests/types.rs
```diff
@@ -1677,3 +1678,103 @@ async fn test_b512() -> Result<()> {
...
+    }
+    {
+        let expected: Vec<Pasta> = vec![
+            (Pasta::Tortelini(Bimbam {
+                bim: 1111,
+                bam: 2222_u32,
+            })),
+            (Pasta::Rigatoni(1987)),
+            (Pasta::Spaghetti(true)),
+        ];
```
segfault-magnet, line 1743: Extra parentheses
```suggestion
        let expected: Vec<Pasta> = vec![
            Pasta::Tortelini(Bimbam {
                bim: 1111,
                bam: 2222_u32,
            }),
            Pasta::Rigatoni(1987),
            Pasta::Spaghetti(true),
        ];

```

segfault-magnet:
Review done alongside @hal3e 

Is returning Vectors from scripts planned as well @digorithm @iqdecay ?

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
```diff
@@ -83,21 +82,6 @@ fn function_arguments(
...
 }
 
-fn resolve_fn_output_type(
-    function: &FullABIFunction,
-    shared_types: &HashSet<FullTypeDeclaration>,
-) -> Result<ResolvedType> {
-    let output_type = resolve_type(function.output(), shared_types)?;
-    if output_type.uses_vectors() {
-        Err(error!(
-            "function '{}' contains a vector in its return type. This currently isn't supported.",
```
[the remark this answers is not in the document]
iqdecay: Disagree: as a user I still want to be able to interact with a contract even if some methods are not supported by the SDK.

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -70,8 +70,33 @@ impl ABIDecoder {
...
+                "The bytes provided do not correspond to a Vec<{:?}> got: {:?}", param_type, bytes
+            ));
+        }
+        let mut results = vec![];
+        let mut bytes_read = 0;
+        while bytes_read < bytes.len() {
+            let res = Self::decode_param(param_type, skip(bytes, bytes_read)?)?;
+            bytes_read += res.bytes_read;
+            results.push(res.token);
+        }
```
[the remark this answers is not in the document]
iqdecay: Great suggestion about the `decode_multiple`. I prefer not to extracte away the decoding logic in yet another function though

packages/fuels-programs/src/contract.rs
iqdecay: I don't prefer this functional approach because IMO it  hinders the readability/understanding you get of the structure of the receipts, and possible debugging steps (you can't really get the same error messages with the functional approach). While implementing `Bytes` especially I really liked having these error messages.

iqdecay:
> Is returning Vectors from scripts planned as well @digorithm @iqdecay ?

Tracking in #848

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -350,6 +327,27 @@ fn skip(slice: &[u8], num_bytes: usize) -> Result<&[u8]> {
...
 
+fn validate_not_nested_vector(param_type: &ParamType) -> Result<()> {
+    match param_type {
+        ParamType::Vector(..) => Err(error!(
+            InvalidData,
+            "Vectors containing vectors are not supported!"
+        )),
+        _ => Ok(()),
+    }
+}
```
segfault-magnet, line 338: This will only detect immediately nested vectors, e.g. `Vec<Vec<u8>>` while not detecting if there is something separating them, for example in Sway: 
```rust
struct SomeStruct {
    a_field: u64,
    another_vector: Vec<u64>,
}
// later on in a contract
    fn bug() -> Vec<SomeStruct> {
        let mut inner_vec: Vec<u64> = Vec::new();
        inner_vec.push(1);
        inner_vec.push(2);
        inner_vec.push(3);

        let some_struct = SomeStruct {
            a_field: 4,
            another_vector: inner_vec,
        };

        let mut vec: Vec<SomeStruct> = Vec::new();
        vec.push(some_struct);

        vec
    }
```
when testing in our harness:
```rust
    {
        let expected: Vec<SomeStruct> = vec![SomeStruct {
            a_field: 4,
            another_vector: vec![1, 2, 3],
        }];
        let response = contract_methods.bug().call().await?.value;
        assert_eq!(response, expected);
    }
```
This would not fail but rather be populated by garbage data for the second vector.

Since vector decoding doesn't work for nested vectors, this validation needs to be fixed.

segfault-magnet (changes requested):

packages/fuels-core/src/abi_decoder.rs
digorithm: Damn, nice catch, @segfault-magnet.

segfault-magnet: 'twas in tandem with @hal3e, which makes PR reviews super fun!

iqdecay: Good catch, implementing a thorough recursive check as a method of `ParamType` and its tests.

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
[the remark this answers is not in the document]
iqdecay: I had _not_ read your comment but it turns out that exactly what I did :smile:

packages/fuels-types/src/param_types.rs
```diff
@@ -68,6 +68,43 @@ impl ParamType {
         }
     }
 
+    pub fn contains_no_nested_vectors(&self) -> bool {
+        match &self {
+            ParamType::Vector(param_type) => {
```
MujkicA, line 73: You can merge the match arms for vectors and arrays:
```rust
ParamType::Vector(param_type) | ParamType::Array(param_type, ..) => {
    !param_type.is_vector() && param_type.contains_no_nested_vectors()
}
```

packages/fuels-types/src/param_types.rs
```diff
@@ -68,6 +68,43 @@ impl ParamType {
...
+                !param_type.is_vector() && param_type.contains_no_nested_vectors()
+            }
+            ParamType::Struct {
+                fields, generics, ..
+            } => {
+                fields.iter().all(|(_, param_type)| {
+                    param_type.contains_no_nested_vectors() && !param_type.is_vector()
```
MujkicA, line 80: Checking `!param_type.is_vector()` before going into recursion might be better.

packages/fuels-programs/src/call_utils.rs
```diff
@@ -38,12 +40,23 @@ pub async fn build_tx_from_contract_calls(
...
+    )
+    .len()
+        * n_vectors_calls;
 
-    // Calculate instructions length for call instructions
-    // Use placeholder for call param offsets, we only care about the length
-    let calls_instructions_len =
-        get_single_call_instructions(&CallOpcodeParamsOffset::default()).len() * calls.len();
-
+    let calls_instructions_len = calls_instructions_len_no_vectors + calls_instructions_len_vectors;
```
MujkicA, line 59: Maybe the logic for calculating this can be moved to a helper. I feel like it's a bit neater if the comment becomes a docstring for the helper.

packages/fuels-programs/src/call_utils.rs
```diff
@@ -230,15 +247,38 @@ pub(crate) fn get_single_call_instructions(offsets: &CallOpcodeParamsOffset) ->
...
         op::movi(0x13, asset_id_offset),
         op::call(0x10, 0x12, 0x13, 0x11),
-    ];
+    ]
+    .to_vec();
+    // The instructions are different if you want to return data that was on the heap
+    if let ParamType::Vector(inner_param_type) = output_param_type {
+        let inner_type_byte_size: u16 =
+            (inner_param_type.compute_encoding_width() * WORD_SIZE) as u16;
+        instructions.extend([
```
MujkicA, line 265: Also a potential candidate for a helper method.

MujkicA (commented):
Left some nits. Hats off for figuring this out.

packages/fuels-programs/src/call_utils.rs
iqdecay: I don't really like yet another helper method inside an helper method.

packages/fuels-types/src/param_types.rs
```diff
@@ -68,6 +68,40 @@ impl ParamType {
...
+            } => {
+                generics.iter().all(|param_type| {
+                    !param_type.is_vector() && param_type.contains_no_nested_vectors()
+                }) && variants.param_types().iter().all(|param_type| {
+                    !param_type.is_vector() && param_type.contains_no_nested_vectors()
+                })
+            }
+            _ => true,
+        }
+    }
```
hal3e, line 99: This could be cleaned up a bit.

```rust
    pub fn contains_nested_vectors(&self) -> bool {
        match &self {
            ParamType::Vector(param_type) => param_type.uses_vectors(),
            _ => self.uses_vectors(),
        }
    }

    fn uses_vectors(&self) -> bool {
        match &self {
            ParamType::Vector(..) => true,
            ParamType::Array(param_type, ..) => param_type.uses_vectors(),
            ParamType::Tuple(param_types, ..) => Self::any_nested_vectors(param_types),
            ParamType::Struct {
                fields, generics, ..
            } => {
                let fields = fields.iter().map(|(_, param_type)| param_type);
                Self::any_nested_vectors(chain!(fields, generics))
            }
            ParamType::Enum {
                generics, variants, ..
            } => {
                let variants = variants.param_types();
                Self::any_nested_vectors(chain!(&variants, generics))
            }
            _ => false,
        }
    }

    fn any_nested_vectors<'a>(param_types: impl IntoIterator<Item = &'a ParamType>) -> bool {
        param_types
            .into_iter()
            .any(|param_type| param_type.uses_vectors())
    }
```

hal3e: This takes in consideration that you changed `contains_no_nested_vectors`

packages/fuels-core/src/abi_decoder.rs
```diff
@@ -325,6 +332,17 @@ fn skip(slice: &[u8], num_bytes: usize) -> Result<&[u8]> {
     }
 }
 
+fn calculate_num_of_elements(param_type: &ParamType, bytes: &[u8]) -> Result<usize> {
+    let memory_size = param_type.compute_encoding_width() * WORD_SIZE;
+    if bytes.len() % memory_size != 0 {
+        return Err(error!(
+            InvalidData,
+            "The bytes provided do not correspond to a Vec<{:?}> got: {:?}", param_type, bytes
```
hal3e, line 340: Update this error message to reflect usage in `decode_raw_slice`.

packages/fuels-programs/src/call_utils.rs
```diff
@@ -124,11 +141,12 @@ pub(crate) fn get_instructions(
...
-    let num_calls = calls.len();
-
     let mut instructions = vec![];
-    for (_, call_offsets) in (0..num_calls).zip(offsets.iter()) {
-        instructions.extend(get_single_call_instructions(call_offsets));
+    for (call, call_offsets) in calls.iter().zip(offsets.iter()) {
+        instructions.extend(get_single_call_instructions(
+            call_offsets,
+            &call.output_param,
+        ));
```
hal3e, line 149: Functional alternative:

```rust
pub(crate) fn get_instructions(
    calls: &[ContractCall],
    offsets: Vec<CallOpcodeParamsOffset>,
) -> Vec<u8> {
    calls
        .iter()
        .zip(&offsets)
        .flat_map(|(call, offset)| get_single_call_instructions(offset, &call.output_param))
        .chain(op::ret(RegId::ONE).to_bytes().into_iter())
        .collect()
}
```

hal3e (commented):
Reviewed alongside @segfault-magnet

packages/fuels-types/src/param_types.rs
```diff
@@ -1215,4 +1249,84 @@ mod tests {
...
+            generics: param_types_no_nested_vec
+        }
+        .contains_no_nested_vectors());
+        assert!(!ParamType::Enum {
+            name: "EnumName".to_string(),
+            variants: EnumVariants::new(tuples_no_nested_vec)?,
+            generics: param_types_nested_vec
+        }
+        .contains_no_nested_vectors());
+        Ok(())
```
segfault-magnet, line 1330: Although this might not qualify as a 100% unit test, it might (or might not) be interesting to you:
```suggestion
        assert!(Vec::<u8>::param_type().contains_no_nested_vectors());
        assert!(!Vec::<Vec<u8>>::param_type().contains_no_nested_vectors());

        assert!(<[u8; 10]>::param_type().contains_no_nested_vectors());
        assert!(!<[Vec<u8>; 10]>::param_type().contains_no_nested_vectors());

        assert!(<(u64, u32)>::param_type().contains_no_nested_vectors());
        assert!(!<((), Byte, Vec<u8>)>::param_type().contains_no_nested_vectors());

        #[derive(Parameterize)]
        #[FuelsTypesPath("crate")]
        struct StructCanHoldAVecInGeneric<T: Parameterize> {
            bim: u16,
            bam: bool,
            generic: T,
        }
        assert!(
            !StructCanHoldAVecInGeneric::<((), Byte, Vec<u8>)>::param_type()
                .contains_no_nested_vectors()
        );
        assert!(
            StructCanHoldAVecInGeneric::<((), Byte, u8)>::param_type().contains_no_nested_vectors()
        );

        #[derive(Parameterize)]
        #[FuelsTypesPath("crate")]
        struct StructHasAVecAsAField {
            bim: u16,
            bam: bool,
            the_vec: Vec<u8>,
        }
        assert!(!StructHasAVecAsAField::param_type().contains_no_nested_vectors());

        #[derive(Parameterize)]
        #[FuelsTypesPath("crate")]
        enum EnumCanHoldAVecInGeneric<T: Parameterize> {
            Bim(u16),
            Bam(bool),
            TheGeneric(T),
        }
        assert!(EnumCanHoldAVecInGeneric::<(u64, u32)>::param_type().contains_no_nested_vectors());
        assert!(!EnumCanHoldAVecInGeneric::<Vec<u8>>::param_type().contains_no_nested_vectors());

        #[derive(Parameterize)]
        #[FuelsTypesPath("crate")]
        enum EnumHasAVecAsAVariant {
            Bim(u16),
            Bam(bool),
            TheVec(Vec<u8>),
        }
        assert!(!EnumHasAVecAsAVariant::param_type().contains_no_nested_vectors());

        Ok(())
```

packages/fuels-core/src/abi_decoder.rs
iqdecay: This particular thing is changed in #868 for a more general purpose.

packages/fuels-programs/src/contract.rs
```diff
@@ -477,13 +478,34 @@ pub fn get_decoded_output(
...
+        ReturnLocation::ReturnData if output_param.is_vector() => {
+            // If the output of the function is a vector, then there are 2 consecutive ReturnData
+            // receipts. The first one is the one that returns the pointer to the vec struct in the
+            // VM memory, the second one contains the actual vector bytes (that the previous receipt
+            // points to).
+            // We ensure to take the right "first" ReturnData receipt by checking for the
+            // contract_id. There are no receipts in between the two ReturnData receipts because of
+            // the way the scripts are built (the calling script adds a RETD just after the CALL
+            // opcode, see `get_single_call_instructions`).
+            // Find the position of the first corresponding ReturnData receipt
```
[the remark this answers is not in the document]
hal3e: I was referring to this line: `// Find the position of the first corresponding ReturnData receipt`

packages/fuels-types/src/param_types.rs
iqdecay: Will keep this form in mind for the `Bytes`, I really just want to get `Vec`s out asap

packages/fuels-types/src/param_types.rs
hal3e: Thanks for updating the code. However, The `is_vector()` check is needlessly repeated and can be avoided if you use a separate function as we suggested above. Also, chaining the iterators is cleaner than calling the `any_nested_vectors` several times.

packages/fuels-types/src/param_types.rs
```diff
@@ -67,6 +68,44 @@ impl ParamType {
...
+                Self::any_nested_vectors(chain!(fields, generics))
+            }
+            _ => false,
+        }
+    }
+
+    fn any_nested_vectors<'a>(param_types: impl IntoIterator<Item = &'a ParamType>) -> bool {
+        param_types
+            .into_iter()
+            .any(|param_type| param_type.is_vector() || param_type.contains_nested_vectors())
```
segfault-magnet, line 102: We're allowing for one vector as the root type in `contains_nested_vectors`. Any further check we do should only detect whether vectors are used. If they are that makes them nested vectors.
```suggestion
            .any(|param_type| param_type.uses_vectors())
```

segfault-magnet (commented):
Reviewed with @hal3e

segfault-magnet (approved):
:tada:

Issue #851, opened 2023-02-21
`fuels-signers` tests fails when run from inside package


segfault-magnet:
When you run `cargo test` from inside `fuels-signers` it fails on master attow indicating that the CI missed this.

Find and fix the `fuels-signers` issue as well as the CI's.

PR #852, opened 2023-02-21
fix: ci individual package testing

segfault-magnet:
closes: #851

Passing `--workspace` will make cargo test each package individually. It is not exactly the same effect as going into the package and running `cargo test` due to workspace-wide feature unification happening before the test.

This is as close as we can get without bumping our CI times (probably) significantly by running `cargo test` for each package and then for the workspace as a whole.


hal3e (approved):
LGTM

digorithm (approved):
Nice!

PR #857, opened 2023-02-23
fix: latest release cannot be used by cargo

segfault-magnet:
If you run 
```bash
 CARGO_LOG=debug cargo update  
```
On a simple project that has `fuels = "0.37.0"` as a dep, among a bunch of other logs you'll see the following:
```
[2023-02-23T15:38:57Z INFO  cargo::sources::registry::index] failed to parse "fu/el/fuels-programs" registry package: feature `std` includes `fuels-test-helpers/std`, but `fuels-test-helpers` is not a dependency
[2023-02-23T15:38:57Z DEBUG cargo::core::registry] load/precise  registry `crates-io`
[2023-02-23T15:38:57Z DEBUG cargo::sources::registry::index] cache missing for "fu/el/fuels-signers" error: No such file or directory (os error 2)
[2023-02-23T15:38:57Z DEBUG cargo::sources::registry::index] slow path for "fu/el/fuels-signers"
[2023-02-23T15:38:57Z INFO  cargo::sources::registry::index] failed to parse "fu/el/fuels-signers" registry package: feature `std` includes `fuels-test-helpers/std`, but `fuels-test-helpers` is not a dependency
[2023-02-23T15:38:57Z DEBUG cargo::core::registry] load/precise  registry `crates-io`
```
`fuels-signers` and `fuels-programs` are the only ones being complained about and they're exactly the ones that don't have a matching version. 

It would seem that enabling features on `dev-dependencies` via a traditional feature is not something you should do.

It will pass on your local machine, it will pass our CI. 

You'll only find out when you release it and stop having the dev dependencies.

we should look into guarding against this if this ends up the reason behind the release issue.

### DIFF
--- a/packages/fuels-core/Cargo.toml
+++ b/packages/fuels-core/Cargo.toml
@@ -19,11 +19,8 @@ itertools = "0.10"
 sha2 = "0.9.5"
 
 [dev-dependencies]
-fuels-test-helpers = { workspace = true }
+fuels-test-helpers = { workspace = true, features = ["std"] }
 
 [features]
 default = ["std"]
-std = [
-    "fuels-test-helpers/std",
-    "fuels-types/std",
-]
+std = ["fuels-types/std"]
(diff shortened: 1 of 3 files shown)


digorithm (approved):
Yikes, Cargo... Yikes.

Nice catch though -- makes sense and I think it'll fix it.

PR #859, opened 2023-02-24
Resolve cyclic deps

### LINKED ISSUE #808: Resolve internal cyclic dependencies. `fuels-*` crates should form a DAG.
I've been talking to @segfault-magnet about how our internal dependencies (i.e., `fuels-*` crates) are organized and depend on each other. @segfault-magnet wrote a simple parser that prints the dependency graph of our inner crates, and we could see many cycles in it. 

Ideally, this should be a Direct Acyclic Graph (DAG). This is both a tech debt work _and_ something affecting our CI/CD. Our publishing action currently has `no-verify: true`, which disables verification of some crate-related info. That's because there's an internal cyclic dependency between the crates we need to publish.

MujkicA:
Closes #808

This PR untangles the cyclic dependency between `fuels-signers`, `fuels-core` and `fuels-test-helpers` by relocating some wallet tests to `fuels` and removing the use of test helpers where they're not relevant for the test itself.


hal3e (approved):
LGTM

Salka1988 (approved):
:+1:

iqdecay (approved):
LGTM

Br1ght0ne (approved):

hal3e:
Please do not forget to update the commit message.

digorithm (approved):
Great stuff. I wonder if we can now remove the `no-verify: true` from the publish step. It was added because of the cyclic dependencies.

segfault-magnet (dismissed):

Issue #855, opened 2023-02-22
Add `DeployConfig` builder


hal3e:
Currently, we have two functions to deploy contracts `deploy` and `deploy_with_parameters`. The later one will, after merging https://github.com/FuelLabs/fuels-rs/pull/844, have 6 arguments and calling it looks like this :
```rust
    let contract_id = Contract::deploy_with_parameters(
        "tests/contracts/configurables/out/debug/configurables.bin",
        &wallet,
        TxParameters::default(),
        StorageConfiguration::default(),
        Configurables::default(),
        Salt::default(),
    )
    .await?;
```
We can improve the UX by using a builder struct `DeployConfig` so the new function with defaults would look like this:
```rust
    let contract_id = Contract::deploy(
        "tests/contracts/configurables/out/debug/configurables.bin",
        &wallet,
        DeployConfig::default(),
    )
    .await?;
```
this would mean that we can remove `deploy_with_parameters` and have `deploy` accept the new config struct.

PR #860, opened 2023-02-24
refactor: contract deployment configuration

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/855

This PR introduces a new builder struct `DeployConfiguration` for contract deployment.  If you want to deploy a contract with the default configuration you can do it like this:

```rust
    &wallet,
    DeployConfiguration::default(),
)
.await?;
```
Alternatively, you can set `TxParameters`, `StorageConfiguration`, `Configurables` and `Salt` like this:

```rust
abigen!(Contract(
    name = "MyContract",
    abi = "packages/fuels/tests/contracts/configurables/out/debug/configurables-abi.json"
));

let tx_parameters = TxParameters::default()
    .set_gas_price(0)
    .set_gas_limit(1_000_000)
    .set_maturity(0);

let key = Bytes32::from([1u8; 32]);
let value = Bytes32::from([2u8; 32]);
let storage_slot = StorageSlot::new(key, value);
let storage_configuration =
    StorageConfiguration::default().set_manual_storage(vec![storage_slot]);

let configurables = MyContractConfigurables::new()
    .set_STR_4("FUEL".try_into()?)
    .set_U8(42u8);

let rng = &mut StdRng::seed_from_u64(2322u64);
let salt: [u8; 32] = rng.gen();

    &wallet,
    DeployConfiguration::default()
        .set_tx_parameters(tx_parameters)
        .set_storage_configuration(storage_configuration)
        .set_configurables(configurables)
        .set_salt(salt),
)
.await?;
```

I have refactored the configuration of `TxParameters`, `StorageConfiguration` to have a consistent way of setting the configurations. Additionally, I refactored the deployment code. Now we only have one `deploy` function and I removed `deploy_with_parameters`. I have also updated the documentation.


docs/src/contracts/configurable-constants.md
```diff
@@ -5,7 +5,7 @@ In Sway, you can define `configurable` constants which can be changed during the
 ```rust,ignore
 {{#include ../../../packages/fuels/tests/contracts/configurables/src/main.sw}}
 ```
-Each of the configurable constants will get a dedicated `set` method in the SDK. For example, the constant `STR_4` will get the `set_str_4` method which accepts the same types as defined in sway. Below is an example where we chain several `set` methods and deploy the contract with the new constants.
+Each of the configurable constants will get a dedicated `set` method in the SDK. For example, the constant `STR_4` will get the `set_STR_4` method which accepts the same types as defined in sway. Below is an example where we chain several `set` methods and deploy the contract with the new constants.
```
MujkicA, line 8: IIRC mention of sway is discouraged since we may support different languages that compile to the vm representation.

```suggestion
Each of the configurable constants will get a dedicated `set` method in the SDK. For example, the constant `STR_4` will get the `set_STR_4` method which accepts the same types as defined in the contract code. Below is an example where we chain several `set` methods and deploy the contract with the new constants.
```

docs/src/getting-started/contracts.md
```diff
@@ -10,12 +10,9 @@ Below is how you can deploy your contracts using the SDK. For more details about
...
 ### The deploy functions
 
-There are two intended ways to deploy a contract
+The `Contract::deploy` function is used to deploy a contract binary to the blockhain. To configure the deployment you can use the `DeployConfiguration` builder.
 
-- `deploy`
-- `deploy_with_parameters`
-
-If you are only interested in a single instance of your contract, then use `deploy`
+If you are only interested in a single instance of your contract, then use the default configuration: `DeployConfiguration::default()`
```
MujkicA, line 15:
```suggestion
If you are only interested in a single instance of your contract, use the default configuration: `DeployConfiguration::default()`
```

Salka1988 (approved):
:+1: nice

segfault-magnet (commented):
Good stuff! I'm not sure about the `with_*` approach though:

The changed structs have no problem having their fields be public. They expose getters and have no logic in their 'setters' (the with_* methods). Nor do they have any other sort of information hiding.

I'm thinking that we might be better off making the fields public. Then if we want to tweak, for example tx_parameters, we do:
```rust
    DeployConfiguration {
        tx_parameters,
        ..Default::default()
    };
```
instead of 
```rust
    DeployConfiguration::new().with_tx_parameters(tx_parameters);
```
I dislike the `new` usage since it doesn't scream default for me. But you kinda should have the `new` to not surprise ppl.

The first example will get rid of `new()` and will force the explicit mention of `default` if somebody wants defaults for the other fields.

Even with only one field the number of characters, eyeballing it, seems to be matched between the two approaches.

When two or more are concerned the ergonomics are in favor of the pub-fields approach:
```rust
    DeployConfiguration {
        tx_parameters,
        storage,
        configurables,
        salt,
    };
```
vs
```rust
    DeployConfiguration::new()
        .with_tx_parameters(tx_parameters)
        .with_storage_configuration(storage)
        .with_configurables(configurables)
        .with_salt(salt);
```

We could then get rid of the getters/setters/macros supporting this. 

On the other hand, the current approach would allow us to change how we actually store the configuration inside. But then again we don't know if we'll ever reap that benefit.

packages/fuels-macros/src/setup_contract_test/code_gen.rs
```diff
@@ -122,15 +119,16 @@ fn contract_deploying_code(
...
-                        )),
-                        Configurables::default(),
-                        Salt::from([#(#salt),*]),
+                        DeployConfiguration::new()
+                            .with_storage_configuration(
+                                StorageConfiguration::new().with_storage_path(
+                                    #storage_path.to_string()
+                                )
+                            )
+                            .with_salt(Salt::from([#(#salt),*]))
```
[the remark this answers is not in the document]
hal3e: Yes but that would bring them into scope and now we just expose the absolute minimum we need

packages/fuels-programs/src/contract.rs
```diff
@@ -74,10 +175,10 @@ impl Contract {
         }
     }
 
-    pub fn compute_contract_id_and_state_root(
+    fn compute_contract_id_and_state_root(
```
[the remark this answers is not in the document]
hal3e: Good point I will have a look!

packages/fuels-programs/src/contract.rs
```diff
@@ -245,115 +308,93 @@ impl Contract {
...
+            ));
+        }
+
+        Ok(())
     }
 
     /// Crafts a transaction used to deploy a contract
-    pub async fn contract_deployment_transaction(
-        compiled_contract: &CompiledContract,
+    fn contract_deployment_transaction(
```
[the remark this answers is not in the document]
hal3e: Good point I will have a look!

packages/fuels-macros/src/setup_contract_test/code_gen.rs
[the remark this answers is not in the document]
segfault-magnet: Oh, sry. Didn't notice that this was the setup contract macro code. Yep, leave it like this.

hal3e:
> Good stuff! I'm not sure about the `with_*` approach though:
> 
> The changed structs have no problem having their fields be public. They expose getters and have no logic in their 'setters' (the with_* methods). Nor do they have any other sort of information hiding.
> 
> I'm thinking that we might be better off making the fields public. Then if we want to tweak, for example tx_parameters, we do:
> 
> ```rust
>     DeployConfiguration {
>         tx_parameters,
>         ..Default::default()
>     };
> ```
> 
> instead of
> 
> ```rust
>     DeployConfiguration::new().with_tx_parameters(tx_parameters);
> ```
> 
> I dislike the `new` usage since it doesn't scream default for me. But you kinda should have the `new` to not surprise ppl.
> 
> The first example will get rid of `new()` and will force the explicit mention of `default` if somebody wants defaults for the other fields.
> 
> Even with only one field the number of characters, eyeballing it, seems to be matched between the two approaches.
> 
> When two or more are concerned the ergonomics are in favor of the pub-fields approach:
> 
> ```rust
>     DeployConfiguration {
>         tx_parameters,
>         storage,
>         configurables,
>         salt,
>     };
> ```
> 
> vs
> 
> ```rust
>     DeployConfiguration::new()
>         .with_tx_parameters(tx_parameters)
>         .with_storage_configuration(storage)
>         .with_configurables(configurables)
>         .with_salt(salt);
> ```
> 
> We could then get rid of the getters/setters/macros supporting this.
> 
> On the other hand, the current approach would allow us to change how we actually store the configuration inside. But then again we don't know if we'll ever reap that benefit.

I like your reasoning and I will change it to use this approach! (The best code is no code :D )

hal3e:
Team @FuelLabs/sdk-rust I did the changes that @segfault-magnet recommended; however now that I have seen it I am not that convinced. We had a lot of instances where we had a one-liner and now we have to intialize the whole struct. For example: 
```rust
         let response = contract_instance_2
             .methods()
             .tx_params(TxParameters::new().with_gas_limit(1_000_000))
             .await?;
```
```rust
         let response = contract_instance_2
             .methods()
             .tx_params(TxParameters {
                 gas_limit: 1_000_000,
                 ..Default::default()
             })
             .await?;
```
or 

```rust
        let tx_params = TxParameters::new().with_gas_limit(10_000);
        let call_params = CallParameters::new().with_gas_forwarded(4300);
```

```rust
        let tx_params = TxParameters {
            gas_limit: 10_000,
            ..Default::default()
        };
        let call_params = CallParameters {
            gas_forwarded: Some(4300),
            ..Default::default()
        };
```
or

```rust
         &wallet,
         DeployConfiguration::new().with_configurables(configurables.into()),
     )
     .await?;
```

```rust
         &wallet,
         DeployConfiguration {
             configurables: configurables.into(),
             ..Default::default()
         },
     )
     .await?;
```

Which version should we use ? Also we could get rid of the `::new()` method and use `::default()` instead. For example:

```rust
        let tx_params = TxParameters::default().with_gas_limit(10_000);
        let call_params = CallParameters::default().with_gas_forwarded(4300);
```
However, that could be confusing for users as @segfault-magnet  said.

MujkicA:
I'm inclined to side with @segfault-magnet s suggestion for two reasons:
 
- One-liners like this are not great for readability anyways IMO. A separate variable would look better.

- Being more explicit about defaults makes sense

That being said, a possible hybrid solution would be to take your `::default()` suggestion and also have `::new()` which accepts all the struct fields.

Another thing I'm unsure about is the use of `with_fieldname()` vs `fieldname()`. The rest of our API usually provides `with_*` methods to return a copy of the instance with a change to that one field. But in this case, it's just a setter.

packages/fuels-programs/src/contract.rs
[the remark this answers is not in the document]
hal3e: Not used outside

packages/fuels-programs/src/contract.rs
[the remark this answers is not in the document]
hal3e: Used in the bridge projects! Good catch!

digorithm:
> I'm inclined to side with @segfault-magnet s suggestion for two reasons:
> 
> * One-liners like this are not great for readability anyways IMO. A separate variable would look better.
> * Being more explicit about defaults makes sense
> 
> A possible hybrid solution would be to take your `::default()` suggestion and also have `::new()` which accepts all the struct fields.
> 
> Another thing I'm unsure about is the use of `with_fieldname()` vs `fieldname()`. The rest of our API usually provides `with_*` methods to return a copy of the instance with a change to that one field. But in this case, it's just a setter.

Yeah, `.with_*()` makes me think it'll return an instance of that thing. 

> Being more explicit about defaults makes sense

Yup. I sleep better at night if I know we're making people think about the default options. IMO the path of least surprise.

segfault-magnet:
> A possible hybrid solution would be to take your `::default()` suggestion and also have `::new()` which accepts all the struct fields.

Another reason in favor of the hybrid might be:

I've seen the `configurables.into()` being written a lot. If you leave the setters approach you can write the setter for the configurables to go something like this:
```rust
pub fn set_configurables(&mut self, configurable: impl Into<Configurables>) -> &mut Self {
self.configurables = configurable.into()
} 
```
later on
```rust
let configurables: MyContractConfigurables = ...; 

config.set_configurables(configurables);
```
and thus move the `.into()` inside the `set_configurables` and have the UX be nicer.

digorithm:
> > A possible hybrid solution would be to take your `::default()` suggestion and also have `::new()` which accepts all the struct fields.
> 
> Another reason in favor of the hybrid setters might be:
> 
> I've seen the `configurables.into()` being written a lot. If you leave the setters approach you can write the setter for the configurables to go something like this:
> 
> ```rust
> pub fn set_configurables(&mut self, configurable: impl Into<Configurables>) -> &mut Self {
>    self.configurables = configurable.into()
>    &mut self
> } 
> ```
> 
> later on
> 
> ```rust
> let configurables: MyContractConfigurables = ...; 
> 
> config.set_configurables(configurables);
> ```
> 
> and thus move the `.into()` inside the `set_configurables` and have the UX be nicer.

That's really neat!

hal3e:
@segfault-magnet  you convinced me to redo my work and add the `..Default::default()` and now you say we should use the setters again :smile:  

I like the setters more even though it is more code. I also like the one-liners. I do not think they are less verbose. You literally say what you do - `set_gas_limit(10)` (we agreed that `with` does not make much sense). That being said, I am ok with both approaches and would like to do what the majority of the teams wants :D So @FuelLabs/sdk-rust lets just agree on one approach and stick to it because **I like the most** when we have one way and use it throughout our sdk. :smile:

segfault-magnet:
> @segfault-magnet  you convinced me to redo my work and add the `..Default::default()` and now you say we should use the setters again :smile:  

No no :D just trying to objective

hal3e:
I have reverted the code to the setters approach and I have added the `impl Into<>` arguments for `salt` and `configurables`.

I have also looked around and people usually set configurations either with
```rust
        let tx_params = TxParameters {
            gas_limit: 10_000,
            ..Default::default()
        };
```
or with
```rust
        let tx_params = TxParameters::new().with_gas_limit(10_000);
```
where `new()` is empty. `TxParameters::default().with_gas_limit(10_000);` is AFAIK not used. Additinally, if you want to set all fields people use `new_from(...)`. So I have added, for all configurations, a `new_from` method that accepts all arguments. For `salt` and `configurables` it is `impl Into<>`

packages/fuels-programs/src/contract.rs
```diff
@@ -49,12 +100,91 @@ pub trait SettableContract {
...
+#[derive(Debug, Clone, Default)]
+pub struct DeployConfiguration {
+    tx_parameters: TxParameters,
+    storage: StorageConfiguration,
+    configurables: Configurables,
+    salt: Salt,
+}
+
+impl DeployConfiguration {
+    pub fn new() -> Self {
```
MujkicA, line 151: IMO it would make more sense for `new()` to be the current `new_from()` and let default be `default()`.
Using `new()` to init a default seems obscure.

MujkicA: I missed the above comment, but I think this still stands. Just because `default()` is not used in the wild doesn't mean we shouldn't encourage it.

hal3e: It is very common to have `new()` without arguments making a default struct. So much so, that clippy has a warning for it https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default

segfault-magnet: I agree with @MujkicA here.

I.e. `new` takes all the struct args and `default` provides sensible defaults.

digorithm: +1

This is relatively minor, IMO. Still, a nice touch 😄

hal3e:
The final version: 
  - `new` with all feilds as arguments
  - a `default` implementation
  - setters and getters (with impl where needed).

packages/fuels-signers/src/lib.rs
```diff
@@ -145,179 +138,4 @@ mod tests {
         Ok(())
         // ANCHOR_END: sign_tx
     }
-
-    #[tokio::test]
-    async fn test_wallet_get_coins() -> fuels_types::errors::Result<()> {
-        const AMOUNT: u64 = 1000;
-        const NUM_COINS: u64 = 3;
-        let mut wallet = WalletUnlocked::new_random(None);
```
segfault-magnet: What happened to these tests? Couldn't find where they were moved. Are they redundant?

segfault-magnet: My bad. Managed to look at a stale version.

segfault-magnet (approved):

segfault-magnet (approved):

digorithm (approved):

Issue #863, opened 2023-02-28
Don't panic in `ParamType::try_from_type_application`


segfault-magnet:
The implementation will panic if an invalid type lookup table is provided. It would seem that the manual creation of this lookup table is a thing and it shouldn't be a panic-causing error to be given an incomplete table.

Change all panics into errors with descriptive messages.

PR #870, opened 2023-03-06
feat: return result from `try_from_type_application`

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/863

`try_from_type_application` will not panic anymore. Instead we propagate an `InvalidData` error.


Br1ght0ne (approved):
Love it!

Salka1988 (approved):
:+1:

digorithm (approved):
💯 

Shouldn't this be a breaking change, though? the current `from` is public, and we're turning it into a `try_from`.

Issue #791, opened 2023-01-20
Name collision on types of different contracts 


supiket:
Example scenario:

_Contract 1_ [imports an enum `AccessError`](https://github.com/FuelLabs/sway-applications/blob/supiket/bump/NFT/project/NFT-contract/src/main.sw#L8). 
This contract [imports _contract 2_](https://github.com/FuelLabs/sway-applications/blob/supiket/bump/NFT/project/NFT-contract/src/main.sw#L26) which in turn [imports a different enum with the same name](https://github.com/FuelLabs/sway-libs/blob/master/libs/nft/src/nft_core.sw#L7).

This results in the following error when using `abigen!` with _contract 1_:
```
the name `AccessError` is defined multiple times
`AccessError` must be defined only once in the type namespace of this module
```

segfault-magnet:
Blocked by https://github.com/FuelLabs/sway/issues/4235

PR #872, opened 2023-03-07
feat: type path support (`forc` flag `--json-abi-with-callpaths`)

segfault-magnet:
closes: #791 

Type paths are currently feature-gated behind the `forc` flag `--json-abi-with-callpaths`. Docs were updated to guide those affected by #791 to build their contracts with the flag above.

The SDK remains compatible with programs compiled with or without the flag. The CI is updated to protect against regressions in this regard. Once this becomes the default in `forc` the backward compatibility will be dropped.

Shared types handling was updated to be less confusing to the user. Shared types are still generated inside `shared_types` but are now reexported to each place where they were originally taken from.

 I.e. types will be shared transparently. If a naming collision happens the user doesn't need to think whether the type was shared but rather just use the original path of the type.


docs/src/abigen/the-abigen-macro.md
```diff
@@ -62,6 +62,22 @@ Finally, `pub use` statements are inserted, so you don't have to fully qualify t
 > **Note:**
 > It is **highly** encouraged that you generate all your bindings in one `abigen!` call. Doing it in this manner will allow type sharing and avoid name collisions you'd normally get when calling `abigen!` multiple times inside the same namespace. If you choose to proceed otherwise, keep in mind the generated code overview presented above and appropriately separate the `abigen!` calls into different modules to resolve the collision.
 
+### Type paths
+Normally when using types from libraries in your contract, script or predicate, they'll be generated directly under the main `mod` of your program bindings, i.e. a type in a contract binding `MyContract` imported from a library `some_library` would be generated under `abigen_bindings::my_contract_mod::SomeLibraryType`.
+
+This can cause problems if you happen to have two types with the same name in different libraries of your program.
+
+This behavior can be changed to include the library path by compiling with:
```
digorithm, line 70: Adding an explicit hint that "compiling" means the Sway code, not the Rust code.

```suggestion
This behavior can be changed to include the library path by compiling your Sway project with the following:
```

docs/src/abigen/the-abigen-macro.md
```diff
@@ -62,6 +62,22 @@ Finally, `pub use` statements are inserted, so you don't have to fully qualify t
...
+```
+forc build --json-abi-with-callpaths
+```
+
+Now the type from the previous example will be generated under `abigen_bindings::my_contract_mod::some_library::SomeLibraryType`.
+
+This might only become relevant to you if your type doesn't get reexported. This can happen, as explained previously, if your type does not have a unique name across all bindings inside one `abigen!` call. You'll then need to fully qualify the access to it.
```
digorithm, line 77: Some clarity nits

```suggestion
This might only become relevant if your type isn't reexported. This can happen, as explained previously if your type does not have a unique name across all bindings inside one `abigen!` call. You'll then need to fully qualify the access to it.
```

examples/rust_bindings/src/rust_bindings_formatted.rs
```diff
@@ -17,15 +23,16 @@ pub mod abigen_bindings {
...
                 let provider = self.wallet.get_provider()?;
                 wallet.set_provider(provider.clone());
                 Ok(Self {
                     contract_id: self.contract_id.clone(),
                     wallet,
+                    log_decoder: self.log_decoder.clone(),
                 })
             }
-            pub async fn get_balances(&self) -> Result<HashMap<String, u64>, Error> {
+            pub async fn get_balances(&self) -> Result<HashMap<String, u64>> {
```
digorithm, line 35: A quick reminder to change the title from `feat` to `feat!` and add the breaking label because this is a breaking change!

segfault-magnet: It does look like a breaking change but this is just the example file of abigen expanded.

The actual breaking changes have long since been merged in master, it is only now that we're updating our docs to show how the generated code might look atm.

packages/fuels-code-gen/src/program_bindings/abigen/logs.rs
```diff
@@ -1,20 +1,15 @@
...
+    abi_types::FullLoggedType, resolved_type::TypeResolver, utils::single_param_type_call,
 };
 
 pub(crate) fn logs_lookup_instantiation_code(
     contract_id: Option<TokenStream>,
     logged_types: &[FullLoggedType],
-    shared_types: &HashSet<FullTypeDeclaration>,
 ) -> TokenStream {
-    let resolved_logs = resolve_logs(logged_types, shared_types);
+    let resolved_logs = resolve_logs(logged_types);
```
digorithm, line 12: I know this isn't the main point of the PR, but I love that you've removed the need to pass types lookup arg all over the codebase. Super clean!

packages/fuels-code-gen/src/program_bindings/resolved_type.rs
```diff
@@ -60,229 +53,217 @@ impl ToTokens for ResolvedType {
...
-    type_application: &FullTypeApplication,
-    shared_types: &HashSet<FullTypeDeclaration>,
-) -> Result<ResolvedType> {
-    let recursively_resolve = |type_applications: &Vec<FullTypeApplication>| {
+pub(crate) struct TypeResolver {
+    relative_to_mod: TypePath,
+}
+
+impl TypeResolver {
+    pub(crate) fn new() -> Self {
```
hal3e, line 68: where you not advertising to use `new` with all fields as arguments and `default()` if you need the default values :)

hal3e (commented):
Beautiful as always! Left some small nits...

hal3e (approved):
Great work! This looks so good!

packages/fuels-code-gen/src/program_bindings/generated_code.rs
```diff
@@ -74,84 +147,199 @@ mod tests {
...
+            pub use some_mod::Unique;
+        };
+
+        assert_eq!(
+            use_statements.to_string(),
+            expected_use_statements.to_string()
+        );
+    }
+
+    fn given_some_struct_code(struct_name: &str) -> GeneratedCode {
```
digorithm, line 335: These tests are amazing!

packages/fuels-test-helpers/src/node.rs
```diff
@@ -281,10 +281,11 @@ pub async fn new_fuel_node(
 pub async fn server_health_check(client: &FuelClient) {
     let mut attempts = 5;
     let mut healthy = client.health().await.unwrap_or(false);
+    let between_attempts = Duration::from_millis(300);
```
digorithm, line 284: Any reason we're increasing this to 300?

Also, unrelated, but: we should pull this from an env var.

packages/fuels/Cargo.toml
```diff
@@ -37,4 +37,5 @@ std = [
     "fuels-test-helpers/std",
     "fuels-types/std",
 ]
+test-type-paths = []
```
digorithm, line 40: Could you remind me again why we need this?

packages/fuels-test-helpers/src/node.rs
segfault-magnet: Previous retry period was too rapid. It seems the node takes longer to become healthy. 300 is just empirically determined to produce repeatable ci runs.

I'll create a task to make it configurable tomorrow.

packages/fuels/Cargo.toml
segfault-magnet: The e2e tests for type paths don't compile unless the sway projects used in them are themselves compiled with the type path forc feature flag.

Using #ignore on the test doesn't help since they still get compiled.

Only practical way was to signal via this flag when the sway projects are compiled with the type paths flag.

So, when our ci uses all features it will use this one as well and test type paths taking care to download the correct sway artifacts.

When testing without the feature the test is going to get disabled and the normal artifacts are going to get downloaded.

This way we should catch regressions.

I couldn't find a flag that will be usable only when running tests. 

It has no effect on the sdk code.

digorithm: Ah, that makes sense. Then this will be used only while the type path feature is feature-flagged in `forc`, right? We should add a `TODO` comment here so we can easily find and remove this once the flag is gone!

packages/fuels/tests/bindings.rs
```diff
@@ -926,3 +926,42 @@ async fn shared_types() -> Result<()> {
...
+        Deploy(
+            name = "contract_a_instance",
+            contract = "ContractA",
+            wallet = "wallet"
+        ),
+    );
+
+    {
+        let contract_a_type =
+            abigen_bindings::contract_a_mod::contract_a_types::VeryCommonNameStruct {
```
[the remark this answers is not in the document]
segfault-magnet: Ideally, and arguably most often, the users' structs will have unique names across all the participants of the `abigen!` call. As such, they will be reexported (`pub use`-ed) and the user will refer to the types by just their names.

Less ideal is the example above where `VeryCommonNameStruct` appears multiple times prohibiting its reexporting. The user must then read the `abigen` titled docs section where this is explained.

There is also a section about enabling forc type paths and how the generated type path will need to look in case the user actually needs to fully qualify the type.

docs/src/abigen/the-abigen-macro.md
```diff
@@ -62,6 +62,22 @@ Finally, `pub use` statements are inserted, so you don't have to fully qualify t
...
+This can cause problems if you happen to have two types with the same name in different libraries of your program.
+
+This behavior can be changed to include the library path by compiling your Sway project with the following:
+```
+forc build --json-abi-with-callpaths
+```
+
+Now the type from the previous example will be generated under `abigen_bindings::my_contract_mod::some_library::SomeLibraryType`.
+
+This might only become relevant if your type isn't reexported. This can happen, as explained previously, if your type does not have a unique name across all bindings inside one `abigen!` call. You'll then need to fully qualify the access to it.
```
segfault-magnet, line 77: @digorithm the part about reexporting with the full type path example just above.

digorithm (approved):
Massive and clean work, thanks for the (many) refactors along the way.

hal3e (dismissed):

digorithm (approved):

Issue #886, opened 2023-03-15
Use cargo sparse protocol to speed up the CI


segfault-magnet:
The sparse protocol is now stable in Rust 1.68.0.
https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#whats-in-1680-stable

Currently, you have to opt-in by configuring the cargo config. It will become default by 1.70.0.

We can opt-in before that so that we may reap the speed-up benefits.

PR #888, opened 2023-03-15
enable cargo sparse protocol

segfault-magnet:
closes: #886 
This should speed up our CI.

### Checklist

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,3 @@
+# Remove once https://github.com/FuelLabs/fuels-rs/issues/887 is unblocked
+[registries.crates-io]
+protocol = "sparse"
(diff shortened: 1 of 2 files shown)


segfault-magnet:
Turns out dtolnay/rust-toolchain does this already by checking if the rust version is compatible. So the CI will use the sparse protocol automatically because we've bumped the rust version to 1.68.0. 

So, CI-wise, the `.cargo/config.toml` is not needed. But I'm thinking about whether to remove it because it will apply to anyone building the project locally. But then again we all can just set this config in our `$HOME/.cargo/config.toml`. 

@FuelLabs/sdk-rust thoughts?

hal3e:
I would not add it. SDK users can set it up locally if they want and it will become a default anyway at some point.

segfault-magnet:
> SDK users can set it up locally if they want 

SDK devs *, this doesn't impact users.

I wouldn't add it as well but figured I better ask.

We still need to bump the rust version for it to work in CI though

hal3e (approved):
Thanks for this! Please do not forget to update the commit message

segfault-magnet:
> Please do not forget to update the commit message
When have I ever :D

Issue #892, opened 2023-03-17
improve UX for getting `contract_id` from binary


hal3e:
Currently users use the `Contract::load_contract` function to create a `CompiledContract` and then they use the `Contract::compute_contract_id_and_state_root` to get the `contract_id`.

PR #899, opened 2023-03-22
refactor!: separate `Contract` loading and deploying 

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/892

__Updated PR description__

After taking in consideration all comments made in this PR, I have decided to do a bit more refactoring on `Contract`. I have tried to move functions not related directly to `Contract` to places where they make more sense. I have:
- Moved `contract_deployment_transaction` to `CreateTransactionBuilder::prepare_contract_deployment`
- Moved `get_storage_slots` and its logic to `StorageConfiguration` - No need for `Contract` to handle this
- Moved `validate_path_and_extension` outside as it is now also used by `StorageConfiguration`
- Moved `method_hash` and `should_compute_custom_input_offset` - No need for `Contract` to handle this
- Deleted `deploy` and renamed `deploy_loaded` to `deploy` - big breaking change will elaborate below
Other refactors:
- Made `DeployConfiguration` into `LoadConfiguration` and took out `TxParametres`
  - The idea is to separate loading and deploying of the contracts. Now the user will add the `TxParameters` when he actually deploy the contract

The biggest change for the user is the way to deploy contracts. Previously the user would use `deploy` directly: 

```rust
        &wallet,
        DeployConfiguration::default(),
    )
    .await?;
```
This function basically did two things: Load and then deploy the contract. I propose we use this instead:

```rust
    let contract_id = Contract::load_from(
        LoadConfiguration::default(),
    )?
    .deploy(&wallet, TxParameters::default())
    .await?;
```

This makes it clear what is being done. Makes the `LoadConfiguration` simpler and also now the user can get the `contract_id` or `state_root` easily.

```rust
    let contract_id = Contract::load_from(
        LoadConfiguration::default(),
    )?
    .contract_id();
```

For reference this is what the user had to do previously to get the `contract_id`

```rust
    let compiled_contract = Contract::load_contract(
        contract_binary_path,
        DeployConfiguration::default(),
    )?;
    let (test_contract_id, _) = Contract::compute_contract_id_and_state_root(&compiled_contract);
```
thanks @segfault-magnet for your suggestions and helping me move code around :smile: 

### Checklist


segfault-magnet (dismissed):
Left a suggestion. 

Love it.

packages/fuels-programs/src/contract.rs
```diff
@@ -326,21 +342,28 @@ impl Contract {
...
 
-        Ok(CompiledContract {
+        Ok(Self {
             binary,
-            salt: configuration.salt,
+            salt,
             storage_slots,
+            contract_id,
+            state_root,
         })
```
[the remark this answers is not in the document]
hal3e: Great advice!

packages/fuels-programs/src/contract.rs
```diff
@@ -175,17 +167,46 @@ impl DeployConfiguration {
 /// compiling, deploying, and running transactions against a contract.
 /// The contract has a wallet attribute, used to pay for transactions and sign them.
 /// It allows doing calls without passing a wallet/signer each time.
```
iqdecay, line 169:
```suggestion
/// It allows doing calls without passing a wallet/signer each time.
#[derive(Debug, Clone, Default)]
```

packages/fuels-programs/src/contract.rs
```diff
@@ -175,17 +167,46 @@ impl DeployConfiguration {
...
+
+    pub fn contract_id(&self) -> ContractId {
+        self.contract_id
+    }
+
+    pub fn state_root(&self) -> Bytes32 {
+        self.state_root
+    }
+
+    fn compute_contract_id_and_state_root(
```
iqdecay, line 200: And this would be a method of `CompiledContract`. I do think we can find a better name.

iqdecay (dismissed):
I think the approach of making `CompiledContract` a good type would improve UX vastly as you did, without loosing semantic richness

hal3e:
> I think the approach of making `CompiledContract` a good type would improve UX vastly as you did, without loosing semantic richness

I did consider that approach. However, `Contract` would not have to hold `CompiledContract` as it is never used. We would just pass `CompiledContract` around. (This is currently the case on master). My reasoning was: Why have `CompliedContract`as a separate struct when our `Contract` can hold the info and we can turn some of the associated functions to methods.

packages/fuels-programs/src/contract.rs
MujkicA: `Contract` doesn't have a wallet attribute

hal3e: Good catch! That is a leftover from the account PR.

packages/fuels-programs/src/contract.rs
```diff
@@ -175,17 +167,46 @@ impl DeployConfiguration {
...
+    salt: Salt,
+    storage_slots: Vec<StorageSlot>,
+    contract_id: ContractId,
+    state_root: Bytes32,
+}
 
 impl Contract {
-    pub fn compute_contract_id_and_state_root(
-        compiled_contract: &CompiledContract,
+    pub fn new(binary: Vec<u8>, salt: Salt, storage_slots: Vec<StorageSlot>) -> Self {
```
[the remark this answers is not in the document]
hal3e: An option would be to make `deploy_loaded` public. Then we could do `Contract::new(...).deploy_loaded(...)`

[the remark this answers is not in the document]
iqdecay: Can we please do this in another PR? This PR is supposed to be here just for the `get_contract_id` UX.

iqdecay:
:pray: :pray: please keep this PR down to a single change. A PR should only fix one thing or introduce one feature, not do two things at the same time.

packages/fuels-programs/src/contract.rs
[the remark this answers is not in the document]
MujkicA: The UX for `get_contract_id ` already leads to a breaking change. I'm leaning towards resolving such things immediately instead of accumulating breaking changes to the same part of our UI in the pipeline.

MujkicA:
> After doing some testing, I realize you are right, it does feel like we're just passing things around. I withdraw my remark, but I feel like my confusion arises from the fact that we are not distinguishing between a Contract deployed at block h and the same Contract deployed at block h+10. They are different objects but have the same base (bytecode + storage slot and salt). I feel like with our current abstraction we don't differentiate between those two objects.

I'm not sure I follow here. From my perspective, they are the same object, a deployed contract. The differentiating factor is only the `salt` which leads to a different `contract_id` being computed.

iqdecay:
> I'm not sure I follow here. From my perspective, they are the same object (albeit not instance), a deployed contract. The difference is only the salt which leads to a different contract_id being computed.

They are not the same "blockchain" objects since they live at two different addresses for me. For instance if I sent fund to `ContractA` which was deployed at block height `h`, I'm not going to have funds in the `ContractB` deployed at block height `h+1000`. To me this shows that they are two different objects, not just two instances of the same object. That's why you need upgrade/proxy functions in the blockchain.

hal3e:
@FuelLabs/sdk-rust I have updated the PR description to reflect the latest changes I have made. Can you please have a look and let me know if you like the new approach.

MujkicA (approved):
The new changes make sense, thanks for revising it

packages/fuels-programs/src/contract.rs
```diff
@@ -173,121 +173,67 @@ impl DeployConfiguration {
 
 /// [`Contract`] is a struct to interface with a contract. That includes things such as
 /// compiling, deploying, and running transactions against a contract.
-/// The contract has a wallet attribute, used to pay for transactions and sign them.
-/// It allows doing calls without passing a wallet/signer each time.
-pub struct Contract;
+#[derive(Debug)]
+pub struct Contract {
+    binary: Vec<u8>,
+    salt: Salt,
```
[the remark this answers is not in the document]
hal3e: That is a very good idea! I will add it.

docs/src/getting-started/contracts.md
```diff
@@ -8,11 +8,7 @@ Once you've written a contract in Sway and compiled it with `forc build` (read [
...
 
-### The deploy functions
-
-The `Contract::deploy` function is used to deploy a contract binary to the blockhain. To configure the deployment you can use the `DeployConfiguration` struct.
-
-If you are only interested in a single instance of your contract, use the default configuration: `DeployConfiguration::default()`
+First, the `Contract::load_from` function is used to load a contract binary with a `LoadConfiguration`. If you are only interested in a single instance of your contract, use the default configuration: `LoadConfiguration::default()`. After the contract binary is loaded you can use the `deploy()` method to deploy the contract to the blockchain.
 
 ```rust,ignore
 {{#include ../../../examples/contracts/src/lib.rs:deploy_contract}}
```
MujkicA, line 14: This snippet includes a call to abigen which isn't relevant to what we are describing here. I would move the abigen call into this anchor `use_deployed_contract` since the binding is being used there.
Then just modify the following sentence in docs 

> You can then use the contract methods very simply:

to not say simply and mention abigen.

docs/src/getting-started/contracts.md
```diff
@@ -24,7 +20,7 @@ You can then use the contract methods very simply:
 {{#include ../../../examples/contracts/src/lib.rs:use_deployed_contract}}
 ```
 
-Alternatively, you can use `DeployConfiguration` to configure the deployment. You can, for example, deploy multiple instances of the same contract, change the storage or `tx_parameters` or update the binary using `configurables`.
+Alternatively, you can use `LoadConfiguration` to configure how the contract is loaded. You can, for example, load the same contract with different `Salt`, change the contract's storage or update the binary using `configurables`.
```
MujkicA, line 23: I'm unsure if we should mention what salt does somewhere or if we rely on upstream docs to convey this. i.e. it leads to a different contract-id being computed, enabling you to deploy the same contract multiple times.

hal3e: Good point. I will mention it here

MujkicA (dismissed):

packages/fuels-programs/src/contract.rs
```diff
@@ -882,6 +756,83 @@ where
...
+/// It returns a prepared [`ContractCall`] that can further be used to
+/// make the actual transaction.
+/// This method is the underlying implementation of the functions
+/// generated from an ABI JSON spec, i.e, this is what's generated:
+///
+/// ```ignore
+/// quote! {
+///     #doc
+///     pub fn #name(&self #input) -> #result {
+///         Contract::method_hash(#tokenized_signature, #arg)
```
segfault-magnet, line 770:
```suggestion
///         contract::method_hash(#tokenized_signature, #arg)
```

segfault-magnet (dismissed):

segfault-magnet (dismissed):

MujkicA (approved):

docs/src/getting-started/contracts.md
```diff
@@ -8,26 +8,26 @@ Once you've written a contract in Sway and compiled it with `forc build` (read [
...
-You can then use the contract methods very simply:
+Additionally, you can set custom `TxParameters` when deploying the loaded contract.
 
 ```rust,ignore
-{{#include ../../../examples/contracts/src/lib.rs:use_deployed_contract}}
+{{#include ../../../examples/contracts/src/lib.rs:deploy_with_parameters}}
 ```
 
-Alternatively, you can use `DeployConfiguration` to configure the deployment. You can, for example, deploy multiple instances of the same contract, change the storage or `tx_parameters` or update the binary using `configurables`.
+After the contract was deployed, you can use the contract's methods like this:
```
digorithm, line 28:
```suggestion
After the contract is deployed, you can use the contract's methods like this:
```

digorithm (approved):
This is an amazing UX improvement; the previous way of loading + deploying was getting way too convoluted. Super clean implementation, too; great stuff, @hal3e. 

Approving a minor grammar-related change request, feel free to merge after!

packages/fuels/tests/contracts.rs
```diff
@@ -1071,53 +1071,26 @@ async fn test_add_custom_assets() -> Result<()> {
 }
 
 #[tokio::test]
-async fn test_contract_raw_slice() -> Result<()> {
```
digorithm: I might be confused by the git situation, but why are we removing this raw_slice test? I think this was updated (or removed) in the raw slice as input PR, but not sure why this is being touched in this PR 🤔

hal3e: That was a merge issue. My branch reintroduced that test somehow so I had to remove it. 
The raw_slice PR moved the test to `types_contracts.rs` https://github.com/FuelLabs/fuels-rs/blob/f5d08d67543ee35fedc561c5c8335bb134ec24fd/packages/fuels/tests/types_contracts.rs#L1828

digorithm (approved):
:shipit: awesome work!

Issue #879, opened 2023-03-10
feat: support `Bytes` as argument


iqdecay:
From @segfault-magnet :
Turns out Byte support isn't difficult as well, very similar to vec support but with the difference that the bytes must not be padded to WORD length:
```rust
    fn encode_bytes(data: &[u8]) -> Result<Vec<Data>> {
        let encoded_data = data
            .iter()
            .map(|&byte| Data::Inline(vec![byte]))
            .collect::<Vec<_>>();
        let cap = data.len() as u64;
        let len = data.len() as u64;

        Ok(vec![
            Data::Dynamic(encoded_data),
            Self::encode_u64(cap),
            Self::encode_u64(len),
        ])
    }
```
This is the selector for Bytes: `s(s(rawptr,u64),u64)`, you can hard code it over in resolve_fn_selector.
Its encoding width is 3 WORDS just like the vector. So `compute_encoding_width` needs to be updated.

Also make sure to update `ParamType::try_from_type_application` to detect `Bytes` since that is what the indexer team uses to get their fn selectors.

And you need to add it to `should_compute_custom_input_offset`. 

And that's all you need to nest Bytes however deep you wish inside other types when used as function arguments.

Issue #878, opened 2023-03-10
feat: support `RawSlice` as argument


iqdecay:
Not sure if Sway supports it though. 
From @segfault-magnet :

I tried implementing RawSlice support, turns out it isn't difficult.

A RawSlice is encoded as two u64 ptr and len. So encoding it would look like this:

    fn encode_raw_slice(data: &[u64]) -> Result<Vec<Data>> {
        let tokens = data
            .iter()
            .map(|&word| word.into_token())
            .collect::<Vec<_>>();
        let encoded_data = Self::encode_tokens(&tokens)?;
        let len = data.len() as u64 * 8;
        Ok(vec![Data::Dynamic(encoded_data), Self::encode_u64(len)])
    }

the function selector is rawslice.

ParamType::try_from_type_application needs to be updated to detect raw slices. This helper is used by the index team to generate fn selectors in runtime from only the json abi file.

Its encoding width is 2 WORDS, so compute_encoding_width needs to be updated.

Also, you need to add it to should_compute_custom_input_offset.

That's all you need to nest raw slices however deep in other types when used as function arguments.

segfault-magnet:
> Not sure if Sway supports it though. 

I figured out the encoding by doing examples in Sway. So passing in raw slices definately works if you're referring to that.

Didn't try nesting it though in inputs, but I nested it in a tuple and returned it and the compiler didn't complain.

Have a feeling if that passed so will nesting when used in fn args

iqdecay:
Closed by #904.

PR #904, opened 2023-03-23
feat: add support for `Bytes` and `RawSlice` inputs

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/879 and https://github.com/FuelLabs/fuels-rs/issues/878
co-developed with: @segfault-magnet 

`Bytes` and `RawSlice` can be nested arbitrarily deep in custom types. We have added e2e test for `contracts`, `scripts` and `predicates`. 

### Checklist


packages/fuels-core/src/abi_encoder.rs
```diff
@@ -139,6 +132,38 @@ impl ABIEncoder {
...
+        Ok(vec![
+            Data::Dynamic(encoded_data),
+            Self::encode_u64(cap),
+            Self::encode_u64(len as u64),
+        ])
+    }
+
+    fn zeropad_to_word_alignment(data: &mut Vec<u8>) {
+        let padded_length = padded_len_usize(data.len());
+        data.resize(padded_length, 0);
```
iqdecay, line 165: I don't think introducing a function just for this is justified here. The `ABIEncoder` struct only contains `encode_xxx` methods, I don't see why we would add a different method just for this. If the meaning is not clear enough, then a comment can be enough.

packages/fuels-types/src/param_types.rs
```diff
@@ -388,6 +385,18 @@ fn try_vector(the_type: &Type) -> Result<Option<ParamType>> {
     Ok(Some(ParamType::Vector(Box::new(vec_elem_type))))
 }
 
+fn try_bytes(the_type: &Type) -> Result<Option<ParamType>> {
+    if !["struct std::bytes::Bytes", "struct Bytes"].contains(&the_type.type_field.as_str()) {
```
iqdecay, line 389: Here, I don't think we should accept `struct Bytes` as a correct option. The `Bytes` type in Sway canonically means `std::bytes::Bytes`, I do think it would be confusing for the user to be able to declare its own `Bytes` type as a mapping to the `std::bytes::Bytes` type, it's a footgun. That's if I understand correctly of course.

packages/fuels/tests/types/scripts/script_raw_slice/src/main.sw
```diff
@@ -1,11 +1,43 @@
...
+    inner: T,
+    inner_enum: SomeEnum<raw_slice>,
+}
+
+fn validate_raw_slice(input: raw_slice) {
+    let vec: Vec<u64> = Vec::from(input);
+    require(vec.len() == 3, "raw slice len is not 3");
+    require(vec.get(2).unwrap() == 42, "expected 3. slice entry to be 42");
+    require(vec.get(1).unwrap() == 41, "expected 2. slice entry to be 41");
+    require(vec.get(0).unwrap() == 40, "expected 1. slice entry to be 40");
```
iqdecay, line 18:
```suggestion
    require(vec.get(2).unwrap() == 42, "expected 3rd slice entry to be 42");
    require(vec.get(1).unwrap() == 41, "expected 2nd slice entry to be 41");
    require(vec.get(0).unwrap() == 40, "expected 1st slice entry to be 40");
```

iqdecay (changes requested):
LGTM, especially  the tests! I have one concern about the `Bytes` type we accept to encode.

packages/fuels-types/src/param_types.rs
segfault-magnet: We have to accept `Bytes` since that is the only way `std::bytes::Bytes` are going to show up without the new `--json-call-paths` (or something similar) feature flag. 

This is the same way we've done it for vectors before. It does mean that if the user has a `Bytes` structure it will match. But then again if the user had a `Vec` structure it would have matched before.

One way out of that predicament is if we match the full declaration of the type, but that would make this code a bit more complicated.

Another way is to just wait until the call paths thing becomes the default.

packages/fuels-core/src/abi_encoder.rs
iqdecay:
```suggestion
    fn encode_bytes(mut data: Vec<u8>) -> Result<Vec<Data>> {
        let len = data.len();
        
       // zeropad to word alignement
        let padded_length = padded_len_usize(data.len());
        data.resize(padded_length, 0);
        
        / --- snip ---/
```

is what I had in mind. I just don't like that we're introducing a one line function that isn't used anywhere else. I believe that for functions that short, yes at least two call sites should be the minimum. To me this is factorization to the extreme and I don't understand it.

packages/fuels-types/src/param_types.rs
iqdecay: Ok so just to make sure I understand : this is temporary until the `--json-call-paths` flag becomes the default? Then we would not match on this?

packages/fuels-core/src/abi_encoder.rs
segfault-magnet: > To me this is factorization to the extreme
 
 To me, it is about hiding details, however small, and talking on a higher abstraction level. The mention of `padded_len_usize`, `resize` are all noise to what I want to communicate to code readers, and that is: we'll zeropad the bytes until word alignment is reached -- go see how if you're that interested.

> we're breaking the nice "all functions inside ABIEncoder are decoding functions"

It's nice to look at, sure. I wouldn't take it as a goal. We're bound to have a helper sooner or later mixed in.

But I've extracted the zeropad fn as a private function after the encoder. I dislike private associated functions anyways. Now we again have all methods start with 'encode_*'.

packages/fuels-types/src/param_types.rs
segfault-magnet: Once the call paths become default we'll always match on the full `std::bytes::Bytes` path. 

But, as I type this, a refactor is underway by @hal3e which will remove all these `try_*` helpers. So it might end up being removed even sooner.

packages/fuels/tests/types_predicates.rs
```diff
@@ -480,3 +480,107 @@ async fn spend_predicate_coins_messages_generics() -> Result<()> {
...
+            predicate_balance,
+            asset_id,
+            TxParameters::default(),
+        )
+        .await?;
+
+    // The predicate has spent the funds
+    assert_address_balance(predicate.address(), &provider, asset_id, 0).await;
+
+    // Funds were transferred
```
MujkicA, line 524:
```suggestion
```

packages/fuels/tests/types_predicates.rs
```diff
@@ -480,3 +480,107 @@ async fn spend_predicate_coins_messages_generics() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn spend_predicate_coins_messages_bytes() -> Result<()> {
```
MujkicA, line 485: Could we make a helper for these tests and just alternate the data we pass to the predicate?

packages/fuels-types/src/param_types.rs
```diff
@@ -388,6 +385,18 @@ fn try_vector(the_type: &Type) -> Result<Option<ParamType>> {
...
+        return Ok(None);
+    }
+
+    Ok(Some(ParamType::Bytes))
+}
+
+fn try_raw_slice(the_type: &Type) -> Result<Option<ParamType>> {
+    Ok((the_type.type_field == "raw untyped slice").then_some(ParamType::RawSlice))
```
digorithm, line 397: Correct me if I'm missing something: if we're only returning an `Ok` here (and not an `Err`), wouldn't it be better if we removed the need for returning `Result<>`? This way, no need to chain two unwraps as we do down there (`try_raw_slice(&the_type).unwrap().unwrap();`). Or making it `Err` if `the_type.type_field == "raw untyped slice"` evals to false. Either way would simplify the use of it (no need for 2 unwraps)

packages/fuels/tests/types/predicates/predicate_bytes/src/main.sw
```diff
@@ -0,0 +1,37 @@
...
+}
+
+fn main(wrapper: Wrapper<Vec<Bytes>>) -> bool {
+    let mut main_result = if let SomeEnum::Second(enum_bytes) = wrapper.inner_enum {
+        validate_bytes(enum_bytes)
+    } else {
+        false
+    };
+
+    main_result = main_result && wrapper.inner.len() == 2;
```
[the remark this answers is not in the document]
digorithm: Yeah, this is very nit-picky, but it seems to be the consensus that early exits are better for reasoning about the code. (and I agree with this consensus)

digorithm (commented):
Great work and awesome tests! I left just a couple of comments, not really a hard request for changes, so this is essentially an approval.

packages/fuels/tests/types/predicates/predicate_bytes/src/main.sw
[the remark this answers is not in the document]
segfault-magnet: > early exits are better

@hal3e and I agree as well. 

> Is there a downside to just saying "if !condition -> return false" for each each condition?

Not that we're aware of. We mainly did the `main_result` thing due to the `forc-fmt` not breaking lines when doing a single boolean sausage. After looking at it again it does seem like a little neck strain is better than chaining the booleans in this way.

We decided on the sausage over the `if/return` for every check. Check the updated code and resolve this discussion if you like it.

packages/fuels-types/src/param_types.rs
segfault-magnet: Although `try_raw_slice` could drop the `Result`, it has to have the same signature as the other `try_*` `fn`s so that the whole resolving logic might look like this:
```rust
        let matched_param_type = [
            try_primitive,
            try_array,
            try_str,
            try_tuple,
            try_vector,
            try_bytes,
            try_raw_slice,
            try_enum,
            try_struct,
        ]
        .into_iter()
        .map(|fun| fun(the_type))
        .flat_map(|result| result.ok().flatten())
        .next();
```

Some of the other resolving functions can indeed match correctly but fail. For example, `try_array` can determine that the format is indeed that of an array `[_; 10]` but the array has `>1` components which makes it invalid. Saying `None` means _it isn't an array_ which isn't true.

But @hal3e and I have already refactored this and will open up a PR in the next few days. The refactor nuked the whole idea with `try_*` and went with something simpler and more robust.

Feel free to resolve the comment if you're satisfied with the answer.

packages/fuels/tests/types_predicates.rs
segfault-magnet: Nice. Shortened the tests considerably. Resolving.

packages/fuels/tests/types_predicates.rs
MujkicA, line 1: Nice

MujkicA (approved):

digorithm (approved):
Nice -- great work!

PR #916, opened 2023-03-29
feat: add receipt decoder

### LINKED ISSUE #914: Unable to get resoponse of a MultiContractCallHandler
`MultiContractCallHandler` allows multiple calls to the same contract, but unable to properly decode the output – for each call with particular `ReturnLocation` it will always try to decode the first data that matches this return location (because receipts array is is always the same and [get_decoded_output](https://github.com/FuelLabs/fuels-rs/blob/c746b6fe78658f48f97018c6e21851c123ca9fe6/packages/fuels-programs/src/contract.rs#L518) will search by the contract id).

https://github.com/FuelLabs/fuels-rs/blob/c746b6fe78658f48f97018c6e21851c123ca9fe6/packages/fuels-programs/src/contract.rs#L1031-L1036

MujkicA:
Closes #914

This PR moves receipt decoding to the new `ReceiptParser`. The key difference is that the `ReceiptParser` owns a copy of the receipts and will consume an entry when a value is generated. 

I've also found that `test_contract_raw_slice` is duplicated so I removed one instance of it.

### Checklist


packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+            .collect();
+
+        Self {
+            receipts: relevant_receipts,
+        }
+    }
+
+    /// Based on receipts returned by a script transaction, the contract ID (in the case of a contract call),
+    /// and the output param, parse the values and return them as Token.
+    pub fn try_parse_output(
```
segfault-magnet, line 57: I'd call it just `parse` since it is probably going to look like this at the call site:
```rust
receipts_parser.parse(...)?;
```

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+        &mut self,
+        contract_id: Option<&Bech32ContractId>,
+        output_param: &ParamType,
+    ) -> Result<Token> {
+        let null_contract_id = ContractId::new([0u8; 32]);
+        let contract_id: ContractId = match contract_id {
+            Some(contract_id) => contract_id.into(),
+            // During a script execution, the script's contract id is the **null** contract id
+            None => null_contract_id,
+        };
```
segfault-magnet, line 67:
```suggestion
        let contract_id = contract_id.map(Into::into).unwrap_or_else(|| {
            // During a script execution, the script's contract id is the **null** contract id
            ContractId::new([0u8; 32])
        });
```

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+
+        let encoded_value = match output_param.get_return_location() {
```
segfault-magnet, line 69: Maybe to hide away the details of how the raw data gets extracted:
```rust
        let Some(data) = self.extract_raw_data(output_param, &contract_id) else {
            return Err(Self::missing_receipts_error(output_param))
        };

        ABIDecoder::decode_single(output_param, &data)
```
```rust
    fn extract_raw_data(
        &mut self,
        output_param: &ParamType,
        contract_id: &ContractId,
    ) -> Option<Vec<u8>> {
        match output_param.get_return_location() {
            ReturnLocation::ReturnData if output_param.is_vm_heap_type() => {
                self.extract_return_data_heap(contract_id)
            }
            ReturnLocation::ReturnData => self.extract_return_data(contract_id),
            ReturnLocation::Return => self.extract_return(contract_id),
        }
    }
```

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+            Ok(decoded_value)
+        } else {
+            Err(Self::parsing_error(output_param))
+        }
+    }
+
+    fn parsing_error(output_param: &ParamType) -> Error {
+        error!(
+            InvalidData,
+            "ReceiptDecoder: failed to find matching receipts entry for {:?}", output_param
```
segfault-magnet, line 89:
```suggestion
            "ReceiptDecoder: failed to find matching receipts entry for {output_param:?}"
```

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+        if let Some(value) = encoded_value {
+            //self.receipts.remove(index);
+            let decoded_value = ABIDecoder::decode_single(output_param, &value)?;
+            Ok(decoded_value)
+        } else {
+            Err(Self::parsing_error(output_param))
+        }
+    }
+
+    fn parsing_error(output_param: &ParamType) -> Error {
```
segfault-magnet, line 86: Maybe calling it `missing_receipts_error` since it is private and only describes that case?

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+                        .to_vec(),
+                ))
+            } else {
+                None
+            }
+        })?;
+
+        self.receipts.remove(index);
+        Some(data)
+    }
```
segfault-magnet, line 112:
```suggestion
    fn extract_return_data(&mut self, contract_id: &ContractId) -> Option<Vec<u8>> {
        for (index, receipt) in self.receipts.iter_mut().enumerate() {
            if let Receipt::ReturnData { id, data, .. } = receipt {
                if id == contract_id {
                    let data = std::mem::take(data);
                    self.receipts.remove(index);
                    return Some(data);
                }
            }
        }
        None
    }
```
A couple of notes about the suggestion: 
* it avoids cloning because it takes the original receipt data out of the `&mut` and replaces it with an empty vec (which doesn't allocate).
* it matches and extracts the data, avoiding the call to `data` which does the matching again
* it avoids handling the impossible case of matching against a `ReturnData` and then having the `.data()` getter fail to match against `ReturnData`

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
```
segfault-magnet, line 134:
```suggestion
    fn extract_return(&mut self, contract_id: &ContractId) -> Option<Vec<u8>> {
        for (index, receipt) in self.receipts.iter().enumerate() {
            if let Receipt::Return { id, val, .. } = receipt {
                if id == contract_id {
                    let data = val.to_be_bytes().to_vec();
                    self.receipts.remove(index);
                    return Some(data);
                }
            }
        }
        None
    }
```
Some of the points of the previous suggestion apply here as well:
* no double match on `Return`
* no handling of the impossible case that a `Return` isn't a `Return`.

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+                Some((i, vec_data.clone()))
+            },
+        ) {
+            self.receipts.remove(i);
+            // the index has adjusted due to the previous removal
+            self.receipts.remove(i);
+            Some(vector_data)
+        } else {
+            None
+        }
```
segfault-magnet, line 157:
```suggestion
        for (index, (current_receipt, next_receipt)) in
            self.receipts.iter().tuple_windows().enumerate()
        {
            if let Some(data) = Self::extract_vec_data(current_receipt, next_receipt, contract_id) {
                let data = data.clone();
                self.receipts.drain(index..=index + 1);
                return Some(data);
            }
        }

        None
```
Notes about the suggestion:
* Removes both elements at once  via `drain`, no need for the comment on why `i` remains the same
* seems simpler to do windows first and then enumerate them instead of the other way around
* for loop seems to cut down on the noise, allows for removing from vectors and immediately returning without worrying about the state of the iterator.

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,323 @@ impl Configurables {
...
+        let mut parser = ReceiptParser::new(&receipts);
+
+        let encoded_data = parser
+            .extract_return_data(target_contract())
+            .expect("This should return data");
+
+        assert_eq!(encoded_data, ENCODED_DATA);
+        assert_eq!(parser.receipts, expected_receipts);
+
+        Ok(())
```
segfault-magnet, line 315: It is better to test through the public interface whenever we can -- makes the component more friendly to refactorings since you won't be battling unit tests that refer to your private innards all the time.

Also inspires more confidence in the component since you're using it in the tests the same way other parts of the code base will use it.
```suggestion
        let expected_receipts = get_relevant_receipts();
        let contract_id = target_contract();

        let mut receipts = expected_receipts.clone();
        receipts.push(get_return_receipt(contract_id, ENCODED_VAL));
        let mut parser = ReceiptParser::new(&receipts);

        let token = parser
            .parse(Some(&contract_id.into()), &u64::param_type())
            .expect("parsing should succeed");

        assert_eq!(u64::from_token(token)?, 225u64);
        assert_eq!(parser.receipts, expected_receipts);

        Ok(())
```

packages/fuels/tests/contracts.rs
```diff
@@ -181,19 +181,75 @@ async fn test_multi_call() -> Result<()> {
...
+    let (val_1, val_2): (u64, u64) = multi_call_handler.call().await?.value;
+
+    assert_eq!(val_1, 7);
+    assert_eq!(val_2, 42);
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_multi_call_pro() -> Result<()> {
```
segfault-magnet, line 202: Do we have a `max` version :smile:

segfault-magnet (commented):
I added suggestions that go with the overall idea of consuming receipts once parsed.

I'm wondering what we gain by removing a receipt once used. It doesn't seem to me like somebody is going to erroneously match against it since we have the contract_id checks. The only case I can think about is the extra receipt we generate for vector returns. If its contract id is all zeroes as well, somebody decoding a script return could maybe match against the vector data. But that would mean being given receipts from both a `Script` and a `Contract` call as well, which I'm not sure we ever do.

It does add complexity. You have to think about iterator invalidation and removing from vectors and avoiding cloning n stuff.

Maybe we should have just kept them as they are and filter through them every time.

If it is about performance, chances are that shuffling elements inside vectors on a remove eats up whatever we gained by not iterating through the full set of receipts every time.

Decoding doesn't need ownership of the bytes so even if great care is taken to avoid cloning it will be for nothing since we could have passed references around anyway.

packages/fuels/tests/contracts.rs
MujkicA: Stay tuned

packages/fuels-programs/src/lib.rs
MujkicA: The goal here was to test the private innards directly, ensuring that small changes to them don't lead to unwanted consequences. These tests can be removed should there be further reaching changes to how we approach receipt parsing.
The correct behaviour of the parser as a whole is tested on a higher level already by our e2e tests.

MujkicA:
> I added suggestions that go with the overall idea of consuming receipts once parsed.
> 
> I'm wondering what we gain by removing a receipt once used. It doesn't seem to me like somebody is going to erroneously match against it since we have the contract_id checks. The only case I can think about is the extra receipt we generate for vector returns. If its contract id is all zeroes as well, somebody decoding a script return could maybe match against the vector data. But that would mean being given receipts from both a `Script` and a `Contract` call as well, which I'm not sure we ever do.
> 
> It does add complexity. You have to think about iterator invalidation and removing from vectors and avoiding cloning n stuff.
> 
> Maybe we should have just kept them as they are and filter through them every time.
> 
> If it is about performance, chances are that shuffling elements inside vectors on a remove eats up whatever we gained by not iterating through the full set of receipts every time.
> 
> Decoding doesn't need ownership of the bytes so even if great care is taken to avoid cloning it will be for nothing since we could have passed references around anyway.

Removing receipts ensures that a multi call doesn't decode the same receipt for two different calls. 
The case you described is also not unlikely if you consider that a user can load a script that contains a contract call. I'm not sure if the latter has some other implications, I will try to test it later.

An alternative would be to somehow track the already consumed receipts without removing them but that would add complexity.

packages/fuels-programs/src/lib.rs
segfault-magnet: > The goal here was to test the private innards directly, ensuring that small changes to them don't lead to unwanted consequences.

I understand, but private innards are implementation details. Caring exactly about how they work damages the refactorability of the code. If you had another idea of how to structure the innards of the receipt parser, the tests are going to fight you rather than help.

A common practice for other languages is to have the unit tests reside in another file. As such, they cannot access the private innards of the thing they're testing. So that forces you into a choice, either test through the public interface or redesign. The reasoning is that if you cannot test through the public interface then you've made your component too big and should consider delegating the work across multiple components.

I'm not innocent of doing this myself when it was a hassle to test through the public interface. But here it requires a little extra effort.

> These tests can be removed should there be further reaching changes to how we approach receipt parsing.

Yes, but by removing them you remove the protection they offer. So you have to write new ones when refactoring. This is the damage to refactorability that I'm referring to.

> The correct behaviour of the parser as a whole is tested on a higher level already by our e2e tests.

I agree that it would be hard to have all the e2e tests pass and not have this component work properly. 

The point I want to get across is that unit tests should completely test the component in question. This offers most of the protection of the e2e tests at a fraction of the cost to run. e2e is to be reserved for validating our expectations against the node or a few select happy paths to see everything integrated correctly. We shouldn't rely on e2e tests to cover what we could have caught in unit tests but didn't. The testing pyramid is a pyramid for a reason. Continuing down the path of needless delegation to e2e testing will have us wait an hour for the suite to complete before long.

MujkicA: Makes sense, thanks for the thorough feedback ❤️

segfault-magnet:
>  The case you described is also not unlikely if you consider that a user can load a script that contains a contract call. I'm not sure if the latter has some other implications, I will try to test it later.

We should probably look into this case more. I'm thinking if a script calls a contract that returns a vector, the receipt for that would only contain the metadata of the vector and it would have the contract_id in it. The additional receipt only happens because we inject the instructions to load from the vm's memory. 

With the contract id in the receipt, it can't be consumed by us parsing the script receipts through the receipts parser. 

> Removing receipts ensures that a multi call doesn't decode the same receipt for two different calls.

Good catch.

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,312 @@ impl Configurables {
...
+    /// and the output param, parse the values and return them as Token.
+    pub fn parse(
+        &mut self,
+        contract_id: Option<&Bech32ContractId>,
+        output_param: &ParamType,
+    ) -> Result<Token> {
+        let contract_id = contract_id.map(Into::into).unwrap_or_else(|| {
+            // During a script execution, the script's contract id is the **null** contract id
+            ContractId::new([0u8; 32])
+        });
```
hal3e, line 65:
```suggestion
        let contract_id = contract_id
            .map(Into::into)
            // During a script execution, the script's contract id is the **null** contract id
            .unwrap_or_else(ContractId::zeroed);
```

packages/fuels-programs/src/lib.rs
```diff
@@ -21,3 +32,312 @@ impl Configurables {
...
+
+        let Some(data) = self.extract_raw_data(output_param, &contract_id) else {
+            return Err(Self::missing_receipts_error(output_param))
+        };
```
hal3e, line 69:
```suggestion
        let data = self
            .extract_raw_data(output_param, &contract_id)
            .ok_or_else(|| Self::missing_receipts_error(output_param))?;
```

packages/fuels-programs/src/contract.rs
```diff
@@ -847,8 +754,9 @@ where
 
     /// Create a [`FuelCallResponse`] from call receipts
     pub fn get_response(&self, receipts: Vec<Receipt>) -> Result<FuelCallResponse<D>> {
-        let token = get_decoded_output(
-            &receipts,
+        let mut receipt_parser = ReceiptParser::new(&receipts);
+
+        let token = receipt_parser.parse(
```
hal3e, line 759:
```suggestion
        let token = ReceiptParser::new(&receipts).parse(
```

hal3e (commented):
This is awesome! Left some small nits...

hal3e:
Maybe it would make sense to move `ReceiptParser` to a separate file.

hal3e (dismissed):
Awesome!

segfault-magnet (approved):

Salka1988 (dismissed):
:+1: LGTM

digorithm:
Can we sync this branch with master, please? 😁

packages/fuels-programs/src/receipt_parser.rs
```diff
@@ -0,0 +1,318 @@
...
+            if let Receipt::ReturnData { id, data, .. } = receipt {
+                if id == contract_id {
+                    let data = std::mem::take(data);
+                    self.receipts.remove(index);
+                    return Some(data);
+                }
+            }
+        }
+        None
+    }
```
digorithm, line 82: How about rewriting this into: 

```suggestion
        fn extract_return_data(&mut self, contract_id: &ContractId) -> Option<Vec<u8>> {
        if let Some(index) = self.receipts.iter().position(|receipt| {
            if let Receipt::ReturnData { id, .. } = receipt {
                id == contract_id
            } else {
                false
            }
        }) {
            if let Receipt::ReturnData { data, .. } = self.receipts.remove(index) {
                Some(data)
            } else {
                None
            }
        } else {
            None
        }
    }
```

Correct if I'm wrong, but, I don't _think_ we need `std::mem::take()` here. These changes make the code more concise and expressive, while also improving its performance by minimizing the number of iterations required to find the relevant receipt. Thoughts?

packages/fuels-programs/src/receipt_parser.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    fn extract_return_data_heap(&mut self, contract_id: &ContractId) -> Option<Vec<u8>> {
+        // If the output of the function is a vector, then there are 2 consecutive ReturnData
+        // receipts. The first one is the one that returns the pointer to the vec struct in the
+        // VM memory, the second one contains the actual vector bytes (that the previous receipt
+        // points to).
+        // We ensure to take the right "first" ReturnData receipt by checking for the
+        // contract_id. There are no receipts in between the two ReturnData receipts because of
+        // the way the scripts are built (the calling script adds a RETD just after the CALL
+        // opcode, see `get_single_call_instructions`).
```
digorithm, line 105: Hoping this assumption never breaks, otherwise we're in a world of pain, haha.

digorithm:
This looks good; I Left just that one nit; feel free to push back, though. Once that discussion is resolved, feel free to merge if I don't come back to approve in time.

packages/fuels-programs/src/receipt_parser.rs
iqdecay: I'm the OP of this assumption. It will not break unless either:
1- Sway doesn't return a vector using `RET` anymore
2- We change the script inside `get_single_call_instructions`. 
There are talks about changing the way vectors and more generally heap types are treated by the compiler so n°1 might happen.
Otherwise, this assumption is actually pretty safe IMO (famous last words lol).

digorithm: > Do you think we should improve the assumption's rationale in-code?

Nah, not much we can do here, really. We just have to keep an eye out for changes in the compiler that could trigger the breakage of this assumption.

packages/fuels-programs/src/receipt_parser.rs
iqdecay: @digorithm I feel that with your suggestion, with the first `if` my brain would need a comment to understand right away. But there is bound to be a lot of receipts parsing so I guess the performance can be important. Maybe just add a comment saying `look for ReturnData receipts with the right contract id`?

digorithm: Yeah. I'm not usually super picky with performance regarding the SDK, but receipt parsing can be one area of performance that could matter.

packages/fuels-programs/src/receipt_parser.rs
iqdecay, line 1: I love that all of this logic is now tested

iqdecay (approved):
noice

packages/fuels-programs/src/receipt_parser.rs
MujkicA: The `receipts` are already filtered at this point to only contain `Return` and `ReturnData`. If we ever measure the performance to be sub par we can come back to this, but right now I think the difference is negligible. 

Since @iqdecay mentioned that this alternative would also warrant a comment, I would rather stick with the current version. 

For my brain it also took a moment to digest that the first if is a filtering step.

hal3e (approved):
Great work!

PR #919, opened 2023-03-31
feat!: add `LogResult` struct

Salka1988:
Renamed `get_logs()` function to `decode_logs()` and modified it to retrieve a `LogResult` structure containing a `results` field that is a vector of `Result<String>` values representing the success or failure of decoding each log.

```rust 
#[derive(Debug)]
pub struct LogResult {
    pub results: Vec<Result<String>>,
}
```
### Checklist


packages/fuels-programs/src/logs.rs
```diff
@@ -144,26 +159,27 @@ pub fn map_revert_error(mut err: Error, log_decoder: &LogDecoder) -> Error {
...
-                return Some(format!(
-                    "assertion failed: `(left == right)`\n left: `{lhs:?}`\n right: `{rhs:?}`"
-                ));
-            }
-            None
-        })
-        .unwrap_or_else(|| "failed to decode logs from assert_eq revert".to_string())
+    let log_result = log_decoder.get_logs(receipts);
+
+    return if let [.., lhs, rhs] = log_result.succeeded.as_slice() {
```
hal3e, line 175: The same comment as above applies here.

packages/fuels/tests/logs.rs
```diff
@@ -1102,7 +1101,7 @@ async fn test_contract_asserts_log() -> Result<()> {
             .call()
             .await
             .expect_err("should return a revert error");
-
+        dbg!(&error);
```
hal3e, line 1104:
```suggestion
```

packages/fuels/tests/logs.rs
```diff
@@ -1121,7 +1120,7 @@ async fn test_contract_asserts_log() -> Result<()> {
...
```
hal3e, line 1123:
```suggestion
```

packages/fuels/tests/logs.rs
```diff
@@ -1259,3 +1258,37 @@ async fn contract_token_ops_error_messages() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_log_results() -> Result<()> {
```
hal3e, line 1263: You can make a wrong log by using assembly and the `log` instruction and putting a invalid log id  in `rb`. This will give you the same result without the two additional projects.

hal3e (commented):
co-reviewed with @segfault-magnet :handshake: 

A slightly different design will solve some of the issues outlined in the review. Namely, if you make the  `LogResult` look something like this:
```rust
LogResult{
   pub results: Vec<Result<String>>
}
```
you can then know for sure which logs succeeded or failed in order. This is paramount for implementing `revert` log decoding (see comments below). Even though users could use the `results` field directly, it would make sense to add some helper methods to hide away some the rust boilerplate involved with extracting data from a `Vec<Results<String>>`. For example, a helper method could be `filter_succeeded` that returns `Vec<String>`. Or another method that returns `Result<Vec<String>>` which will give you all logs or an `Error`. The `Error` could either be the first error or a concatenation of all errors.

packages/fuels-programs/src/logs.rs
```diff
@@ -62,18 +64,31 @@ pub struct LogDecoder {
...
+    pub failed: Vec<Error>,
+}
+
+impl LogResult {}
+
 impl LogDecoder {
     /// Get all decoded logs from the given receipts as `String`
-    pub fn get_logs(&self, receipts: &[Receipt]) -> Result<Vec<String>> {
-        receipts
+    pub fn get_logs(&self, receipts: &[Receipt]) -> LogResult {
```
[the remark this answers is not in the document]
Salka1988: I'll do it :)

packages/fuels/tests/logs.rs
MujkicA: AFAIK the log decoder is bound to a contract response. So he would need to make a dummy contract call just to get access to a log decoder which he can use on the receipts of his custom script with `log `. Is there a more straight up way to do this? Otherwise I would keep it as it is

Salka1988: Maybe we should leave this as it is for now until Ahmed moves the log decoder?

digorithm: > On a related note, I'm thinking of moving the log decoder to the contract instance so that user can decode logs just from receipts that they retrieved via a tx id.

@MujkicA This is an excellent idea! Are you on it already, and/or do we already have an issue documenting this?

hal3e: > AFAIK the log decoder is bound to a contract response. So he would need to make a dummy contract call just to get access to a log decoder which he can use on the receipts of his custom script with `log `. Is there a more straight up way to do this? Otherwise I would keep it as it is.
> 
> On a related note, I'm thinking of moving the log decoder to the contract instance so that user can decode logs just from receipts that they retrieved via a tx id.

We already have `contaract_logs_abi` and `contract_logs` sway project. He updated both to have a `produce_logs_bad_abi` method. He would just need to add some `asm` in `contract_logs::produce_logs_bad_abi`. Then there would not be any need for `contract_logs_one` and `contract_logs_two`.

hal3e: FYI `LogDecoder` is already inside all contract instances.. we just have to add a getter

https://github.com/FuelLabs/fuels-rs/blob/00e2ada480a2c903b2133cca1c20dadf08c53a5b/packages/fuels-code-gen/src/program_bindings/abigen/bindings/contract.rs#L43

MujkicA: @hal3e That makes sense. I thought your suggestion was to use a custom script.

packages/fuels-programs/src/call_response.rs
iqdecay, line 1: nice

packages/fuels-programs/src/logs.rs
```diff
@@ -62,23 +63,81 @@ pub struct LogDecoder {
     pub log_formatters: HashMap<LogId, LogFormatter>,
 }
 
+#[derive(Debug)]
+pub struct LogResult {
+    pub results: Vec<Result<String>>,
+}
+
+impl LogResult {
+    pub fn filter_succeeded(&self) -> Vec<&str> {
```
iqdecay, line 72:
```suggestion
    pub fn successful_logs(&self) -> Vec<&str> {
```

It's a simple getter so I think you can use this convention https://rust-lang.github.io/api-guidelines/naming.html#c-getter

packages/fuels-programs/src/logs.rs
```diff
@@ -62,23 +63,81 @@ pub struct LogDecoder {
...
+        self.results
+            .iter()
+            .filter_map(|result| result.as_deref().ok())
+            .collect()
+    }
+
+    pub fn filter_failed(&self) -> Vec<&Error> {
```
iqdecay, line 79: Same here: 

```suggestion
    pub fn failed_logs(&self) -> Vec<&Error> {
```

packages/fuels-programs/src/logs.rs
```diff
@@ -62,23 +63,81 @@ pub struct LogDecoder {
...
+        self.results
+            .iter()
+            .filter_map(|result| result.as_ref().err())
+            .collect()
+    }
+}
+
 impl LogDecoder {
     /// Get all decoded logs from the given receipts as `String`
```
iqdecay, line 88: Update the docstring

packages/fuels-programs/src/logs.rs
```diff
@@ -62,23 +63,81 @@ pub struct LogDecoder {
...
+        receipts
+            .iter()
+            .rev()
+            .extract_log_id_and_data()
+            .next()
+            .ok_or_else(|| error!(InvalidData, "nani"))
+            .and_then(|(log_id, data)| self.format_log(&log_id, &data))
+    }
+
+    pub fn decode_last_two_logs(&self, receipts: &[Receipt]) -> Result<(String, String)> {
```
iqdecay, line 118:
```suggestion 
fn decode_last_two_logs(&self, receipts: &[Receipt]) -> Result<(String, String)> {
```

This method doesn't need to be public I think.

packages/fuels-programs/src/logs.rs
```diff
@@ -62,23 +63,81 @@ pub struct LogDecoder {
...
+            .and_then(|log_formatter| log_formatter.format(data))
+    }
+
+    pub fn decode_last_log(&self, receipts: &[Receipt]) -> Result<String> {
+        receipts
+            .iter()
+            .rev()
+            .extract_log_id_and_data()
+            .next()
+            .ok_or_else(|| error!(InvalidData, "nani"))
```
iqdecay, line 114: Fix this error message

iqdecay (changes requested):
Left some remarks

docs/src/calling-contracts/logs.md
```diff
@@ -14,8 +14,7 @@ You can access the logged values in Rust by calling `logs_with_type::<T>` from a
 {{#include ../../../packages/fuels/tests/logs.rs:produce_logs}}
 ```
 
-You can also get a vector of all the logged values as strings using `get_logs()`:
-
+You can use the `get_logs()` function to retrieve a `LogResult` struct containing a `succeeded` field with a vector of successfully decoded logs, and a `failed` field with a vector of Errors indicating any log that could not be decoded.
```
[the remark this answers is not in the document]
Salka1988: I think the explanation is better than the structure alone, and if someone wants to look at the structure, they will find it easily.

packages/fuels-programs/src/logs.rs
Salka1988: Missed that one!

packages/fuels-programs/src/logs.rs
Salka1988: It's no ordinary getter. Values from `results: Vec<Result<String>>` are filtered. `log_result.filter_succeeded()` is a more accurate description of what it does.

packages/fuels-programs/src/logs.rs
Salka1988: You're right.

packages/fuels-programs/src/logs.rs
Salka1988: https://github.com/FuelLabs/fuels-rs/pull/919#discussion_r1161777651

docs/src/calling-contracts/logs.md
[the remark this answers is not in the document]
iqdecay: Just realized that this description does not match the implementation so you should change the description.

iqdecay (changes requested):

iqdecay (approved):
:ship: it

PR #921, opened 2023-04-03
chore: default-features warnings on cargo nightly

segfault-magnet:
closes: #918 

From [the cargo book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html):
> Other than optional and features, inherited dependencies cannot use any other dependency key (such as version or default-features).

We've been inheriting workspace deps and then trying to disable default features. Cargo doesn't support this. If a dep has default features enabled when defined at the workspace level, those features are there to stay. You can only add more to them via `features` or make the dep itself optional in the place you're using it via `optional=true`.

The check for this still hasn't landed in cargo stable. I opted not to add a CI step to build `fuels-rs` with the nightly toolchain to check for regressions in this regard.

Functionally nothing has changed. The changes are primarily so we don't mislead ourselves and so we don't run into an error when this does reach stable.

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,12 +42,12 @@ fuel-crypto = "0.26"
 fuel-merkle = "0.26"
 fuel-storage = "0.26"
 fuel-tx = "0.26"
-fuel-types = "0.26"
+fuel-types = { version = "0.26", default-features = false }
 fuel-abi-types = "0.2.1"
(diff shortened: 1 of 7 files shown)


Salka1988 (approved):
:+1:

digorithm (approved):

PR #929, opened 2023-04-12
Bump versions to 0.40.0

digorithm:


rvmelkonian:
How long normally typically after a version bump is the new version published to crates.io?

PR #930, opened 2023-04-12
feat: cache `get_spendable_resources`

### LINKED ISSUE #903: Transactions getting rejected due to "UTXO does not exist"
Transactions on the beta network will occasionally be rejected with the following reason "UTXO does not exist". 

One possible cause could be that resources_to_spend API is returning coins that have already been included in other transactions they've submitted (due to the delay between submitting a transaction and marking the inputs as spent). 

If the cause is confirmed to be the above, we can consider the following solution suggested by @Voxelot:

The SDK should maintain a cache of recently used UTXOs, which are to be excluded calling resources_to_spend. When a transaction is confirmed or rejected, these UTXOs can be removed from the cache. This may require users to wait for their change outputs to become available before making subsequent transactions, but the process can be optimized by spending the change outputs of pending transactions in-flight. The SDK can predict the change outputs of previously submitted transactions during dry-run, which can aid in selecting pending outputs.

MujkicA:
Closes #903

As proposed, the PR introduces a caching mechanism for accounts (wallet and predicate). The cache holds UTXOs consumed by a recent transaction as well as UTXOs that are expected to be produced. The expected UTXOs are currently only supported for transfers as that is the only scenario where we can calculate a minimum change amount that can be fixed and assigned to an `Output::Coin` instead of `Output::Change`. The calculated change output can then be used as input for a new transaction because we know the exact amount.
The cache uses interior mutability to avoid introducing the need for mutable references on the public side of our API.

Some caveats:
- had to move `get_spendable_resources` to `Account` from `ViewOnlyAccount`, the alternative was to move the cache access methods to `ViewOnlyAccount` but that makes even less sense
- still missing a reliable way to test this as the issue in question appears sporadically on the beta network

### Checklist


MujkicA:
Currently blocked by failing wasm tests

packages/fuels/tests/contracts.rs
```diff
@@ -275,7 +275,7 @@ async fn test_contract_call_fee_estimation() -> Result<()> {
     let tolerance = 0.2;
 
     let expected_min_gas_price = 0; // This is the default min_gas_price from the ConsensusParameters
-    let expected_gas_used = 606;
+    let expected_gas_used = 516;
```
MujkicA, line 278:
```suggestion
    let expected_gas_used = 606;
```

iqdecay:
@FuelLabs/sdk-rust what should we do with this?

Voxelot:
any updates? :)

MujkicA:
On it's way, updating the branch

MujkicA:
Found a bug where consecutive transactions won't fetch the right coins because of our two step process when retrieving coins (add_fee_resource) and how the cache is invalidated. It get's invalidated in the first step which means that after the second, final step the resources can still be wrong. Working on a better concept...

PR #946, opened 2023-04-28
add section comments for devrel

sarahschwartz:
Adds section comments to the documentation so that the devrel team can easily pull content from this book


iqdecay:
@sarahschwartz can you please give us a link to the documentation of those tags? So we can know how to maintain them in the future.

sarahschwartz:
> @sarahschwartz can you please give us a link to the documentation of those tags? So we can know how to maintain them in the future.

Yes! We are working on a frontend for guides that uses git submodules to pull in all of the fuel docs, and created a component to pull in content based on the comments. For example, we would have a component that looks like this:

```jsx
<MDXImport
file="../../path/to/index.md"
comment="fuels_rs"
/>
```

This will pull in the content in between matching comments like this:

```
<!-- This example should include a summary of fuels_rs -->
<!-- fuels_rs:example:start -->
CONTENT
<!-- fuels_rs:example:end -->
```

The first comment is meant to explain the purpose of this content snippet, so that as the docs are updated, the core intent of these sections are maintained and we can simply update the submodules to update the guides. (If something needs to be deleted or moved in the docs, that is still fine as long as we are informed.)

iqdecay:
Ok great! Is there some tags we would need to add/modify when updating the docs?

docs/src/calling-contracts/call-params.md
```diff
@@ -20,7 +24,11 @@ Then, in Rust, after setting up and deploying the above contract, you can config
 {{#include ../../../examples/contracts/src/lib.rs:call_parameters}}
 ```
 
-`call_params` returns a result to ensure you don't forward assets to a contract method that isn't payable. In the following example, we try to forward an amount of 100 of the base asset to `non_payable`. As its name suggests, `non_payable` isn't annotated with `#[payable]` in the contract code. Passing `CallParameters` with an amount other than 0 leads to an `InvalidCallParameters` error:
+<!-- This section should why `call_params` returns a result -->
```
iqdecay, line 27:
```suggestion
<!-- This section should explain why `call_params` returns a result -->
```

sarahschwartz:
> Ok great! So when modifiying the docs that fall under a tag, we would just ping devrel?

Only if you need to remove the content, move to another file, or change the overall intent of the content so that the comment description doesn't apply anymore. For general updates, we don't need to be pinged.

iqdecay (approved):
found one nit, otherwise lgtm, looking forward to seeing this in action!

docs/src/calling-contracts/call-params.md
sarahschwartz: thanks!

digorithm (approved):
Very nice, can't wait to see this in action 😍

PR #955, opened 2023-05-05
feat: add trimming utilities to `SizedAsciiString`

### LINKED ISSUE #953: Support `.trim()` for `SizedAsciiString`
- Here is the issue I created on the indexer project to track this https://github.com/FuelLabs/fuel-indexer/issues/853
  - This issue has all the context, happy to provide more if needed

iqdecay:
This closes #953 by implementing some trimming and padding utilities for `SizedAsciiString` types.

### DIFF
--- a/packages/fuels-types/src/core/sized_ascii_string.rs
+++ b/packages/fuels-types/src/core/sized_ascii_string.rs
@@ -24,6 +24,30 @@ impl<const LEN: usize> SizedAsciiString<LEN> {
         }
         Ok(Self { data })
     }
+
+    pub fn to_trimmed_string(&self) -> String {
+        self.data.trim().to_string()
+    }
+    pub fn to_left_trimmed_string(&self) -> String {
+        self.data.trim_start().to_string()
+    }
+    pub fn to_right_trimmed_string(&self) -> String {
+        self.data.trim_end().to_string()
+    }
+
+    /// Pad `data` string with whitespace characters on the right to fit into the `SizedAsciiString`
+    pub fn new_with_right_whitespace_padding(data: String) -> Result<Self> {
+        if data.len() > LEN {
+            let l = data.len();
+            return Err(error!(
+                InvalidData,
+                "SizedAsciiString<{LEN}> cannot be constructed from a string of size {l}"
+            ));
+        }
+        Ok(Self {
+            data: format!("{:width$}", data, width = LEN),
+        })
+    }
 }
 
 impl<const LEN: usize> TryFrom<&str> for SizedAsciiString<LEN> {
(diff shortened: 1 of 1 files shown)


packages/fuels-types/src/core/sized_ascii_string.rs
```diff
@@ -24,6 +24,30 @@ impl<const LEN: usize> SizedAsciiString<LEN> {
...
+            let l = data.len();
+            return Err(error!(
+                InvalidData,
+                "SizedAsciiString<{LEN}> cannot be constructed from a string of size {l}"
+            ));
+        }
+        Ok(Self {
+            data: format!("{:width$}", data, width = LEN),
+        })
+    }
```
hal3e, line 50:
```suggestion
    pub fn new_with_right_whitespace_padding(data: String) -> Result<Self> {
        if data.len() > LEN {
            return Err(error!(
                InvalidData,
                "SizedAsciiString<{LEN}> cannot be constructed from a string of size {}",
                data.len()
            ));
        }

        Ok(Self {
            data: format!("{:LEN$}", data),
        })
    }
```

hal3e (changes requested):
Nice! I have some sugestions

packages/fuels-types/src/core/sized_ascii_string.rs
```diff
@@ -138,4 +162,20 @@ mod tests {
...
+        // Every whitespace in those literals is made of a single whitespace
+        let untrimmed = SizedAsciiString::<9>::new(" est abc ".to_string())?;
+
+        assert_eq!("est abc ", untrimmed.to_left_trimmed_string());
+        assert_eq!(" est abc", untrimmed.to_right_trimmed_string());
+        assert_eq!("est abc", untrimmed.to_trimmed_string());
+
+        let padded =
+            SizedAsciiString::<100>::new_with_right_whitespace_padding("victor".to_string())?;
+        assert_eq!(&*format!("{:width$}", "victor", width = 100), padded);
```
segfault-magnet, line 177: Hard coding the expectation would be more readable. 
e.g
```suggestion
        let padded =
            SizedAsciiString::<3>::new_with_right_whitespace_padding("victor".to_string())?;
        assert_eq!("victor   ", padded);
```

Also, it avoids using the code we're testing to produce expectations for the test itself. This can lead to a never-failing test.

packages/fuels-types/src/core/sized_ascii_string.rs
```diff
@@ -24,6 +24,30 @@ impl<const LEN: usize> SizedAsciiString<LEN> {
         }
         Ok(Self { data })
     }
+
+    pub fn to_trimmed_string(&self) -> String {
+        self.data.trim().to_string()
+    }
```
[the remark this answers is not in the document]
iqdecay: I disagree with naming it `trim`. Using the `trim` naming would imply that we are returning the same base type, whereas here it's a conversion to a base type that's quite different.

hal3e (approved):

PR #958, opened 2023-05-09
feat: `low level call`support

Salka1988:
This PR adds two macros to the project: `fn_selector!()` and `calldata()`.
These macros simplify the usage of the `std::low_level_call::call_with_function_selector` function by providing a more convenient way to generate the function_selector and calldata arguments.

contracts/low_level_call/src/main.sw
```sw
fn call_low_level_call(
    target: ContractId,
    function_selector: Bytes,
    calldata: Bytes,
    single_value_type_arg: bool,
) {
    let call_params = CallParams {
        coins: 0,
        asset_id: BASE_ASSET_ID,
        gas: 10_000,
    };

    call_with_function_selector(target, function_selector, calldata, single_value_type_arg, call_params);
}
```
low_level_cal test
```rust
let function_selector = fn_selector!(set_value(u64));
let call_data = calldata!(42u64);

contract_instance
    .methods()
    .call_low_level_call(
        contract_id.clone().into(),
        Bytes(function_selector),
        Bytes(call_data),
        true,
    )
    .call()
    .await?;
```

### Checklist


Salka1988:
I faced a challenge while attempting to enable `return` for `call_with_function_selector` function for dynamic data types. `Vec<_>`
I sought the opinion of @MujkicA  and together we concluded that returning `Vec<>` would not be feasible without support from the SWAY team. @nfurfaro later confirmed that supporting `low_level_call return` may not be necessary at all. In light of this, I decided to remove support for returning static data for the time being, with the possibility of adding it later if required.

nfurfaro (commented):
This looks good. As you mentioned, a future enhancement could be to handle dynamic return data if needed.
One other thing that could be good to check here would be to deploy a second contract "Target" which has all the abi functions currently in the test contract except for the low_level_call function.
The idea would be to call `Contract_A.low_level_call`, passing the `ContractId` of the Target contract, the function selector and call params.
Everything should continue to work the same, but that's more representative of how this will be used. The key difference here being that we don't need to use the abigen! macro for the target contract; we just need its `ContractId` and a function selector.

nfurfaro (approved):
Looks good to me!

packages/fuels/tests/contracts.rs
```diff
@@ -1234,3 +1236,106 @@ async fn multi_call_from_calls_with_different_account_types() -> Result<()> {
...
+        .set_contracts(&[&target_contract_instance])
+        .call()
+        .await?;
+
+    let response = target_contract_instance
+        .methods()
+        .get_counter()
+        .call()
+        .await?;
+    assert_eq!(response.value, 42);
```
MujkicA, line 1288: I would move this assertion to the test in `/examples` and have this test focused on `set_value_multiple_complex`

MujkicA (approved):

packages/fuels-core/src/function_selector.rs
```diff
@@ -75,6 +75,24 @@ fn resolve_arg(arg: &ParamType) -> String {
...
+pub use fn_selector;
+
+#[macro_export]
+macro_rules! calldata {
+    ( $($arg: expr),* ) => {
+        ::fuels::core::abi_encoder::ABIEncoder::encode(&[$(::fuels::types::traits::Tokenizable::into_token($arg)),*]).unwrap().resolve(0)
+    }
+}
+
+pub use calldata;
```
iqdecay, line 94: Can we add some basic unit tests for those macros?

iqdecay (changes requested):
Can you just please add unit tests that showcase what the generated code looks like

iqdecay (approved):
lgtm

PR #966, opened 2023-05-15
try wrapping ConsensusParameters

iqdecay:
This is just to show what it would "look like" if we wrapped ConsensusParameters in their own type

### DIFF
--- a/examples/cookbook/src/lib.rs
+++ b/examples/cookbook/src/lib.rs
@@ -89,8 +89,7 @@ mod tests {
     async fn custom_chain() -> Result<()> {
         use fuels::prelude::*;
         // ANCHOR: custom_chain_import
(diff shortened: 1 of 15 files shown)


segfault-magnet (commented):
I'm for reexposing the type. Doesn't seem like there is anything we can simplify by wrapping it.

iqdecay:
> I'm for reexposing the type. Doesn't seem like there is anything we can simplify by wrapping it.

Agreed, + I can't make it work properly anyway. It's just that I really dislike the interface for `ChainConfig` and `ConsensusParameters`, there are way to many fields, no nice defaults, and most of all, the `ConsensusParameters` are placed in the `transaction_parameters` field of the `ChainConfig`.

iqdecay:
Closing as this is no longer relevant.

Issue #962, opened 2023-05-13
Change contract methods to accept Into<Address> and Into<ContractId> instead of Address and ContractId


SwayStar123:
As wallets return a Bech32Address instead of an Address on calling .address() on them, this would simplify the interface for most people

PR #967, opened 2023-05-15
feat: add `Into` for `Address` and `ContractId` fn arguments

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/962

Our `FunctionGenerator` now generates `impl Into<Address>` and `impl Into<ContractId>` when those types are used as function arguments.

This is reflected on all `contract` methods, `predicate` encode functions, and `script` main functions.

### Checklist

### DIFF
--- a/packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
+++ b/packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
@@ -66,7 +66,17 @@ impl FunctionGenerator {
     }
 
     pub fn tokenized_args(&self) -> TokenStream {
-        let arg_names = self.args.iter().map(|component| &component.field_name);
+        let arg_names = self.args.iter().map(|component| {
+            let field_name = &component.field_name;
+
+            if component.wrap_into {
+                let field_type = &component.field_type;
+
+                quote! {::core::convert::Into::<#field_type>::into(#field_name)}
+            } else {
+                quote! {#field_name}
+            }
+        });
         quote! {[#(::fuels::types::traits::Tokenizable::into_token(#arg_names)),*]}
     }
 
@@ -108,7 +118,12 @@ impl From<&FunctionGenerator> for TokenStream {
         let arg_declarations = fun.args.iter().map(|component| {
             let name = &component.field_name;
             let field_type = &component.field_type;
-            quote! { #name: #field_type }
+
+            if component.wrap_into {
+                quote! { #name: impl ::core::convert::Into<#field_type> }
+            } else {
+                quote! { #name: #field_type }
+            }
         });
 
         let output_type = fun.output_type();
--- a/packages/fuels-code-gen/src/program_bindings/utils.rs
+++ b/packages/fuels-code-gen/src/program_bindings/utils.rs
@@ -19,6 +19,7 @@ use crate::{
 pub(crate) struct Component {
     pub field_name: Ident,
     pub field_type: ResolvedType,
+    pub wrap_into: bool,
 }
 
 impl Component {
@@ -36,6 +37,7 @@ impl Component {
         Ok(Component {
             field_name: safe_ident(&field_name),
             field_type: TypeResolver::new(mod_of_component).resolve(component)?,
+            wrap_into: component.type_decl.wrap_type_into(),
         })
     }
 }
--- a/examples/contracts/src/lib.rs
+++ b/examples/contracts/src/lib.rs
@@ -350,7 +350,7 @@ mod tests {
 
         // withdraw some tokens to wallet
         let response = contract_methods
-            .transfer_coins_to_output(1_000_000, contract_id.into(), address.into())
+            .transfer_coins_to_output(1_000_000, contract_id, address)
             .append_variable_outputs(1)
             .call()
             .await?;
@@ -435,7 +435,7 @@ mod tests {
         let amount = 100;
 
         let response = contract_methods
-            .increment_from_contract_then_mint(called_contract_id, amount, address.into())
+            .increment_from_contract_then_mint(called_contract_id, amount, address)
             .call()
             .await;
 
@@ -447,7 +447,7 @@ mod tests {
 
         // ANCHOR: dependency_estimation_manual
         let response = contract_methods
-            .increment_from_contract_then_mint(called_contract_id, amount, address.into())
+            .increment_from_contract_then_mint(called_contract_id, amount, address)
             .append_variable_outputs(1)
             .set_contract_ids(&[called_contract_id.into()])
             .call()
@@ -460,7 +460,7 @@ mod tests {
 
         // ANCHOR: dependency_estimation
         let response = contract_methods
-            .increment_from_contract_then_mint(called_contract_id, amount, address.into())
+            .increment_from_contract_then_mint(called_contract_id, amount, address)
             .estimate_tx_dependencies(Some(2))
             .await?
             .call()
--- a/examples/cookbook/src/lib.rs
+++ b/examples/cookbook/src/lib.rs
@@ -57,7 +57,7 @@ mod tests {
             .set_asset_id(base_asset_id);
 
         contract_methods
-            .deposit(wallet.address().into())
+            .deposit(wallet.address())
             .call_params(call_params)?
             .append_variable_outputs(1)
             .call()
@@ -73,7 +73,7 @@ mod tests {
             .set_asset_id(lp_asset_id);
 
         contract_methods
-            .withdraw(wallet.address().into())
+            .withdraw(wallet.address())
             .call_params(call_params)?
             .append_variable_outputs(1)
             .call()
--- a/packages/fuels-code-gen/src/program_bindings/abi_types.rs
+++ b/packages/fuels-code-gen/src/program_bindings/abi_types.rs
@@ -259,13 +259,15 @@ impl FullTypeDeclaration {
     }
 
     pub fn is_enum_type(&self) -> bool {
-        let type_field = &self.type_field;
-        type_field.starts_with("enum ")
+        self.type_field.starts_with("enum ")
     }
 
     pub fn is_struct_type(&self) -> bool {
-        let type_field = &self.type_field;
-        type_field.starts_with("struct ")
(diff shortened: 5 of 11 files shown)


Br1ght0ne (dismissed):

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
```diff
@@ -66,7 +66,17 @@ impl FunctionGenerator {
...
+        let arg_names = self.args.iter().map(|component| {
+            let field_name = &component.field_name;
+
+            if component.wrap_into {
+                let field_type = &component.field_type;
+
+                quote! {::core::convert::Into::<#field_type>::into(#field_name)}
+            } else {
+                quote! {#field_name}
+            }
```
iqdecay, line 78: I think this should be replaced with just an associated function on `Component`: 
`component.wraps_into()`

packages/fuels-code-gen/src/program_bindings/utils.rs
```diff
@@ -19,6 +19,7 @@ use crate::{
 pub(crate) struct Component {
     pub field_name: Ident,
     pub field_type: ResolvedType,
+    pub wrap_into: bool,
```
iqdecay, line 22: I think this field should have some documentation as to why it's there and what purposes it serves. Also I am not sure I understand, if this is only for `ContractId` and `Address` types, why have this field at all and not just use a boolean `Component::wraps_into()` method? And you would have an enum that hold all the strings that need to be wrapped into (for now, just `Address` and `ContractId`).

iqdecay (changes requested):
A question on design

iqdecay (commented):
I feel like this would enable the user to pass in an `Address` to a method that normally only accepts `ContractId` without knowing, and this is not something we want. The difference between `Address` and `ContractId` is fundamental. Am I getting this wrong?

hal3e:
> I feel like this would enable the user to pass in an `Address` to a method that normally only accepts `ContractId` without being explicit about it, and this is not something we want. It needs to be explicit, because the difference between `Address` and `ContractId` is fundamental and can lead to a loss of fund IIUC. Am I getting this wrong?

You are right. Normally it would not be a problem to have a `impl Into<ContractId>` as I would imagine that `into` would be implemented only types that make sense. However, we will get into implementations for conversion between `Address` and `ContractId` and then we do not have type safety anymore. 

The only other thing that comes to mind is to have methods accept `Bech32Address` and `Bech32ContractId`. However, this solution is not as ergonomic as `impl Into<ContractId>` as you would have to turn a already existing `ContractID` into `Bech32ContractId`

@FuelLabs/sdk-rust what are you thoughts on this? I am leaning towards leaving it as it is and closing this PR.

segfault-magnet:
> > I feel like this would enable the user to pass in an `Address` to a method that normally only accepts `ContractId` without being explicit about it, and this is not something we want. It needs to be explicit, because the difference between `Address` and `ContractId` is fundamental and can lead to a loss of fund IIUC. Am I getting this wrong?
> 
> You are right. Normally it would not be a problem to have a `impl Into<ContractId>` as I would imagine that `into` would be implemented only for types that make sense. However, we will get into implementations for conversion between `Address` and `ContractId` and then we do not have type safety anymore.
> 
> The only other thing that comes to mind is to have methods accept `Bech32Address` and `Bech32ContractId`. However, this solution is not as ergonomic as `impl Into<ContractId>` as you would have to turn an already existing `ContractID` into `Bech32ContractId` (same applies to `Address`)
> 
> @FuelLabs/sdk-rust what are you thoughts on this? I am leaning towards leaving it as is and closing this PR.

What about replacing the usages of `Address` with `impl Into<Bech32Address>` in the generated fn bindings?

So you can then give both an `Address` and its bech32 equivalent. Same for the `ContractId`s.

You'd retain type safety since neither `Address`  satisfies `impl Into<Bech32ContractId>` nor does `ContractId` satisfy `impl Into<Bech32Address>`. 

You'd still have pure addresses nested in generated structs and enums, since using `impl` there would cause generic headaches.

iqdecay:
I would like the `Into<Bech32Address>` and `Into<Bech32ContractId>`, sending something to a `Bech32Address` vs its corresponding `Address` doesn't work I think. If they are not functionaly equivalent, I'm for keeping type safety above all, otherwise users would probably shoot themselves in the foot. 
But it does ask the question of why we are bothering users with implementation details of exposing the `Bech32` format at all, IMO we should discuss around how to abstract this away.

hal3e:
> What about replacing the usages of `Address` with `impl Into<Bech32Address>` in the generated fn bindings?
> 
> So you can then give both an `Address` and its bech32 equivalent. Same for the `ContractId`s.
> 
> You'd retain type safety since neither `Address` satisfies `impl Into<Bech32ContractId>` nor does `ContractId` satisfy `impl Into<Bech32Address>`.
> 
> You'd still have pure addresses nested in generated structs and enums, since using `impl` there would cause generic headaches.

This is a great idea! :heart:  I did a quick first implementation and it works as expected. We retain the type safety as you said.

@iqdecay you said:  but sending something to a Bech32Address vs its corresponding Address doesn't work I think. 
We turn it back to an `Address` under the hood so everything works the same. The interface is just a bit nicer as you can give either a `Address` or an `Bech32Address`

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
hal3e: resolving as the code changed. See: https://github.com/FuelLabs/fuels-rs/pull/967#issuecomment-1556131371

packages/fuels-code-gen/src/program_bindings/utils.rs
hal3e: resolving as the code changed. See: https://github.com/FuelLabs/fuels-rs/pull/967#issuecomment-1556131371

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
```diff
@@ -108,7 +117,13 @@ impl From<&FunctionGenerator> for TokenStream {
         let arg_declarations = fun.args.iter().map(|component| {
             let name = &component.field_name;
             let field_type = &component.field_type;
-            quote! { #name: #field_type }
+
+            find_into_impl_type(&field_type.type_name.to_string())
+                .map(|new_type| {
+                    let new_type = new_type;
```
segfault-magnet, line 123: Seems odd, the compiler gave you trouble?

packages/fuels-code-gen/src/program_bindings/utils.rs
```diff
@@ -155,3 +155,13 @@ pub(crate) fn sdk_provided_custom_types_lookup() -> HashMap<TypePath, TypePath>
     })
     .collect()
 }
+
+pub(crate) fn find_into_impl_type(ttype: &str) -> Option<TokenStream> {
```
segfault-magnet, line 159: It might be more readable to step down the generality, and make the name more specific until we need it to be more general.

E.g. `get_equivalent_bech32_type` 

If we ever `Into` something else we can revert.

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
hal3e: good catch!

packages/fuels-types/src/bech32.rs
```diff
@@ -79,6 +79,11 @@ bech32type!(Bech32Address);
 bech32type!(Bech32ContractId);
 
 // Bech32Address - Address conversion
+impl From<&Bech32Address> for Bech32Address {
+    fn from(data: &Bech32Address) -> Bech32Address {
+        data.clone()
+    }
+}
```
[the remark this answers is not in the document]
hal3e: Just that there is no implementation for `From<&Bech32Address>`. We have `From<&Address>` so I added the same for `&Bech32Address` to make it "consistent".

hal3e:
> > 
> 
> What I meant is I think that rather than changing this implementation detail, we should re-explore why we are exposing two types of addresses at all. Fixing this seems like a bandaid to the more general problem of exposing two types of addresses instead of handling them both under the hood. This particular fix is moreover not the one requested by user (because the one requested by users is a gateway to a footgun), so we don't actuallly need to do it IMO.

I agree but what you say is above this request. The user wanted a nice UI where he can give either `Address` or `Bech32Address` (same for `ContractId`) to contract methods. This PR makes that possible without loosing type safety. So until we have an unified `Address` and `ContractId` IMO it makes sense to have this.

iqdecay:
Ah yes you're right mb

segfault-magnet (approved):
Good stuff.

Salka1988 (approved):
:+1:

iqdecay (approved):
nice!

SwayStar123:
Thank you hal3e, very cool

diyahir:
Could we not determine if it's a contract or address at run-time and avoid casting it altogether? 

Casting it only at runtime for the pieces that must know

diyahir:
If I give you an address 0x123....789, and I say to send it 10 ETH, is it fair to say you could get an error depending on if it's a contractId or an address, so you would need to ask me a follow-up, is it an address or a contract. Well a multisig is a contract, but I could see someone making that mistake. 

It seems like we're building a 'gotcha' into the sway stack that could be avoided entirely if we just abstract all of this into a run-time check (if possible, if not possible currently then add a flag to each address?)

iqdecay:
This is a good argument I agree, but this is not the right place to discuss it, https://github.com/FuelLabs/sway/issues/4569. This design decision is not ours, and even if we wanted we cannot change it, it's in Sway.

diyahir:
> This is a good argument I agree, but this is not the right place to discuss it, [this issue](https://github.com/FuelLabs/sway/issues/4569) is. This design decision is not ours, and even if we wanted we cannot change it, it's in Sway.

This development would be superfluous if we did a run-time check (eliminating Address and ContractId distinction), so It does feel like a decent place to discuss

PR #968, opened 2023-05-16
feat: add `encode` method to `ContractCallHandler`

### LINKED ISSUE #709: `encode` method to return the encoded data (as `bytes` type) with function selector for a contract call
Sway supports a tightly packed `bytes` [type](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/bytes.sw), which allows for [low-level calls](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/low_level_call.sw) when the ABI is not known at compile time (e.g. smart contract wallet calls, flashloan calls, etc). 

To support this, the SDK will need tools for encoding data according to a particular function signature. I'm imagining this could be part of the contract API, something like

```
// encode calldata with function selector for callEE contract
let a : u8 = 1;
let b : u8 = 2;
let c : u8 = 3;
let calldata = callee_contract.methods().my_method(a, b, c).encode()?;
```

Now `calldata` is (for this example) `0x00000000b18477cf000100020003`, which can be passed to a function expecting `bytes`

```
// Now I can use my general call method on my callER contract
caller_contract.methods().general_call(callee_contract_id, calldata).call().await?;
```

Related: #775

Br1ght0ne:
Close #709.

### Checklist


segfault-magnet (commented):
Doesn't seem to handle value types? 

I.e. any type whose `should compute custom offset` (in the SDK) returns `false`. 

Any idea how to handle that? Technically, when encoding, we already have enough info to deduce if we're loading the value or a ptr in the register. Maybe you can prepend that into the bytes to be later used in Sway? Ugly but it works. 

Or ditch the Bytes and go with an enum clearly separating the two cases. But then you'd probably need to have it in the Sway std lib somewhere.

iqdecay:
@Br1ght0ne if you are not ready to implement the value-types handling, maybe we can pass this to draft?

digorithm (commented):
Any updates on this? 👀 The branch is a bit stale.

Br1ght0ne:
Closing this for now, needs triage/double-take.

PR #985, opened 2023-05-28
docs: show example of block time travel

### LINKED ISSUE #849: Sway `timestamp()` seems to not be shifted by `produce_blocks()`
[Here](https://github.com/Braqzen/tmp_timestamp) is a link to the repro.

It makes a call to some function which logs the time, the SDK shifts some blocks at some interval of time and calls the same function again to log the new timestamp. The timestamp printed to the standard output is the same.

iqdecay:
This PR closes #849 (actually it was closed by #950) by demonstrating that the Sway timestamp is indeed shifted by the `produce_blocks` API.


packages/fuels-accounts/src/provider.rs
```diff
@@ -253,7 +253,7 @@ impl Provider {
             .map_err(Into::into)
     }
 
-    #[cfg(feature = "fuel-core")]
```
digorithm, line 256: I'm out of context: why are we making this change here?

digorithm (commented):
Very neat and clear tests. Looks good to me, but haven't approved it yet because I need some more clarification on the question I left.

packages/fuels-accounts/src/provider.rs
iqdecay: Oh it's just that the feature is called `fuel-core-lib` and that this one is called `fuel-core`. Unless we happen to have a `fuel-core` feature that's different but in that case it's not documented.

iqdecay: Btw this kind of proves a fear I had which is that we are pretty much not testing anything related to the `fuel-core-lib` feature.

digorithm: Hmm, that's true. I can't seem to find any feature called just `"fuel-core"`, shouldn't it fail if we conditionally compile something with a feature that doesn't exist? 🤔 In any case, lgtm.

digorithm (approved):

Salka1988 (approved):
👍

PR #991, opened 2023-06-06
chore: use `#[allow(dead_code)]` in forc projects

hal3e:
`#[allow(dead_code)]` was recently added to `sway` so I have updated our test projects to lower the number of warnings we get when building them.

I have also removed some other warnings like unused variables etc.


digorithm (approved):
💯

iqdecay (approved):
lgtm

MujkicA (approved):
Looks good!

PR #998, opened 2023-06-11
feat:  add support for DbType::RocksDb

Salka1988:
### Checklist


Cargo.toml
```diff
@@ -45,7 +45,7 @@ elliptic-curve = { version = "0.13.4", default-features = false }
 eth-keystore = "0.5.0"
 fuel-abi-types = "0.2.1"
 fuel-asm = "0.31.1"
-fuel-core = { version = "0.18.2", default-features = false }
+fuel-core = { version = "0.18.2", default-features = false,  features = ["rocksdb"] }
```
Voxelot, line 48: this will likely increase build times a lot, might be worth putting under a feature flag?

Salka1988: ok

packages/fuels/.DS_Store
MujkicA, line 1: We probably don't need this 😄

packages/fuels-test-helpers/src/node.rs
```diff
@@ -230,12 +244,36 @@ pub async fn new_fuel_node(
...
+        args.extend(vec![
+            "--db-type".to_string(),
+            match config.database_type {
+                DbType::InMemory => "in-memory",
+                DbType::RocksDb => "rocks-db",
+            }
+            .to_string(),
+        ]);
+
+        if let DbType::RocksDb = config.database_type {
```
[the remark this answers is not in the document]
Salka1988: you are right. Forgot that

packages/fuels-test-helpers/src/node.rs
```diff
@@ -40,9 +41,18 @@ pub enum Trigger {
     },
 }
 
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum DbType {
+    InMemory,
+    RocksDb,
```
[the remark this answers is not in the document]
Salka1988: I aim to align with Fuel core's approach for consistency.  It would be beneficial if they could export the "Config" for us to use.

packages/.DS_Store
Braqzen, line 1: Excuse me good sir, there appears to be a file here which should not exist. Would you be so kind as to do to it what we need to do to the IRS?

packages/fuels/Cargo.toml
```diff
@@ -45,3 +45,4 @@ std = [
 # TODO: To be removed once https://github.com/FuelLabs/fuels-rs/issues/881 is unblocked.
 test-type-paths = []
 fuel-core-lib = ["fuels-test-helpers?/fuel-core-lib", "dep:fuel-core"]
+rocksdb = ["fuels-test-helpers?/rocksdb"]
```
Braqzen, line 48:
```suggestion
rocksdb = ["fuels-test-helpers?/rocksdb"]

```

GitHub likes its new lines

packages/fuels/.DS_Store
Braqzen, line 1: Oh no, sir, the disease is spreading. We must nuke it before it pollutes the environment any further!

.github/workflows/ci.yml
```diff
@@ -213,7 +213,20 @@ jobs:
...
-        run: cargo ${{ matrix.cargo_command }} -p ${{ matrix.package }} ${{ matrix.args }}
+        run: |
+         if [[ ${{ matrix.package }} == "fuels" ]]; then
+           if [[ ${{ matrix.cargo_command }} == "test" ]]; then
+             cargo test -p fuels --all-targets --features "default fuel-core-lib test-type-paths" --workspace
+           fi
+           if [[ ${{ matrix.cargo_command }} == "clippy" ]]; then
+             cargo clippy -p fuels --all-targets --features "default fuel-core-lib test-type-paths"
+           fi
+           if [[ ${{ matrix.cargo_command }} == "check" ]]; then
```
Braqzen, line 224: Should these be changed to use `elif` instead? I believe all `if` expressions are evaluated; however, when using `elif` if one condition is met then the rest will not be evaluated which saves a bit of computation

examples/contracts/src/lib.rs
```diff
@@ -790,4 +790,72 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    #[ignore]
```
Braqzen, line 795: Should the test be ignored?

examples/contracts/src/lib.rs
```diff
@@ -790,4 +790,72 @@ mod tests {
...
+            SecretKey::from_str(
+                "0x4433d156e8c53bf5b50af07aa95a29436f29a94e0ccc5d58df8e57bdc8583c32",
+            )
+            .unwrap(),
+            None,
+        );
+
+        const NUM_ASSETS: u64 = 2;
+        const AMOUNT: u64 = 225883;
+        const NUM_COINS: u64 = 1;
```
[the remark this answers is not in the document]
Braqzen: I'm also unsure about the value used for `AMOUNT`. Is that a magic number, is it important to use that number? I would either change the value to be something that can be intuited e.g. `100` ("oh that's probably arbitrary") or add a comment to explain the purpose of `225883`.

examples/contracts/src/lib.rs
```diff
@@ -790,4 +790,72 @@ mod tests {
...
+        let (coins, _) =
+            setup_multiple_assets_coins(wallet.address(), NUM_ASSETS, NUM_COINS, AMOUNT);
+        let (provider, _) =
+            setup_test_provider(coins.clone(), vec![], Some(node_config), Some(chain_config)).await;
+
+        // ANCHOR_END: create_rocksdb
+
+        wallet.set_provider(provider);
+
+        let _ = Contract::load_from(
```
Braqzen, line 845: Unsure about the purpose of loading contracts here. We do use the `wallet` which is manipulated above so I guess we want to prove that continuing to use it works as expected.

Thoughts? Should we remove those 2 lines?

examples/cookbook/src/lib.rs
```diff
@@ -186,4 +187,65 @@ mod tests {
...
+        let mut wallet = WalletUnlocked::new_from_private_key(
+            SecretKey::from_str(
+                "0x4433d156e8c53bf5b50af07aa95a29436f29a94e0ccc5d58df8e57bdc8583c32",
+            )
+            .unwrap(),
+            None,
+        );
+
+        wallet.set_provider(provider.clone());
+
```
Braqzen, line 223:
```suggestion
```

examples/cookbook/src/lib.rs
```diff
@@ -186,4 +187,65 @@ mod tests {
...
+            *wallet.get_balances().await?.iter().next().unwrap().1,
+            225883
+        );
+        assert_eq!(
+            *wallet.get_balances().await?.iter().next().unwrap().1,
+            225883
+        );
+        assert_eq!(wallet.get_balances().await?.len(), 2);
+
+        fs::remove_dir_all(path.parent().expect("Db parend folder do not exist"))?;
```
Braqzen, line 247:
```suggestion
        fs::remove_dir_all(path.parent().expect("Db parent folder do not exist"))?;
```

docs/src/providers/rocksdb.md
```diff
@@ -0,0 +1,18 @@
...
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:create_rocksdb}}
+```
+
+To utilize an existing database, follow these instructions:
+
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:use_created_rocksdb}}
+```
+Please note that if the specified database doesn't exist, the provided code will create a new database at the specified path.
```
Braqzen, line 16:
```suggestion

> Note 1: If the specified database does not exist a new database will be created at that path.
```

Braqzen (commented):
If we have multiple tests pointing at the same database does each test have its own isolated version or could 1 test pollute another test by interacting with that db?

examples/contracts/src/lib.rs
Salka1988: #[ignore] can be deleted. The 'rocksdb' flag will not be tested due to its significant impact on build time. Additionally, the testing for 'rocksdb' has already been performed by 'fuel-core'. Thus, there is no need to repeat the testing.

.github/workflows/ci.yml
Salka1988: why not :) Ill do it

packages/fuels/.DS_Store
Salka1988: I didn't know it regenerated. It goes to .gitignore

examples/contracts/src/lib.rs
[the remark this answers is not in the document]
Salka1988: Constants require a type. I can tell you that the number 225883 is quite magical in the Balkans :smiley:. I changed the values, so take a look.

examples/contracts/src/lib.rs
Salka1988: I wanted to test the number of blocks.

Salka1988:
> If we have multiple tests pointing at the same database does each test have its own isolated version or could 1 test pollute another test by interacting with that db?

Each test will use the same database it points to. I'm not sure if you can pollute it. As far as I know, you cannot change blocks or deploy the same contracts. You can increase the number of blocks, but that is not pollution.

examples/cookbook/src/lib.rs
```diff
@@ -236,15 +233,15 @@ mod tests {
         assert_eq!(blocks.len(), 3);
         assert_eq!(
             *wallet.get_balances().await?.iter().next().unwrap().1,
-            225883
+            1000000000
         );
         assert_eq!(
             *wallet.get_balances().await?.iter().next().unwrap().1,
-            225883
+            1000000000
```
Braqzen: I would import the constant instead of hardcoding a large value here

Salka1988: done

Braqzen (approved):
Tentative ACK

packages/fuels/Cargo.toml
segfault-magnet, line 48: This is a bit convoluted. This line means the following:
If `fuel-test-helpers` is enabled, then enable its `rocksdb` feature. 

If we take a look at the `rocksdb` feature definition over in `fuel-test-helpers` it says:
```toml
rocksdb = ["fuel-core", "fuel-core/rocksdb"]
```

So this creates the following situation:
* If you enable `fuel-core-lib` and `rocksdb` features but don't enable `fuels-test-helpers` you're not going to get the `rocksdb` feature enabled on `fuel-core` because `fuels-test-helpers` was in charge of that.

The fix depends on what you decide to do with regard to feature flags in one of my previous comments.

packages/fuels-test-helpers/Cargo.toml
```diff
@@ -34,3 +34,4 @@ which = { workspace = true, default-features = false }
 default = ["fuels-accounts", "std"]
 std = ["fuels-accounts?/std", "fuels-core/std"]
 fuel-core-lib = ["fuel-core"]
+rocksdb = ["fuel-core", "fuel-core/rocksdb"]
```
Voxelot, line 37: You can use a `?` so that rocksdb only takes affect when fuel-core-lib is enabled.

cc @MujkicA 

```suggestion
rocksdb = ["fuel-core?/rocksdb"]
```

examples/cookbook/src/lib.rs
```diff
@@ -186,4 +187,64 @@ mod tests {
...
+        use fuels::accounts::fuel_crypto::SecretKey;
+        use fuels::accounts::wallet::WalletUnlocked;
+        use fuels::client::{PageDirection, PaginationRequest};
+        use fuels::prelude::{setup_test_provider, Config, DbType, ViewOnlyAccount};
+        use std::fs;
+        use std::path::PathBuf;
+        use std::str::FromStr;
+
+        // ANCHOR: use_created_rocksdb
+        let path =
```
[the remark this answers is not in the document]
Salka1988: Maybe we should talk about this because I'm not really sure about this. I did some tests and I don't think this should be a problem. Of course, we can always delete example for loading an old db.

docs/src/providers/short-lived.md
```diff
@@ -29,3 +29,11 @@ The `fuel-core-lib` is a feature defined in the `fuels` library, allowing us to
 ```rust,ignore
 fuels = { version = "0.43.0", features = ["fuel-core-lib"] }
 ```
+
+## RocksDb
+
+The `rocksdb` is a feature defined in the `fuels` library, allowing us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine and creating rocksDb database. Using the `rocksdb` feature flag entails downloading all the dependencies needed to run the fuel-core node and rocksdb.
```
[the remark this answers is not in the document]
Salka1988: Let's talk about this too :calling:

examples/contracts/src/lib.rs
iqdecay: If the point is to test the number of blocks I think it's more straightforward to use the `produce_blocks` method. Also I don't see where you are actually testing it. Otherwise I would remove those lines as well, added maintenance.

examples/cookbook/src/lib.rs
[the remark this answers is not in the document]
iqdecay: Agreed with @segfault-magnet here. But I do think that proving that an old DB can be reloaded and "prove" that it keeps the state is good, just that it should be done in the same test.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -6,8 +6,11 @@ use std::net::SocketAddr;
 #[cfg(feature = "fuels-accounts")]
 pub use accounts::*;
 #[cfg(feature = "fuel-core-lib")]
+pub use fuel_core::service::DbType;
+#[cfg(feature = "fuel-core-lib")]
 use fuel_core::service::FuelService;
 #[cfg(feature = "fuel-core-lib")]
+#[cfg(feature = "fuel-core-lib")]
```
iqdecay, line 13:
```suggestion
```

packages/fuels-test-helpers/src/node.rs
```diff
@@ -26,6 +27,8 @@ use tokio::{process::Command, sync::oneshot};
 
 use crate::utils::{into_coin_configs, into_message_configs};
 
+pub const DEFAULT_CACHE_SIZE: usize = 10 * 1024 * 1024;
```
iqdecay, line 30: What's the unit of this cache size, and where does this default come from?

iqdecay (changes requested):
left some questions

packages/fuels-test-helpers/src/node.rs
Salka1988: It comes from fuel-core

Salka1988:
https://github.com/FuelLabs/fuels-rs/pull/998#discussion_r1238235180

`use_created_db_rocksdb` is testing Db.
We're thinking about removing the `use_created_db_rocksdb` test because it requires running the `create_rocksdb` test first. This order of execution can potentially lead to errors, so we're considering dropping the test altogether.

.github/workflows/ci.yml
```diff
@@ -213,7 +213,18 @@ jobs:
...
+           if [[ ${{ matrix.cargo_command }} == "test" ]]; then
+             cargo test -p fuels --all-targets --features "default fuel-core-lib test-type-paths" --workspace
+           elif [[ ${{ matrix.cargo_command }} == "clippy" ]]; then
+             cargo clippy -p fuels --all-targets --features "default fuel-core-lib test-type-paths"
+           else
+             cargo check -p fuels --features "default fuel-core-lib test-type-paths"
+           fi
+         else
+           cargo ${{ matrix.cargo_command }} -p ${{ matrix.package }} ${{ matrix.args }}
+         fi
```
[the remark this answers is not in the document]
Salka1988: I reverted the CI to the old version and changed --all-features to --features "default fuel-core-lib test-type-paths". The check command is the only one that still uses --all-features, but it only slightly adds to the build time (about 1 minute), which is acceptable.

packages/fuels-test-helpers/src/node.rs
iqdecay: What I mean is that : 
- If this default comes from `fuel-core` then we should use the constant from the `fuel-core` repository. If it's not possible, a comment should explain where this default comes from and why it's there.

Salka1988: The explanation will suffice. We can't import anything from fuel-core because of fuel-core.bin.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -40,9 +43,18 @@ pub enum Trigger {
...
+pub enum DbType {
+    InMemory,
+    RocksDb,
+}
+
 #[derive(Clone, Debug)]
 pub struct Config {
     pub addr: SocketAddr,
+    pub max_database_cache_size: usize,
+    pub database_path: PathBuf,
```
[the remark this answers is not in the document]
Salka1988: Maybe we could do that in new PR?

docs/src/connecting/short-lived.md
```diff
@@ -24,8 +24,17 @@ let wallet = launch_provider_and_get_wallet().await;
...
-The `fuel-core-lib` is a feature defined in the `fuels` library, allowing us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine. Using the `fuel-core-lib` feature flag entails downloading all the dependencies needed to run the fuel-core node.
+The `fuel-core-lib` is a feature that allows us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine. Using the `fuel-core-lib` feature flag entails downloading all the dependencies needed to run the fuel-core node.
 
 ```rust,ignore
 fuels = { version = "0.43.0", features = ["fuel-core-lib"] }
 ```
+
+## RocksDb
+
+The `rocksdb` is a feature that allows us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine and creating rocksDb database.
```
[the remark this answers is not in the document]
Salka1988: You are right. Forgot to update this.

.github/workflows/ci.yml
```diff
@@ -161,10 +161,10 @@ jobs:
             args: --all-targets
             download_sway_artifacts: sway-examples
           - cargo_command: clippy
-            args: --all-targets --all-features
+            args: --all-targets --features "default fuel-core-lib test-type-paths"
             download_sway_artifacts: sway-examples-w-type-paths
           - cargo_command: nextest
-            args: run --all-targets --all-features --workspace
+            args: run --all-targets --features "default fuel-core-lib test-type-paths" --workspace
```
iqdecay, line 167: I don't see why we should stop testing all the other features.

docs/src/connecting/rocksdb.md
```diff
@@ -2,18 +2,12 @@
...
-```rust,ignore
-{{#include ../../../examples/cookbook/src/lib.rs:use_created_rocksdb}}
+{{#include ../../../examples/cookbook/src/lib.rs:create_or_use_rocksdb}}
 ```
 
 > Note 1: If the specified database does not exist a new database will be created at that path.
 
-> Note 2: To utilize the code snippets above, either the `fuel-core` binary must be present or the `rocksdb` feature needs to be used.
\ No newline at end of file
+> Note 2: To utilize the code snippets above, either the `fuel-core` binary must be present or the `fuel-core-lib, rocksdb` features need to be enabled.
```
iqdecay, line 13:
```suggestion
> Note 2: To utilize the code snippets above, either:
- the `fuel-core` binary must be present
or
- both the `fuel-core-lib` and `rocksdb` features need to be enabled.
```

iqdecay: It does sound like you can use the `rocksdb` code without enabling the `rocksdb` feature if you have a `fuel-core` binary present. If that's true, I think we should prevent this, and make it only available if `rocksdb` is enabled.

examples/cookbook/src/lib.rs
```diff
@@ -190,60 +187,21 @@ mod tests {
...
-            DEFAULT_COIN_AMOUNT
-        );
-        assert_eq!(
-            *wallet.get_balances().await?.iter().next().unwrap().1,
-            DEFAULT_COIN_AMOUNT
-        );
-        assert_eq!(wallet.get_balances().await?.len(), 2);
-
-        fs::remove_dir_all(path.parent().expect("Db parent folder does not exist"))?;
+        let _wallets =
```
iqdecay, line 202: This part of the test does not do anything, I think you can remove it.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -26,7 +26,7 @@ use tempfile::NamedTempFile;
 use tokio::{process::Command, sync::oneshot};
 
 use crate::utils::{into_coin_configs, into_message_configs};
-
+// Set the cache for tests = 10MB
```
iqdecay, line 29:
```suggestion
// Set the cache for tests = 10MB, which is the default size in `fuel-core`.
```

packages/fuels/tests/contracts.rs
iqdecay, line 1: I think you can reorg the test better with some comments about the different parts of the test.

iqdecay (changes requested):
Left some remarks, looking good!

.github/workflows/ci.yml
Salka1988: https://github.com/FuelLabs/fuels-rs/pull/998#discussion_r1226057805

Salka1988: I think that all other features are being tested because each of our packages has a `default` and exports all its features through it. 🤔

examples/cookbook/src/lib.rs
Salka1988: Forgot to delete that from example.

docs/src/connecting/rocksdb.md
Salka1988: Why?We explained when to use `rocksdb` feature.

docs/src/connecting/rocksdb.md
```diff
@@ -0,0 +1,13 @@
+## RocksDb
+
+RocksDb enables the preservation of the blockchain's state locally, facilitating its future utilization.
+
+To create or to use a local database, follow these instructions:
```
digorithm, line 5:
```suggestion
To create or use a local database, follow these instructions:
```

docs/src/connecting/rocksdb.md
```diff
@@ -0,0 +1,13 @@
...
+
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:create_or_use_rocksdb}}
+```
+
+> Note 1: If the specified database does not exist a new database will be created at that path.
```
digorithm, line 11:
```suggestion
> Note 1: If the specified database does not exist, a new database will be created at that path.
```

docs/src/connecting/short-lived.md
```diff
@@ -24,8 +24,16 @@ let wallet = launch_provider_and_get_wallet().await;
 
 ## Fuel-core lib
 
-The `fuel-core-lib` is a feature defined in the `fuels` library, allowing us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine. Using the `fuel-core-lib` feature flag entails downloading all the dependencies needed to run the fuel-core node.
+The `fuel-core-lib` is a feature that allows us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine. Using the `fuel-core-lib` feature flag entails downloading all the dependencies needed to run the fuel-core node.
```
digorithm, line 27: Conciseness nit.

```suggestion
The `fuel-core-lib` feature allows us to run a `fuel-core` node without installing the `fuel-core` binary on the local machine. Using the `fuel-core-lib` feature flag entails downloading all the dependencies needed to run the fuel-core node.
```

packages/fuels-test-helpers/src/node.rs
```diff
@@ -25,6 +26,8 @@ use tempfile::NamedTempFile;
 use tokio::{process::Command, sync::oneshot};
 
 use crate::utils::{into_coin_configs, into_message_configs};
+// Set the cache for tests = 10MB, , which is the default size in `fuel-core`.
```
digorithm, line 29:
```suggestion
// Set the cache for tests to 10MB, which is the default size in `fuel-core`.
```

packages/fuels/tests/contracts.rs
digorithm: This might be a good suggestion, but could you make it more concrete?

packages/fuels/tests/contracts.rs
```diff
@@ -1321,3 +1321,124 @@ async fn low_level_call() -> Result<()> {
...
+                setup_test_provider(coins.clone(), vec![], Some(node_config), Some(chain_config))
+                    .await;
+
+            provider.produce_blocks(2, None).await?;
+
+            Ok::<(), Box<dyn std::error::Error>>(())
+        })
+        .unwrap();
+
+    // I had to terminate the runtime as I encountered an issue while attempting to start two nodes within the same runtime that utilize the same RocksDB.
```
digorithm, line 1391: Do we need to keep this comment?

digorithm (changes requested):

.github/workflows/ci.yml
```diff
@@ -161,10 +161,10 @@ jobs:
             args: --all-targets
             download_sway_artifacts: sway-examples
           - cargo_command: clippy
-            args: --all-targets --all-features
+            args: --all-targets --features "default fuel-core-lib test-type-paths"
             download_sway_artifacts: sway-examples-w-type-paths
           - cargo_command: test
-            args: --all-targets --all-features --workspace
+            args: --all-targets --features "default fuel-core-lib test-type-paths" --workspace
```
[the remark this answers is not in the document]
digorithm: I gave some thought to this after our call... Not testing it ever is just... not good. But adding this much time to our CI runtime will hinder productivity. Let's test this at _release_ time.

packages/fuels/tests/contracts.rs
Salka1988: https://github.com/FuelLabs/fuels-rs/pull/998#discussion_r1252143393
Maybe it is better to keep it.

docs/src/connecting/rocksdb.md
iqdecay: What I mean is that the current phrasing is ambiguous, because it sounds like if I just have `the fuel-core binary present` then I can use `rocksdb` without enabling the `rocksdb` feature.

Salka1988: Yes, it should behave that way. There is no need to add a flag if we don't need it.

iqdecay: Oh ok I didn't understand that was expected behavior.

docs/src/connecting/rocksdb.md
```diff
@@ -0,0 +1,16 @@
...
+```rust,ignore
+{{#include ../../../examples/cookbook/src/lib.rs:create_or_use_rocksdb}}
+```
+
+> Note 1: If the specified database does not exist, a new database will be created at that path.
+
+> Note 2: To utilize the code snippets above, either:
+- the `fuel-core` binary must be present
+or
+- both the `fuel-core-lib` and `rocksdb` features need to be enabled.
```
segfault-magnet, line 16: 2023-07-11-144045_828x157_scrot

The note doesn't span the whole suggestion.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -230,12 +246,41 @@ pub async fn new_fuel_node(
...
+        if let DbType::RocksDb = config.database_type {
+            if config.database_path.as_os_str().is_empty() {
+                args.extend(vec![
+                    "--db-path".to_string(),
+                    PathBuf::from(std::env::var("HOME").expect("HOME env var missing"))
+                        .join(".fuel/db")
+                        .to_string_lossy()
+                        .to_string(),
+                ]);
+            } else if !config.database_path.as_os_str().is_empty() {
```
segfault-magnet, line 271: You don't have to check the condition again, just leave the `else`.

packages/fuels-test-helpers/src/node.rs
```diff
@@ -230,12 +246,41 @@ pub async fn new_fuel_node(
...
+                        .to_string_lossy()
+                        .to_string(),
+                ]);
+            } else {
+                args.extend(vec![
+                    "--db-path".to_string(),
+                    config.database_path.to_string_lossy().to_string(),
+                ]);
+            }
+        }
```
segfault-magnet, line 277:
```rust
        if let DbType::RocksDb = config.database_type {
            let path = if config.database_path.as_os_str().is_empty() {
                PathBuf::from(std::env::var("HOME").expect("HOME env var missing")).join(".fuel/db")
            } else {
                config.database_path
            };
            args.extend(vec![
                "--db-path".to_string(),
                path.to_string_lossy().to_string(),
            ]);
        }

```

segfault-magnet (dismissed):
a small suggestion

segfault-magnet (approved):

Br1ght0ne (approved):

digorithm (approved):
Great work, @Salka1988!

hal3e (approved):

PR #1012, opened 2023-06-22
feat: check fuel-core dependency version

Salka1988:
Added a function that checks the `fuel-core` version in the published `fuels` crate against our current version. If they don't match, a message prompts the user to align the versions to avoid potential errors. 
At the moment, I'm only comparing the major and minor, and I left out the patch.

### Checklist


packages/fuels-test-helpers/src/utils.rs
```diff
@@ -14,3 +18,180 @@ pub fn into_message_configs(messages: Vec<Message>) -> Vec<MessageConfig> {
...
+                };
+
+                let fuel_core_req = fuel_core_req_vec
+                    .split('.')
+                    .take(2)
+                    .collect::<Vec<&str>>()
+                    .join(".");
+
+                if fuel_core_req != used_fuel_version {
+                    eprintln!(
```
segfault-magnet, line 183: We shouldn't spam `stderr`. Any message you wish to relay is best done through logging at the appropriate level.

segfault-magnet (commented):
I'd dump the support for `fuel-core-lib` or, better yet, implement the suggestions I made in the comments regarding getting a version out of `fuel-core`. 

This will cover only the users that are testing locally. Those testing against the testnet will still continue to ask questions.

We can do one of the following:
1. implement some kind of heuristic to try and give helpful suggestions to the user in the case of a version mismatch. Something as simple as `if err.contains( 'no field in request' )` might go a long way towards reducing user confusion.
2. program some sort of compatibility check in the SDK with an endpoint on the node. We might look up the version returned from fuel core and cross-reference our own compatibility table. 

Although we have no such table, a simple script that walks backward through the git history of every release can extract which pairs are known to work together. That alongside relying a bit on `fuel-core` respecting semver (most of the time) might be good enough.

iqdecay:
Waiting on some refactoring first before reviewing.

Salka1988:
> Waiting on some refactoring first before reviewing.

My mistake. This was supposed to be a draft through which I would exchange some ideas with you.

iqdecay:
@FuelLabs/sdk-rust what do we do with this

Salka1988:
> @FuelLabs/sdk-rust what do we do with this

We need this
 https://github.com/FuelLabs/fuels-rs/pull/1067

PR #1019, opened 2023-06-27
Docs Nits

SilentCicero:
A set of small docs nits.

1. `.into` for ContractID


digorithm (commented):
I _think_ that shouldn't be the case anymore. In the latest version (https://github.com/FuelLabs/fuels-rs/releases/tag/v0.43.0) we changed it from needing `.into()` to not needing it at all: 
image

Maybe because you had to use `0.41` you ran into this issue, since you were reading the latest docs!

But keep the docs nits coming if you've got some more!

PR #1022, opened 2023-06-29
docs: update testnet to beta-3 version

Salka1988:
### Checklist

### DIFF
--- a/docs/src/connecting/external-node.md
+++ b/docs/src/connecting/external-node.md
@@ -5,6 +5,10 @@ We can interact with the `Testnet` node by using the following example.
 ```rust,ignore
 {{#include ../../../examples/providers/src/lib.rs:connect_to_testnet}}
 ```
+>
+> For detailed information about various testnet networks and their optimal toolchain configurations for your project, please visit the following link:
+>
+> [networks](https://fuelbook.fuel.network/master/networks/networks.html)
 
 In the code example, we connected a new provider to the Testnet node and created a new wallet from a private key.
 
@@ -17,6 +21,8 @@ In the code example, we connected a new provider to the Testnet node and created
 > In addition to the faucet, there is a block explorer for the Tesnet at
 >
 > [block-explorer](https://fuellabs.github.io/block-explorer-v2)
+
+
 
 If you want to connect to another node just change the url or IP and port. For example, to connect to a local node that was created with `fuel-core` you can use:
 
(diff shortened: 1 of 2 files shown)


digorithm (approved):

hal3e (dismissed):

segfault-magnet (dismissed):

iqdecay (changes requested):
AFAIK the endopoint currently does not work with the SDK because the endpoint uses `fuel-core 0.17`, unless it was updated. If it doesn't work, we should document it in the docs. Also there are several other instances of the `node-beta-2.fuel.network` string accross the codebase, those should be changed too (even if they don't work).

segfault-magnet:
This doesn't seem blocked? 

I'd update all references to beta-3 and add a note specifying what is the latest version that does work with beta-3, as @iqdecay said. That way whomever reads the latest docs will be guided to the right `fuels` for the testnet.

docs/src/connecting/external-node.md
```diff
@@ -5,19 +5,25 @@ We can interact with the `Testnet` node by using the following example.
 ```rust,ignore
 {{#include ../../../examples/providers/src/lib.rs:connect_to_testnet}}
 ```
+>
+> For detailed information about various testnet networks and their optimal toolchain configurations for your project, please visit the following link:
+>
+> [networks](https://fuelbook.fuel.network/master/networks/networks.html)
```
iqdecay, line 11:
```suggestion
> **Note:** 
This network will only work with the versions of the toolchain specified [here](https://fuelbook.fuel.network/master/networks/networks.html).
> It will **not** work with versions other than the ones listed above. 
```

iqdecay (changes requested):
I think we need to make it more clear that it's not going to work unless you use the proper toolchain version. I added a suggestion, feel free to change it.

docs/src/connecting/external-node.md
Salka1988: There are certain compatible versions that are not included in the list provided on the Fuel Network's website at https://fuelbook.fuel.network/master/networks/networks.html.

iqdecay: Can you please just write a sentence that explains what the user should do in order to have something that works.

Salka1988: Isn't this part enough:`... their optimal toolchain configurations for your project , please visit the following link`.
If user clicks on the link, he will find information about all testnets and the recommended crate versions to use.
Versions specified on the websites are correct.

iqdecay: I agree that it does say the truth, only we should double check the user understands that it won't work unless she uses the versions specified. Anyway I feel like I'm nitpicking on wording lol.

iqdecay (approved):

digorithm (approved):

hal3e (approved):

Issue #1020, opened 2023-06-28
Retry mechanism


digorithm:
We should support an optional, configurable parameter to enable retries when sending a transaction to the node. This mechanism's design and API are part of this work; nothing in mind yet!

PR #1035, opened 2023-07-09
feat: add retry mechanism

Salka1988:
Closes https://github.com/FuelLabs/fuels-rs/issues/1020. 

This code is an implementation of a retry mechanism that allows a function to be retried a certain number of times with a specified interval between attempts.

`RetryConfig`: This is a configuration struct that holds the retry configuration. It contains the following fields:

- max_attempts: The maximum number of times the action should be retried.
- interval:  Determines the backoff strategy and timing to be used during retry attempts. It allows you to configure how much time a system should wait before retrying an action after a failure. 

`retry<Fut, T, ShouldRetry>`: This is the main function responsible for implementing the retry logic.

This function takes several parameters:

- action: A closure that performs the action you want to retry. It returns a Future of the result.
- retry_config: A reference to a `RetryConfig` struct that specifies the retry behavior.
- should_retry: A closure that takes a reference to the result of the action and returns a boolean indicating whether the action should be retried based on the result.

Aditional Work:
I've separated sending a transaction from getting back its value:
```rust
  let response = contract_methods.get(1, 2).submit().await?;
  let value = response.response().await?;
```

Added `RetryableClient` a wrapper around the `FuelClient` that provides a convenient way to interact with a remote service. It also enables automatic retry functionality in scenarios where network issues may occur.

```rust
#[derive(Debug, Clone)]
pub(crate) struct RetryableClient {
    client: FuelClient,
    url: String,
    retry_config: RetryConfig,
}
```

### Checklist


Br1ght0ne (changes requested):
When designing a retry mechanism for anything with a notion of “network call” in some way, I like implementing a delay to allow the network to self-regulate its throughput and often requiring less attempts overall.

I usually go for [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) - delays start off small and ramp up steadily. There are crates like [`backon`](https://lib.rs/crates/backon) with full support of sync/async functions.

Alternatives:
- hand-roll our own delayed retrying (only if adding a dependency is a problem)
- skip having delays altogether (I hereby request your comments on it)

Salka1988:
> Great UX and overall approach! I have a small consideration regarding the logic itself.
> 
> When designing a retry mechanism for anything with a notion of “network call” in some way, I like implementing a delay to allow the network to self-regulate its throughput when congested and often requiring less attempts overall.
> 
> I usually go for [exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) - delays start off small and ramp up steadily. There are crates like [`backon`](https://lib.rs/crates/backon) and [`backoff`](https://lib.rs/crates/backoff) with full support of sync/async functions.
> 
> Extra considerations:
> 
> * what are acceptable minimum/maximum delays for our use case?
> 
> Alternatives:
> 
> * hand-roll our own delayed retrying (only if adding a dependency is a problem)
> * skip having delays altogether (I hereby request your comments on it)

I will use one of the crates you recommended. If that's a problem, I'll implement `exponential backoff'.

segfault-magnet (commented):
Not all errors benefit from retrying. Usually, you want to wait out a transient error (network or some other kind) and not actually offer a fully-fledged retry mechanism to the user especially when retrying again so fast.

For example, if the tx is out of gas it will be reverted and you'll retry again most probably changing nothing (and more importantly - wasting gas).

segfault-magnet:
> > Not all errors benefit from retrying. Usually, you want to wait out a transient error (network or some other kind) and not actually offer a fully-fledged retry mechanism to the user especially when retrying again so fast.
> > For example, if the tx is out of gas it will be reverted and you'll retry again most probably changing nothing (and more importantly - wasting gas).
> 
> You are right, I understand you but will we waste gas if we fail?

Yep. You lose the gas used to execute the contract up until the point of reverting. The rest will be refunded if you provided an adequate change output.

segfault-magnet:
> > Not all errors benefit from retrying. Usually, you want to wait out a transient error (network or some other kind) and not actually offer a fully-fledged retry mechanism to the user especially when retrying again so fast.
> > For example, if the tx is out of gas it will be reverted and you'll retry again most probably changing nothing (and more importantly - wasting gas).
> 
> You are right, I understand you but will we waste gas if we fail?

Yep. You lose the gas used to execute the contract up until the point of reverting. The rest will be refunded if you provided an adequate change output.

Br1ght0ne:
Thanks for your consideration @Salka1988! I agree with @segfault-magnet - saving on gas is important, and a proper `max_attempts` should be set to avoid excessive retrying. I dismissed my blocking review, and will approve whatever logic we agree on (if the code is correct 😉)

Salka1988:
> > > Not all errors benefit from retrying. Usually, you want to wait out a transient error (network or some other kind) and not actually offer a fully-fledged retry mechanism to the user especially when retrying again so fast.
> > > For example, if the tx is out of gas it will be reverted and you'll retry again most probably changing nothing (and more importantly - wasting gas).
> > 
> > 
> > You are right, I understand you but will we waste gas if we fail?
> 
> Yep. You lose the gas used to execute the contract up until the point of reverting. The rest will be refunded if you provided an adequate change output.

Yea! I forgot about `simulate()`

packages/fuels-programs/src/contract.rs
```diff
@@ -476,6 +479,29 @@ where
             .map_err(|err| map_revert_error(err, &self.log_decoder))
     }
 
+    pub async fn call_with_retry(
```
iqdecay, line 482: these functions should have docstrings

packages/fuels-test-helpers/src/node.rs
```diff
@@ -60,6 +60,7 @@ pub struct Config {
     pub block_production: Trigger,
     pub vm_backtrace: bool,
     pub silent: bool,
+    pub chain_conf: ChainConfig,
```
iqdecay, line 63: call it `chain_config` I think, we are not using `conf` anywhere in the codebase.

packages/fuels-programs/src/contract.rs
```diff
@@ -693,6 +719,29 @@ impl<T: Account> MultiContractCallHandler<T> {
             .map_err(|err| map_revert_error(err, &self.log_decoder))
     }
 
+    pub async fn call_call_with_retry<D: Tokenizable + Debug>(
```
iqdecay, line 722: why not just `call_with_retry`? For `MultiContractCallHandler` we use the `call` method, the same as `ContractCallHandler` so I don't see why we would have different names here.

packages/fuels-programs/src/contract.rs
```diff
@@ -476,6 +479,29 @@ where
...
+        max_attempts: Option<u64>,
+    ) -> Result<FuelCallResponse<D>> {
+        let attempts = max_attempts.unwrap_or(0);
+        let mut delay = Duration::from_secs(1);
+
+        for _ in 1..=attempts {
+            match self.call_or_simulate(false).await {
+                Ok(response) => return Ok(response),
+                Err(ProviderError(description))
+                    if description.starts_with("Client request error") =>
```
iqdecay, line 493: I think we're better off matching with the `ClientRequestError` variant rather than a string description that can change.

packages/fuels/tests/contracts.rs
iqdecay, line 1: why are those tests commented out?

iqdecay (changes requested):
looking good, left a few questions.

packages/fuels-accounts/src/provider.rs
```diff
@@ -355,14 +355,15 @@ impl Provider {
...
+        let res: Vec<_> = self
             .client
             .coins_to_spend(
                 &filter.owner(),
                 queries.spend_query(),
                 queries.exclusion_query(),
             )
-            .await?
+            .await
+            .map_err(ProviderError::ClientRequestError)?
```
segfault-magnet, line 366:
```suggestion
            .await?
```

packages/fuels-programs/src/contract.rs
segfault-magnet: I'd also create a `RetryConfig` with some defaults and place the now hardcoded interval inside of there alongside the max attempts.

packages/fuels-programs/src/contract.rs
segfault-magnet: Yep. I'd leave a match statement that retries only if the provider error was network related. 

Not sure how fine-tuned you can make that. For example, schema incompatibility between our gql client and the server might also appear as the same type of error (since fuel gql client uses `anyhow`). I suggest digging a bit to see whether you can make any sort of robust claim as to the transitive nature of the error.

You can always send a PR over to the client for more library-friendly errors.

packages/fuels-programs/src/contract.rs
segfault-magnet: Probably a paste-ing mishap.

packages/fuels-programs/src/contract.rs
Salka1988: I did that, but I tried to add some functionality like  `on_retry: Option<Box<dyn Fn(usize) -> std::pin::Pin<Box<dyn std::future::Future<Output=()>>> + Send + Sync>>` function that would allow us to change certain contract parameters in a certain repetition or similar. I added this primarily to test certain things, but everything got a little complicated.:goberserk:
 Does it even make sense for me to add this functionality

packages/fuels-test-helpers/src/node.rs
Salka1988: This was added for testing purposes. I don't think it will be necessary.

packages/fuels-programs/src/contract.rs
Salka1988: You're right, but this is currently the "only" way to do matching. I should talk to the client team.

Salka1988:
We should wait for #1073 to merge before we continue with the implementation.

packages/fuels-accounts/src/provider.rs
```diff
@@ -144,6 +144,13 @@ impl Provider {
     }
 
     /// Sends a transaction to the underlying Provider's client.
+    pub async fn send_transaction_and_wait_to_commit<T: Transaction>(&self, tx: T) -> Result<TxId> {
```
MujkicA, line 147: Nit: it seems a bit wordy, maybe `send_transaction_and_await`?

packages/fuels-programs/src/contract.rs
```diff
@@ -574,12 +615,27 @@ where
             .map_err(|err| map_revert_error(err, &self.log_decoder))
     }
 
-    pub async fn submit(mut self) -> Result<ContractCallHandler<T, D>> {
+    pub async fn submit(mut self) -> Result<SubmitResponse<T, D>> {
```
MujkicA, line 618: If you go with the suggestion about send/submit in the provider, we should align this to say send as well

packages/fuels-test-helpers/src/node.rs
MujkicA: Reminder to remove it

docs/src/running-scripts.md
```diff
@@ -37,16 +45,3 @@ And this is an example that uses `set_contract_ids(&[&contract_id, ...])`.
 ```rust,ignore
 {{#include ../../packages/fuels/tests/logs.rs:external_contract_ids}}
 ```
-
-## Configurable constants
```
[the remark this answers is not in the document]
Salka1988: mistake!

packages/fuels-programs/src/contract.rs
```diff
@@ -462,13 +465,51 @@ pub struct ContractCallHandler<T: Account, D> {
     pub account: T,
     pub datatype: PhantomData<D>,
     pub log_decoder: LogDecoder,
+    pub retry_config: RetryConfig,
+}
+
+// Implement Clone for ContractCallHandler once all types are provided.
+impl<T: Account, D> Clone for ContractCallHandler<T, D> {
```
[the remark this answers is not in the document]
Salka1988: I needed this until @hal3e  joined one of his PRs. It stayed like this because it didn't cause me a problem, so I forgot about it. Now we can do as you said. `Clone`

packages/fuels-programs/src/retry.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+}
+
+impl Backoff {
+    pub fn wait_duration(&self, attempt: usize) -> Duration {
+        match self {
+            Backoff::Linear(base_duration) => *base_duration * (attempt) as u32,
+            Backoff::Exponential(base_duration) => {
+                *base_duration * (2_usize.pow((attempt) as u32)) as u32
+            }
```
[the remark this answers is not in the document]
Salka1988: Nice! Thanx :)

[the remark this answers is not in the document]
MujkicA: What should be the expected behaviour if someone enters 0?

packages/fuels-programs/src/submit_response.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+    }
+
+    pub fn with_retry_config(mut self, retry_config: RetryConfig) -> Self {
+        self.retry_config = retry_config;
+        self
+    }
+
+    pub async fn value(self) -> Result<D> {
+        self.process_value().await
```
MujkicA, line 85: Why do we need this indirection

packages/fuels-programs/src/retry.rs
```diff
@@ -0,0 +1,444 @@
...
+) -> Result<T, K>
+where
+    T: Clone + Debug,
+    Fut: Future<Output = Result<T, K>>,
+    K: Clone + Error + 'static,
+    ShouldRetry: Fn(&Result<T, K>) -> bool,
+{
+    let mut last_err = None;
+    let max_attempts = retry_config.max_attempts;
+    for attempt in 0..max_attempts {
```
MujkicA, line 163: If the count starts at 0 here, and the strategy is `Linear`, the first retry will occur with no delay. Is this expected of linear backoff?

packages/fuels-programs/src/submit_response.rs
Salka1988: We dont need it.

Br1ght0ne (approved):
Very thorough and well-designed PR. LGTM!

packages/fuels-programs/src/retry.rs
```diff
@@ -0,0 +1,455 @@
...
+            Ok(value) if !should_retry(&result) => return Ok(value),
+            Err(error) if should_retry(&result) => last_err = Some(error.clone()),
+            Err(error) => return Err(error),
+            _ => (),
+        }
+
+        tokio::time::sleep(retry_config.interval.wait_duration(attempt)).await;
+    }
+
+    last_err.map_or_else(
```
MujkicA, line 167: Can we have something like if attempts == 0 -> Error on the top of the method.
IMO it's more readable that way

Salka1988: Ok

packages/fuels-programs/src/retry.rs
```diff
@@ -0,0 +1,455 @@
...
+
+            Ok(())
+        }
+
+        #[tokio::test]
+        async fn retry_on_io_error() -> anyhow::Result<()> {
+            let values = Mutex::new(vec![
+                Ok(TxId::from_str(
+                    "0x98f01c73c2062b55bba70966917a0839995e86abfadfff24534262d1c8b7a64e",
+                )),
```
hal3e, line 310: maybe we can extract this to a `tx_id` variable and later just use `assert_eq!(ok, tx_id);`
```suggestion
            let tx_id = TxId::from_str(
                "0x98f01c73c2062b55bba70966917a0839995e86abfadfff24534262d1c8b7a64e",
            );
            let values = Mutex::new(vec![
                Ok(tx_id),
```

packages/fuels-programs/src/retry.rs
```diff
@@ -0,0 +1,455 @@
...
+            let retry_options = RetryConfig::new(3, Backoff::Linear(Duration::from_millis(10)));
+
+            let ok = retry(will_always_fail, &retry_options, should_retry_fn).await?;
+
+            assert_eq!(
+                ok,
+                TxId::from_str(
+                    "0x98f01c73c2062b55bba70966917a0839995e86abfadfff24534262d1c8b7a64e"
+                )
+            );
```
hal3e, line 333:
```suggestion
            assert_eq!(ok, tx_id);
```

packages/fuels-programs/src/script_calls.rs
```diff
@@ -242,11 +264,28 @@ where
...
+        self.cached_tx_id = Some(if self.retry_config.max_attempts != 0 {
+            retry(
+                || async { provider.send_transaction(tx.clone()).await },
+                &self.retry_config,
+                should_retry_fn,
+            )
+            .await?
+        } else {
+            provider.send_transaction(tx).await?
+        });
```
hal3e, line 283: take it if you like it:
```suggestion
        self.cached_tx_id = match self.retry_config.max_attempts {
            0 => provider.send_transaction(tx).await,
            _ => {
                retry(
                    || async { provider.send_transaction(tx.clone()).await },
                    &self.retry_config,
                    should_retry_fn,
                )
                .await
            }
        }
        .map(Some)?;
```

hal3e (changes requested):
This is such a great improvement! I love the fact that we can configure how we retry our calls. The code and test are awesome! Great work!

I have mad many suggestions but they are actually just small things that, if you decide, can be added very easily. Additionally, I have added all suggestions on all lines because I wanted to make it easier for you to accept them if you decided. I did this because I was already making the changes and checking if everything works so instead of making a new PR/commit I have just added them as suggestions.

Br1ght0ne (approved):

segfault-magnet (dismissed):
Reviewed and gave feedback live with @Salka1988 .

Salka1988:
> Reviewed and gave feedback live with @Salka1988 .

@segfault-magnet  creativity and expertise truly elevated the outcome. Thanx!

Br1ght0ne (approved):

hal3e (approved):
Awesome work!

PR #997, opened 2023-06-11
deps: upgrade `fuel-core` and `fuel-vm` crates versions

iqdecay:
This PR is here to adapt the SDK to :
- `fuel-core@0.19` 
- `fuel-vm@0.34`

Issue #881, opened 2023-03-12
Remove backward compatibility for ABI custom types without paths


segfault-magnet:
Currently, type-paths are feature-gated behind the `forc` flag `--json-abi-with-callpaths`. 

Once the TS SDK implements support for type paths, `forc` will make this behavior default. 

We can then remove the backward compatibility code from `fuels-rs`. Most, if not all of the places that should be looked at have a `TODO` referencing this issue.

hal3e:
Is there an issue on the TS side that we can track ?

hal3e:
Issue to track on the ts side: https://github.com/FuelLabs/fuels-ts/issues/2162

PR #1042, opened 2023-07-11
feat!: add `String` support

### LINKED ISSUE #1011: feat: support `String` type from `sway-libs`
The [sway std-lib crate](https://github.com/FuelLabs/sway/blob/master/sway-lib-std/src/string.sw) added support for dynamically-sized string types, we should support it as a type in the SDK.

iqdecay:
**Note:** this PR is dependent on Sway `v0.42`, and therefore will need #997 before being effective (`fuel-core 0.19`).

This PR closes #1011 by adding support for Sway's `std::string::String` type.
- The `ParamType` introduced here is called `ParamType::StdString` because `ParamType::String` represents statically-sized Sway strings. I think it would be better to use `String` for the dynamic type, and `StaticString` for statically-sized ones, but I think this refactoring would make the PR less clear.
- <s>I need to figure out what's wrong with how I encode the `String` data, this is why the test is weird. </s>


segfault-magnet (commented):
Beautifully done!!

As for the open sway issue, you can test your understanding of the encoding by leaving the SDK out and using a plain ol' sway script calling a sway contract via `forc`. 

Log the received data and do these assertions and see if they hold up.

Not approving yet since the issue might have to do with our encoding (albeit that seems unlikely if String really just uses `Bytes` under the hood).

iqdecay:
Marking as draft until I figure out what is wrong with my encoding.

hal3e (approved):
Great work!

Salka1988 (approved):

packages/fuels-core/src/codec/abi_encoder.rs
```diff
@@ -1171,4 +1173,27 @@ mod tests {
...
+        let string = String::from("This ");
+        let token = Token::StdString(string);
+        let offset = 40;
+
+        // act
+        let encoded_std_string = ABIEncoder::encode(&[token])?.resolve(offset);
+
+        // assert
+        let ptr = [0, 0, 0, 0, 0, 0, 0, 64];
+        let cap = [0, 0, 0, 0, 0, 0, 0, 8];
```
MujkicA, line 1189: Curious, why is the cap 8 here?

packages/fuels-core/src/types/param_types.rs
```diff
@@ -1452,6 +1468,25 @@ mod tests {
...
+    fn try_std_string_correctly_resolves_param_type() {
+        let the_type = given_type_with_path("std::string::String");
+
+        let param_type = try_std_string(&the_type).unwrap().unwrap();
+
+        assert_eq!(param_type, ParamType::StdString);
+    }
+
+    #[test]
+    fn try_std_string_is_type_path_backward_compatible() {
```
MujkicA, line 1481: #881 
Just mentioning this here so it is visible on the relevant issue

MujkicA (approved):
LGTM

PR #1044, opened 2023-07-17
feat!: add support for string slices

IGI-111:
Add support for `str` in prevision for https://github.com/FuelLabs/sway/pull/4778

### Checklist

### DIFF
--- a/packages/fuels-core/src/codec/abi_encoder.rs
+++ b/packages/fuels-core/src/codec/abi_encoder.rs
@@ -42,7 +42,8 @@ impl ABIEncoder {
             Token::B256(arg_bits256) => vec![Self::encode_b256(arg_bits256)],
             Token::Array(arg_array) => Self::encode_array(arg_array)?,
             Token::Vector(data) => Self::encode_vector(data)?,
-            Token::String(arg_string) => vec![Self::encode_string(arg_string)?],
+            Token::StringSlice(arg_string) => Self::encode_string_slice(arg_string)?,
+            Token::StringArray(arg_string) => vec![Self::encode_string_array(arg_string)?],
             Token::Struct(arg_struct) => Self::encode_struct(arg_struct)?,
             Token::Enum(arg_enum) => Self::encode_enum(arg_enum)?,
             Token::Tuple(arg_tuple) => Self::encode_tuple(arg_tuple)?,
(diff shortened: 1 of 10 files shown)


packages/fuels-core/src/codec/abi_encoder.rs
```diff
@@ -513,6 +522,48 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn encode_function_with_string_slice_type() -> Result<()> {
```
digorithm, line 526: Just to make it clear: right now, we cannot have tests that rely on `forc`-generated bytecode (e.g., a contract using this new `str` slice) because this is still a WIP in the compiler, right? I guess these types of tests would be a future work, once this feature is released in the compiler.

IGI-111: That's right. Unfortunately SDK and Sway tests are mutually dependent, so we have to introduce something backwards compatible here first to complete work on the compiler side.

The longer term idea is to deprecate string arrays in favor of string slices anyhow, so the tests will indeed have to be updated here once string slices land on the compiler side.

digorithm (approved):
LGTM -- thanks for taking the time to do this!

Also, we should update [this section of the documentation](https://rust.fuel.network/v0.44.0/types/string.html) with this new distinction between string slices and arrays. Can be done in this PR or follow-up work.

iqdecay:
How does this interact with/nullify #1042?

digorithm:
> How does this interact with/nullify #1042?

I was also wondering about this. It seems like your work on #1042 is for a feature that already exists in the language. And what @IGI-111 is working on is for a future feature. Not sure how these two features interact in the compiler, or if one replaces the other. 

@IGI-111, let us know your thoughts.

iqdecay (approved):
very neat, thanks a lot!

packages/fuels-core/src/types/core/sized_ascii_string.rs
```diff
@@ -2,6 +2,73 @@ use std::fmt::{Debug, Display, Formatter};
...
+#[derive(Debug, PartialEq, Clone, Eq)]
+pub struct AsciiString {
+    data: String,
+}
+
+impl AsciiString {
+    pub fn new(data: String) -> Result<Self> {
+        if !data.is_ascii() {
+            return Err(error!(InvalidData,
+                "SizedAsciiString must be constructed from a string containing only ascii encodable characters. Got: {data}"
```
segfault-magnet, line 16:
```suggestion
                "AsciiString must be constructed from a string containing only ascii encodable characters. Got: {data}"
```

digorithm (approved):

Salka1988 (approved):

hal3e (approved):

Issue #1060, opened 2023-07-25
`abigen` and `setup_program_test` macros don't detect file changes


segfault-magnet:
Incremental rust compilation will not recompile macros when the files used inside are changed. 

Although this is still an open issue in Rust (see [here](https://github.com/rust-lang/rust/issues/99515)) there might be a workaround to tie us over.

Unless fixed this might prove frustrating to users who are changing the contract/script code and have their compile fail because the macro didn't reevaluate (missing types, wrong contract ids, etc).

PR #1061, opened 2023-07-25
feat: macro recompile trigger

segfault-magnet:
closes: #1060 

The solution (until Rust has an official way of doing it) is to call `include_bytes` inside your macro. This is detected by the compiler and it will watch the path for changes. 

I've assigned it to a `const _: &[u8] = include_bytes!(...)` and it seems it gets optimized away. 

Couldn't find a trace of the JSON file in the resulting executable (even under debug build). 

So we might have gotten away with a no-downsides hack.

Tested via a separate e2e suite. Didn't include it in the PR since it would require compiling the whole SDK twice during testing. 

### Checklist

### DIFF
--- a/packages/fuels-code-gen/src/utils/source.rs
+++ b/packages/fuels-code-gen/src/utils/source.rs
@@ -71,6 +71,14 @@ impl Source {
             Source::String(abi) => Ok(abi.clone()),
         }
     }
+
+    /// If the abigen is given a path, will return it here.
+    pub fn path(&self) -> Option<PathBuf> {
+        match self {
+            Source::Local(path) => Some(path.clone()),
+            _ => None
+        }
+    }
 }
 
 fn get_local_contract(path: &Path) -> Result<String> {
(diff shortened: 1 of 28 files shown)


Br1ght0ne (approved):
Awesome engineering, LGTM! Maybe we should open a related issue for removing this later.  Doesn't seem like rust-lang/rust#99515 will land any time soon, though. :)

segfault-magnet:
> Awesome engineering, LGTM! Maybe we should open a related issue for removing this later. Doesn't seem like [rust-lang/rust#99515](https://github.com/rust-lang/rust/issues/99515) will land any time soon, though. :)

Made an issue [here](https://github.com/FuelLabs/fuels-rs/issues/1062)

packages/fuels-code-gen/src/utils/source.rs
```diff
@@ -71,6 +71,14 @@ impl Source {
             Source::String(abi) => Ok(abi.clone()),
         }
     }
+
+    /// If the abigen is given a path, will return it here.
```
hal3e, line 75: IMO the comment is not necessary. If we still want to keep the comment, maybe we say `if the source is a path, return it` rather than `if the abigen is given a path...`

hal3e (dismissed):
Awesome! Left an optional nit.

hal3e (approved):

Br1ght0ne (approved):

digorithm (approved):
Very cool. Hacky, yes, but very cool 🤠

Issue #1058, opened 2023-07-25
TOB-FUEL-15: Out-of-memory when decoding certain ABI types


Dentosal:
## Description
The following test causes extensive memory usage in the ABI decoder. Because ABI descriptions can come from untrusted sources, the parser should be hardened to stop parsing after a certain threshold.

#### Figure 15.1: Test case which causes extensive memory usage.

```rust
 fn oom_test() -> Result<()> { 
    let the_tuple = ABIDecoder::decode_single( 
        &ParamType::Array( 
            Box::new(ParamType::Array( 
                Box::new(ParamType::Unit), 
                0xFFFFFFFF
            )),
        0xFFFFFFFF),
        &hex::decode("00080008000800080008000800080008").unwrap()
    ); 
    Ok(())
}
```

The reason for this issue is that the decoder supports zero-sized types which do not result in progress during parsing. Therefore, the parser might run for a long time and eventually run out of memory.

## Exploit Scenario
A user executes a contract and interprets the returned date according to an unverified ABI specification, provided by a third party. If the ABI specification includes a type definition like shown in the above figure, then the user might experience a hang and run out of memory.

## Recommendations

Short term, abort the parsing after a certain amount of iterations.
Long term, avoid zero-sized types like empty static arrays, empty tuples, empty structs or the unit types. Zero-sized-types are a common issue in parsing libraries, which might cause a denial-of-service.

PR #1066, opened 2023-07-27
bug: fix for memory issue

segfault-magnet:
closes: #1058 
codeveloped with @Salka1988 
Special tnx to @nedsalk for brainstorming with us.

### The issue from the example
There was a memory problem in the example but not for the reasons listed in the issue. The issue (quoting the audit) says:

> The reason for this issue is that the decoder supports zero-sized types which do not result in progress during parsing. Therefore, the parser might run for a long time and eventually run out of memory.

Units in Sway are encoded as 1WORD of zeroes. So the given example didn't experience the zero-sized-type issue. What it did experience is a caveat of how we decoded multiple tokens of the same type (by cloning the param type). This is now replaced with a `std::iter::repeat` which will copy (as needed) the reference to a single `ParamType` solving the memory issue (by only having 1 reference to the param type at any moment).

### Real zero-sized types
A real zero-sized type would be an empty struct or an empty string array (i.e. `str[0]`). 

Enums don't come into play since empty enums cannot be instantiated. Also, an Enum with a ZST variant would have been at least 1 WORD since it encodes the discriminant, making it not a ZST.

You also cannot define an empty tuple (as far as we're aware).

Empty structs could also eventually be used as markers (e.g. `Tcp<Connected>` where `struct Connected{}`).

## The solution
The decoder now can be configured to accept two upper limits: one for the maximum tolerable depth, the second for the maximum tokens allowed.

The depth limit protects our stack (think call-frames and recursions) against types such as these:
```rust
struct Parent {
   child1: Child1 { 
      child2: Child2 {
          .... 
            child3: Child3 { ...
                  child7812341: Child7812341
      }
}
```
The whole structure would be zero-sized and contain so many levels of nested zero-size structs that we might run out of stack trying to decode it.

The second limit (`max_tokens`) protects against:
* collections of ZSTs
* structures with endless fields of ZSTs
* basically any other combination we could conceive of. 

We've set the defaults to 45 (levels of nesting) and 10k max tokens. 

The user can tweak the decoder used for contract and script calls. Tests have been added and the docs updated.

### Checklist


Salka1988 (approved):

digorithm (approved):
Alright, so correct if I'm wrong: this is solving the memory issue only for _collections_ and not the other cases you mentioned in the description related to other ZSTs, right? If that's the case, the title should mention it, and the memory issue is partially solved (maybe create a separate issue for the follow-up work).

As for handling the other ZSTs... banning them would probably not be a good approach. I'd rule this one out.

Defining a timeout mechanism that's mostly implicit and only explicit when the user needs it (e.g., they _just_ experienced the OOM issue and need to act on it) sounds like a good idea. 

Now... onto defining the limits, yeah... token based seems weird. Not sure. Time-based seems like a safe bet. I might be off here, but this type of decoding seems like it's either very fast (e.g., by and large, less than a handful of seconds most of the time) or something ill-defined or malicious that's straight-up undesirable that would take more than, say 30 seconds. This could cover the vast majority of cases. 

Then, in case of slower hardware struggling to decode reasonable and valid data for whatever reason, this user will see a timeout error, and the error should direct the user to increase the configurable timeout limit. What do y'all think?

segfault-magnet:
> Alright, so correct if I'm wrong: this is solving the memory issue only for collections and not the other cases you mentioned in the description related to other ZSTs, right? 

Right.

> If that's the case, the title should mention it, and the memory issue is partially solved (maybe create a separate issue for the follow-up work).

I'll change the title. Will create an issue as soon as we agree on further steps.

> Defining a timeout mechanism that's mostly implicit and only explicit when the user needs it (e.g., they just experienced the OOM issue and need to act on it) sounds like a good idea.

> Now... onto defining the limits, yeah... token based seems weird. Not sure. Time-based seems like a safe bet. I might be off here, but this type of decoding seems like it's either very fast (e.g., by and large, less than a handful of seconds most of the time) or something ill-defined or malicious that's straight-up undesirable that would take more than, say 30 seconds. This could cover the vast majority of cases.

Didn't play around with it but my thinking was that, due to our recursive approach to decoding, every nesting level will incur a call frame penalty to the stack along with any other data we may leave on the stack. Some of the recursions might be tail-optimized, but I'm not sure all are.

If the attack was in the form of a struct with a lot of fields, that isn't a problem since we'd be freeing and reserving the stack during the decoding. 

But if the attack was in the form of an enormously nested struct, then for each nesting level we'd go deeper in the recursion. This might cause us to hit the stack limit of the thread.

So limiting decoding by elapsed time might prove finicky, how fast will the recursion burrow, i.e. how loaded is the CPU at the moment? The user might be happy with a limit in normal times, only to cause him an OOM if the attack happened while his CPU load was down. 

Not sure that untangling the recursion would help by that much either since we need to keep state for every level.

I'll play around, try and recreate what this attack might look like, and get some numbers.

Ideally, we'd limit by time and memory used, but I'm not sure how practical this is to implement.

digorithm:
Yeah, it sounds like we gotta do a spike to iron out these details.

Dentosal:
> The whole structure would be zero-sized and contain so many levels of nested zero-size structs that we might run out of stack trying to decode it.

Typically you wouldn't run out of stack space, since the parsed structure would be placed into the heap. Then you could have a memory size limit, similar to what e.g. [regex crate does](https://docs.rs/regex/latest/regex/struct.RegexBuilder.html#method.size_limit).

packages/fuels-programs/src/contract.rs
```diff
@@ -360,6 +360,7 @@ pub struct ContractCallHandler<T: Account, D> {
     pub contract_call: ContractCall,
     pub tx_parameters: TxParameters,
     // Initially `None`, gets set to the right tx id after the transaction is submitted
+    decoder_config: DecoderConfig,
```
hal3e, line 363: should be placed above the comment. The comment relates to the `cached_tx_id`

packages/fuels-programs/src/logs.rs
```diff
@@ -84,6 +88,19 @@ impl LogResult {
...
 
 impl LogDecoder {
+    pub fn new(log_formatters: HashMap<LogId, LogFormatter>) -> Self {
+        Self {
+            log_formatters,
+            decoder_config: Default::default(),
+        }
+    }
+
+    // Set the `DecoderConfig` to be used when decoding logs
```
hal3e, line 98:
```suggestion
    // Set the `DecoderConfig` used when decoding logs
```
or we can delete this comment completely.

packages/fuels-core/src/codec.rs
```diff
@@ -43,11 +44,19 @@ mod tests {
...
+        macro_rules! test_decode {
+            ($($for_type: ident),*) => {
+                $(
+                    assert_eq!(
+                        try_from_bytes::<$for_type>(&bytes, DecoderConfig::default())?,
+                        $for_type::MAX
+                    );
+                )*
+            };
+        }
```
hal3e, line 56: This is so @segfault-magnet. Lets add 11 lines to remove 4 :smile: 

IMO we can remove one indentation level:
```suggestion
        macro_rules! test_decode {
            ($($for_type: ident),*) => {
                $(assert_eq!(
                        try_from_bytes::<$for_type>(&bytes, DecoderConfig::default())?,
                        $for_type::MAX
                );)*
            };
        }
```

packages/fuels-core/src/codec.rs
```diff
@@ -56,18 +65,20 @@ mod tests {
...
+                        try_from_bytes::<$for_type>(&bytes, DecoderConfig::default())?,
+                        $for_type::new(bytes.as_slice().try_into()?)
+                    );
+                )*
+
+            };
+        }
+
+        test_decode!(Address, ContractId, AssetId);
+
```
hal3e, line 81:
```suggestion
        macro_rules! test_decode {
            ($($for_type: ident),*) => {
                $(assert_eq!(
                        try_from_bytes::<$for_type>(&bytes, DecoderConfig::default())?,
                        $for_type::new(bytes.as_slice().try_into()?)
                );)*
            };
        }

        test_decode!(Address, ContractId, AssetId);
        
```

hal3e (commented):
Reviewing this PR was a joy. Code, docs, tests are awesome! Well done!

The only real suggestion I have is to add a similar limit to the `abi_encoder`. What to you think @FuelLabs/sdk-rust ? I can imagine some scenarios where someone could try to exploit our encoder. If we decide to do, I suggest we do it in a separate PR.

segfault-magnet:
> The only real suggestion I have is to add a similar limit to the `abi_encoder`

I'll add it as a discussion on the project. We can continue there so that this PR might come in.

hal3e (dismissed):
Awesome work!

Dentosal (dismissed):
The taken approach looks good, and the code is very readable albeit a bit verbose. Good documentation and tests. I'd appreciate some fuzzing-style tests as well, but those can always be added later.

Dentosal (approved):

hal3e (dismissed):

Br1ght0ne (approved):

MujkicA (approved):

docs/src/codec/decoding.md
```diff
@@ -0,0 +1,49 @@
...
+
+First into a [`Token`](https://docs.rs/fuels/latest/fuels/types/enum.Token.html), then via the [`Tokenizable`](https://docs.rs/fuels/latest/fuels/core/traits/trait.Tokenizable.html) trait, into the desired type.
+
+If the type came from [`abigen!`](../abigen/index.md) (or uses the [`::fuels::macros::TryFrom`](https://docs.rs/fuels/latest/fuels/macros/derive.TryFrom.html) derivation) then you can also use `try_into` to convert bytes into a type that implements both [`Parameterize`](https://docs.rs/fuels/latest/fuels/core/traits/trait.Parameterize.html) and [`Tokenizable`](https://docs.rs/fuels/latest/fuels/core/traits/trait.Tokenizable.html):
+
+```rust,ignore
+{{#include ../../../examples/codec/src/lib.rs:decoding_example_try_into}}
+```
+
+Under the hood [`try_from_bytes`](https://docs.rs/fuels/latest/fuels/core/codec/fn.try_from_bytes.html) is being called which basically does what the preceding example did.
```
digorithm, line 19:
```suggestion
Under the hood, [`try_from_bytes`](https://docs.rs/fuels/latest/fuels/core/codec/fn.try_from_bytes.html) is being called, which does what the preceding example did.
```

docs/src/codec/encoding.md
```diff
@@ -0,0 +1,20 @@
...
+
+Be sure to read the [prerequisites](./index.md#prerequisites-for-decodingencoding) to encoding.
+
+Encoding is done via the [`ABIEncoder`](https://docs.rs/fuels/latest/fuels/core/codec/struct.ABIEncoder.html):
+
+```rust,ignore
+{{#include ../../../examples/codec/src/lib.rs:encoding_example}}
+```
+
+Note that the return type of `encode` is `UnresolvedBytes`. The encoding cannot be finished until we know at which memory address this data is to be loaded on. If you don't use heap types (`::std::vec::Vec`, `::fuels::types::Bytes`, `::std::string::String`) then you can safely `.resolve(0)` to get the encoded bytes.
```
digorithm, line 11:
```suggestion
Note that the return type of `encode` is `UnresolvedBytes`. The encoding cannot be finished until we know at which memory address this data is to be loaded. If you don't use heap types (`::std::vec::Vec`, `::fuels::types::Bytes`, `::std::string::String`), then you can safely `.resolve(0)` to get the encoded bytes.
```

docs/src/codec/index.md
```diff
@@ -0,0 +1,45 @@
+# Codec
+
+Encoding and decoding are done as per [the fuel spec](https://specs.fuel.network/master/abi/argument-encoding.html). To this end `fuels` makes use of the [ABIEncoder](https://docs.rs/fuels/latest/fuels/core/codec/struct.ABIEncoder.html) and the [ABIDecoder](https://docs.rs/fuels/latest/fuels/core/codec/struct.ABIDecoder.html).
```
digorithm, line 3:
```suggestion
Encoding and decoding are done as per [the fuel spec](https://specs.fuel.network/master/abi/argument-encoding.html). To this end, `fuels` makes use of the [ABIEncoder](https://docs.rs/fuels/latest/fuels/core/codec/struct.ABIEncoder.html) and the [ABIDecoder](https://docs.rs/fuels/latest/fuels/core/codec/struct.ABIDecoder.html).
```

digorithm (changes requested):
I love the new encoding/decoding session; we badly needed that. I just left a couple of grammar nits to improve flow/clarity.

I'm still going through the actual implementation, but I'd like to send these suggestions now rather than later.

examples/codec/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+        };
+
+        #[derive(Tokenizable)]
+        struct MyStruct {
+            field: u64,
+        }
+
+        let instance = MyStruct { field: 101 };
+        let encoded: UnresolvedBytes = ABIEncoder::encode(&[instance.into_token()])?;
+        let load_memory_address: u64 = 0x100;
```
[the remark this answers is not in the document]
segfault-magnet: We should probably explain... but the thing is it's not simple to explain. 

You'd have to start with how the VM gets initialized and proceed to other implementation details pointing to various files over in the `fuel-*` crates.

I'd rather wait for the encoding support before burdening our docs. Or, maybe, we provide two encoders -- one capable of encoding heap types and one that is not.

We make the incapable one public and remove the `.resolve(addr)` part of the API. 

The result is that users can't encode heap types anymore but you also then don't have to document it.

But, to be honest, were users ever able to properly encode heap types? I mean getting the correct offset is involved enough for us, let alone the users.

Maybe we should make this into a discussion, so as not to block this long-overdue PR. @FuelLabs/sdk-rust ?

packages/fuels/tests/logs/script_needs_custom_decoder_logging/src/main.sw
```diff
@@ -0,0 +1,6 @@
+script;
+
+fn main() {
```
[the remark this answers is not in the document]
segfault-magnet: Depending on the final API, it might well be possible to remove it. We'll have to wait for it to land. I'd rather leave a TODO here than make a small tech-dept issue block the PR.

MujkicA (approved):

digorithm (approved):
Thanks for the patience; this was a big one. The new tests helped me a lot to understand these changes. Great job, dude!

PR #1067, opened 2023-07-28
feat!: make functions return Result and implement Node Killer

Salka1988:
Refactored functions to return Result for better error handling.

+ launch_provider_and_get_wallet()
+ launch_custom_provider_and_get_wallets()
+ setup_test_provider()
+ setup_test_client()

I've implemented FuelService and added support for the fuel_core_services::Service trait. This allows us to have finer control over the behavior of FuelNode;

```rust
pub struct FuelService {
    service: CoreFuelService,
    #[cfg(not(feature = "fuel-core-lib"))]
    service: BinFuelService,
    bound_address: SocketAddr,
}
```

### Checklist


packages/fuels-test-helpers/src/lib.rs
```diff
@@ -142,7 +143,7 @@ pub async fn setup_test_client(
     messages: Vec<Message>,
     node_config: Option<Config>,
     chain_config: Option<ChainConfig>,
-) -> (FuelClient, SocketAddr, ConsensusParameters) {
+) -> Result<(FuelClient, SocketAddr, ConsensusParameters), Error> {
```
hal3e, line 146: When spawning a new node with 
```rust
let srv = FuelService::new_node(config).await.unwrap();
```
we could map and propagate the error now that we return a result.

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -181,7 +177,7 @@ pub async fn setup_test_client(
...
```
hal3e, line 180: When we try to get a `bound_address` 
```rust
    let bound_address = if requested_port == 0 {
        get_socket_address()
    } else if is_free(requested_port) {
        config.addr
    } else {
        panic!("Error: Address already in use");
    };
```
we could return an error instead of panicking.

packages/fuels/tests/bindings.rs
```diff
@@ -14,7 +14,7 @@ pub fn null_contract_id() -> Bech32ContractId {
 }
 
 #[tokio::test]
-async fn compile_bindings_from_contract_file() {
+async fn compile_bindings_from_contract_file() -> Result<()> {
```
hal3e, line 17: This test does not have any `unwraps` nor returns results so it does not benefit from adding a `Result` return type.

```suggestion
async fn compile_bindings_from_contract_file() {
```

hal3e (changes requested):
Thanks for the nice refactor and devX improvement!

packages/fuels/tests/bindings.rs
Salka1988:
```rust
fn wallet_initialization_code(maybe_command: Option<InitializeWalletCommand>) -> TokenStream {
.
.
.
    quote! {
        let [#(#wallet_names),*]: [_; #num_wallets] = launch_custom_provider_and_get_wallets(
            WalletsConfig::new(Some(#num_wallets as u64), None, None),
            None,
            None,
        )
        .await?
        .try_into()
        .expect("Should have the exact number of wallets");
    }
}
```
"Since I added `.await?` to the `wallet_initialization_code`, the test now needs to return `Result` to handle the potential errors."

hal3e: We avoided passing the errors from the macro. This forces everybody that uses this macro to return a result. Instead you should `expect` with a nice message.

Salka1988: done

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -157,22 +162,20 @@ pub async fn setup_test_client(
     let mut config = node_config.unwrap_or_else(Config::local_node);
     config.chain_conf = chain_conf;
 
-    let srv = FuelService::new_node(config).await.unwrap();
+    let srv = FuelService::new_node(config)
+        .await
+        .map_err(|err| fuels_core::types::errors::Error::InfrastructureError(err.to_string()))?;
```
hal3e, line 167: you can import `fuels_core::types::errors::{error, Error}` and do 
```suggestion
        .map_err(|err| error!(InfrastructureError, "{err}"))?;
```

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -190,7 +193,9 @@ pub async fn setup_test_client(
     } else if is_free(requested_port) {
         config.addr
     } else {
-        panic!("Error: Address already in use");
+        return Err(fuels_core::types::errors::Error::IOError(
+            std::io::ErrorKind::AddrInUse.into(),
+        ));
```
hal3e, line 198: If you import `Error` you can shorten this to:
```suggestion
        return Err(Error::IOError(std::io::ErrorKind::AddrInUse.into()));
```

packages/fuels-test-helpers/src/lib.rs
Salka1988: I saw that this was the practice and tried to do it like this, but I ran into problems. Maybe because of the flags we have. I'll try again

packages/fuels-test-helpers/src/lib.rs
Salka1988: had to add allow(unused_imports) to to silence the cargo check and clippy

segfault-magnet (commented):
This is going to break every test out there. It should definitely be fixed while we have the time, my thinking is just whether we want to make some more changes to this so that we may at least have something more to show for it besides better panic handling. 

For example, one thing we've been putting off for so long is the ability to kill the spawned node. Useful when users want to test the behavior of their app when the node is down.

Currently, they need to kill the tokio runtime since a handle was hidden somewhere inside that keeps the node alive.

Ideally, we'd ship this out with that feature (and maybe any other pressing ones) and at least make a feature-rich-breaking change.

@FuelLabs/sdk-rust

Salka1988:
> This is going to break every test out there. It should definitely be fixed while we have the time, my thinking is just whether we want to make some more changes to this so that we may at least have something more to show for it besides better panic handling.
> 
> For example, one thing we've been putting off for so long is the ability to kill the spawned node. Useful when users want to test the behavior of their app when the node is down.
> 
> Currently, they need to kill the tokio runtime since a handle was hidden somewhere inside that keeps the node alive.
> 
> Ideally, we'd ship this out with that feature (and maybe any other pressing ones) and at least make a feature-rich-breaking change.
> 
> @FuelLabs/sdk-rust

I could do it in this PR, but I think it is better to make a new PR. I will do as @FuelLabs/sdk-rust decides. Anyway, I'm going to start implementing `the spawned node killer`.

Br1ght0ne (approved):

packages/fuels-macros/src/setup_program_test/code_gen.rs
```diff
@@ -88,6 +88,7 @@ fn wallet_initialization_code(maybe_command: Option<InitializeWalletCommand>) ->
             None,
         )
         .await
+        .expect("Error while trying to fetch wallets from the custom provider")
```
iqdecay, line 91: why not use `?` here

packages/fuels-test-helpers/src/fuel_service.rs
iqdecay, line 1: normally this feature should be in a separate PR. If I think we need some time to back and forth on this feature then I'm blocking the refactoring.

packages/fuels/tests/types_contracts.rs
```diff
@@ -883,7 +883,7 @@ async fn strings_must_have_all_ascii_chars() {
         "#,
     ));
 
-    let wallet = launch_provider_and_get_wallet().await;
+    let wallet = launch_provider_and_get_wallet().await.unwrap();
     let contract_instance = SimpleContract::new(null_contract_id(), wallet);
```
iqdecay, line 887: make the function return result and use `?`, as we conventionnaly do.

packages/fuels/tests/types_contracts.rs
```diff
@@ -969,7 +969,7 @@ async fn strings_must_have_correct_length_custom_types() {
         "#,
     ));
 
-    let wallet = launch_provider_and_get_wallet().await;
+    let wallet = launch_provider_and_get_wallet().await.unwrap();
```
iqdecay, line 972: same here

packages/fuels/tests/types_contracts.rs
```diff
@@ -1063,7 +1063,7 @@ async fn strings_must_have_all_ascii_chars_custom_types() {
         foo: inner_struct,
     };
 
-    let wallet = launch_provider_and_get_wallet().await;
+    let wallet = launch_provider_and_get_wallet().await.unwrap();
```
iqdecay, line 1066: same here

iqdecay (changes requested):
Left some remarks. Will review the node part later, but would much prefer if it was a separate PR.

packages/fuels-test-helpers/src/fuel_service.rs
```diff
@@ -0,0 +1,323 @@
...
+            args.push("--vm-backtrace".to_string());
+        }
+
+        Ok(args)
+    }
+
+    pub fn write_temp_config_file(&mut self, config: Value) -> FuelResult<()> {
+        writeln!(self.config_file, "{}", &config.to_string())?;
+        Ok(())
+    }
```
hal3e, line 120:
```suggestion
    pub fn write_chain_config_file(&mut self) -> FuelResult<()> {
        Ok(serde_json::to_writer(
            &mut self.config_file,
            &self.config.chain_conf,
        )?)
    }
```

hal3e (commented):
Reviewed with @segfault-magnet. The `Service` trait from `fuel-core-services` is an overkill for our needs. We suggest a simpler approach that will also clean up the `cfg!` usage:
- Create and expose only one `FuelService` type where the implementation differs depending whether `fuel-core-lib` is enabled or not.
- The interface of `FuelService` should not change depending on the implementation and should have methods to start and stop the node.
- Create and expose only one `Config` that can be used by both implementations (either by converting to a `FuelCore` config or creating a `json` file)

Br1ght0ne (dismissed):

packages/fuels/tests/contracts.rs
```diff
@@ -1670,3 +1671,20 @@ async fn heap_types_correctly_offset_in_create_transactions_w_storage_slots() ->
 
     Ok(())
 }
+
+#[tokio::test]
+async fn bla_bla() -> Result<()> {
```
MujkicA, line 1676: I guess this is not needed anymore

MujkicA (commented):
Code looks good, there are some inconsistencies witch docs:

- There is a mention of `launch_provider_and_get_wallet().await` in the docs that should include `?` now. 

- Also there is mentions of `setup_test_client()` in some comments even though it doesn't exist anymore.

MujkicA (dismissed):

MujkicA (approved):

Br1ght0ne (approved):

iqdecay (approved):

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -153,53 +152,18 @@ pub async fn setup_test_provider(
...
-    )
-    .await;
-
-    let client = FuelClient::from(bound_address);
-    server_health_check(&client).await;
-
-    Provider::from(bound_address)
-        .await
-        .expect("Could not connect to node!")
+        .expect("Could not connect to node"))
```
hal3e, line 166:
```suggestion
```

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -153,53 +152,18 @@ pub async fn setup_test_provider(
...
+
+    let address = srv.bound_address();
+
     tokio::spawn(async move {
         let _own_the_handle = srv;
         let () = futures::future::pending().await;
     });
 
-    Provider::from(address)
+    Ok(Provider::from(address)
```
hal3e, line 164: We can return the error. No need to expect
```suggestion
    Provider::from(address).await
```

packages/fuels-test-helpers/src/service.rs
```diff
@@ -0,0 +1,75 @@
...
+        self.service
+            .stop()
+            .await
+            .map_err(|err| error!(InfrastructureError, "{}", err))
+    }
+
+    pub fn bound_address(&self) -> SocketAddr {
+        self.bound_address
+    }
+}
```
hal3e, line 75: As they are almost identical I would combine the code so:

```suggestion
pub struct FuelService {
    #[cfg(feature = "fuel-core-lib")]
    service: CoreFuelService,
    #[cfg(not(feature = "fuel-core-lib"))]
    service: BinFuelService,
    bound_address: SocketAddr,
}

impl FuelService {
    pub async fn start(config: Config) -> Result<Self> {
        #[cfg(feature = "fuel-core-lib")]
        let service = CoreFuelService::new_node(config.into())
            .await
            .map_err(|err| error!(InfrastructureError, "{}", err))?;

        #[cfg(not(feature = "fuel-core-lib"))]
        let service = BinFuelService::new_node(config)
            .await
            .map_err(|err| error!(InfrastructureError, "{}", err))?;

        let bound_address = service.bound_address;

        Ok(FuelService {
            service,
            bound_address,
        })
    }

    pub async fn stop(&self) -> Result<State> {
        #[cfg(feature = "fuel-core-lib")]
        let result = self.service.stop_and_await().await;

        #[cfg(not(feature = "fuel-core-lib"))]
        let result = self.service.stop();

        result.map_err(|err| error!(InfrastructureError, "{}", err))
    }

    pub fn bound_address(&self) -> SocketAddr {
        self.bound_address
    }
}
```

packages/fuels-test-helpers/src/fuel_bin_service.rs
```diff
@@ -0,0 +1,222 @@
...
+    let between_attempts = Duration::from_millis(300);
+
+    while attempts > 0 && !healthy {
+        healthy = client.health().await.unwrap_or(false);
+        sleep(between_attempts).await;
+        attempts -= 1;
+    }
+
+    if !healthy {
+        panic!("error: Could not connect to fuel core server.")
```
hal3e, line 173: IMO we should return an error here. The function already returns a result

packages/fuels-test-helpers/src/fuel_bin_service.rs
```diff
@@ -0,0 +1,222 @@
...
+
+async fn run_node(params: ServerParams) -> FuelResult<JoinHandle<()>> {
+    let ServerParams {
+        mut extended_config,
+    } = params;
+    let args = extended_config.config_to_args_vec()?;
+
+    let binary_name = "fuel-core";
+    let paths = which::which_all(binary_name)
+        .unwrap_or_else(|_| panic!("failed to list '{binary_name}' binaries"))
```
hal3e, line 192: IMO we should return an error. `or_else` could be used here with our error

packages/fuels-test-helpers/src/fuel_bin_service.rs
```diff
@@ -0,0 +1,222 @@
...
+        .collect::<Vec<_>>();
+    let path = paths
+        .first()
+        .unwrap_or_else(|| panic!("no '{binary_name}' in PATH"));
```
hal3e, line 196: same here

hal3e (changes requested):
I like the way this looks now! I have some change requests though...

packages/fuels/Cargo.toml
```diff
@@ -10,6 +10,9 @@ repository = { workspace = true }
 rust-version = { workspace = true }
 description = "Fuel Rust SDK."
 
+[package.metadata.cargo-machete]
+ignored = ["fuel-core"]
```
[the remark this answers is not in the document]
Salka1988: machete fails without this :/

hal3e (approved):
Great work!

Br1ght0ne (approved):

MujkicA (approved):

Issue #965, opened 2023-05-15
Abstract away the need for setting witness indexes


MujkicA:
Currently, our `Input` type requires you to set a `witness_index`. This is propagates to the methods `get_asset_inputs_for_amount`, `add_fee_resources` which is especially painful because they are defined on the `Account` trait. Predicates as an implementor of `Account` have no use for the `witness_index`.

`witness_index` is the index in the `witnesses` field of a transaction where the signature of the wallet that owns the input is expected to be found. It's quite tricky to avoid exposing this on a higher level as we can't ignore the order of signatures because  every signer also signs over the `witnesses` field and thereby includes every previous signature.

@segfault-magnet mentioned the possibility of storing calls to `sign_transaction` as intentions which will be applied in a post-processing step, similarly to how we handle dynamic data encoding for predicates.

Another option might be to always default to idx 0 (or 1 in the case of Contract txs) and have a special interface for creating txs with multiple signatures since they are more of a niche use-case.

PR #1069, opened 2023-07-31
feat!: resolve witness indexes automatically

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/965

Previously, when spending signed resource, the user had to manually set `witness` indexes. This meant that the user had to pay attention on the signing order and  make sure to offset the index if some witnesses where already set.

This PR automates the whole process.

Proposed workflow:
- wallet's `sign_transaction` now accepts a `TransactionBuilder`
- when a wallet signs a `TransactionBuilder` the intention is saved in `UnresolvedSignatures`
- when the user builds the transaction all signatures are resolved
- after building the transaction can not be changed except with `append_witness` - this allows the user to add custom witnesses and it will not affect the `TxId`

Additionally, building a transaction will fail if there are signed resources that are missing corresponding signatures.

BREAKING CHANGE: 
- `send_transaction`, `dry_run`, `dry_run_no_validation` and `estimate_transaction_cost` need ownership of the `tx`.
- removed witness index from `Input` and corresponding functions
- wallet's `sign_transaction` now accepts a `TransactionBuilder`

### Checklist


packages/fuels-test-helpers/src/node.rs
```diff
@@ -70,7 +70,7 @@ impl Config {
             max_database_cache_size: DEFAULT_CACHE_SIZE,
             database_path: Default::default(),
             database_type: DbType::InMemory,
-            utxo_validation: false,
+            utxo_validation: true,
```
MujkicA, line 73: Nice! This has been preventing us to catch errors with the CI in the past.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -205,7 +183,7 @@ impl_tx_trait!(ScriptTransactionBuilder, ScriptTransaction);
 impl_tx_trait!(CreateTransactionBuilder, CreateTransaction);
 
 impl ScriptTransactionBuilder {
-    fn convert_to_fuel_tx(self, base_offset: usize) -> ScriptTransaction {
+    fn convert_to_fuel_type(self, base_offset: usize) -> Script {
```
MujkicA, line 186: I don't mind the renaming, but maybe the Builders should follow the naming convention then

packages/fuels-accounts/src/provider.rs
```diff
@@ -294,20 +298,18 @@ impl Provider {
             .unwrap_or(Ok(()))
     }
 
-    pub async fn dry_run<T: Transaction + Clone>(&self, tx: &T) -> Result<Vec<Receipt>> {
-        let receipts = self.client.dry_run(&tx.clone().into()).await?;
+    pub async fn dry_run<T: Transaction>(&self, mut tx: T) -> Result<Vec<Receipt>> {
+        tx.resolve_transaction()?;
```
MujkicA, line 302: I was wondering if we should protect against the case where we don't know if a tx has been already resolved by introducing another type like `ResolvedTransaction` that results from this method call.
@hal3e considered it but was more leaned towards introducing less types into our workflow which already consists of builders, wrappers etc.

packages/fuels-core/src/types/transaction_builders.rs
hal3e: I needed a way to construct the underlying fuel's type (in this case `Script`) and then later I make the `ScriptTransaction`

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -245,32 +208,30 @@ impl Transaction for TransactionType {
...
+            pub(crate) consensus_parameters: ConsensusParameters,
         }
 
         impl From<$wrapped> for $wrapper {
             fn from(tx: $wrapped) -> Self {
-                $wrapper { tx }
+                $wrapper {
+                    tx,
+                    unresolved_signatures: Default::default(),
+                    consensus_parameters: Default::default(),
```
MujkicA, line 233: @hal3e and I concluded that this could potentially lead to cases where the users holds a tx with wrong params.
When we fetch txs from the node, we convert them to our wrapper type before exposing it to the user. A potential risk here is that the user will use the default parameters somewhere down the line.

hal3e: Awesome catch! @FuelLabs/sdk-rust what do you think we make a dedicated `from_fuel_tx` that accepts a `fuel_tx` and the `consensus parameters` ?

MujkicA:
Regarding tests, we could include a case with multiple inputs coming from different wallets for a simple transfer and one for a create transaction

segfault-magnet (commented):
Love it. Left suggestions around macros.

segfault-magnet (approved):

Salka1988 (approved):
👍

Br1ght0ne (dismissed):

hal3e:
I will turn this PR into a draft. Even though I implemented the requested feature, we still have some use-cases with witnesses we need to cover. I will soon make a new version and then we can discuss it.

Br1ght0ne (approved):
Great, LGTM!

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -67,48 +66,35 @@ pub enum TransactionType {
...
     fn metered_bytes_size(&self) -> usize;
 
-    fn inputs(&self) -> &Vec<FuelInput>;
+    fn inputs(&self) -> &Vec<Input>;
 
     fn outputs(&self) -> &Vec<Output>;
 
     fn witnesses(&self) -> &Vec<Witness>;
 
-    fn witnesses_mut(&mut self) -> &mut Vec<Witness>;
```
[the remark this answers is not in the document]
hal3e: The idea is that `ScriptTransaction`  `CreateTransaction` can not be changed after we have built them from the builders. The only exception is `append_witness`

Br1ght0ne (dismissed):

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -443,17 +529,17 @@ fn convert_to_fuel_inputs(inputs: &[Input], offset: usize) -> Vec<FuelInput> {
...
+                    utxo_id,
+                    balance_root,
+                    state_root,
+                    tx_pointer,
+                    contract_id,
+                ))
             }
         })
-        .collect::<Vec<FuelInput>>()
+        .collect::<Result<Vec<FuelInput>>>()
```
segfault-magnet, line 542:
```suggestion
        .collect()
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -392,48 +450,76 @@ impl CreateTransactionBuilder {
     }
 }
 
-fn convert_to_fuel_inputs(inputs: &[Input], offset: usize) -> Vec<FuelInput> {
-    let mut new_offset = offset;
-
+/// Resolve SDK Inputs to fuel_tx Inputs. This function will calculate the right
+/// data offsets for predicates and set witness indexes for signed coins.
+fn resolve_fuel_inputs(
```
segfault-magnet, line 455: The fn is starting to be a bit long (82 lines). 

Wdyt about splitting it up a bit? Maybe flattening the nesting level by removing the inner match on `resource`?

packages/fuels/tests/wallets.rs
```diff
@@ -403,3 +404,49 @@ async fn transfer_coins_of_non_base_asset() -> Result<()> {
...
+    let wallets = launch_custom_provider_and_get_wallets(wallet_config, None, None).await;
+    let provider = wallets[0].try_provider()?;
+
+    let mut receiver = WalletUnlocked::new_random(None);
+    receiver.set_provider(provider.clone());
+
+    let amount_to_transfer = 20;
+
+    let mut inputs = vec![];
+    for wallet in wallets.iter() {
```
segfault-magnet, line 420:
```suggestion
    for wallet in &wallets {
```

segfault-magnet (commented):
Ideally we should communicate that nothing is signed until `build` is called. 

Although the ABI is not stable, someone might go and design an app that accepts `dylib`s. If they then think give the unfinished tx to a malicious `dynlib` to do some finishing touches (thinking that the signature will not be valid if the actor tries to change anything that was signed over) then they might find that the actor changed the outputs of the tx :D 

Although it is super far fetched, we should still communicate that the object carries the signing keys in it until the tx is finalized.

packages/fuels-accounts/src/provider.rs
```diff
@@ -521,49 +536,41 @@ impl Provider {
...
+            .get_gas_used_with_tolerance(new_tx.clone(), tolerance)
             .await?;
-        let gas_price = std::cmp::max(tx.gas_price(), min_gas_price);
 
-        // Update the dry_run_tx with estimated gas_used and correct gas price to calculate the total_fee
-        dry_run_tx
-            .with_gas_price(gas_price)
-            .with_gas_limit(gas_used);
+        // Update the tx with estimated gas_used and correct gas price to calculate the total_fee
+        let new_tx = new_tx.to_dry_run_tx(gas_price, gas_used);
```
MujkicA, line 555: IMO `to_dry_run_tx` feels a bit obscure to be a user facing method. It just set's two fields at once, while the name implies there is a higher purpose to it. The previous `generate_dry_run_tx` applied specific values to these fields so the name made more sense.
What do you think about just using two setters or going back to the old implementation?

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -190,42 +157,21 @@ impl Transaction for TransactionType {
         }
     }
 
-    fn with_gas_price(self, gas_price: u64) -> Self {
```
MujkicA, line 193: Why was this removed?

packages/fuels-core/src/types/wrappers/transaction_response.rs
```diff
@@ -27,8 +27,11 @@ pub enum TransactionStatus {
     SqueezedOut(),
 }
 
-impl From<ClientTransactionResponse> for TransactionResponse {
-    fn from(client_response: ClientTransactionResponse) -> Self {
+impl TransactionResponse {
+    pub fn from_fuel_response(
+        client_response: ClientTransactionResponse,
+        consensus_parameters: ConsensusParameters,
```
MujkicA, line 33: Could this lead to bugs? I imagine a scenario where the params changed at some point T during the growth of the chain. Now, some time after T a user wants to inspect a tx that happened before T. They retrieve the tx but the attached params correspond to the current ones instead of the ones at the time of execution. 

I don't suppose there is a realistic scenario where `chain_id` could change so calling `id` should be safe regardless.  However, calling `fee_checked_from_tx` could lead to wrong results. It's a weird use case to calculate the fee on an old tx but still a bug.

I'm not yet sure how we can tackle this but for now I just wanted to confirm the above assumptions are true.

MujkicA (commented):
I left some questions.
Nice addition to the UX 👍

packages/fuels-accounts/src/provider.rs
hal3e: Thanks for the comment! I went back to the old implementation.

packages/fuels-core/src/types/wrappers/transaction.rs
hal3e: I thought it would change the tx id so I did not want to user to think that he can just use `with_gas_price` after the tx was built and the ID was set. 

However, I was wrong! :D these methods do not change the id so I have returned all of them (maturity, gas_price, gas_limit )

packages/fuels-core/src/types/wrappers/transaction_response.rs
hal3e: Good point. I do not see a better way to deal with this for now so I have reverted both `fee_checked_from_tx` and `id` to accept `ConsensusParameters` and `ChainId` respectively.

segfault-magnet (dismissed):

Salka1988 (approved):

MujkicA (approved):

MujkicA (dismissed):

MujkicA (approved):

segfault-magnet (approved):

Salka1988 (approved):

PR #1073, opened 2023-08-01
separate send_transaction and get_receipts

Salka1988:
Refactored the `submit_and_await_commit_with_receipts` function to separate transaction submission and receipt retrieval into individual functions: `submit_tx` and `get_receipts`.

### Checklist


Br1ght0ne (dismissed):

segfault-magnet (commented):
A step in the right direction. This divide between tx submittal and response querying needs to reach up to the contract/script call. 

So you can have the option of submitting a call and awaiting its response separately. 

Something like:
```rust
let handle = contract_instance.methods().some_method(10, 20).submit().await;

let tx_id = handle.tx_id();

let response = handle.retry_config(...).response().await?;
```

So that we may give more control to the user as to how long they're willing to wait for the tx or how often they want to check for results, etc. 

Was originally discussed (with @MujkicA) when we had the `None` returned for receipts issue because the node we queried didn't have them at the moment.

Ideally, leave `call` as well to anyone who doesn't want to customize the behavior.

Salka1988:
> Looks good, but agree that we must let the user the option to call directly `call` or `simulate` without worrying about submission vs receipts

You can do that. These functions exist and are the same as before.

Br1ght0ne (approved):

hal3e (approved):
lgtm

PR #1090, opened 2023-08-17
style!: unify string type names

### LINKED ISSUE #1088: Discrepancy between `ParamType` and `Token` names for `str[x]`
The param type for `str[15]` is `ParamType::String(15)` while the token is: `Token::StringArray(...)`. 

These names should be the same.

iqdecay:
Closes #1088.
- The `String` type in Sway is the perfect equivalent of the Rust `String`
type, therefore this naming is more coherent. It was not possible before
because the `String` name was already taken for statically-sized
strings, which are now called `StringArray`s.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,7 +227,7 @@ jobs:
         if: ${{ matrix.command == 'test_wasm' }}
         uses: actions/setup-node@v3
         with:
-          node-version:  18
+          node-version: 18
 
       - name: Test WASM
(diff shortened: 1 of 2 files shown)


digorithm (approved):
LGTM -- this is a breaking change though; some devs (e.g., Indexer) use these `Token`s and `ParamType`s directly!

Br1ght0ne (approved):

Salka1988 (approved):

hal3e (approved):
lgtm

segfault-magnet (approved):

PR #1094, opened 2023-08-19
chore: move `fuels/tests` to `fuels-e2e-tests` crate

### LINKED ISSUE #1093: e2e tests to be moved from `fuels`
By having the `e2e` tests inside `fuels` we're prone to `use`ing stuff from `fuels` dependencies rather than `fuels` itself. This then leads to issues like "we didn't know xyz hadn't been reexported throught `fuels`. The test worked fine" only to find that we included xyz from something like `fuel-tx` in the test.

This cannot be avoided as long as the tests are in `fuels` because we cannot make prod deps unavailable for the integration tests.

One idea is to have a new `e2e` (or whatever) package that would contain all the e2e tests. The package itself should not be published. The tests there would only be dependent on `fuels`.

Br1ght0ne:
Close #1093.

Points of interest:
- I added some public imports to [packages/fuels/src/lib.rs](https://github.com/FuelLabs/fuels-rs/blob/ba915939057900e56d1453839fbb5323557fe3f1/packages/fuels/src/lib.rs), namely `tx::{CheckError, TxPointer, UtxoId}`
- `db_rocksdb` test runs only when `not(feature = "fuel-core-lib")`
-  `test_network_error` runs only when `feature = "fuel-core-lib"`. That's because the test requires `impl ServiceTrait for FuelService`, which is not implemented by `fuels_test_helpers` (yet). An issue can be created for this, so at some point we can drop `fuel-core-lib` altogether.

### Checklist


segfault-magnet (commented):
Not sure about the comment about removing `fuel-core` as a dep. My understanding is that we want that feature for easier testing.

But then again maybe removing it isn't such a bad idea. 

The good side is that you get a locked version of fuel-core that works with the SDK. 

The bad side is the maintenance overhead of configuring both the binary and the lib.

@FuelLabs/sdk-rust thoughts?

PR #1096, opened 2023-08-21
style!: rename 'set_' to 'with_'

### LINKED ISSUE #1089: `Configurables`| `LoadConfiguration` misleading naming wrt ownership
Configurables have methods for each configurable that start with `set_`. 

They also unexpectedly consume `self` which is unexpected coming from a setter.

Maybe we should rename this to `with_xyz` instead.

We can also still generate the `set` methods but make them deprecated with a note pointing to the `with_xyz` new alternatives.

The same is present on `LoadConfiguration` where `set_` methods take ownership of the object.

Also it seems the ContractCallHandler does this. We might be having a setter crisis :)

MujkicA:
Closes #1089

### Checklist

### DIFF
--- a/packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
+++ b/packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
@@ -38,7 +38,7 @@ impl FunctionGenerator {
         })
     }
 
-    pub fn with_name(&mut self, name: String) -> &mut Self {
+    pub fn name(&mut self, name: String) -> &mut Self {
         self.name = name;
         self
     }
@@ -48,12 +48,12 @@ impl FunctionGenerator {
         self
     }
 
-    pub fn with_body(&mut self, body: TokenStream) -> &mut Self {
+    pub fn body(&mut self, body: TokenStream) -> &mut Self {
         self.body = body;
         self
     }
 
-    pub fn with_doc(&mut self, text: String) -> &mut Self {
+    pub fn doc(&mut self, text: String) -> &mut Self {
         self.doc = Some(text);
         self
     }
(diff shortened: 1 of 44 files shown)


packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
```diff
@@ -38,7 +38,7 @@ impl FunctionGenerator {
         })
     }
 
-    pub fn set_name(&mut self, name: String) -> &mut Self {
+    pub fn with_name(&mut self, name: String) -> &mut Self {
```
hal3e, line 41: This seems to be a mix of what we try to separate:  `set_(&mut self) -> ()` and `with(self) -> Self`.
IMO I would not use `with` here as it is not consistent with the rest of our code and could be confusing.

Here are two have 2 options that come to mind: 
- use a different name (not `set` and not  `with` ). Maybe only the field name like `name`.
- use either `set` or  `with` and change the code to fit this

hal3e (commented):
Thanks for going through the whole code base and fixing this.

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
MujkicA: Changed it to `name`. 
I figured it's too much of an impact since it's not a public facing API.

MujkicA: Decided to leave it at  `set_` since there is some getters that occupy the `name` convention already.

hal3e (approved):

segfault-magnet (approved):
We thank you for your service to standardization. 

Also, for future reference, `with_*` methods don't have a monopoly on creating builder-like interfaces. We can have `set_*` return `&mut Self`. Writing because I think I saw a few places where `-> &mut Self` was being dropped from a `set_*` method. Didn't write a comment since, IIRC, it was the provider, which we don't need to configure too much anyway.

PR #1105, opened 2023-08-28
feat: caching of recently used coins

MujkicA:
### Checklist

Adds a cache to the `Provider` for recently used coins. The cache is behind an optional feature flag coin-cache.

The cache holds a map with entries of `(Bech32Address, AssetId)` -> `HashSet<CoinTypeId>` where `CoinTypeId` is either a utxo or a nonce and `Bech32Address` is the address of the account that owns the coins.

Whenever a tx is submitted, every `CoinTypeId` is added to the cache entry corresponding to it's owner and asset. 
When an account requests spendable_coins of a specific asset, the node is queried with recently used coins from the cache as a filtering condition.
The cache specifies a TTL for the items, once it expires the corresponding entries will be ignored and evicted the next time active entries are retrieved. 

`CoinCacheItem` is a wrapper for a `CoinTypeId` that is stored in the cache. It includes the creation time so it can be verified against the TTL. It implements the Hash and PartialEq traits so that inserting them into the Set only compares against the id and not the creation time. This means that if a coin somehow is used again before it was evicted, the corresponding entry will just be updated with a new timestamp.
Tests make use of tokio::time to advance time manually instead of using sleep().

**Other changes:**
The strategy for adding fee coins is changed from completly recomputing the
base asset input set to just adding an amount of base asset that will cover the fee.
While this might lead to a bit more dust it gives us more flexibility and we can allow the user to specify base asset inputs at will.


digorithm (commented):
Oh no, the branch's got some conflicts!

Br1ght0ne (approved):
LGTM, sweet cache impl!

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -0,0 +1,160 @@
...
+    pub fn new(ttl: Duration) -> Self {
+        Self {
+            ttl,
+            items: HashMap::default(),
+        }
+    }
+
+    pub fn insert_multiple(&mut self, coin_ids: HashMap<CoinCacheKey, Vec<CoinTypeId>>) {
+        coin_ids.into_iter().for_each(|(key, ids)| {
+            let new_items = ids.into_iter().map(|coin_id| CoinCacheItem::new(coin_id));
```
hal3e, line 35:
```suggestion
            let new_items = ids.into_iter().map(CoinCacheItem::new);
```

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -0,0 +1,160 @@
...
+            .unwrap_or_default()
+            .into_iter()
+            .map(|item| item.id)
+            .collect()
+    }
+
+    fn remove_expired_entries(&mut self, key: &CoinCacheKey) {
+        self.items
+            .get_mut(key)
+            .map(|entry| entry.retain(|item| item.is_valid(self.ttl)));
```
hal3e, line 57:
```suggestion
        if let Some(entry) = self.items.get_mut(key) {
            entry.retain(|item| item.is_valid(self.ttl));
        }
```

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -0,0 +1,160 @@
...
+    #[test]
+    fn test_insert_and_get_active() {
+        let mut cache = CoinsCache::new(Duration::from_secs(60));
+
+        let key: CoinCacheKey = Default::default();
+        let (item1, item2) = get_items();
+        let items = HashMap::from(HashMap::from([(
+            key.clone(),
+            vec![item1.clone(), item2.clone()],
+        )]));
```
hal3e, line 115:
```suggestion
        let items = HashMap::from([(key.clone(), vec![item1.clone(), item2.clone()])]);
```

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -0,0 +1,160 @@
...
+        assert!(active_coins.contains(&item2));
+    }
+
+    #[tokio::test]
+    async fn test_insert_and_expire_items() {
+        let mut cache = CoinsCache::new(Duration::from_secs(10));
+
+        let key: CoinCacheKey = Default::default();
+        let (item1, _) = get_items();
+        let items = HashMap::from(HashMap::from([(key.clone(), vec![item1.clone()])]));
```
hal3e, line 132:
```suggestion
        let items = HashMap::from([(key.clone(), vec![item1.clone()])]);
```

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        cache.insert_multiple(items);
+
+        // Advance time by more than the cache's TTL
+        tokio::time::pause();
+        tokio::time::advance(Duration::from_secs(12)).await;
+
+        let (_, item2) = get_items();
+        let items = HashMap::from(HashMap::from([(key.clone(), vec![item2.clone()])]));
```
hal3e, line 141:
```suggestion
        let items = HashMap::from([(key.clone(), vec![item2.clone()])]);
```

packages/fuels-accounts/src/provider.rs
```diff
@@ -422,6 +448,59 @@ impl Provider {
...
+        mut filter: ResourceFilter,
+    ) -> ProviderResult<Vec<CoinType>> {
+        self.extend_filter_with_cached(&mut filter).await;
+        self.request_coins_to_spend(filter).await
+    }
+
+    #[cfg(feature = "coin-cache")]
+    async fn extend_filter_with_cached(&self, filter: &mut ResourceFilter) {
+        let mut cache = self.cache.lock().await;
+        let used_coins = cache.get_active(&(filter.from.clone(), filter.asset_id.clone()));
```
hal3e, line 478:
```suggestion
        let used_coins = cache.get_active(&(filter.from.clone(), filter.asset_id));
```

packages/fuels-programs/src/contract.rs
```diff
@@ -299,8 +299,9 @@ impl Contract {
         tx_parameters: TxParameters,
     ) -> Result<Bech32ContractId> {
         let network_info = account.try_provider()?.network_info().await?;
+        let _chain_id = network_info.chain_id();
```
hal3e, line 302:
```suggestion
```

packages/fuels/Cargo.toml
```diff
@@ -32,10 +32,12 @@ fuels-test-helpers = { workspace = true }
 hex = { workspace = true, default-features = false }
 sha2 = { workspace = true }
 tempfile = { workspace = true }
-tokio = { workspace = true }
+tokio = { workspace = true, features = ["time", "test-util"] }
 
 [features]
 default = ["std", "fuels-test-helpers?/fuels-accounts"]
+coin-cache = ["fuels-accounts/coin-cache"]
```
hal3e, line 39: should this be used by default ?

hal3e (changes requested):
Great devX improvement! Nicely done!

I have some minor change requests...

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -109,22 +111,26 @@ macro_rules! impl_tx_trait {
...
+                *tx.gas_price_mut() = 0;
+                *tx.gas_limit_mut() = self.network_info.consensus_parameters.max_gas_per_tx;
+
+                tx.estimate_predicates(
+                    &self.network_info.consensus_parameters,
+                    &self.network_info.gas_costs,
+                )?;
+
+                *tx.gas_price_mut() = gas_price;
+                *tx.gas_limit_mut() = gas_limit;
```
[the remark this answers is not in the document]
MujkicA: I mainly did this because using `build()` leads to an error if signatures are not present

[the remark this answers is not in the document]
hal3e: Makes sense, but the current implementation does not remove the error. `build()` just calls the `resolve_fuel_tx` and setts a boolean. If you want to fix this here, I suggest you separate converting the inputs and appending missing witnesses (this is where the error originates -> there is a SignedRedource without an unresolved signature )

packages/fuels/Cargo.toml
MujkicA: @digorithm

packages/fuels-core/src/types/transaction_builders.rs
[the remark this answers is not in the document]
MujkicA: Right. I think I'll just go with build then. Estimating without signatures doesn't make too much sense, the tx can't be executed without the signatures.

segfault-magnet (commented):
Great work. Clean overall. 

I have a suggestion for the Provider.

Rather than riddle it with `cfg` compile-time conditional code why not separate the logic into a strategy? It would concern itself with allowing or disallowing resources. Something like this:
```rust
#[async_trait::async_trait]
pub trait ResourcePicker {
    async fn register_used_resources(
        &self,
        coin_ids: impl IntoIterator<Item = (CoinCacheKey, Vec<CoinTypeId>)> + Send + Sync,
    );

    async fn modify_filter(&self, filter: &mut ResourceFilter);
}
```

And have two implementations, choose the correct one over in the provider:
```rust
    #[cfg(not(feature = "coin-cache"))]
    resource_picker: NoRestrictions,
    resource_picker: ReusageAvoider,
```

`NoRestrictions` would just be a No-op impl of the trait. ReusageAvoider can house the logic now present in the provider:
```rust
#[derive(Default, Clone, Debug)]
pub struct ReusageAvoider {
    cache: Arc<Mutex<CoinsCache>>,
}

#[async_trait::async_trait]
impl ResourcePicker for ReusageAvoider {
    async fn register_used_resources(
        &self,
        coin_ids: impl IntoIterator<Item = (CoinCacheKey, Vec<CoinTypeId>)> + Send + Sync,
    ) {
        self.cache.lock().await.insert_multiple(coin_ids);
    }

    async fn modify_filter(&self, filter: &mut ResourceFilter) {
        let mut cache = self.cache.lock().await;
        let used_coins = cache.get_active(&(filter.from.clone(), filter.asset_id));

        let excluded_utxos = used_coins
            .iter()
            .filter_map(|coin_id| match coin_id {
                CoinTypeId::UtxoId(utxo_id) => Some(utxo_id),
                _ => None,
            })
            .cloned()
            .collect::<Vec<_>>();

        let excluded_message_nonces = used_coins
            .iter()
            .filter_map(|coin_id| match coin_id {
                CoinTypeId::Nonce(nonce) => Some(nonce),
                _ => None,
            })
            .cloned()
            .collect::<Vec<_>>();

        filter.excluded_utxos.extend(excluded_utxos);
        filter
            .excluded_message_nonces
            .extend(excluded_message_nonces);
    }
}
```

Since the choice is a compile-time one, we should feel very little if any impact when using the noop implementation (after optimizations).

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -117,6 +127,8 @@ pub trait Transaction: Into<FuelTransaction> + Clone {
 
     /// Append witness and return the corresponding witness index
     fn append_witness(&mut self, witness: Witness) -> usize;
+
+    fn used_coins(&self) -> HashMap<(Bech32Address, AssetId), Vec<CoinTypeId>>;
```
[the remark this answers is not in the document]
MujkicA: I was thinking to move away from the "Resource" naming scheme in a future PR completely since core refers to both messages and coins as CoinType now.

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -0,0 +1,157 @@
...
+use std::hash::{Hash, Hasher};
+use tokio::time::{Duration, Instant};
+
+use fuel_types::AssetId;
+use fuels_core::types::bech32::Bech32Address;
+use fuels_core::types::coin_type_id::CoinTypeId;
+
+type CoinCacheKey = (Bech32Address, AssetId);
+
+const DEFAULT_TTL: u64 = 30;
```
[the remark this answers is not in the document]
MujkicA: I wanted to make it more apparent in case we decide to play around with this parameter

packages/fuels-accounts/src/account.rs
```diff
@@ -166,11 +146,50 @@ pub trait Account: ViewOnlyAccount {
         ]
     }
 
-    async fn add_fee_resources<Tb: TransactionBuilder>(
+    /// Get some spendable resources (coins and messages) of asset `asset_id` owned by the account
+    /// that add up at least to amount `amount`. The returned coins (UTXOs) are actual coins that
+    /// can be spent. The number of UXTOs is optimized to prevent dust accumulation.
```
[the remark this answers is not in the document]
MujkicA: Now that I think about it, "The returned coins (UTXOs) are actual coins that can be spent" seems a bit excessive given the previous sentence. I will omit it and leave the full consistency regarding the coin/resource terminology to a separate PR.

Voxelot:
I didn't review the full PR, but from a quick read of the PR description it seems like we should maybe try to invalidate cached UTXOs in the event a transaction is unsuccessful. Otherwise users may need to wait for the TTL before being able to try again. This would only work if the user submits_and_awaits so that the sdk can check the response. Alternatively, a short TTL such as `5 * block_time_ms` might be good enough in most cases.

digorithm (commented):
> I didn't review the full PR, but from a quick read of the PR description it seems like we should maybe try to invalidate cached UTXOs in the event a transaction is unsuccessful. Otherwise users may need to wait for the TTL before being able to try again. This would only work if the user submits_and_awaits so that the sdk can check the response.
> 
> Alternatively, a short TTL such as `5 * block_time_ms` might be good enough in most cases (since we currently use 1s blocks).

Coincidentally, we discussed this scenario yesterday (where the tx failed, and the user would wait for the invalidation). 

@MujkicA How hard is it to adapt these changes to support invalidating an entry upon tx failure? If it's too hard, maybe we should go with a short TTL and do this later. If not, it might be worth doing it now.

packages/fuels/tests/providers.rs
```diff
@@ -848,3 +849,63 @@ async fn test_sway_timestamp() -> Result<()> {
...
+        network_info,
+    );
+    wallet.sign_transaction(&mut tb);
+    wallet.adjust_for_fee(&mut tb, amount).await?;
+    tb.build()
+}
+
+#[cfg(feature = "coin-cache")]
+#[tokio::test]
+async fn test_caching() -> Result<()> {
```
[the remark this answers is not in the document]
MujkicA: Good catch, the test missing some stuff. I've updated it now

MujkicA:
> I didn't review the full PR, but from a quick read of the PR description it seems like we should maybe try to invalidate cached UTXOs in the event a transaction is unsuccessful. Otherwise users may need to wait for the TTL before being able to try again. This would only work if the user submits_and_awaits so that the sdk can check the response.
> 
> Alternatively, a short TTL such as `5 * block_time_ms` might be good enough in most cases (since we currently use 1s blocks).

I thought about it but decided against it because it works only partially, i.e. when they use send_and_await. I imagine using send_and_await is more likely in local tests then out in the wild or if it is used on a live network, I think it's unlikely they will encounter the problem of using the same inputs since they wait for completion anyways.
It doesn't seem like a big win, but it would add a bit more complexity. Let me know if my assumptions are wrong @Voxelot.

On the other hand, we could probably benefit from invalidating at places where we call submit_and_await internally like contract calls. So maybe it's worth it.

MujkicA:
**Update**

I've adjusted it to invalidate on failed txs during `submit_and_await` and included a test. I've also fixed the general test for caching.

packages/fuels/Cargo.toml
digorithm: I think so, yes. My rationale is that this feature is mostly to protect nodes against heavy querying. I don't see application builders being too mindful of this, so having this on by default should be safer. However, these users might run into caching-related issues, and then they will be mindful of checking this feature and, if needed, disabling it.

digorithm (commented):
I did a first pass, and it's looking solid. I'll do another pass once the CI is green (I believe this is test-related so it shouldn't change much).

packages/fuels-accounts/Cargo.toml
```diff
@@ -25,7 +25,7 @@ rand = { workspace = true, default-features = false }
 semver = { workspace = true }
 tai64 = { workspace = true, features = ["serde"] }
 thiserror = { workspace = true, default-features = false }
-tokio = { workspace = true, features = ["full"], optional = true}
+tokio = { workspace = true, features = ["full", "time", "test-util"], optional = true}
```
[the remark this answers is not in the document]
digorithm: Yeah, I agree with this approach. But I also agree with @MujkicA: we should leave it for a follow-up PR. We should get this out ASAP. Let's not forget to open an issue for this once this is merged, though.

packages/fuels-accounts/src/coin_cache.rs
```diff
@@ -32,7 +30,7 @@ impl CoinsCache {
 
     pub fn insert_multiple(
         &mut self,
-        coin_ids: impl IntoIterator<Item = (CoinCacheKey, Vec<CoinTypeId>)>,
+        coin_ids: impl IntoIterator<Item = ((Bech32Address, AssetId), Vec<CoinTypeId>)>,
```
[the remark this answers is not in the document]
MujkicA: Quoting @segfault-magnet from an earlier comment:

> I wouldn't use a private alias in a public method, it will work but LSPs are going to annoy users by plugging in a CoinCacheKey they have no hope of importing.

Public methods don't make use of `CoinCacheKey `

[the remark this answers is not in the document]
MujkicA: On second thought, I think we had a misunderstanding there. Users are never supposed to directly interact with the cache or it's methods. I will use `CoinCacheKey` in all places. 
Good catch!

segfault-magnet (dismissed):

packages/fuels-accounts/src/provider.rs
```diff
@@ -188,7 +203,17 @@ impl Provider {
...
+
+        #[cfg(feature = "coin-cache")]
+        {
+            if matches!(
+                _status,
+                TransactionStatus::SqueezedOut { .. } | TransactionStatus::Failure { .. }
+            ) {
+                self.cache.lock().await.remove_items(tx.used_coins())
+            }
+        }
```
digorithm, line 216: Is there any way we can test this at this layer (as opposed to only testing the cache implementation)? E.g., forcing a tx failure and checking the cached utxo. It doesn't have to be done in this PR, though. But it would be nice to have a proper check at this level.

digorithm (approved):

Salka1988 (approved):

hal3e (approved):
:rocket:

PR #1106, opened 2023-08-28
feat: return Enums containing heap types at depth 1

### LINKED ISSUE #940: Returning a wrapped `Bytes` type from contracts
Returning of a `Bytes` type has recently been introduced and is playing a vital part in developer's contracts, libraries, and Fuel standards. This has enabled a number of different features such as the `String` type. 
However, returning a wrapped `Bytes` type is currently prohibited. This is key for certain features, such as returning a `String` or `Option<Bytes>`.

The current definition of the `String` type is the following. This has enabled UTF-8 encoded strings in the Sway language and opened the possibility for NFT's URIs, Fuel Name Service, and more.
```rust
pub struct String {
    bytes: Bytes,
}
```

An example where this is needed is in the [FRC-20 and FRC-721 standards](https://github.com/FuelLabs/sway-libs/tree/master/standards) that have been introduced. There has recently been a push to create standards for Fuel and having the ability to return a `String` is **critical**.  We currently have the following ABI definitions for our standards:

```rust
    /// Get the name of the token
    /// Example (with trailing padding): "MY_TOKEN                                                        "
    fn name() -> str[64];
```
```rust
    /// A distinct Uniform Resource Identifier (URI) for a given asset.
    /// URIs are defined in RFC-3986.
    fn uri(token_id: u64) -> str[2048];
```

These definitions suffer from having a static length, requiring lots of white space or possibly not being long enough in certain cases. Instead, we would like to update these to the following:
```rust
    /// Get the name of the token
    fn name() -> String;
```
```rust
    /// A distinct Uniform Resource Identifier (URI) for a given asset.
    /// URIs are defined in RFC-3986.
    fn uri(token_id: u64) -> String;
```

As of today, we are able to pass a `String` to a contract by manually constructing the struct and passing it to the contract. While this is less than ideal, it still offers the ability to pass a `String` to a contract. For example, here we construct the String struct and pass it to the contract. This works as expected.
```rust
#[tokio::test]
async fn takes_string_test() {
    let instance = setup().await;

    let string = "Fuel is blazingly fast!";
    let input = String {
        bytes: Bytes(string.as_bytes().to_vec()),
    };

    instance.methods().takes_string(input).call().await.unwrap();
}
```
 However, returning of a `String` is incompatible with the rust sdk. While the following compiles, it will give a runtime error using the rust sdk:

```rust 
    fn returns_string() -> String {
        let mut my_string = String::new();

        my_string.bytes.push(0u8);
        my_string.bytes.push(1u8);
        my_string.bytes.push(2u8);

        my_string
    }

    fn returns_bytes_option() -> Option<Bytes> {
        let mut my_bytes = Bytes::new();

        my_bytes.push(0u8);
        my_bytes.push(1u8);
        my_bytes.push(2u8);

        Option::Some(my_bytes)
    }
```
Screenshot 2023-04-18 131334

Returning a wrapped `Bytes` type is **necessary** for URIs, token names, FNS, symbols, FRC standards, and more within the Fuel ecosystem. 

This has been tested with forc v0.37.0 and fuels-rs v0.41.0

iqdecay:
This PR enables returning heap types surrounded in an `Enum` type. It notably enables returning
a `Result<Vec>` for instance from a contract call.

Closes #940.

Main disavantages from this approach: 
- We are injecting more custom bytecode
- The interdependencies of the decoder and the bytecode make the code hard to navigate
- Enum can contain only one variant that uses a heap type inside. It could be extended to having an arbitrary number of them, but they would all have to have the same encoding width. This is a personal design choice, open to criticism.


roshnidhanjal21:
Preparing review...

roshnidhanjal21:
Preparing review...

roshnidhanjal21:
Preparing review...

packages/fuels-programs/src/call_utils.rs
```diff
@@ -143,11 +144,31 @@ pub(crate) async fn build_tx_from_contract_calls(
...
+                && matches!(c.output_param, ParamType::Enum { .. })
+        })
         .count();
-    let n_stack_type_calls = calls.len() - n_heap_type_calls;
+
+    let n_heap_type_calls = calls
+        .iter()
+        .filter(|c| c.output_param.heap_inner_element_size().is_some())
+        .count()
+        - n_enum_heap_type_calls;
```
segfault-magnet, line 159: An alternative:
```rust
    let heap_calls = || {
        calls
            .iter()
            .map(|call| &call.output_param)
            .filter(|param| param.uses_heap_types(false))
    };

    let n_enum_heap_type_calls = heap_calls()
        .filter(|param| matches!(param, ParamType::Enum { .. }))
        .count();

    let n_heap_type_calls = heap_calls().count() - n_enum_heap_type_calls;
```
It would seem `uses_heap_types(false)` is equivalent to `heap_inner_element_size().is_some()` but more concise in meaning.

packages/fuels-programs/src/call_utils.rs
```diff
@@ -334,24 +357,51 @@ pub(crate) fn get_single_call_instructions(
...
+                // The RET register contains the pointer address of the `CALL` return (a stack
+                // address).
+                // The RETL register contains the length of the `CALL` return (=24 in the case of
+                // Vec/Bytes/String because their struct takes 3 WORDs). We don't actually need it
+                // unless the Vec/Bytes/String struct encoding changes in the compiler.
+                // Load the word located at the address contained in RET, it's a word that translates to
+                // a heap address. We use the offset because in the case of an enum containing a heap
+                // type, we need to get the pointer that is located at the heap address pointed to by
+                // the RET register.
+                op::lw(0x15, RegId::RET, offset),
```
segfault-magnet, line 380: This load must be hidden behind the `jnef` check below.

Consider the following sway fn:
```sway
    fn a_problem() -> Result<Bytes, b256> {
        Result::Err(0x1111111111111111111111111111111111111111111111111111111111111111)
    }
```

This will cause a `MemoryOverflow` panic when executed. 

This is because we read the memory here without checking whether the discriminant is pointed towards the heap type. If it is not then we're reading whatever *is* there as a ptr.

This wasn't caught by tests because the error types used, when interpreted as a ptr, pointed to allocated memory. In order to showcase the bug I made the error a `b256` with a high value, so that we will index out-of-bounds memory.

packages/fuels-programs/src/receipt_parser.rs
```diff
@@ -69,6 +75,24 @@ impl ReceiptParser {
         }
     }
 
+    fn extract_enum_heap_type_data(&mut self, contract_id: &ContractId) -> Option<Vec<u8>> {
+        for (index, (current_receipt, next_receipt)) in
+            self.receipts.iter().tuple_windows().enumerate()
+        {
+            if let (Some(first_data), Some(second_data)) =
+                Self::extract_vec_data(current_receipt, next_receipt, contract_id)
```
segfault-magnet, line 83: I'd rename this as it no longer pertains to only vectors.

packages/fuels/tests/types_contracts.rs
```diff
@@ -1,9 +1,14 @@
+#![allow(unused_imports, unused_variables, unused_mut, dead_code)]
```
segfault-magnet, line 1: Should be removed.

segfault-magnet (changes requested):
Must have been a pain to get it working. Kudos for sticking through it. 

Requesting changes due to:
1. the bug with memory overflow
2. Using Result<MyStruct { offset: u8, bytes: Bytes }, ...> causes `ContractNotFound` when called. We should either adapt the bytecode generation to be able to call such methods (even though decoding is going to fail) or prohibit calling them altogether (less flexible). But `ContractNotFound` is probably a bad error message for not supporting nested heap types.
3. Regression in enums as arguments.

packages/fuels-core/src/types/enum_variants.rs
```diff
@@ -34,6 +33,15 @@ impl EnumVariants {
         })
     }
 
+    pub fn heap_type_variant(&self) -> Option<(u8, &ParamType)> {
+        // TODO: segfault this needs to look at only the parent heap types. Or maybe to extract
+        // this from here because this limitation is not present when enums are used as args.
```
segfault-magnet, line 38:
```suggestion
```

packages/fuels-core/src/types/param_types.rs
```diff
@@ -84,57 +84,85 @@ impl ParamType {
...
+                if grandchildren_need_receipts {
+                    return Err(error!(
+                        InvalidType,
+                        "Enums currently support only one level deep heap types."
+                    ));
+                }
+
+                let num_of_children_needing_receipts = all_param_types
+                    .filter(|param_type| param_type.needs_extra_data_receipt(false))
+                    .count();
```
segfault-magnet, line 122: No need for `chain!` and cloning of the param types:
```suggestion
                let all_param_types = variants.param_types();
                let grandchildren_need_receipts = all_param_types
                    .iter()
                    .any(|child| child.children_need_extra_receipts());
                if grandchildren_need_receipts {
                    return Err(error!(
                        InvalidType,
                        "Enums currently support only one level deep heap types."
                    ));
                }

                let num_of_children_needing_receipts = all_param_types
                    .iter()
                    .filter(|param_type| param_type.needs_extra_data_receipt(false))
                    .count();
```

packages/fuels-core/src/types/param_types.rs
```diff
@@ -84,57 +84,86 @@ impl ParamType {
...
+            _ => Ok(()),
+        }
     }
 
-    pub fn is_vm_heap_type(&self) -> bool {
-        matches!(
-            self,
-            ParamType::Vector(..) | ParamType::Bytes | ParamType::String
-        )
+    pub fn needs_extra_data_receipt(&self, top_level_type: bool) -> bool {
```
MujkicA, line 142: Small nit here, seeing `needs_extra_data_receipt(true)` makes it seem like a setter. Would be a bit more readable if we go for somethign like `is_extra_receipt_needed(true)` or similar.

MujkicA (approved):

hal3e (approved):

Br1ght0ne (approved):

Issue #1117, opened 2023-08-30
Add storage slot initialization for contract deployment 


cold-briu:
As seen in [sway/4634](https://github.com/FuelLabs/sway/issues/4634)

Storage initialization values won't be automatically set by the sdk since the initialization values are output into a separate json file from the abi. 

The sdk should automatically include storage slots. Otherwise, an additional config step is required to deploy the contract as seen [here](https://github.com/FuelLabs/fuel-docs/pull/102).

This is the [fix](https://github.com/FuelLabs/fuels-ts/pull/1220) for the same issue on fuels-ts.

Issue #1118, opened 2023-08-30
Paying for contract deployment with vector-in-args predicates fails if storage_slots.len() > 0


segfault-magnet:
Predicates that accept heap types, when being used to pay for a contract deployment, fail to validate if any storage slots are used.

PR #1120, opened 2023-08-31
feat!: autoload storage slots

segfault-magnet:
closes: #1117 #1118 

Storage slots are now loaded automatically by default (an opt-out exists). 

If the `StorageConfiguration` indicates that autoload should take place, `Contract::load_from` will try and discover the storage slots file in the same directory as the contract binary to be loaded.

If it is for some reason missing this constitutes an error. The user will be guided to either provide the file or disable autoloading.

Any additional storage slots configured in `StorageConfiguration` act as an override taking precedence over what was loaded from the storage slots file. 

Enabling the autoloading uncovered a bug where we didn't account for storage slots when calculating the heap data offset for predicates.

### Breaking
The storage configuration interface changed
Autoloading by default might change behavior.
### Checklist

### DIFF
--- a/docs/src/deploying/storage-slots.md
+++ b/docs/src/deploying/storage-slots.md
@@ -1,6 +1,6 @@
 # Overriding storage slots
 
-If you use storage in your contract the default storage values are going to be generated in a JSON file (e.g. `my_contract-storage_slots.json`) by the Sway compiler. These are loaded automatically for you when you load a contract binary. If you wish to override some of the defaults you need to provide the corresponding storage slots manually:
+If you use storage in your contract, the default storage values will be generated in a JSON file (e.g. `my_contract-storage_slots.json`) by the Sway compiler. These are loaded automatically for you when you load a contract binary. If you wish to override some of the defaults, you need to provide the corresponding storage slots manually:
 
 ```rust,ignore
 {{#include ../../../examples/contracts/src/lib.rs:storage_slots_override}}
--- a/docs/src/deploying/storage-slots.md
+++ b/docs/src/deploying/storage-slots.md
@@ -1,0 +1,13 @@
+# Overriding storage slots
+
+If you use storage in your contract the default storage values are going to be generated in a JSON file (e.g. `my_contract-storage_slots.json`) by the Sway compiler. These are loaded automatically for you when you load a contract binary. If you wish to override some of the defaults you need to provide the corresponding storage slots manually:
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:storage_slots_override}}
+```
+
+If you don't have the slot storage file (`my_contract-storage_slots.json` example from above) for some reason, or you don't wish to load any of the default values, you can disable the auto-loading of storage slots:
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:storage_slots_disable_autoload}}
+```
--- a/examples/contracts/src/lib.rs
+++ b/examples/contracts/src/lib.rs
@@ -1,8 +1,11 @@
 #[cfg(test)]
 mod tests {
-    use fuels::types::{
-        errors::{error, Error, Result},
-        Bits256,
+    use fuels::{
+        prelude::{LoadConfiguration, StorageConfiguration},
+        types::{
+            errors::{error, Error, Result},
+            Bits256,
+        },
     };
 
     #[tokio::test]
@@ -791,6 +794,35 @@ mod tests {
         assert_eq!(result_uint, 2);
         assert!(result_bool);
         assert_eq!(result_str, "fuel");
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn storage_slots_override() -> Result<()> {
+        {
+            // ANCHOR: storage_slots_override
+            use fuels::{programs::contract::Contract, tx::StorageSlot};
+            let slot_override = StorageSlot::new([1; 32].into(), [2; 32].into());
+            let storage_config =
+                StorageConfiguration::default().add_slot_overrides([slot_override]);
+
+            let load_config =
+                LoadConfiguration::default().with_storage_configuration(storage_config);
+            let _: Result<Contract> = Contract::load_from("...", load_config);
+            // ANCHOR_END: storage_slots_override
+        }
+
+        {
+            // ANCHOR: storage_slots_disable_autoload
+            use fuels::programs::contract::Contract;
+            let storage_config = StorageConfiguration::default().with_autoload(false);
+
+            let load_config =
+                LoadConfiguration::default().with_storage_configuration(storage_config);
+            let _: Result<Contract> = Contract::load_from("...", load_config);
+            // ANCHOR_END: storage_slots_disable_autoload
+        }
 
         Ok(())
     }
--- a/examples/contracts/src/lib.rs
+++ b/examples/contracts/src/lib.rs
@@ -138,7 +138,8 @@ mod tests {
         let key = Bytes32::from([1u8; 32]);
         let value = Bytes32::from([2u8; 32]);
         let storage_slot = StorageSlot::new(key, value);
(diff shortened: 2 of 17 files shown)


docs/src/deploying/storage-slots.md
```diff
@@ -0,0 +1,13 @@
+# Overriding storage slots
+
+If you use storage in your contract the default storage values are going to be generated in a JSON file (e.g. `my_contract-storage_slots.json`) by the Sway compiler. These are loaded automatically for you when you load a contract binary. If you wish to override some of the defaults you need to provide the corresponding storage slots manually:
```
digorithm, line 3: Some tiny grammar nits:

```suggestion
If you use storage in your contract, the default storage values will be generated in a JSON file (e.g. `my_contract-storage_slots.json`) by the Sway compiler. These are loaded automatically for you when you load a contract binary. If you wish to override some of the defaults, you need to provide the corresponding storage slots manually:
```

examples/contracts/src/lib.rs
```diff
@@ -791,4 +797,33 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn storage_slots_override() -> Result<()> {
+        {
+            // ANCHOR: storage_slots_override
+            use fuels::{programs::contract::Contract, tx::StorageSlot};
+            let slot_override = StorageSlot::new([1; 32].into(), [2; 32].into());
```
digorithm, line 806: Minor UX-related suggestion: how hard would it be to change this signature from `[1; 32].into()` to `[1; 32]` (i.e., directly accepting the slice)?

digorithm (changes requested):
Very clean implementation and good UX choices. Left just a couple of nits/suggestions.

hal3e (commented):
IMO it is a bit weird that we try to get the storage slots path from the bin path. What do you think that instead we ask for the sway project folder. From there we can get both the bin and the storage slots and we can generate nice error messages. Something  like this:

```rust
    let contract = Contract::load_from(
        "some/path/sway_project_name",
        LoadConfiguration::default().
    )?;
```

segfault-magnet:
> IMO it is a bit weird that we try to get the storage slots path from the bin path. What do you think that instead we ask for the sway project folder. From there we can get both the bin and the storage slots and we can generate nice error messages. Something like this:
> 
> ```rust
>     let contract = Contract::load_from(
>         "some/path/sway_project_name",
>         LoadConfiguration::default().
>     )?;
> ```

I share your feelings, but have the following reservations:

If we start giving project paths to the `Contract::load_from` then we're implicitly expecting a certain layout. The bin file must be named as the dir is, the ABI has `-abi.json` appended etc.

We would expect that the user gave us the root dir of the project (the one where `Forc.toml` lives) because we need it to extract the project name (then to get the contract bin name etc). 

But is the build a `debug` or a `release` one? We would need to add a flag for that as well due to the artifacts landing in `out/debug` and `out/release` respectively. 

Is the project titled the same as the dir? The `name` property of the `Forc.toml` can differ from the project rootdir name. From the `name` property all the generated files derive their names. 

What if somebody just wants to load a contract binary from outside a project? We would need to support them and provide an alternative, custom-path-loading mechanism. Or they would need to recreate this folder structure by hand.

If we start accepting project paths, then we would probably want to be consistent in the `abigen!` as well. So that users might have the same experience of "give the thing a path to your project". Otherwise, they would remember `abigen!` gets the full path to the JSON, contracts get the project, and predicates again get the custom path (or they get the project). 

Although it seems `scripty` (for lack of a better word), the fact remains that for projects it will detect the storage slots reliably, for one-off contract binaries the autoloading can be disabled and both parties are content.

examples/contracts/src/lib.rs
segfault-magnet: We'd need to have our own wrapper around `StorageSlot`. Currently, it is just being exported from `fuel_tx`.

Some care would also be needed to check that we consistently use our wrapper and not mix-match with the `fuel_tx` one.

We might also want to consider at some point if we have any benefit in the SDK from the wrapper around `[u8; 32]` that `Bytes32` is. They're causing this `into` ux snag here.

I'd rate it a 4/10 on a PITA scale. Drop a msg and I'll implement it.

MujkicA (approved):

examples/contracts/src/lib.rs
hal3e: I would not add the wrapper

hal3e:
> > IMO it is a bit weird that we try to get the storage slots path from the bin path. What do you think that instead we ask for the sway project folder. From there we can get both the bin and the storage slots and we can generate nice error messages. Something like this:
> > ```rust
> >     let contract = Contract::load_from(
> >         "some/path/sway_project_name",
> >         LoadConfiguration::default().
> >     )?;
> > ```
> 
> I share your feelings, but have the following reservations:
> 
> If we start giving project paths to the `Contract::load_from` then we're implicitly expecting a certain layout. The bin file must be named as the dir is, the ABI has `-abi.json` appended etc.
> 
> We would expect that the user gave us the root dir of the project (the one where `Forc.toml` lives) because we need it to extract the project name (then to get the contract bin name etc).
> 
> But is the build a `debug` or a `release` one? We would need to add a flag for that as well due to the artifacts landing in `out/debug` and `out/release` respectively.
> 
> Is the project titled the same as the dir? The `name` property of the `Forc.toml` can differ from the project rootdir name. From the `name` property all the generated files derive their names.
> 
> What if somebody just wants to load a contract binary from outside a project? We would need to support them and provide an alternative, custom-path-loading mechanism. Or they would need to recreate this folder structure by hand.
> 
> If we start accepting project paths, then we would probably want to be consistent in the `abigen!` as well. So that users might have the same experience of "give the thing a path to your project". Otherwise, they would remember `abigen!` gets the full path to the JSON, contracts get the project, and predicates again get the custom path (or they get the project).
> 
> Although it seems `scripty` (for lack of a better word), the fact remains that for projects it will detect the storage slots reliably, for one-off contract binaries the autoloading can be disabled and both parties are content.

I agree with this and I do not mind going with this option if we do not find a better one. 

> > IMO it is a bit weird that we try to get the storage slots path from the bin path. What do you think that instead we ask for the sway project folder. From there we can get both the bin and the storage slots and we can generate nice error messages. Something like this:
> > ```rust
> >     let contract = Contract::load_from(
> >         "some/path/sway_project_name",
> >         LoadConfiguration::default().
> >     )?;
> > ```
> 
> I share your feelings, but have the following reservations:
> 
> If we start giving project paths to the `Contract::load_from` then we're implicitly expecting a certain layout. The bin file must be named as the dir is, the ABI has `-abi.json` appended etc.
> 
> We would expect that the user gave us the root dir of the project (the one where `Forc.toml` lives) because we need it to extract the project name (then to get the contract bin name etc).
> 
> But is the build a `debug` or a `release` one? We would need to add a flag for that as well due to the artifacts landing in `out/debug` and `out/release` respectively.
> 
> Is the project titled the same as the dir? The `name` property of the `Forc.toml` can differ from the project rootdir name. From the `name` property all the generated files derive their names.
> 
> What if somebody just wants to load a contract binary from outside a project? We would need to support them and provide an alternative, custom-path-loading mechanism. Or they would need to recreate this folder structure by hand.
> 
> If we start accepting project paths, then we would probably want to be consistent in the `abigen!` as well. So that users might have the same experience of "give the thing a path to your project". Otherwise, they would remember `abigen!` gets the full path to the JSON, contracts get the project, and predicates again get the custom path (or they get the project).
> 
> Although it seems `scripty` (for lack of a better word), the fact remains that for projects it will detect the storage slots reliably, for one-off contract binaries the autoloading can be disabled and both parties are content.

I agree. Let's use this until we have something better!

hal3e (approved):

examples/contracts/src/lib.rs
digorithm: I'm pushing this for later; I don't want to stall this work because of this discussion.

digorithm (approved):

PR #1121, opened 2023-09-05
feat: pull tx param defaults from network

### LINKED ISSUE #1113: Fetch the default values for the transaction from the network instead of hardcoding
Fetch the default values for the transaction from the network endpoint instead of hardcoding them in the code directly.

As an example, the current default value is more than allowed by beta 4, and it causes errors like:

```bash
>>> forc deploy --node-url http://beta-4.fuel.network/graphql --gas-price 1 
Error: Decode error: Custom { kind: Other, error: "Response errors; TransactionGasLimit" }
```

MujkicA:
Closes #1113

Changes `gas_limit` and `gas_price` in `TxParams` to be optional. `TransactionBuilder`s use the corresponding values from `ConsensusParameters` and `NodeInfo` during `build()` if these fields were set to `None`.

`Transaction` still holds the actual values, at that point only manual changes of the parameters are possible. 

### Checklist


packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -41,8 +41,8 @@ pub trait TransactionBuilder: Send {
     fn add_unresolved_signature(&mut self, owner: Bech32Address, secret_key: SecretKey);
     fn fee_checked_from_tx(&self, params: &ConsensusParameters) -> Option<TransactionFee>;
     fn with_maturity(self, maturity: u32) -> Self;
-    fn with_gas_price(self, gas_price: u64) -> Self;
-    fn with_gas_limit(self, gas_limit: u64) -> Self;
+    fn with_gas_price(self, gas_price: Option<u64>) -> Self;
```
segfault-magnet, line 44: We should decide on our approach to optional values. For example, in `CallParameters` we store `gas_forwarded` as an `Option<u64>`. 

But the 'setter' looks like this:
```rust
    pub fn with_gas_forwarded(mut self, gas_forwarded: u64) -> Self {
        self.gas_forwarded = Some(gas_forwarded);
```

So if you don't call the setter you'd get the default.

The downside is once you override the gas price there is no turning back to the 'use gas price from consensus params.

If we decide that reverting is something that we wish to support then I suggest a different approach: 

Have the setter resemble the one from `CallParameters`:
```rust
pub fn with_gas_price(mut self, gas_price: u64) -> Self {
     self.gas_price = Some(gas_price);
     // ....
```

But also have another method for reverting:
```rust
pub fn use_gas_price_from_network(mut self) -> Self {
  self.gas_price = None;
  self
  }
  ```
Is more explicit than explaining that 'None' means consensus params and Some(...) will actually set it.

Another approach would be an enum:
```rust
enum GasPriceSource {
ConsensusParameters,
Fixed(u64)
}
```

And give that to the `with` method. 

Whatever we choose, we should be consistent in the SDK.

segfault-magnet (commented):
Also tests failing, OutOfGas. @MujkicA

packages/fuels-core/src/types/transaction_builders.rs
MujkicA: The `gas_forwarded` by `CallParameters` is not limited by network parameters, it defaults to `None`. That is, on default we won't be forwarding gas to a call, if the user decided to do so it is their responsibility. And even then, if it exceed the `max_gas_per_tx` threshold, the CALL opcode will uses all remaining gas if the parameter is more than what is available.

I assume the lack of revert support for `CallParameters` hasn't been a problem so far since it is usually created immediately before a call and undergoes no changes. I can change the setter for it in this PR to

```rust
pub fn with_gas_forwarded(mut self, gas_forwarded: Option<u64>) -> Self {
        self.gas_forwarded = Some(gas_forwarded);
```

I think the resulting interface is self explanatory:

```rust
call_params.with_gas_forwarded(None); // No gas will be forwarded
```

But given that `None` communicates different things in these two cases, I see no need to keep consistency. The only remaining question is whether we want to document that `None` in `TxParameters ` means network default or be more explicit in code.

segfault-magnet: I have a feeling we may have a slight misunderstanding. Let me try and reiterate without any examples from our code:

When we have a field:
```rust
some_field: Option<u64>
```
That is initialized as `None` and afterward we want the user to be able to set it to something we can do one of two things:
1. `set_some_field(&mut self, value: u64) { self.some_field = Some(value); }`
2. `set_some_field(&mut self, value: Option<u64>) { self.some_field = value; }`

num 1 is more ergonomic for the user, if they didn't call the setter, then `None` will remain. The UI is simpler by accepting only u64 for the `Some` case and having `None` represented by not calling the setter at all.

The downside is that once set, if you wish to unset it you have to have a separate method, like:
``` rust
fn unset_some_field(&mut self) { self.some_field = None; }
```
Besides the generic `unset_*` name you can give it more meaning like `use_gas_price_from_network`.

Number 2 is the basic setter. More cumbersome but you get all the functionality. 

Having that in mind, there are some parts of our codebase where we do 1 and others where we do 2. I gave the example of `gas_forwarded` as an example of 1, your gas price here is an example of 2.

Seeing as the setter design can be lifted out of the meaning of the field itself, we needn't go into what `gas_forwarded` or `gas_price` means. 

I was suggesting that we think about adopting a coding style guide for such situations. 

My preference, ideally, is to not have `Option`s in structs if they are there to provide a fallback (default) mechanism. For example, do not store `Option<SomeConfig>` if later on `None` means we will use `SomeConfig::default()`, rather store `SomeConfig` in the struct and initialize it with the default value instead. 

Other cases such as yours do not represent a fallback-to-`Default::default()`. Here we can be more explicit in code by either passing in an enum and having the code do the speaking for us OR document what happens if the field isn't set. 

If we go the documentation route we can model the setter as a 'number-1' or a 'number-2'.

I was just pointing out that in some places we do 1, in others 2, without any noticeable pattern :)

hal3e: IMHO I would take option 1. and leave `None` as default and that it means that it is going to take it from the network. I would also not add this `unset` or use option 2. Here is my reasoning:

- If I do not set anything I expect default values. The only difference is that now we take the defaults from another place.
- If I need to set something I will create it and pass it. If I need another param I would just create a new one with the values.
here is an example from our test:
```rust
    let response = contract_methods
        .payable()
        .call_params(
            CallParameters::default()
                .with_amount(100)
                .with_gas_forwarded(20_000),
        )?
        .call()

    let err = contract_methods
        .non_payable()
        .call_params(CallParameters::default().with_amount(100))
        .expect_err("Should return call params error.");

    let response = contract_methods
        .non_payable()
        .call_params(CallParameters::default().with_gas_forwarded(20_000))?
        .call()
        .await?;
 ```
 for me this is ok and verbose enough.
 
- The enum approach seems cumbersome.  I need to create an enum and pass this now as an argument for something that was done easily with option 1.

I agree that whatever we choose we should be consistent :)

Comment (MujkicA) [reply packages/fuels-core/src/types/transaction_builders.rs:44]: @segfault-magnet I see, I was confused because consensus_parameters was mentioned in the context of call params here:
> The downside is once you override the gas price there is no turning back to the 'use gas price from consensus params.
--------
> Having that in mind, there are some parts of our codebase where we do 1 and others where we do 2. I gave the example of gas_forwarded as an example of 1, your gas price here is an example of 2.

My point was that I don't see this as a consistency issue. When the situation allows it, we hide the underlying use of `Option` from the user with 1, making it seem like a normal field from the outside. An example would be `Predicate::with_provider()` where once the provider is set, it doesn't make much sense to go back to `None`.
 And in other cases, we go with 2.

At any rate, I agree with @hal3e suggestion.

Comment (segfault-magnet) [reply packages/fuels-core/src/types/transaction_builders.rs:44]: Let's go with @hal3e's suggestion here and on the new changes to `gas_forwarded`.

**Comment by MujkicA** [comment]
For tx builders to be able to set gas price automatically they also need to access to `NodeInfo`. This only applies when there was no gas price was provided. We already have a similar case with `ConsensusParameters` where they are only required if the tx uses predicates. 

These requirements are awkward since we try to abstract away some of the complexities of creating a tx but end up pushing it on the user to understand when certain parameters are required. 
So the next idea was to make builders hold a `Provider` that let's them retrieve these parameters as needed but as @segfault-magnet suggested, this makes the interface impure and could cause trouble for users that build off of our interface.

Another solution would be to provide tx parameters with the network defaults via a `Provider` method while instantiating `TxParameters` could only be done by providing all 3 values. The downside here is that it again requires some more awareness from the users side. 

I tried to strike a balance by only updating `gas_limit` to the network value since  `ConsensusParameters` is already part of the builders and by using the hardcoded constant for `gas_limit`. AFAIK this should be enough to protect us from further changes to these parameters in beta updates.

**Comment by MujkicA** [comment]
Since the builder requires both `ConsensusParameters` and `NodeInfo` I've created a new struct `NetworkInfo` to hold this data for a tx builder. It can be retrieved via a `Provider`.

This removes the need for `ConsensusParameters` in builders, provides `GasCosts` from network instead of using defaults and helps resolve the tx parameters.

File: packages/fuels-core/src/types/transaction_builders.rs [hunk 2]
```diff
@@ -62,27 +86,20 @@ macro_rules! impl_tx_trait {
...
-                        .consensus_parameters
-                        .ok_or(error!(
-                                TransactionBuildError,
-                                "predicate inputs require consensus parameters. Use `.set_consensus_parameters()`."))?;
-                    (self.base_offset(&consensus_params), consensus_params)
-                } else {
-                    // If no ConsensusParameters have been set, we can use the default instead of
-                    // erroring out since the tx doesn't use predicates
-                    (0, self.consensus_parameters.unwrap_or_default())
+                let base_offset = match uses_predicates {
```
segfault-magnet, line 89: matching on a boolean is a `if-else`
```rust
                let base_offset = if uses_predicates {
                    self.base_offset()
                } else {
                    0
                };
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -216,23 +232,50 @@ pub struct CreateTransactionBuilder {
...
+            outputs: vec![],
+            witnesses: vec![],
+            network_info,
+            unresolved_signatures: Default::default(),
+        }
+    }
+
+    fn resolve_fuel_tx(self, base_offset: usize, num_witnesses: u8) -> Result<Script> {
+        let gas_price = resolve_gas_price(self.gas_price, self.network_info.min_gas_price);
+        let gas_limit = resolve_gas_limit(self.gas_limit, self.network_info.max_gas_per_tx);
```
segfault-magnet, line 274:
```suggestion
        let gas_price = self.gas_price.unwrap_or(self.network_info.min_gas_price);
        let gas_limit = self.gas_limit.unwrap_or(self.network_info.max_gas_per_tx);
```

Salka1988 (dismissed):
👍

Salka1988 (dismissed):

hal3e (commented):
Looks very good! Left a tiny comment...

Salka1988 (approved):

Br1ght0ne (approved):

hal3e (approved):

Issue #1141, opened 2023-09-27
Contract deployment is not being checked for failure


segfault-magnet:
When deploying contracts we only check that the tx status became something other than `Submitted`. 

Since that can mean `SquuezedOut` or `Failed` as well, we need to check the status in more detail.

PR #1145, opened 2023-09-27
feat: check deploy tx status

hal3e:
close: https://github.com/FuelLabs/fuels-rs/issues/1141

This PR adds a check of the tx status when deploying contracts.
I have refactored the code a bit and added a new error variant `SqueezedOutTransactionError`. I am not sure if this should be an error. What do you think @FuelLabs/sdk-rust ?

### Checklist

### DIFF
--- a/packages/fuels-core/src/types/errors.rs
+++ b/packages/fuels-core/src/types/errors.rs
@@ -35,8 +35,8 @@ pub enum Error {
         revert_id: u64,
         receipts: Vec<Receipt>,
     },
-    #[error("Transaction was squeezed out")]
-    SqueezedOutTransactionError,
+    #[error("Transaction was squeezed out. Reason: `{0}`")]
+    SqueezedOutTransactionError(String),
     #[error("Transaction build error: {0}")]
     TransactionBuildError(String),
 }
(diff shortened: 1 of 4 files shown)


digorithm (approved):
Yeah, I believe this new error should be bubbled up to the SDK user.

Br1ght0ne (dismissed):

packages/fuels-core/src/types/errors.rs
```diff
@@ -35,6 +35,8 @@ pub enum Error {
         revert_id: u64,
         receipts: Vec<Receipt>,
     },
+    #[error("Transaction was squeezed out")]
+    SqueezedOutTransactionError,
```
MujkicA, line 39: Can we propagate the reason for why the tx was squeezed out?

hal3e: great catch!

digorithm (approved):
:shipit:

iqdecay (approved):
I think it's ok to make it an SDK error, it is the same class of mistake as a revert for instance.

Salka1988 (approved):

MujkicA (approved):

Issue #1155, opened 2023-10-05
Predicate can not read manually added witness data


hal3e:
Predicates that try to read a manually added witness to a SDK Transaction will panic.

PR #1156, opened 2023-10-05
fix: predicate witness data access 

hal3e:
close:  https://github.com/FuelLabs/fuels-rs/issues/1155

Transaction builders do not precompute or estimate predicates anymore when `build()` is called. Instead, this is done before sending the tx to the client. In addition, `ScriptTransaction` and `CreateTransaction` have a new field: `is_using_predicates` to avoid running the estimation code if no predicates are used.

### Checklist


Salka1988 (approved):

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -312,6 +362,32 @@ macro_rules! impl_tx_wrapper {
...
+            ) -> Result<()> {
+                let gas_price = *self.tx.gas_price();
+                let gas_limit = *self.tx.gas_limit();
+                *self.tx.gas_price_mut() = 0;
+                *self.tx.gas_limit_mut() = consensus_parameters.max_gas_per_tx;
+
+                self.tx
+                    .estimate_predicates(consensus_parameters, gas_costs)?;
+                *self.tx.gas_price_mut() = gas_price;
+                *self.tx.gas_limit_mut() = gas_limit;
```
xgreenx, line 386:
```suggestion
                self.tx
                    .estimate_predicates(consensus_parameters, gas_costs)?;
```

Should have the same effect=)

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -65,7 +64,7 @@ pub trait TransactionBuilder: Send {
 
     fn build(self) -> Result<Self::TxType>;
     fn add_unresolved_signature(&mut self, owner: Bech32Address, secret_key: SecretKey);
-    fn fee_checked_from_tx(&self, params: &ConsensusParameters) -> Result<Option<TransactionFee>>;
+    fn fee_checked_from_tx(&self, networ_info: &NetworkInfo) -> Result<Option<TransactionFee>>;
```
iqdecay, line 67: typo

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -86,25 +85,20 @@ macro_rules! impl_tx_trait {
         impl TransactionBuilder for $ty {
             type TxType = $tx_ty;
             fn build(self) -> Result<$tx_ty> {
-                let uses_predicates = self.is_using_predicates();
-                let base_offset = if uses_predicates {
+                let is_using_predicates = self.is_using_predicates();
+                let base_offset = if is_using_predicates {
```
iqdecay, line 89:
```suggestion
                let base_offset = if self.is_using_predicates() {
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -86,25 +85,20 @@ macro_rules! impl_tx_trait {
...
-
-                if uses_predicates {
-                    estimate_predicates(&mut tx, &network_info)?;
-                };
+                let tx = self.resolve_fuel_tx(base_offset, num_witnesses)?;
 
-                Ok($tx_ty { tx })
+                Ok($tx_ty {
+                    tx,
+                    is_using_predicates,
```
iqdecay, line 100:
```suggestion
                    self.is_using_predicates(),
```

iqdecay (changes requested):
left some minor remarks

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -117,10 +111,23 @@ macro_rules! impl_tx_trait {
...
-                params: &ConsensusParameters,
+                network_info: &NetworkInfo,
             ) -> Result<Option<TransactionFee>> {
-                let tx = self.clone().build()?.tx;
-                Ok(TransactionFee::checked_from_tx(params, &tx))
+                let mut tx = self.clone().build()?;
+
+                tx.precompute(&network_info.chain_id())?;
+
+                if tx.is_using_predicates() {
```
digorithm, line 120: Let's be kind to our future selves (and other devs) and put a comment around here explaining _why_ we need to precompute and why we're estimating predicates the way that we're doing here. This kind of thing makes a lot of sense to us now that we have the context, but it would be an absolute nightmare to reason about without this context.

packages/fuels-core/src/types/wrappers/transaction.rs
digorithm: @xgreenx is this ☝️ expected behaviour?

digorithm (commented):
It's looking really good. Just left a minor ask for comments (+ the good suggestions made by everyone else).

packages/fuels-core/src/types/transaction_builders.rs
hal3e: Thanks for the suggestion. As the above is a sturct, I would have to write 
```rust
                Ok($tx_ty {
                    tx,
                    is_using_predicates: self.is_using_predicates(),
                })
```
IMHO the old version is nicer.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -117,10 +111,23 @@ macro_rules! impl_tx_trait {
...
             fn fee_checked_from_tx(
                 &self,
-                params: &ConsensusParameters,
+                network_info: &NetworkInfo,
             ) -> Result<Option<TransactionFee>> {
-                let tx = self.clone().build()?.tx;
-                Ok(TransactionFee::checked_from_tx(params, &tx))
+                let mut tx = self.clone().build()?;
+
+                tx.precompute(&network_info.chain_id())?;
```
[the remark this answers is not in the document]
hal3e: Added a new test: `tx_id_not_changed_after_adding_witnesses` that checks the tx_id after building, after adding witnesses and the tx_id returned from the provider.

MujkicA (approved):

packages/fuels-core/src/types/wrappers/transaction.rs
xgreenx: Yes, because you should have some coins at least to cover `min_fee` that already is known before predicate estimation. Estimating predicates also will increase the `min_fee` later=)

Br1ght0ne (approved):

PR #1172, opened 2023-10-24
test: add option to run against live node

iqdecay:
This PR adds a `cargo` feature called `test-against-live-node`, which enables running some test against a testnet node.

- This PR **does not** add the CI step, as it is still faulty with some weird errors of `status not found for transaction` (see https://github.com/FuelLabs/fuels-rs/actions/runs/6881505027/job/18718373777). These errors are being investigated as part of #1188.
- From comments on this PR, final design is this one:
If you want a test to be able to be run against a live node, remove the `Wallets` command from the `setup_test_program`, and in place use the `maybe_live_wallet` helper, as follows: 
```rust
    let [wallet]: [WalletUnlocked; 1] = maybe_live_wallet(1)
        .await?
        .try_into()
        .expect("Vec can be converted to an array");
    setup_program_test!(
        Abigen(Contract(
            name = "TestContract",
            project = "packages/fuels/tests/contracts/contract_test"
        )),
        Deploy(
            name = "contract_instance",
            contract = "TestContract",
            wallet = "wallet"
        ),
    );
```

Previously, this would have looked like this:
```rust
    setup_program_test!(
        Wallets("wallet"),
        Abigen(Contract(
            name = "TestContract",
            project = "packages/fuels/tests/contracts/contract_test"
        )),
        Deploy(
            name = "contract_instance",
            contract = "TestContract",
            wallet = "wallet"
        ),
    );
```

- This makes the code a little wordy but it's better than having to write the same tests twice. 
- I added a helper that generates up to 3 wallets connected to a testnet live node. The secret keys to these wallets are stored in github secrets.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,41 +256,6 @@ jobs:
         run: |
           ! cargo doc --document-private-items |& grep -A 6  "warning: unresolved link to"
 
-
-  run-live-node:
-
-    # is commented out for now so we can test the CI works properly
-    #    if: github.event_name == "push"
-    needs:
-      - setup-test-projects
-      - verify-rust-version
-      - get-workspace-members
-      - publish-crates-check
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
-        with:
-          ref: ${{ github.event.pull_request.head.sha }}
-      - uses: dtolnay/rust-toolchain@master
-        with:
-          toolchain: ${{ env.RUST_VERSION }}
-          components: clippy,rustfmt
-      - name: Install Fuel Core
-        run: |
-          curl -sSLf https://github.com/FuelLabs/fuel-core/releases/download/v${{ env.FUEL_CORE_VERSION }}/fuel-core-${{ env.FUEL_CORE_VERSION }}-x86_64-unknown-linux-gnu.tar.gz -L -o fuel-core.tar.gz
-          tar -xvf fuel-core.tar.gz
-          chmod +x fuel-core-${{ env.FUEL_CORE_VERSION }}-x86_64-unknown-linux-gnu/fuel-core
-          mv fuel-core-${{ env.FUEL_CORE_VERSION }}-x86_64-unknown-linux-gnu/fuel-core /usr/local/bin/fuel-core
-      - name: Download sway example artifacts
-        uses: actions/download-artifact@v3
-        with:
-          name: sway-examples-w-type-paths
-          path: packages/fuels/tests/
-      - name: Test against live node
-        run: cargo test --features test-against-live-node
-        continue-on-error: true
-
   publish:
     needs:
       - cargo-verifications
--- a/packages/fuels-macros/src/setup_program_test/code_gen.rs
+++ b/packages/fuels-macros/src/setup_program_test/code_gen.rs
@@ -76,6 +76,9 @@ fn wallet_initialization_code(
     };
 
     let wallet_names = extract_wallet_names(&command);
+    if wallet_names.is_empty() {
+        return Ok(Default::default());
+    }
     let num_wallets = wallet_names.len();
 
     Ok(quote! {


hal3e (changes requested):
Thanks for the devx improvement!

I have made some comments below on the current implementation. While I was testing it locally, I made a lot small changes so I added it to another branch. In the branch, `RunOnLiveNode` can be used without arguments and only one time etc. Branch: `hal3e/test-against-live-node`

Maybe it is worth to explore another implementation possibility. We could avoid adding a new command and instead rely on `env` variables. If the variable is set we can use beta wallets. So instead of checking if the command was used we could just use something like `option_env!` https://doc.rust-lang.org/stable/std/macro.option_env.html

Pseudo code:
in `code_gen`:
- check if `RUN_ON_LIVE_NODE` env variable is `Some("true")` -> use beta wallets
- else use random wallets

hal3e:
> @hal3e thanks a lot for your suggestion! I really like the idea of environment variables. However, we want to give the option to run only some relevant tests on the live nodes, first of all because it's slower, and potentially expensive. The env variables approach does not allow for this flexibility, as the pseudocode you wrote would apply each time we call `setup_program_test` which occurs >100 times in the codebase.
> 
> Thanks a lot for the branch, it's much better this way!

That is a good point!

segfault-magnet (commented):
Cleanly done!

I have one worry that I'd like to talk about:
`setup_program_test` besides being useful to us, is also marketed to users to help them bootstrap their tests quicker.

`RunOnLiveNode` isn't very flexible:
1. You cannot say where the node is live at, an old beta? Some custom user node/network?
2. It complicates the `Wallets` command. Previously it meant: "if a wallet is mentioned here I will setup the node and fund the wallet". Now it has a spin of "but if it is a live node, I will read ENV variables with some predetermined name format and connect to beta 4". 

Currently, the following is possible without changes to the code:

User connecting to a node of his choice, with wallets of his choice:
```rust
    let provider = Provider::connect("my.provider.com").await?;
    let wallet = WalletUnlocked::new_from_private_key("private_key".parse().unwrap(), Some(provider))

    setup_program_test!(
        Abigen(Contract(
            name = "TestContract",
            project = "packages/fuels/tests/contracts/contract_test"
        )),
        Deploy(
            name = "contract_instance",
            contract = "TestContract",
            wallet = "wallet"
        ),
    );
```

Us having a helper controlled by a feature:
```rust
async fn maybe_live_wallet() -> Result<WalletUnlocked> {
    if cfg!(feature = "live") {
        let provider = Provider::connect("my.provider.com").await?;
        Ok(WalletUnlocked::new_from_private_key(
            "private_key".parse().unwrap(),
            Some(provider),
        ))
    } else {
        launch_provider_and_get_wallet().await
    }
}

#[tokio::test]
async fn some_test() -> Result<()> {
    let wallet = maybe_live_wallet().await?;
    setup_program_test!(
        Abigen(Contract(
            name = "TestContract",
            project = "packages/fuels/tests/contracts/contract_test"
        )),
        Deploy(
            name = "contract_instance",
            contract = "TestContract",
            wallet = "wallet"
        ),
    );
```

This way we can keep complexity out of the macro while still being able to achieve our live tests with one extra line.

hal3e:
> Cleanly done!
> 
> I have one worry that I'd like to talk about: `setup_program_test` besides being useful to us, is also marketed to users to help them bootstrap their tests quicker.
> 
> `RunOnLiveNode` isn't very flexible:
> 
>     1. You cannot say where the node is live at, an old beta? Some custom user node/network?
> 
>     2. It complicates the `Wallets` command. Previously it meant: "if a wallet is mentioned here I will setup the node and fund the wallet". Now it has a spin of "but if it is a live node, I will read ENV variables with some predetermined name format and connect to beta 4".
> 
> 
> Currently, the following is possible without changes to the code:
> 
> User connecting to a node of his choice, with wallets of his choice:
> 
> ```rust
>     let provider = Provider::connect("my.provider.com").await?;
>     let wallet = WalletUnlocked::new_from_private_key("private_key".parse().unwrap(), Some(provider))
> 
>     setup_program_test!(
>         Abigen(Contract(
>             name = "TestContract",
>             project = "packages/fuels/tests/contracts/contract_test"
>         )),
>         Deploy(
>             name = "contract_instance",
>             contract = "TestContract",
>             wallet = "wallet"
>         ),
>     );
> ```
> 
> Us having a helper controlled by a feature:
> 
> ```rust
> async fn maybe_live_wallet() -> Result<WalletUnlocked> {
>     if cfg!(feature = "live") {
>         let provider = Provider::connect("my.provider.com").await?;
>         Ok(WalletUnlocked::new_from_private_key(
>             "private_key".parse().unwrap(),
>             Some(provider),
>         ))
>     } else {
>         launch_provider_and_get_wallet().await
>     }
> }
> 
> #[tokio::test]
> async fn some_test() -> Result<()> {
>     let wallet = maybe_live_wallet().await?;
>     setup_program_test!(
>         Abigen(Contract(
>             name = "TestContract",
>             project = "packages/fuels/tests/contracts/contract_test"
>         )),
>         Deploy(
>             name = "contract_instance",
>             contract = "TestContract",
>             wallet = "wallet"
>         ),
>     );
> ```
> 
> This way we can keep complexity out of the macro while still being able to achieve our live tests with one extra line.

I like this approach. As this is for our CI test only, a helper controlled with a flag seems to be the best solution.

PR #1122, opened 2023-09-05
feat: use `REG_CGAS` as default forwarded gas

hal3e:
Use `REG_CGAS` register if no `gas_forwarded` was not set for a specific call.

### Checklist
- [x] I have added necessary labels.
- [x] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [x] I have requested a review from the relevant team or maintainers.

PR #1173, opened 2023-10-26
fix: encode contract arguments with `gas_forwarded`

hal3e:
In [1122](https://github.com/FuelLabs/fuels-rs/pull/1122) I introduced a bug where contract arguments are not put at the right location in the script data. We did not catch this because we did not have a test that has arguments and uses `gas_forwarded`. I have fixed the issue and added a test to cover this use-case.

### Checklist


digorithm (approved):
Thanks for the quick response; we dropped the ball by not having a specific test for this, but no harm. We should get this released asap!

Br1ght0ne (approved):

segfault-magnet (approved):

PR #1177, opened 2023-10-30
fix: generate salt inside macro call to avoid contract id collision

iqdecay:
This PR fixes an issue where two contracts would have the same contract id because the salt was not generated properly.

### DIFF
--- a/packages/fuels-macros/src/setup_program_test/code_gen.rs
+++ b/packages/fuels-macros/src/setup_program_test/code_gen.rs
@@ -119,6 +119,9 @@ fn contract_deploying_code(
             let bin_path = project.bin_path();
 
             quote! {
+                // Generate random salt for contract deployment.
+                // These lines must be inside the `quote!` macro, otherwise the salt remains
+                // identical between macro compilation, causing contract id collision.
                 let mut rng = <::rand::rngs::StdRng as ::rand::SeedableRng>::from_entropy();
                 let salt: [u8; 32] = <::rand::rngs::StdRng as ::rand::Rng>::gen(&mut rng);
 


digorithm (approved):
Excellent catch. I'm approving this, but we should have a test for this: a test that deploys the same contract twice and it checks for the different generated salts. This should prevent us from having this same incident in the future.

packages/fuels-macros/src/setup_program_test/code_gen.rs
```diff
@@ -31,6 +30,8 @@ pub(crate) fn generate_setup_program_test_code(
     let script_code = script_loading_code(&load_scripts, &project_lookup);
 
     Ok(quote! {
+        // Necessary for salt generation
+        use rand::{rngs::StdRng, Rng, SeedableRng};
```
hal3e, line 34: When reading the code it is not obvious why this line is here. The comment helps but still you could ask what `salt` generation. I propose we instead move this line closer to the actual usage.
```suggestion
```

packages/fuels-macros/src/setup_program_test/code_gen.rs
```diff
@@ -123,8 +121,13 @@ fn contract_deploying_code(
             let bin_path = project.bin_path();
 
             quote! {
+                // Generate random salt for contract deployment.
+                // These lines must be inside the `quote!` macro, otherwise the salt remains
+                // identical between macro compilation, causing contract id collision.
+                let mut rng = StdRng::from_entropy();
+                let salt: [u8; 32] = rng.gen();
```
hal3e, line 128: I did not find a nice place to add `use rand::{rngs::StdRng, Rng, SeedableRng};` so instead I propose we use fully qualified syntax.
```suggestion
                // Generate random salt for contract deployment.
                let mut rng = <rand::rngs::StdRng as rand::SeedableRng>::from_entropy();
                let salt: [u8; 32] = <rand::rngs::StdRng as rand::Rng>::gen(&mut rng);

```

hal3e (approved):

Salka1988 (approved):

segfault-magnet (approved):

PR #1185, opened 2023-11-02
fix: enum discriminant overflow

Salka1988:
### Checklist


packages/fuels-macros/src/derive/utils.rs
```diff
@@ -75,15 +75,12 @@ pub(crate) fn extract_variants(
...
-                let current_discriminant = discriminant;
-                discriminant += 1;
-
                 Ok(ExtractedVariant::Normal {
                     info: VariantInfo {
                         name: variant.ident,
                         is_unit,
                     },
-                    discriminant: current_discriminant,
+                    discriminant: discriminant as u8,
```
hal3e, line 83: Even though, sway uses `u8` for the discriminant, I would still guard against this conversion with an `try_into`. Someone could manually make an `Enum` or a json abi with more than 256 variants.

```rust
  let discriminant = discriminant.try_into().map_err(|_| {
      Error::new_spanned(&variant.ident, "Enums can not have more than 256 variants")
  })?;

  Ok(ExtractedVariant::Normal {
      info: VariantInfo {
          name: variant.ident,
          is_unit,
      },
      discriminant,
  })
```

We could also write an UI test for this.

hal3e (changes requested):
Nice!

packages/fuels-macros/src/derive/utils.rs
Salka1988: You're right, I'll do it.

MujkicA (approved):

hal3e (approved):

digorithm (approved):
Good job @Salka1988

PR #1186, opened 2023-11-02
refactor: use a single type for transaction status

### LINKED ISSUE #1179: refactor: use a single type for transaction status
Currently, there are two co-existing, incompatible types for a transaction status:
- `TxStatus` living inside `packages/fuels-core/src/types/tx_status.rs`
- `TransactionStatus` living inside `packages/fuels-core/src/types/wrappers/transaction_response.rs`.

Those two types should be unified.

iqdecay:
This PR closes #1179 by removing the `TransactionStatus` type that was used in only one place. 

Further work is needed because the current `TxStatus` holds receipts which the client `TransactionStatus` does not,
and doesn't hold some information the client's type does (block id, time, etc).


MujkicA (approved):
LGTM

Maybe we want to deprecate receipts in the status at some point since it's separate on the node client side as well

digorithm (approved):

hal3e (approved):

Br1ght0ne (approved):

Issue #1128, opened 2023-09-11
Use `client.submit_and_await_commit` in `Provider`


segfault-magnet:
Currently, we have our own `send_transaction_and_await` where we call `submit` and then `await_transaction_commit`. 

We can just delegate to `submit_and_await_commit` instead. 

Some refactoring is needed as we do some checks before submitting a tx.

PR #1187, opened 2023-11-05
refactor!: use `submit_and_await_commit` API

### LINKED ISSUE #1180: refactor: use the `submit_and_await_commit` API
While exploring #1135, we realised that there is an issue where we don't call the `submit_and_await_commit`  `fuel-core` API, but rather first submit, then await for commit (see `provider.rs` l189). 
This can create issues where the tx id returned is not synchronized properly. 
This needs to be refactored, but first we need a unified `TxStatus` type as raised in https://github.com/FuelLabs/fuels-rs/issues/1179.

iqdecay:
Closes #1180,  #1128.
This PR uses the correct API for sending a transaction and ensuring that we wait for the commit of the transaction before returning.
This is needed because otherwise there might be a disconnect between submitting the transaction and awaiting for its commit, if it is done using the two separate APIs as before. 
Further work will be needed as part of #1188

BREAKING CHANGE: 
- `send_transaction_and_await_commit` function now returns a `TxStatus` instead of `TxId`.

### DIFF
--- a/packages/fuels-accounts/src/account.rs
+++ b/packages/fuels-accounts/src/account.rs
@@ -213,7 +213,10 @@ pub trait Account: ViewOnlyAccount {
             ScriptTransactionBuilder::prepare_transfer(inputs, outputs, tx_policies);
 
         self.add_witnessses(&mut tx_builder);
-        self.adjust_for_fee(&mut tx_builder, amount).await?;
+
+        let used_base_amount = if asset_id == AssetId::BASE { amount } else { 0 };
+        self.adjust_for_fee(&mut tx_builder, used_base_amount)
+            .await?;
 
         let tx = tx_builder.build(provider).await?;
         let tx_id = tx.id(provider.chain_id());


MujkicA (commented):
Could you elaborate a bit more on what problem inspired this change?
Why are we deleting `await_transaction_commit`? Then we are no longer able to call `submit` and await at a later point in the code.

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -63,6 +63,11 @@ impl RetryableClient {
             .await
     }
 
+    pub async fn submit_and_await_commit(&self, tx: &Transaction) -> io::Result<TransactionStatus> {
```
segfault-magnet, line 66: Maybe we should turn this into our `TxStatus`? If it is to be used by us further up the call tree?

Br1ght0ne (dismissed):

packages/fuels-accounts/src/provider.rs
```diff
@@ -197,24 +197,29 @@ impl Provider {
     }
 
     /// Sends a transaction to the underlying Provider's client.
-    pub async fn send_transaction_and_await_commit<T: Transaction>(&self, tx: T) -> Result<TxId> {
-        let tx_id = self.send_transaction(tx.clone()).await?;
-        let _status = self.client.await_transaction_commit(&tx_id).await?;
-
+    pub async fn send_transaction_and_await_commit<T: Transaction>(
```
hal3e, line 200: IMO this function should return a `TxStatus`. We have the status here and the user can calculate the `TxId` if he needs it. Also if we make this change in our `call_or_simulate` we could use only one call instead of using two:
```rust
let tx_id = self.provider.send_transaction_and_await_commit(tx).await?;
self.provider.tx_status(&tx_id).await?
```
This I how I recommend we write the fn:
```rust
pub async fn send_transaction_and_await_commit<T: Transaction>(
    &self,
    mut tx: T,
) -> Result<TxStatus> {
    self.prepare_transaction_for_sending(&mut tx).await?;
    let tx_status = self
        .client
        .submit_and_await_commit(&tx.clone().into())
        .await?
        .into();

    if matches!(
        tx_status,
        TxStatus::SqueezedOut { .. } | TxStatus::Revert { .. }
    ) {
        self.cache.lock().await.remove_items(tx.used_coins())
    }

    Ok(tx_status)
}
```
There were also some small changes in other parts of the codebase and I have fixed them on this branch: `hal3e/refactor-await-commit-api`

hal3e (changes requested):

packages/fuels-accounts/src/provider.rs
hal3e: Thanks for changing it. This is now a breaking change so please do not forget to add this to the PR description.

hal3e (approved):
Great works! Thanks!

Br1ght0ne (approved):

segfault-magnet (approved):

Salka1988 (approved):

Issue #1189, opened 2023-11-07
Use dynamic supported `fuel-core` version in version check


Br1ght0ne:
#1166 should be extended by using actual `fuel-core` version from `Cargo.toml`.

PR #1190, opened 2023-11-07
Use dynamic supported fuel-core version

Br1ght0ne:
Close https://github.com/FuelLabs/fuels-rs/issues/1189. Waiting on a new `fuel-core` release incorporating https://github.com/FuelLabs/fuel-core/pull/1473.

### Checklist

### DIFF
--- a/packages/fuels-accounts/Cargo.toml
+++ b/packages/fuels-accounts/Cargo.toml
@@ -14,6 +14,7 @@ async-trait = { workspace = true, default-features = false }
 chrono = { workspace = true }
 elliptic-curve = { workspace = true, default-features = false }
 eth-keystore =  { workspace = true, optional = true }
+fuel-core = { workspace = true }
 fuel-crypto = { workspace = true, features = ["random"] }
 fuel-tx = { workspace = true }
 fuel-types = { workspace = true, features = ["random"] }
--- a/packages/fuels-accounts/src/provider/supported_versions.rs
+++ b/packages/fuels-accounts/src/provider/supported_versions.rs
@@ -1,7 +1,9 @@
 use semver::Version;
 
 fn get_supported_fuel_core_version() -> Version {
-    "0.20.6".parse().unwrap()
+    fuel_core::VERSION
+        .parse()
+        .expect("failed to parse fuel-core version")
 }
 
 #[derive(Debug, PartialEq, Eq)]


packages/fuels-accounts/Cargo.toml
```diff
@@ -36,4 +37,4 @@ tokio = { workspace = true, features = ["test-util"]}
 [features]
 default = ["std"]
 coin-cache = ["tokio?/time"]
-std = ["fuels-core/std", "dep:tokio", "fuel-core-client/default", "dep:eth-keystore"]
+std = ["fuels-core/std", "dep:tokio", "fuel-core-client/default", "dep:fuel-core", "dep:eth-keystore"]
```
segfault-magnet, line 40: It is worrysome to have `fuel-core` as a `std` dep.

`std` is enabled by default but `fuel-core-lib` (feature from `fuels`) is disabled by default.

Since `std` from `fuels` will enable `std` from `fuels-accounts` and that will enable `fuel-core`, we then compile `fuel-core` whether we use `fuel-core-lib` or not. That might not be ideal since the reason behind the feature was to cut down on compile times for demo-like environments.

Br1ght0ne: Thanks for the suggestion - I'll try forwarding the `fuel-core-lib` feature to `fuels-accounts` then

packages/fuels-accounts/src/provider.rs
```diff
@@ -2,6 +2,7 @@ use std::{collections::HashMap, fmt::Debug, io, net::SocketAddr};
 
 mod retry_util;
 mod retryable_client;
+#[cfg(feature = "fuel-core-lib")]
```
hal3e, line 5: If I understand correctly, the supported version will only be checked when `fuel-core-lib`is used. If `fuel-core-lib` is used we compile `fuel-core` ourselfs which means that we will use the correct versione from our `Cargo.toml`.

Would it be possible to hardcore the supported version to a file during the publish step of our CI ?

hal3e (changes requested):

packages/fuels-accounts/src/provider.rs
digorithm: > If I understand correctly, the supported version will only be checked when fuel-core-libis used. If fuel-core-lib is used we compile fuel-core ourselfs which means that we will use the correct versione from our Cargo.toml.

Yeah, this type of check seems less critical than needing to do something similar for when you're _not_ using the `fuel-core-lib`. e.g., when you're talking to an external (external to the SDK code) node.

What are the current blockers/difficulties in doing something like that?

> Would it be possible to hardcore the supported version to a file during the publish step of our CI ?

How would that look and what are the benefits of doing it? At first this sounds like having the `fuel-core` lib version pinned in our `Cargo.toml` 🤔 but I'm probably missing something.

digorithm (commented):
I know this hasn't been your focus lately, but is there any updates on this?

segfault-magnet (changes requested):
Some notes:
* Using a `build.rs` is problematic because this will cause our users to always depend on `fuel-core` since we're packaging a `build.rs` dependent on `fuel-core` that will run when users build their software. One of the motivations for the `fuel-core-lib` is to lower compile times for live demos. 
* We need a CI check that when the fuel-core version in the workspace cargo toml is updated the corresponding version file matches the version in the toml.
* If we're to generate version files we might as well generate a rust file that has a `pub const version: ... = VERSION_HERE`. Better yet if there is a `const` version structure so that we don't have to do any parsing in runtime.

Maybe it is enough to have that `version.rs` file with a unpublished binary that checks if the versions match. Something like the docs checker binary currently run in the CI. 

Also keep in mind that sometimes our `fuel-core` versions are git checkouts, when there are circularly dependent breaking changes between the compiler, core and us. The CI should allow for this (unless we're publishing).

segfault-magnet (dismissed):

hal3e (approved):
Nicely done!

The only thing I was not sure about is using `include_str!` instead of `include`. With `include` we need to write valid rust code. Let's see what others say

Cargo.toml
```diff
@@ -45,6 +46,8 @@ async-trait = { version = "0.1.74", default-features = false }
 bech32 = "0.9.1"
 bytes = { version = "1.5.0", default-features = false }
 chrono = "0.4.31"
+clap = { version = "4.5.3", features = ["derive"] }
+color-eyre = "0.6.2"
```
MujkicA, line 50: Can we confine these deps to the local Cargo.toml since they are only used by the new script?

MujkicA (approved):

scripts/fuel-core-version/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+license = { workspace = true }
+publish = false
+repository = { workspace = true }
+rust-version = { workspace = true }
+
+[dependencies]
+clap = { version = "4.5.3", features = ["derive"] }
+color-eyre = "0.6.2"
+semver = { workspace = true }
+versions-replacer = { path = "../versions-replacer" }
```
segfault-magnet, line 16: can we use `workspace = true` ? Don't use path deps elsewhere except in the workspace toml.

segfault-magnet (approved):

Salka1988 (approved):

segfault-magnet (approved):

digorithm (approved):

hal3e (approved):
LGTM

PR #1200, opened 2023-11-11
feat!: `beta-5` support

hal3e:
This PR adds sdk support for `beta-5`.

- TODO: update docs and describe breaking changes

BREAKING CHANGE:
- `TxParameters` are replaced with `TxPolicies`
- `GasPrice` and `Maturity` fields are optional 
- `TxPolicies` introduced new fields: 
  - `WitnessLimit` - allows the limitation of the maximum size of witnesses in bytes.
  - `MaxFee` - allows the upper bound for the maximum fee that users agree to pay for the transaction.
- The `ScriptGasLimit` only limits the script execution. Previously, the `ScriptGasLimit` also limited the predicate execution time, but it is not valid anymore. So, it is not possible to use the `ScriptGasLimit` for transaction cost limitations. A new `MaxFee` policy is a way to do that. The `GasLimit` field was removed from the `Create` transaction because it only relates to the script execution(which the `Create` transaction doesn't have).
- The new `WhitnessLimit` also impacts the `max_gas` and `max_fee` calculation along with the `ScriptGasLimit`(in the case of `Create` transaction only `WitnessLimit` affects the `max_gas` and `max_fee`).
- The minimal gas also charges the user for transaction ID calculation.
- Each transaction requires setting the `GasPrice` policy.
- Previously, `GasLimit` should be less than the `MAX_GAS_PER_TX` constant. After removing this field from the `Create` transaction, it is impossible to require it. Instead, it requires that `max_gas <= MAX_GAS_PER_TX` for any transaction. Consequently, any `Script` transaction that uses `MAX_GAS_PER_TX` as a `GasLimit` will always fail because of a new rule. Setting the estimated gas usage instead solves the problem.
- If the `max_fee > policies.max_fee`, then transaction will be rejected.
- If the `witnessses_size > policies.witness_limit`, then transaction will be rejected.
- `get_message_proof` not uses `Nonce` instead of the message_id
- predicates do not use `ChainId` for address calculation
- `manual_blocks_enabled` is replaced with `debug` in local chain config
- `fee_checked_from_tx` uses `FeeParameters`
- `fuel_tx::ConsensusParameters` were refactored - this also affects us
- when building a transaction_builder the `BuildableTransacion` trait needs to be in scope
- `utxo_validation` and `manual_blocks` are enabled by default for test providers
- node config does not have `local_node` anymore use `default` - `let node_config = Config::default();`

Thanks @MujkicA for updating the documentation.

### Checklist


packages/fuels-core/src/utils/constants.rs
```diff
@@ -11,3 +11,8 @@ pub const BASE_ASSET_ID: AssetId = AssetId::BASE;
 // ANCHOR_END: default_call_parameters
 
 pub const DEFAULT_GAS_ESTIMATION_TOLERANCE: f64 = 0.2;
+
+// Supports 10 signatures
+pub const DEFAULT_SCRIPT_WITNESS_LIMIT: u64 = 720;
```
MujkicA, line 16: Is there a constant on the fuel-core size that defines the size of the signature?
It would be nice to have something like 
```rust
pub const DEFAULT_SCRIPT_WITNESS_LIMIT: u64 = 10 * SIG_SIZE;
```

packages/fuels-test-helpers/src/node_types.rs
```diff
@@ -70,10 +72,10 @@ impl Config {
     pub fn local_node() -> Self {
```
MujkicA, line 72: `local_node()` and `default()` lead to the same config, we can deduplicate a little

packages/fuels/Forc.toml
```diff
@@ -101,3 +101,6 @@ members = [
   'tests/types/scripts/script_u256',
   'tests/types/scripts/script_vectors',
 ]
+
+[patch.'https://github.com/fuellabs/sway']
+std = { git = "https://github.com/fuellabs/sway", branch = "hal3e/fuel-core-master2" }
```
MujkicA, line 106: Do we want to track updating this and the other pathes this with an issue?

MujkicA (commented):
Just some minor remarks for now, feel free to discard them

packages/fuels/Forc.toml
hal3e: https://github.com/FuelLabs/fuels-rs/issues/1201

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -76,12 +106,31 @@ pub enum TransactionType {
     Create(CreateTransaction),
 }
 
-pub trait Transaction: Into<FuelTransaction> + Clone {
+pub trait UpdateGas {
+    fn with(self, gas_price: u64, gas_limit: u64) -> Self;
```
MujkicA, line 110: Maybe we can go for something more descriptive. `with_gas`?

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -451,7 +480,9 @@ macro_rules! impl_tx_wrapper {
                         _ => {
                             // not a contract, it's safe to unwrap
                             let owner = extract_owner_or_recipient(input).unwrap();
-                            let asset_id = input.asset_id().unwrap().to_owned();
+                            // Can we just use `BASE_ASSET_ID` here ? cc @AhmedM
```
MujkicA, line 483: From looking at the fuel-core impl of `asset_id` it seems that the arguments is used the return the asset id of message coins which according to protocol is always the baset asset. 
So I would say yes, but I wonder why it wasn't hardcoded on the core side. @xgreenx should know

xgreenx: In the future, we plan not to support different base asset ids.

Cargo.toml
```diff
@@ -94,3 +96,18 @@ fuels-core = { version = "0.50.1", path = "./packages/fuels-core", default-featu
...
+
+[patch.crates-io]
+fuel-asm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-crypto = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-merkle = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-storage = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-tx = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-types = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-vm = { git = "https://github.com/FuelLabs/fuel-vm/", branch = "master" }
+fuel-core = { git = "https://github.com/FuelLabs/fuel-core", branch = "feature/transaction-policies" }
```
digorithm, line 108: It seems like this `fuel-core` branch has been merged; maybe we can point it to master for now?

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,48 +245,139 @@ impl ScriptTransactionBuilder {
...
+            .resolve_fuel_tx_provider(base_offset, num_witnesses, provider)
+            .await?;
+
+        Ok(ScriptTransaction {
+            tx,
+            is_using_predicates,
+        })
+    }
+
+    fn create_dry_run_witnesses(&self, num_witnesses: u8) -> Vec<Witness> {
```
digorithm, line 281: A comment here explaining _why_ this is done the way it is would be nice for future-us.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,48 +245,139 @@ impl ScriptTransactionBuilder {
...
+        let unresolved_witnesses_len = self.unresolved_signatures.addr_idx_offset_map.len();
+        repeat(Default::default())
+            // Add one in case there is no witnesses at all
+            .take(num_witnesses as usize + unresolved_witnesses_len + 1)
+            .collect()
+    }
+
+    async fn set_script_gas_limit_to_gas_used(
```
digorithm, line 289: This whole fn could use some explanatory comments; It's a bit confusing without proper context.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,48 +245,139 @@ impl ScriptTransactionBuilder {
...
+        provider: &impl DryRunner,
+        network_info: &NetworkInfo,
+        tolerance: f32,
+    ) -> Result<()> {
+        let consensus_params = &network_info.consensus_parameters;
+        // Add `1` because of rounding
+        let max_gas = tx.max_gas(consensus_params.gas_costs(), consensus_params.fee_params()) + 1;
+
+        // If `gas_limit` was not set use `max_gas_per_tx` but subtract the tx's base maximum cost
+        // TODO: @xgreenx why do I need to / 2
```
digorithm, line 300: @xgreenx question for you!

digorithm (commented):
Sending some comments as I go through this PR; I'm almost halfway through it.

packages/fuels-core/src/types/transaction_builders.rs
xgreenx: I guess because you first need to add an input and only set the gas limit after=)

Cargo.toml
hal3e: We point to the `rc` now.

examples/contracts/src/lib.rs
```diff
@@ -273,47 +273,46 @@ mod tests {
...
 
         println!("Contract deployed @ {contract_id}");
-        // ANCHOR: tx_parameters
+        // ANCHOR: tx_policies
         let contract_methods = MyContract::new(contract_id.clone(), wallet.clone()).methods();
 
-        let my_tx_parameters = TxParameters::default()
+        let tx_policies = TxPolicies::default()
             .with_gas_price(1)
-            .with_gas_limit(1_000_000)
```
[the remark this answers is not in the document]
hal3e: Well `gas_limit` is not the limit for the whole tx anymore. IHMO it makes sense to rename it so that the user immediately knows that this is something new.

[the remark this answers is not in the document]
digorithm: I get it. However:

> Well gas_limit is not the limit for the whole tx anymore

Coming from other blockchain ecosystems/platforms, this sentence alone is confusing. I'm trying to put myself in the user's shoes here. When I set the gas limit for a transaction, that's it. Having to think about multiple gas limits is confusing. I'm still going through the changes so I don't know this: will the user ever be required to set _both_ gas limits in a single transaction?

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,48 +245,139 @@ impl ScriptTransactionBuilder {
...
+            self.witness_limit.or(Some(DEFAULT_SCRIPT_WITNESS_LIMIT)),
+        );
+
+        let scritp_len = self.script.len();
+        let dry_run_witnesses = self.create_dry_run_witnesses(num_witnesses);
         let mut tx = FuelTransaction::script(
-            gas_price,
-            gas_limit,
-            self.maturity.into(),
+            0, // use temporarily
```
digorithm, line 341: If we're temporarily keeping this `0` here, we should create an issue and add a `TODO` here so we don't forget about it.

examples/contracts/src/lib.rs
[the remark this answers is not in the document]
hal3e: The sdk is just reflecting the changes from `fuel-core` and `fuel-vm`. `script_gas_limit` now limits only script gas. and `max_fee` limits the whole `tx`. The user might set both limits in one `tx`. cc @xgreenx

packages/fuels-core/src/types/transaction_builders.rs
hal3e: I will improve the comments. 0 is used for the `dry_run`

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -23,27 +27,34 @@ use crate::{
...
     pub min_gas_price: u64,
-    pub gas_costs: GasCosts,
 }
 
 impl NetworkInfo {
     pub fn new(node_info: NodeInfo, chain_info: ChainInfo) -> Self {
         Self {
-            max_gas_per_tx: chain_info.consensus_parameters.max_gas_per_tx,
-            consensus_parameters: chain_info.consensus_parameters.into(),
+            max_gas_per_tx: chain_info.consensus_parameters.tx_params().max_gas_per_tx,
```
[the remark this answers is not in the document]
digorithm: Seems like it's still `tx_params` in `fuel-tx`: https://github.com/FuelLabs/fuel-vm/blob/master/fuel-tx/src/transaction/consensus_parameters.rs#L93-L96

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -186,12 +161,26 @@ macro_rules! impl_tx_trait {
...
 
-            fn consensus_parameters(&self) -> ConsensusParameters {
-                self.network_info.consensus_parameters
+            fn consensus_parameters(&self) -> &ConsensusParameters {
+                &self.network_info.consensus_parameters
             }
         }
 
         impl $ty {
+            fn generate_common_fuel_policies(&self) -> Policies {
```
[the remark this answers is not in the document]
hal3e: Shared. I will change the fn name to make it more clear :)

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -421,22 +567,45 @@ impl CreateTransactionBuilder {
         }
     }
 
-    fn resolve_fuel_tx(self, base_offset: usize, num_witnesses: u8) -> Result<Create> {
-        let num_of_storage_slots = self.storage_slots.len();
+    async fn build_with_provider(self, _: &impl DryRunner) -> Result<CreateTransaction> {
+        self.build()
+    }
```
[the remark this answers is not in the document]
digorithm: Yeah, I agree it isn't very clear. We should write a comment explaining it if there's a proper reason behind it.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -421,22 +574,45 @@ impl CreateTransactionBuilder {
...
+
+    fn resolve_fuel_tx(self, mut base_offset: usize, num_witnesses: u8) -> Result<Create> {
+        let mut policies = self.generate_common_fuel_policies();
+        policies.set(
+            PolicyType::WitnessLimit,
+            self.witness_limit.or(Some(DEFAULT_CREATE_WITNESS_LIMIT)),
+        );
+
+        let storage_slots_offset = self.storage_slots.len() * StorageSlot::SLOT_SIZE;
+        base_offset += storage_slots_offset + policies.size_dynamic();
```
digorithm, line 606: This bit here could also use some _why_ comments.

packages/fuels-core/src/types/transaction_builders.rs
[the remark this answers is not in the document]
hal3e: I changed te code a bit. Should be clearer now. The main reason for the confusion is that the `CreateTransaction` does not need the `provider` and `ScriptTransaction` does. `CreateTransaction` does not have `script_gas_limit`

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -76,12 +106,27 @@ pub enum TransactionType {
     Create(CreateTransaction),
 }
 
-pub trait Transaction: Into<FuelTransaction> + Clone {
+pub trait EstimablePredicates {
+    /// If a transactions contains predicates, we have to estimate them
+    /// before sending the transaction to the node. The estimation will check
+    /// all predicates and set the `predicate_gas_used` to the actual consumed gas.
```
digorithm, line 112:
```suggestion
    /// If a transaction contains predicates, we have to estimate them
    /// before sending the transaction to the node. The estimation will check
    /// all predicates and set the `predicate_gas_used` to the actual consumed gas. 
```

Excellent comment btw!

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -451,7 +467,8 @@ macro_rules! impl_tx_wrapper {
                         _ => {
                             // not a contract, it's safe to unwrap
                             let owner = extract_owner_or_recipient(input).unwrap();
-                            let asset_id = input.asset_id().unwrap().to_owned();
+                            let asset_id = input.asset_id(&BASE_ASSET_ID).unwrap().to_owned();
```
digorithm, line 470: I guess we missed this `unwrap` here and on the line above. Maybe a good opportunity to properly handle these two errors?

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -1,40 +1,52 @@
...
-use crate::types::{bech32::Bech32Address, Result};
+use crate::{
+    constants::BASE_ASSET_ID,
+    types::{bech32::Bech32Address, errors::error, Result},
+};
 
 #[derive(Default, Debug, Copy, Clone)]
-pub struct TxParameters {
+pub struct TxPolicies {
     gas_price: Option<u64>,
```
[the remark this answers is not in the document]
hal3e: Good catch. This is a mistake.

packages/fuels-test-helpers/src/fuel_bin_service.rs
```diff
@@ -93,7 +93,7 @@ impl ExtendedConfig {
             [
                 (self.config.vm_backtrace, "--vm-backtrace"),
                 (self.config.utxo_validation, "--utxo-validation"),
-                (self.config.manual_blocks_enabled, "--manual_blocks_enabled"),
```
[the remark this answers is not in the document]
hal3e: It is on by default with the `debug` flag.

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -483,6 +508,51 @@ impl CreateTransaction {
     }
 }
 
+impl GasValidation for CreateTransaction {
+    fn validate_gas(&self, min_gas_price: u64, _gas_used: u64) -> Result<()> {
```
digorithm, line 512: Good opportunity to mention that `CreateTransaction` doesn't consume gas:

```suggestion
    // We're not using `gas_used` in this implementation because `CreateTransaction` consumes no gas.
    fn validate_gas(&self, min_gas_price: u64, _gas_used: u64) -> Result<()> {
```

packages/fuels-core/src/types/transaction_builders.rs
hal3e: Nothing to comment. That is just how the `vm` expects the offset. base_offset + storage_slot_offset + whatever was set with policies.

packages/fuels/tests/providers.rs
```diff
@@ -308,46 +286,19 @@ async fn contract_deployment_respects_maturity() -> Result<()> {
...
+    assert!(matches!(err, Error::ValidationError(s) if s == "TransactionMaturity"));
 
     provider.produce_blocks(1, None).await?;
     deploy_w_maturity(1u32)?
         .await
-        .expect("Should be able to deploy now since maturity (1) is <= than the block height (1)");
-    Ok(())
-}
-
-#[tokio::test]
```
digorithm, line 328: Why are we deleting this test?

packages/fuels/tests/scripts.rs
```diff
@@ -192,40 +199,6 @@ async fn test_script_call_with_non_default_max_input() -> Result<()> {
     Ok(())
 }
 
-#[tokio::test]
```
digorithm, line 195: Why are we deleting this test?

digorithm (commented):
First pass done! It looks great so far. Not approving it yet because there are some open discussions yet to be resolved

packages/fuels/tests/scripts.rs
hal3e: This tested that the script tx signature is ok. `utxo_validation` is turned on by default in tests,  which means this test doesn't test anything new. We have this script call several times in our tests.

packages/fuels-core/src/types/wrappers/transaction.rs
hal3e: they are safe to unwrap as we exclude `Contract`

packages/fuels/tests/providers.rs
hal3e: Because this is not the case anymore. the `tx.script_gas_limit` should not match `max_gas_per_tx`

hal3e:
>     * I don't understand why the `build` function now needs the `provider` argument, even though said argument is not used.
> 
>     * I'm not sure I understand the logic behind renaming everything. To me, we are exposing implementation details to the user. As a user, using `TxPolicies` instead of `TxParameters` feels unusual at best. Also I feel like our responsability as the SDK is to shield the user from all the little internal changes the `fuel-core` or `fuel-vm` repos go through. So I'm not sure why we are replicating all these breaking changes in our API.

- To keep the interface consistent between `ScriptTransaction` and `CreateTransaction` we need to pass the `provider` even though `CreateTransaction` does not need it.

- It was decided to switch to `TxPolicies`. Guarding the user from this is not possible IMHO.

.github/workflows/ci.yml
```diff
@@ -61,7 +62,7 @@ jobs:
         working-directory: packages/fuels
 
       - name: Build Sway test projects
-        run: forc build --terse --error-on-warnings
+        run: forc build --terse
```
hal3e, line 65: revert this when `sway` fixes warnings in the `std`(this will probably happen before merging this PR)
```suggestion
        run: forc build --terse --error-on-warnings
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -55,20 +66,48 @@ impl NetworkInfo {
...
     secret_keys: Vec<SecretKey>,
 }
 
-pub trait TransactionBuilder: Send {
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait BuildableTransaction {
+    type TxType: Transaction;
+    /// Build a `Transaction` from the `TransactionBuilder`. `DryRunner` is
+    /// used to return the actual `gas_used` which is set as the `script_gas_limit`.
+    /// `CreateTransaction`s do not have `gas_limit` so the `DryRunner` is not used
```
segfault-magnet, line 85: The part of the comment that references the `CreateTransaction` should be moved there.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,48 +288,146 @@ impl ScriptTransactionBuilder {
...
+        num_witnesses: u8,
+        provider: &impl DryRunner,
+    ) -> Result<Script> {
+        let mut policies = self.generate_shared_fuel_policies();
+        policies.set(
+            PolicyType::WitnessLimit,
+            self.witness_limit.or(Some(DEFAULT_SCRIPT_WITNESS_LIMIT)),
+        );
+
+        let script_len = self.script.len();
```
segfault-magnet, line 392: Maybe `let no_code_present = self.script.is_empty()`

would eliminate the need for the comment below.

packages/fuels/tests/contracts/revert_transaction_error/src/main.sw
```diff
@@ -1,15 +1,15 @@
...
+    fn make_transaction_fail(fail: bool) -> u64;
 }
 
 impl MyContract for Contract {
-    fn make_transaction_fail(input: u64) -> u64 {
-        // Work around DCE and revert
-        if true {
-            revert(input);
+    fn make_transaction_fail(fail: bool) -> u64 {
+        if fail {
```
segfault-magnet, line 9: The previous comment mentioned working around the `DCE`. Does this change still work around the `DCE`, whatever that is?

packages/fuels/tests/providers.rs
```diff
@@ -308,46 +286,19 @@ async fn contract_deployment_respects_maturity() -> Result<()> {
...
 
-    let err = deploy_w_maturity(1u32)?.await.expect_err("Should not have been able to deploy the contract since the block height (0) is less than the requested maturity (1)");
-    assert!(matches!(
-        err,
-        Error::ValidationError(fuel_tx::CheckError::TransactionMaturity)
-    ));
+    let err = deploy_w_maturity(1u32)?.await.expect_err(
+        "Should not deploy contract since block height (0) is less than the requested maturity (1)",
+    );
+    assert!(matches!(err, Error::ValidationError(s) if s == "TransactionMaturity"));
```
segfault-magnet, line 296: Although it is easier to write, we should probably try and avoid `assert!(matches!` assertions -- they don't make for great diagnostics when the test fails -- you just get that an assertion had failed. If we had done something like this:
```rust
    let Error::ValidationError(s) = err else {
        panic!("Expected a ValidationError, got: {err}");
    };
    assert_eq!(s, "TransactionMaturity");
```

The error would be much clearer. Maybe there is a crate somewhere for better messages when matching against enums. In any case I don't suggest you change anything, just thought I would share.

docs/src/calling-contracts/tx-params.md
```diff
@@ -3,6 +3,7 @@
 <!-- This section should explain what the tx params are and how to configure them -->
 <!-- tx_params:example:start -->
 The parameters for a transaction are:
+//TODO: Fix docs to use Policies
```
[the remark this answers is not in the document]
hal3e: Yes it is. @MujkicA and I will update the documentation.

packages/fuels/tests/contracts/revert_transaction_error/src/main.sw
Br1ght0ne: FYI, DCE seems to mean [dead code elimination](https://en.wikipedia.org/wiki/Dead-code_elimination).

hal3e: I changed the input to be a boolean so we cat decide what behavior we want.

digorithm:
Adding to @iqdecay's

> * I'm not sure I understand the logic behind renaming everything. To me, we are exposing implementation details to the user. As a user, using `TxPolicies` instead of `TxParameters` feels unusual at best. Also I feel like our responsability as the SDK is to shield the user from all the little internal changes the `fuel-core` or `fuel-vm` repos go through. So I'm not sure why we are replicating all these breaking changes in our API

Think of "transaction policies" as the new domain language here. This is coming from upstream (fuel-core), meaning that if we somehow wrap it around our own `TxParameters" language, we will be causing a _lot_ of confusion in the future when this abstraction inevitably leaks. Sticking with one language is best for communicating. A lot of users will be talking about the configurations for transaction policies. If the SDK forces it to be called transaction parameters, it'll just be confusing!

docs/src/calling-contracts/tx-policies.md
```diff
@@ -0,0 +1,46 @@
...
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:tx_policies_struct}}
+```
+
+Where: 
+
+1. **Gas Price** - Maximum gas price for transaction
+2. **Witness Limit** - The maximum amount of witness data 
+allowed for the transaction
+3. **Maturity** - Block until which tx cannot be included.
```
digorithm, line 16:
```suggestion
3. **Maturity** - Block until which transaction cannot be included.
```

docs/src/calling-contracts/tx-policies.md
```diff
@@ -0,0 +1,46 @@
...
+You can also use `TxPolicies::default()` to use the default values.
+<!-- tx_policies_default:example:end -->
+
+This way:
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:tx_policies_default}}
+```
+
+As you might have noticed already, `TxPolicies` can also be specified when deploying contracts or transferring assets by passing it to the respective methods.
```
digorithm, line 46:
```suggestion
As you might have noticed, `TxPolicies` can be specified when deploying contracts or transferring assets by passing it to the respective methods.
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,48 +298,144 @@ impl ScriptTransactionBuilder {
...
 
+        Ok(ScriptTransaction {
+            tx,
+            is_using_predicates,
+        })
+    }
+
+    // When dry running a tx with `utxo_validation` off, the node will not validate signatures.
+    // However, the node will check if the right number of witnesses is present.
+    // This function will create empty witnesses such that the total length matches the expected one.
```
digorithm, line 336: 💯

digorithm (changes requested):
Thanks @MujkicA for the doc updates.

Code comments are also much better now.

Just a few nits left.

digorithm (approved):
LFG! 🚀

segfault-magnet (approved):

PR #1202, opened 2023-11-13
feat: allow trasaction builder to be configured

xunilrj:
### Checklist
- [ ] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).

### DIFF
--- a/packages/fuels-core/src/types/transaction_builders.rs
+++ b/packages/fuels-core/src/types/transaction_builders.rs
@@ -193,9 +193,15 @@ macro_rules! impl_tx_trait {
 
         impl $ty {
             fn is_using_predicates(&self) -> bool {
-                self.inputs()
-                    .iter()
-                    .any(|input| matches!(input, Input::ResourcePredicate { .. }))
+                self.inputs().iter().any(|input| {
+                    matches!(
+                        input,
+                        Input::ResourcePredicate { .. }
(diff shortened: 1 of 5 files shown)


packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -193,9 +193,15 @@ macro_rules! impl_tx_trait {
...
         impl $ty {
             fn is_using_predicates(&self) -> bool {
-                self.inputs()
-                    .iter()
-                    .any(|input| matches!(input, Input::ResourcePredicate { .. }))
+                self.inputs().iter().any(|input| {
+                    matches!(
+                        input,
+                        Input::ResourcePredicate { .. }
+                            | Input::FuelInput(FuelInput::CoinPredicate(_))
```
hal3e, line 200: We should not be mixing the SDK `Input` with the `FuelInput` there is a reason why we made this separation

hal3e (changes requested):
@xunilrj thank you for the contribution. Unfortunately, this will probably not be merged. There is a reason why we created the sdk `Input` wrapper and we are not using the `fuel_tx::Input`. To not go into much detail here, we try to abstract some implementation details from the user and try to help them create those inputs with correct `offsets` and `witness` indexes.

Issue #1204, opened 2023-11-15
Write some docs for the `Bytes` support


digorithm:
There's no explicit mention of it in our docs.

PR #1205, opened 2023-11-16
feat: add `from_hex_str` for `Bytes`

hal3e:
close: https://github.com/FuelLabs/fuels-rs/issues/1204

co-developed with: @segfault-magnet 

Adds `Bytes::from_hex_str(hex: &str)`. Adds test and updates documentation.

### Checklist


packages/fuels-core/src/types/core/bytes.rs
```diff
@@ -1,6 +1,29 @@
...
 
+impl Bytes {
+    /// Create a new `Bytes` from a string representation of a hex.
+    /// Accepts both `0x` prefixed and non-prefixed hex strings.
+    pub fn from_hex_str(hex: &str) -> Result<Self> {
+        let hex = if let Some(stripped_hex) = hex.strip_prefix("0x") {
+            stripped_hex
+        } else {
+            hex
+        };
```
Br1ght0ne, line 14:
```suggestion
        let hex = hex.strip_prefix("0x").unwrap_or(hex);
```

Br1ght0ne (approved):
LGTM! Left a small nit, but it may be ignored if it's not too obvious for the reader.

segfault-magnet (approved):

packages/fuels-core/src/types/core/bytes.rs
segfault-magnet: A release is coming up and it would be great if this was part of it. Since we need to do it now, we'll merge without this suggestion and apply it afterward. 

@Br1ght0ne

PR #1209, opened 2023-11-20
refactor: separate receipts from `TxStatus`

iqdecay:
This PR took me a while to figure out, but in order to catch the errors at the
right time, and to further align with the types from `fuel-core`, it seemed
necessary to separate checking for a transaction status and checking for the
receipts of this transaction.

### DIFF
--- a/examples/contracts/src/lib.rs
+++ b/examples/contracts/src/lib.rs
@@ -904,9 +904,11 @@ mod tests {
         let tx = tb.build(provider).await?;
 
         let tx_id = provider.send_transaction(tx).await?;
-        let tx_status = provider.tx_status(&tx_id).await?;
+        let receipts = provider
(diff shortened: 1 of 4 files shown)


Br1ght0ne (changes requested):
Needs to merge in `master` to include https://github.com/FuelLabs/fuels-rs/pull/1200. As this targets another PR, could pose an inconvenience.

digorithm (commented):
Seems like we're still using `take_receipts_checked` in some places (e.g., https://github.com/FuelLabs/fuels-rs/pull/1209/files#diff-e2d910d4db2123b78ea4a2e8fa4971ac2a87c8b90c9a94678b91ab6b668cbcd5R673 and since it doesn't exist anymore it's breaking the CI 👀

Br1ght0ne (approved):

PR #1211, opened 2023-11-22
chore: bump rust version

hal3e:


Salka1988 (approved):

hal3e:
We cannot bump rust because our docs workflows use `ubuntu-latest` which does not yet have rust version `1.74.0`

PR #1212, opened 2023-11-23
feat!: improve transaction builders wrt `TxPolicies`

hal3e:
This PR improves the usage of `TxPolicies` and fixes some of the issues that remained after their were introduced.

BREAKING CHANGE:
- `ScriptTransactionBuilder` and `CreateTransactionBuilder` do not have `with_maturity` and `with_gas_price`, (`with_gas_limit`). They are set through the `with_tx_policies` method.
- `Maturity` is not set to `0` by default - instead we use an `Option<u64>` and `None` if it was not set. This saves bytes as it is not serialized if `None`
- `WitnessLimit` is set to the size of all witnesses in the builder. If the user uses `append_witnesses` after finalizing the transaction a new error will be returned that helps the user set the witness limit manually.

### Checklist

### DIFF
--- a/packages/fuels-core/src/types/transaction_builders.rs
+++ b/packages/fuels-core/src/types/transaction_builders.rs
@@ -344,12 +344,12 @@ impl ScriptTransactionBuilder {
         })
     }
 
-    async fn gas_used(
+    async fn set_script_gas_limit_to_gas_used(
         tx: &mut Script,
         provider: &impl DryRunner,
         network_info: &NetworkInfo,
         tolerance: f32,
-    ) -> Result<u64> {
+    ) -> Result<()> {
         let consensus_params = &network_info.consensus_parameters;
 
         // The `dry_run` validation will check if there is an input present that can cover
@@ -372,8 +372,9 @@ impl ScriptTransactionBuilder {
             tx.set_witness_limit(tx.witness_limit() + WITNESS_STATIC_SIZE as u64);
         }
 
-        // Add `1` because of rounding
+        // Get `max_gas` used by everything except the script execution. Add `1` because of rounding.
         let max_gas = tx.max_gas(consensus_params.gas_costs(), consensus_params.fee_params()) + 1;
+        // Increase `script_gas_limit` to the maximum allowed value.
         tx.set_script_gas_limit(network_info.max_gas_per_tx() - max_gas);
 
         let gas_used = provider
@@ -388,7 +389,9 @@ impl ScriptTransactionBuilder {
             tx.set_witness_limit(tx.witness_limit() - WITNESS_STATIC_SIZE as u64);
         }
 
-        Ok(gas_used)
+        tx.set_script_gas_limit(gas_used);
+
+        Ok(())
     }
 
     async fn resolve_fuel_tx_provider(
@@ -416,16 +419,17 @@ impl ScriptTransactionBuilder {
             dry_run_witnesses,
         );
 
-        let gas_limit = if has_no_code {
-            0
-        // Use the user defined value even if it makes the tx revert
+        if has_no_code {
+            tx.set_script_gas_limit(0);
+
+        // Use the user defined value even if it makes the transaction revert.
         } else if let Some(gas_limit) = self.tx_policies.script_gas_limit() {
-            gas_limit
+            tx.set_script_gas_limit(gas_limit);
 
         // If the `script_gas_limit` was not set by the user,
-        // `dry_run` the tx to get the `gas_used`
+        // dry-run the tx to get the `gas_used`
         } else {
-            Self::gas_used(
+            Self::set_script_gas_limit_to_gas_used(
                 &mut tx,
                 provider,
                 &self.network_info,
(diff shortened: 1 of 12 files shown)


packages/fuels-core/src/utils.rs
```diff
@@ -1,7 +1,18 @@
...
 pub const fn round_up_to_word_alignment(bytes_len: usize) -> usize {
     (bytes_len + (WORD_SIZE - 1)) - ((bytes_len + (WORD_SIZE - 1)) % WORD_SIZE)
 }
+
+pub(crate) fn calculate_witnesses_size<'a, I: IntoIterator<Item = &'a Witness>>(
+    witnesses: I,
+) -> usize {
+    witnesses
+        .into_iter()
+        .map(|w| w.as_ref().len() + WITNESS_STATIC_SIZE)
```
MujkicA, line 16: Could you explain why `WITNESS_STATIC_SIZE` is added

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -486,3 +499,36 @@ impl ScriptTransaction {
...
+                vec![],
+            ),
+            is_using_predicates: false,
+        };
+
+        let witness = vec![0, 1, 2].into();
+        let err = tx.append_witness(witness).expect_err("should error");
+
+        let expected_err_str = "Validation error: Witness limit exceeded. \
+                                Consider setting the limit manually with a transaction builder. \
```
MujkicA, line 529: Is `witness_limit ` a mealable field? Could it be adjusted on the transaction level instead of builder level?

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -252,39 +246,40 @@ macro_rules! impl_tx_trait {
...
                 Ok(num_witnesses as u8)
             }
+
+            fn calculate_witnesses_size(&self) -> Option<u64> {
+                let witnesse_size = calculate_witnesses_size(&self.witnesses);
+                let signature_size =
+                    SIGNATURE_WITNESS_SIZE * self.unresolved_signatures.secret_keys.len();
+
+                Some(padded_len_usize(witnesse_size + signature_size) as u64)
+            }
```
MujkicA, line 256:
```suggestion
            fn calculate_witnesses_size(&self) -> Option<u64> {
                let witnesses_size = calculate_witnesses_size(&self.witnesses);
                let signature_size =
                    SIGNATURE_WITNESS_SIZE * self.unresolved_signatures.secret_keys.len();

                Some(padded_len_usize(witnesses_size + signature_size) as u64)
            }
```

packages/fuels-core/src/utils.rs
hal3e: AFAIK that is how the witnesses are serialized. There is a static and dynamic part and the static size is constant and needs to be added when calculating the witness size.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -337,48 +328,64 @@ impl ScriptTransactionBuilder {
     fn create_dry_run_witnesses(&self, num_witnesses: u8) -> Vec<Witness> {
```
MujkicA, line 328: A comment here would be useful so we don't forget why we do this

packages/fuels-core/src/types/wrappers/transaction.rs
hal3e: `witness_limit` is not malleable. So it has to be adjusted on the builder level.

packages/fuels-core/src/types/transaction_builders.rs
hal3e: There is a comment just above the `fn`

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -337,48 +328,64 @@ impl ScriptTransactionBuilder {
...
             .dry_run_and_get_used_gas(tx.clone().into(), tolerance)
             .await?;
 
-        // Remove the temporary coin
-        tx.inputs_mut().pop();
-
-        tx.set_script_gas_limit(gas_used);
+        // Remove dry_run input
+        if no_input_for_fee {
+            tx.inputs_mut().pop();
```
MujkicA, line 385: Shouldn't we also reset the witnesses, witness limit and script_gas_limit here?

hal3e: We do not have to as we used the temporary `dry_run` witnesses. I will add a comment

MujkicA (commented):
I left some questions

packages/fuels-core/src/types/transaction_builders.rs
hal3e: I just realized that we should readjust the witness_limit! So thanks for the question!

hal3e: Refactored the code again. Please have a look

Br1ght0ne (approved):

digorithm (approved):
Super clean, great job!

Salka1988 (approved):

PR #1218, opened 2023-11-27
Bump versions to 0.52.0

digorithm:


digorithm:
> Isn't there an md file where it says "which is the latest fuel version" that you maybe forgot to change? Iirc it contains only `0.51` and not `0.51.0`.

It was here https://github.com/FuelLabs/fuels-rs/pull/1218/files#diff-4c08e6586ec5cb2cef9f5daa67e1ba40efbcb055cec06a9521ab167c2bbb1f5fR30 but we've moved to `x.y.z` a while back

PR #1224, opened 2023-12-05
docs: configure spell-check

sarahschwartz:
This PR configures the spell-check added to the docs workflow in the github-actions repo: https://github.com/FuelLabs/github-actions/pull/23

The files checked are configured in `.spellcheck.yml`.  This is also where you can configure what types of elements are ignored.

Right now it ignores:
 - All code blocks that have a language (and will check code blocks that do not have a language)
- Anything in between backticks
- Words in `spell-check-custom-words.txt` (case sensitive, only exact match)
- Numbers, even if they are attached to a word
- Links in markdown link format

#### Handling errors:

If the test fails:
- look up the word in the question to verify it is a real word and is correctly spelled
- If it is a file name or is code, use backticks to ignore the word.
- If it is a real word that is spelled correctly, or an acronym that is either common or is defined already, add it to `spell-check-custom-words.txt`.
- If needed, rewrite the sentence. Ex: DON'T use  "`lock`ing" and add "ing" to the custom words list. Instead, rewrite the sentence as "locking with the `lock` method".
- If it otherwise should be ignored, you can configure the pipeline in `.spellcheck.yml`.


digorithm (approved):
This is awesome, thanks for doing this @sarahschwartz!

Salka1988 (approved):

PR #1227, opened 2023-12-08
fix: prevent overflow while decoding param types

### LINKED ISSUE #1215: bug: `decode_multiple` can overflow
During a fuzzing campaign, it turned out that this test panics with an `attempt to add with overflow`:
```rust
        let param_types = vec![ParamType::Enum {
            variants: EnumVariants::new(vec![ParamType::Struct {
                fields: vec![
                    ParamType::StringArray(12514849900987264429),
                    ParamType::StringArray(7017071859781709229),
                ],
                generics: vec![],
            }])
            .unwrap(),
            generics: vec![],
        }];
        let bytes = vec![];
        let a = ABIDecoder::default().decode_multiple(&param_types, &bytes);
```

at the `decode_multiple` step.
We should probably enforce some upper limit on the length of `StringArray` and `Array`.

iqdecay:
Closes #1215 by fixing the way encoding size is computed

### DIFF
--- a/packages/fuels-core/src/codec/abi_decoder.rs
+++ b/packages/fuels-core/src/codec/abi_decoder.rs
@@ -60,11 +60,6 @@ impl ABIDecoder {
         BoundedDecoder::new(self.config).decode(param_type, bytes)
     }
 
-    /// Decode data from one of the receipt returns.
-    pub fn decode_receipt_return(&self, param_type: &ParamType, bytes: &[u8]) -> Result<Token> {
-        BoundedDecoder::new(self.config).decode(param_type, bytes)
-    }
-
     /// Same as `decode` but decodes multiple `ParamType`s in one go.
     /// # Examples
     /// ```
--- a/packages/fuels-core/src/codec/logs.rs
+++ b/packages/fuels-core/src/codec/logs.rs
@@ -35,8 +35,7 @@ impl LogFormatter {
         bytes: &[u8],
     ) -> Result<String> {
         Self::can_decode_log_with_type::<T>()?;
-        let token =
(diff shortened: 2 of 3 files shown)


Br1ght0ne (approved):
LGTM!

packages/fuels-core/src/types/param_types.rs
```diff
@@ -164,36 +159,68 @@ impl ParamType {
...
+            }
             _ => None,
-        }
+        };
+        Ok(heap_bytes_size)
     }
 
     /// Calculates the number of bytes the VM expects this parameter to be encoded in.
-    pub fn compute_encoding_in_bytes(&self) -> Option<usize> {
+    pub fn compute_encoding_in_bytes(&self) -> Result<usize> {
```
MujkicA, line 181: Maybe we can make this and the test a bit cleaner by adding a closure:

```rust
let overflow_error = || {
            error!(
                InvalidType,
                "Reached overflow while computing encoding size for {:?}",
                self
            )
        };
```

Then used it like this:

```rust
ParamType::Array(param, count) => param
                .compute_encoding_in_bytes()?
                .checked_mul(*count)
                .ok_or_else(overflow_error),
```

If the `Debug` representation contains more info than needed we can go for `Display`. Maybe the `Display` implementation will also be useful in other places where we return `InvalidType`.

digorithm (commented):
Great job finding this issue and fixing it. LGTM, but I agree with @MujkicA's suggestions; it'll make a lot nicer to the eyes.

Br1ght0ne (approved):

digorithm (approved):

MujkicA (approved):

Salka1988 (approved):

PR #1231, opened 2023-12-11
refactor!: remove `NetworkInfo`

hal3e:
When building a transaction with the SDK we currently ask for both `NetworkInfo` and a `provider` when building. 
`NetworkInfo` is redundant as we can fetch all necessary information from the `provider` while building.

BREAKING CHANGE:
- removed `NetworkInfo` and all related functions/methods
- renamed `DryRunner` trait and added trait methods
- removed `ScritpTransactionBuilder::new` and `CreateTransactionBuilder::new`
- removed `Provider::new`

### Checklist
- [x] I have updated the 

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ readme = "README.md"
 license = "Apache-2.0"
 repository = "https://github.com/FuelLabs/fuels-rs"
 rust-version = "1.74.0"
-version = "0.53.0"
+version = "0.54.0"
 
 [workspace.dependencies]
 Inflector = "0.11.4"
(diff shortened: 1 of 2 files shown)


packages/fuels-accounts/src/provider/supported_versions.rs
```diff
@@ -1,7 +1,7 @@
 use semver::Version;
 
 fn get_supported_fuel_core_version() -> Version {
-    "0.21.0-rc.1".parse().unwrap()
```
digorithm, line 4: Nice catch -- how can we make sure this value stays up-to-date? I mean, is there a way to automate that in our CI? I think it will be tricky to remember to update this constantly.

digorithm (approved):
Great work, @hal3e. This is an awesome UX simplification!

packages/fuels-accounts/src/provider/supported_versions.rs
segfault-magnet: I think @Br1ght0ne has something in mind over on #1190 . 

I'll review it today. Something caught my eye that might be a problem.

segfault-magnet (dismissed):
I'd rather keep the old name even if we sneak one more method under. 

This one is too generic. Might be a sign that we're looking at two traits: `DryRunner`, `ChainInfoProvider`.

digorithm:
> I'd rather keep the old name even if we sneak one more method under.
> 
> This one is too generic. Might be a sign that we're looking at two traits: `DryRunner`, `ChainInfoProvider`.

Fair enough

MujkicA:
The main point of the trait is to provide the used amount of gas.
Whether that is being done with a dry run or through some other means doesn't need to be exposed IMO. 
I would vote for calling it something like `GasEstimator`

packages/fuels-core/src/types/wrappers/node_info.rs
```diff
@@ -2,7 +2,7 @@
 
 use fuel_core_client::client::types::node_info::NodeInfo as ClientNodeInfo;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Default)]
```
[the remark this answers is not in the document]
hal3e: great catch! It is not needed.

hal3e:
I have decided to leave the name `DryRunner` for now.

packages/fuels-accounts/src/provider/supported_versions.rs
Br1ght0ne: #1190 should be good to go!

digorithm (approved):

segfault-magnet (approved):

PR #1232, opened 2023-12-11
test: add fuzzing targets

### LINKED ISSUE #1210: Explore adding fuzzing tests in the SDK
TBD.

iqdecay:
This closes #1210 partially.

Issues encountered while trying to set-up fuzzing on a larger scale:
- Due to the fact that fuzzing happens in its own crate, we can only fuzz public
  structs. 
  This severely limits the scale. See https://github.com/rust-fuzz/cargo-fuzz/issues/156.
- The structured fuzzing can only happen on types we declare ourselves. It means
  that we cannot fuzz function using types that contains any type that we do not
  own. 
  This severely limits the scale.
- Other tools for fuzzing are very hard to setup to do structured fuzzing. I
  took a look at https://llvm.org/docs/LibFuzzer.html. 
- Unstructured fuzzing seems more doable, but it seems less relevant.
- I might take a look at AFL next https://rust-fuzz.github.io/book/afl/tutorial.html.


fuzz/Cargo.toml
```diff
@@ -0,0 +1,46 @@
...
+[[bin]]
+name = "vector_of_param_type"
+path = "fuzz_targets/vector_of_param_type.rs"
+test = false
+doc = false
+[[bin]]
+name = "boxed_param_type"
+path = "fuzz_targets/boxed_param_type.rs"
+test = false
+doc = false
```
segfault-magnet, line 46: The files for these targets don't exist.
```suggestion
```

fuzz/fuzz_targets/boxed_param_type.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    let _ = vector.get_return_location();
+    let _ = vector.compute_encoding_in_bytes();
+    let _ = vector.children_need_extra_receipts();
+    let _ = vector.validate_is_decodable(max_depth);
+    let _ = vector.is_extra_receipt_needed(true);
+    let _ = vector.is_extra_receipt_needed(false);
+    let _ = vector.heap_inner_element_size(true);
+    let _ = vector.heap_inner_element_size(false);
+});
```
segfault-magnet, line 30: You can extract the repeating logic into a helper e.g. `utils.rs`. Also if you accept the `edition = "2021"` suggestion you can clean this up like so:
```suggestion
#![no_main]
mod utils;

use fuels::{core::codec::DecoderConfig, types::param_types::ParamType};
use libfuzzer_sys::fuzz_target;

fuzz_target!(|param_type: ParamType| {
    let max_depth = DecoderConfig::default().max_depth;
    let array = ParamType::Array(Box::new(param_type.clone()), 10_000);
    utils::exercise_param_type(array, max_depth);

    let vector = ParamType::Vector(Box::new(param_type));
    utils::exercise_param_type(vector, max_depth);
});
```

and `utils.rs`:
```rust
use fuels::types::param_types::ParamType;

pub fn exercise_param_type(param_type: ParamType, max_depth: usize) {
    let _ = param_type.children_need_extra_receipts();
    let _ = param_type.compute_encoding_in_bytes();
    let _ = param_type.get_return_location();
    let _ = param_type.heap_inner_element_size(false);
    let _ = param_type.heap_inner_element_size(true);
    let _ = param_type.is_extra_receipt_needed(false);
    let _ = param_type.is_extra_receipt_needed(true);
    let _ = param_type.validate_is_decodable(max_depth);
}
```

fuzz/fuzz_targets/vector_of_param_type.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    let _ = enum_of_param_types.get_return_location();
+    let _ = enum_of_param_types.compute_encoding_in_bytes();
+    let _ = enum_of_param_types.children_need_extra_receipts();
+    let _ = enum_of_param_types.validate_is_decodable(max_depth);
+    let _ = enum_of_param_types.is_extra_receipt_needed(true);
+    let _ = enum_of_param_types.is_extra_receipt_needed(false);
+    let _ = enum_of_param_types.heap_inner_element_size(true);
+    let _ = enum_of_param_types.heap_inner_element_size(false);
+});
```
segfault-magnet, line 47: Same suggestion here. 
```suggestion
#![no_main]

mod utils;

use fuels::types::param_types::ParamType;
use fuels::{core::codec::DecoderConfig, types::enum_variants::EnumVariants};
use libfuzzer_sys::fuzz_target;

fuzz_target!(|param_types: Vec<ParamType>| {
    let max_depth = DecoderConfig::default().max_depth;

    let struct_of_param_types = ParamType::Struct {
        fields: param_types.clone(),
        generics: vec![],
    };
    utils::exercise_param_type(struct_of_param_types, max_depth);

    let tuple = ParamType::Tuple(param_types.clone());
    utils::exercise_param_type(tuple, max_depth);

    let enum_of_param_types = ParamType::Enum {
        variants: EnumVariants::new(param_types).unwrap(),
        generics: vec![],
    };
    utils::exercise_param_type(enum_of_param_types, max_depth);
});
```

packages/fuels-core/src/codec/abi_decoder/bounded_decoder.rs
```diff
@@ -333,24 +333,23 @@ impl BoundedDecoder {
...
+        let skip_extra_in_bytes = match variants.heap_type_variant() {
+            Some((heap_type_discriminant, heap_type)) => {
+                if heap_type_discriminant == discriminant {
+                    heap_type.compute_encoding_in_bytes()?
+                } else {
+                    0
+                }
+            }
+            None => 0,
+        };
```
segfault-magnet, line 350:
```suggestion
        let skip_extra_in_bytes = match variants.heap_type_variant() {
            Some((heap_type_discriminant, heap_type)) if heap_type_discriminant == discriminant => {
                heap_type.compute_encoding_in_bytes()?
            }
            _ => 0,
        };
```

packages/fuels-core/src/types/enum_variants.rs
```diff
@@ -48,32 +49,27 @@ impl EnumVariants {
...
         if self.only_units_inside() {
-            return Some(ENUM_DISCRIMINANT_BYTE_WIDTH);
+            return Ok(ENUM_DISCRIMINANT_BYTE_WIDTH);
         }
 
         let width = self.param_types().iter().try_fold(0, |a, p| {
             let size = p.compute_encoding_in_bytes()?;
-            Some(a.max(size))
+            Result::<usize>::Ok(a.max(size))
         })?;
```
segfault-magnet, line 60:
```rust
        let width = self.param_types().iter().try_fold(0, |a, p| -> Result<_> {
            let size = p.compute_encoding_in_bytes()?;
            Ok(a.max(size))
        })?;
```

packages/fuels-core/src/types/param_types.rs
```diff
@@ -128,21 +126,20 @@ impl ParamType {
...
-                    Err(error!(
+                    return Err(error!(
                         InvalidType,
                         "Enums currently support only one heap-type variant. Found: \
                         {num_of_children_needing_receipts}"
                     ))
-                } else {
-                    Ok(())
-                }
+                };
```
segfault-magnet, line 134: So you don't repeat the `compute_encoding_in_bytes` validation:
```rust
    pub fn validate_is_decodable(&self, max_depth: usize) -> Result<()> {
        if let ParamType::Enum { variants, .. } = self {
            let all_param_types = variants.param_types();
            let grandchildren_need_receipts = all_param_types
                .iter()
                .any(|child| child.children_need_extra_receipts());
            if grandchildren_need_receipts {
                return Err(error!(
                    InvalidType,
                    "Enums currently support only one level deep heap types."
                ));
            }

            let num_of_children_needing_receipts = all_param_types
                .iter()
                .filter(|param_type| param_type.is_extra_receipt_needed(false))
                .count();

            if num_of_children_needing_receipts > 1 {
                return Err(error!(
                    InvalidType,
                    "Enums currently support only one heap-type variant. Found: \
                        {num_of_children_needing_receipts}"
                ));
            }
        } else if self.children_need_extra_receipts() {
            return Err(error!(
                InvalidType,
                "type {:?} is not decodable: nested heap types are currently not supported except in Enums.",
                DebugWithDepth::new(self, max_depth)
            ));
        }

        self.compute_encoding_in_bytes()?;
        Ok(())
    }
```

packages/fuels-core/src/types/param_types.rs
```diff
@@ -164,36 +161,67 @@ impl ParamType {
...
-            _ => None,
+            ParamType::Bytes | ParamType::String => Ok(Some(std::mem::size_of::<u8>())),
+            ParamType::StringSlice if !top_level_type => {
+                Ok(Some(ParamType::U8.compute_encoding_in_bytes()?))
+            }
+            ParamType::RawSlice if !top_level_type => {
+                Ok(Some(ParamType::U64.compute_encoding_in_bytes()?))
+            }
+            _ => Ok(None),
         }
```
segfault-magnet, line 178: A way to cut down on the `Ok` repetition:
```rust
        let heap_bytes = match &self {
            ParamType::Vector(inner_param_type) => {
                Some(inner_param_type.compute_encoding_in_bytes()?)
            }
            // `Bytes` type is byte-packed in the VM, so it's the size of an u8
            ParamType::Bytes | ParamType::String => Some(std::mem::size_of::<u8>()),
            ParamType::StringSlice if !top_level_type => {
                Some(ParamType::U8.compute_encoding_in_bytes()?)
            }
            ParamType::RawSlice if !top_level_type => {
                Some(ParamType::U64.compute_encoding_in_bytes()?)
            }
            _ => None,
        };

        Ok(heap_bytes)

```

packages/fuels-core/src/utils.rs
```diff
@@ -1,13 +1,31 @@
...
+            InvalidType,
+            "Remainder overflow while rounding up {bytes_len} bytes to word alignment"
+        )
+    })?;
+    lhs.checked_sub(rhs).ok_or_else(|| {
+        error!(
+            InvalidType,
+            "Substraction overflow while rounding up {bytes_len} bytes to word alignment"
+        )
+    })
```
segfault-magnet, line 27: I'm not sure this level of detail is useful to the user.

Maybe to be vague here, catch it somewhere higher up and give an error message like "Cannot handle xyz of this size"

segfault-magnet (changes requested):
The fuzzing tests don't build. It seems some are missing, others are missing imports, etc. 

We should think about making this a part of our CI, if not the fuzzing part then at least to have this code pass `cargo check` so that we can fuzz when we want to.

Ideally a manually runnable action on the CI as well.

Issue #1239, opened 2023-12-15
Error transfering custom assets of amount greater than the amount of owned base asset


mpoplavkov:
The `transfer` function throws an error on attempt to transfer a custom asset, when the amount of base asset is lower than the transfer amount. 

For example:
- wallet has 100 USDC (let's say decimals = 0)
- and 0.00000001 ETH (which is 10 if we don't take decimals into account)

An attempt to transfer 100USDC will result int the error "not enough coins to fit the target"

PR #1240, opened 2023-12-15
fix: use only base asset amount when calculating fee

mpoplavkov:
Resolves the issue https://github.com/FuelLabs/fuels-rs/issues/1239

### DIFF
--- a/packages/fuels-accounts/src/account.rs
+++ b/packages/fuels-accounts/src/account.rs
@@ -213,7 +213,9 @@ pub trait Account: ViewOnlyAccount {
             ScriptTransactionBuilder::prepare_transfer(inputs, outputs, tx_policies);
 
         self.add_witnessses(&mut tx_builder);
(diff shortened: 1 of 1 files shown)


digorithm (changes requested):
Nice catch. The solution seems reasonable to me; can't think of anything wrong that this could lead up to, at the moment. 

The CI is failing because of a tiny formatting issue; make sure to `rustfmt` the code. 

Invoking some of the guys that worked on fees to double-check this solution: @MujkicA @hal3e @segfault-magnet

MujkicA (approved):
LGTM, thanks!

digorithm (approved):
Many thanks, @mpoplavkov! We'll cut a new release with this change asap.

mpoplavkov:
Awesome, thanks!

Salka1988 (approved):

PR #1241, opened 2023-12-18
refactor!: signing and the `Signer` trait

hal3e:
This PR refactors the `Signer` trait and the way we sign transactions. 

The new `Signer` trait has two methods:
- `sign` - signs a `Message` and returns a `Signature` asynchronously
- `address` (public key)

When signing built `Transactions` we use the `sign_with(& impl Signer, ChainId)` method.
For `TransactionBuilder` we use the `add_signer(impl Signer)` method. The `Signer` (not the `SecretKey` as before) is stored on the heap and the `sign` method is called while building the tx.

In addition, this PR adds the ability to build transaction without signatures. This is useful if the user wants to estimate costs or wants to simulate the TX on a node without setting the right signatures.

Changes:
- Refactor `Signer` trait
- Added a new method called `build_without_signatures(provider: &impl DryRunner)` to `TransactionBuilders`
- Replaced the Transaction `check_without_signatures` method with `check`
- `fee_checked_from_tx` now builds without signatures which means we can use `adjust_for_fee` without signing first cc @MujkicA
- Added documentation

NOTE:  If the user builds without signatures, it is their responsibility to sign the transaction in the right order. For example, if we have `CoinSignedA` then `CoinSignedB`, we need to sign with `WalletA` and then with `WalletB`. This comes from the fact that we need to set the witness indexes while building the transaction.

BREAKING CHANGE:
- Removed `sign_message` and `sign_transaction` from the `Signer` trait
- Added `sign` and `address` methods to the `Signer` trait
- `Signer` trait moved do `fuels::core::traits:::Signer`
- `Message`, `PublicKey`, `SecretKey` and `Signature` moved to `fuels::crypto::`
- Replaced `Transaction`'s `check_without_signatures` with `check`
- Renamed `Account`s `add_witnessses` to `add_witnesses`
- Removed `Clone` for `TransactionBuilder`s

### 


docs/src/custom-transactions/transaction-builders.md
```diff
@@ -83,3 +83,13 @@ Finally, we verify the transaction succeeded and that the cold storage indeed ho
 ```rust,ignore
 {{#include ../../../examples/cookbook/src/lib.rs:custom_tx_verify}}
 ```
+
+## Building a transaction without signatures
+
+If you need to build the transaction without signatures - useful when estimating transaction costs or simulations,  you can use the `build_without_signatures(&provider)` method and later sign the built transaction.
```
digorithm, line 89:
```suggestion
If you need to build the transaction without signatures, which is useful when estimating transaction costs or simulations, you can use the `build_without_signatures(&provider)` method and later sign the built transaction.
```

docs/src/custom-transactions/transaction-builders.md
```diff
@@ -83,3 +83,13 @@ Finally, we verify the transaction succeeded and that the cold storage indeed ho
...
+
+```rust,ignore
+{{#include ../../../packages/fuels/tests/contracts.rs:tb_build_without_signatures}}
+```
+
+> **Note** In contrast to signing a transaction builder, when signing a built transaction, you will have to make sure that the order of signatures matches the order of signed inputs.
```
digorithm, line 95: Clarity nit.

```suggestion
> **Note** In contrast to signing a transaction builder, when signing a built transaction, you must ensure that the order of signatures matches the order of signed inputs.
```

packages/fuels-accounts/src/wallet.rs
```diff
@@ -263,19 +264,32 @@ impl Account for WalletUnlocked {
...
         let sig = Signature::sign(&self.private_key, &message);
+
         Ok(sig)
     }
 
     fn sign_transaction(&self, tb: &mut impl TransactionBuilder) {
         tb.add_unresolved_signature(self.address().clone(), self.private_key);
     }
+
+    fn sign_built_transaction(&self, tx: &mut impl Transaction) -> Result<Signature> {
```
digorithm, line 279: I know we already mentioned it in the docs, but I believe we should add a comment here, again, about calling this method in the right witness order; I can see this `tx.append_witness` (along with tx being `mut`) causing some issues due to ordering and being something hard to debug. Official docs + method-level docs should make things easier.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -906,4 +941,110 @@ mod tests {
...
+        fn consensus_parameters(&self) -> &ConsensusParameters {
+            &self.c_param
+        }
+        async fn min_gas_price(&self) -> Result<u64> {
+            Ok(0)
+        }
+    }
+
+    #[tokio::test]
+    async fn create_tb_builder_witness_indexes_set_correctly() -> Result<()> {
```
digorithm, line 982: Shouldn't it be tx builder instead of tb builder?
```suggestion
    async fn create_tx_builder_witness_indexes_set_correctly() -> Result<()> {
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -906,4 +941,110 @@ mod tests {
...
+    async fn scritp_tb_builder_witness_indexes_set_correctly() -> Result<()> {
+        // given
+        let num_witnesses = 6;
+        let num_inputs = 4;
+
+        let tb = ScriptTransactionBuilder::default()
+            .with_witnesses(given_witnesses(num_witnesses))
+            .with_inputs(given_inputs(num_inputs));
+
+        //when
```
digorithm, line 1026:
```suggestion
        // when
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -906,4 +941,110 @@ mod tests {
...
+        let tx = tb
+            .build_without_signatures(&MockDryRunner::default())
+            .await?;
+
+        //then
```
digorithm, line 1031:
```suggestion
        // then
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -906,4 +941,110 @@ mod tests {
...
+    async fn create_tb_builder_witness_indexes_set_correctly() -> Result<()> {
+        // given
+        let num_witnesses = 2;
+        let num_inputs = 3;
+
+        let tb = CreateTransactionBuilder::default()
+            .with_witnesses(given_witnesses(num_witnesses))
+            .with_inputs(given_inputs(num_inputs));
+
+        //when
```
digorithm, line 991:
```suggestion
        // when
```

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -906,4 +941,110 @@ mod tests {
...
+        let tx = tb
+            .build_without_signatures(&MockDryRunner::default())
+            .await?;
+
+        //then
```
digorithm, line 996:
```suggestion
        // then
```

digorithm (changes requested):
Great work; very clean new API! Left some nits/suggestions.

segfault-magnet (commented):
Just some random thoughts that came with the morning coffee as I was reviewing this:

The names of `sign_build_transaction` and `sign_transaction` sound like we're injecting the fn argument type into the function name to workaround rust not having function overloading. Something akin to overloading can be achieved via traits. WDYT about having `Signable` and `Signer` traits (we already have the latter). Then have `Signer` sign anything that's `Signable`:
```rust
pub trait Signable {
    fn sign(&mut self, owner: &Bech32Address, key: &SecretKey, chain_id: ChainId) -> Result<()>;
}
```

Change the definitions of `Transaction` and `TransactionBuilder` traits to include the new trait as a bound:
```rust
pub trait Transaction:
    Signable + Into<FuelTransaction> + EstimablePredicates + GasValidation + Clone + Debug
// ...

pub trait TransactionBuilder: Signable + BuildableTransaction + Send + Clone {
```

Then implement for all tx types
```rust
        impl Signable for $wrapper {
            fn sign(
                &mut self,
                _owner: &Bech32Address,
                key: &::fuel_crypto::SecretKey,
                chain_id: ChainId,
            ) -> Result<()> {
                let id = self.id(chain_id);

                let message = Message::from_bytes(*id);
                let sig = Signature::sign(&key, &message);

                self.append_witness(sig.as_ref().into())?;

                Ok(())
            }
        }
```

and all tx builder types:
```rust
        impl Signable for $ty {
            fn sign(
                &mut self,
                owner: &Bech32Address,
                key: &SecretKey,
                _chain_id: ChainId,
            ) -> Result<()> {
                self.add_unresolved_signature(owner.clone(), key.clone());
                Ok(())
            }
        }
```

Then you can change `Signer` to have a single `sign` method that signs anything that's `Signable`:
```rust
pub trait Signer: std::fmt::Debug + Send + Sync {
    type Error: std::error::Error + Send + Sync;
    // ...
    fn sign<S: Signable>(&self, signable: &mut S) -> std::result::Result<(), Self::Error>;
}
```

When signing for both builders and transactions looks like this:
```rust
let tx = ...;
wallet.sign(&mut tx);

let tx_builder = ... ;
wallet.sign(&mut tx_builder); 
```


Some more morning thoughts: 

Because of the whole witness index resolving thing we were forced to capture private keys until the tx is built and witnesses settled down (mostly). 

This is why the `Signable` interface requires the private key. Ideally we should think about capturing a collection of `Signer` implementations to future-proof against hardware wallets as `Signer`s -- in that case we won't be able to extract the private key -- we will be able only to provide a `Message` for the hardware wallet to sign.

So maybe prepare for that in the near future and have `TxBuilder`s keep something like `Vec<Box<dyn Signer>>`.

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -56,18 +57,21 @@ impl<T: DryRunner> DryRunner for &T {
...
 
 #[cfg_attr(not(target_arch = "wasm32"), async_trait)]
 pub trait BuildableTransaction {
     type TxType: Transaction;
 
     async fn build(self, provider: &impl DryRunner) -> Result<Self::TxType>;
+
+    /// Building without signatures will set the witness indexes of signed coins in the
+    /// order as they appear in the inputs. Multiple coins with the same owner will have
+    /// the same witness index. Make sure you sign the  built transaction in the expected order.
```
digorithm, line 73:
```suggestion
    /// the same witness index. Make sure you sign the built transaction in the expected order.
```

digorithm (changes requested):
Did a first pass, looking great. Excellent work! Just left a couple of doc-related nits.

packages/fuels-programs/src/script_calls.rs
```diff
@@ -214,8 +214,8 @@ where
     pub async fn build_tx(&self) -> Result<ScriptTransaction> {
         let mut tb = self.transaction_builder().await?;
 
-        self.account.add_witnessses(&mut tb);
         self.account.adjust_for_fee(&mut tb, 0).await?;
+        self.account.add_witnesses(&mut tb);
```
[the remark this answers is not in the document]
hal3e: Previously, we had to first add the signers before `adjusting_for_fee` as we would internally use `build` which required the signatures to be there. Now that we can `build_without_signatures` there is no need to sign first and then estimate and adjust for fees. 

However, If there are many signers that could impact the fee estimation and therefore it is better to set the `Signers` before the fee estimation. Thanks!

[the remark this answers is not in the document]
digorithm: Great point

digorithm (approved):

docs/src/custom-transactions/transaction-builders.md
```diff
@@ -52,20 +52,22 @@ We combine all of the inputs and outputs and set them on the builder:
...
+We need to do one more thing before we stop thinking about transaction inputs. Executing the transaction also incurs a fee that is paid with the base asset. Our base asset inputs need to be large enough so that the total amount covers the transaction fee and any other operations we are doing. The `Account` trait lets us use `adjust_for_fee()` for adjusting the transaction inputs if needed to cover the fee. The second argument to `adjust_for_fee()` is the total amount of the base asset that we expect our transaction to spend regardless of fees. In our case, this is the **ask_amount** we are transferring to the predicate.
 
 ```rust,ignore
 {{#include ../../../examples/cookbook/src/lib.rs:custom_tx_adjust}}
 ```
 
+> **Note** It is recommended to add signers before adjusting for fee as the estimation will include the witnesses fee.
```
MujkicA, line 69:
```suggestion
> **Note** It is recommended to add signers before calling `adjust_for_fee()` as the estimation will include the size of the witnesses.
```

MujkicA (dismissed):

MujkicA (dismissed):

digorithm (dismissed):
:shipit:

Salka1988 (dismissed):

packages/fuels-core/src/traits/signer.rs
```diff
@@ -0,0 +1,14 @@
...
+use fuel_crypto::{Message, Signature};
+
+use crate::types::{bech32::Bech32Address, errors::Result};
+
+/// Trait for signing transactions and messages
+///
+/// Implement this trait to support different signing modes, e.g. hardware wallet, hosted etc.
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait Signer: std::fmt::Debug + Send + Sync + 'static {
```
segfault-magnet, line 11: I'm not sure about the `Debug` bound.

 It is needed because our transaction builders are debug. 

Maybe we should do a custom impl of `Debug` for the TBs and remove this bound here? 

In our custom impl on TBs we can just say `[Signer(PUB_ADDR), Signer(PUB_ADDR), ...]`.

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -170,8 +173,14 @@ pub trait GasValidation {
     fn validate_gas(&self, min_gas_price: u64, gas_used: u64) -> Result<()>;
 }
 
+#[cfg_attr(target_arch = "wasm32", async_trait(?Send))]
+#[cfg_attr(not(target_arch = "wasm32"), async_trait)]
+pub trait SignableTransaction {
```
segfault-magnet, line 178: Do we have some plan on how this is to be used? Don't see a usage in our code currently. Otherwise I'd just make it part of the Transaction trait.

hal3e: The idea was not to make the whole `Transaction` trait depend on `async_trait` because of one method.

packages/fuels-core/src/traits/signer.rs
hal3e: Added custom `Debug` for `dyn Signer`.

segfault-magnet (approved):

digorithm (approved):

PR #1243, opened 2024-01-06
chore!: remove duplicate function

iqdecay:
The `decode_receipt_return` function was a duplicate of the `decode` function, this PR removes it.


digorithm (approved):
Nice catch! Since this is a pub fn we should keep in mind this is a breaking change

hal3e (approved):
Nice catch!

MujkicA (approved):

Salka1988 (approved):

Br1ght0ne (approved):

PR #1249, opened 2024-01-14
fix!: encoding capacity overflow

### LINKED ISSUE #1228: bug: `capacity overflow` while encoding param types
Encoding param types without checking for their size can lead to capacity overflow.

iqdecay:
- This PR closes #1228 by adding the `EncoderConfig` similarly to what was done in #1066.
- We bubble up the encoding error only when `call()` or `simulate()` is called on contract or script calls. 

BREAKING CHANGE: 
- `Configurables` structs now need to be instantiated through a `::new(encoder_config)` or `::default()` method.
- `Configurables::with_some_string_config(some_string)` methods now return a `Result<Configurables>` instead of `Configurables`.
- `Predicates::encode_data` now returns a `Result<UnresolvedBytes>` instead of `UnresolvedBytes`.
- `PredicateEncoder` structs must be instantiated through a `::new(encoder_config)` or `::default()`  method.

### DIFF
--- a/packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
+++ b/packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
@@ -88,8 +88,8 @@ fn generate_builder_methods(resolved_configurables: &[ResolvedConfigurable]) -> 
                 #[allow(non_snake_case)]
                 // Generate the `with_XXX` methods for setting the configurables
                 pub fn #name(mut self, value: #ttype) -> ::fuels::prelude::Result<Self> {
-                    let encoder = #encoder_code?.resolve(0);
-                    self.offsets_with_data.push((#offset, encoder));
+                    let encoded = #encoder_code?.resolve(0);
+                    self.offsets_with_data.push((#offset, encoded));
                     ::fuels::prelude::Result::Ok(self)
                 }
             }


Br1ght0ne (approved):
LGTM!

packages/fuels-programs/src/contract.rs
```diff
@@ -732,7 +739,7 @@ pub fn method_hash<D: Tokenizable + Parameterize + Debug, T: Account>(
 
     let compute_custom_input_offset = should_compute_custom_input_offset(args);
 
-    let unresolved_bytes = ABIEncoder::encode(args)?;
+    let unresolved_bytes = ABIEncoder::default().encode(args)?;
```
hal3e, line 742: The `method_hash` function should have the `encoder_configuration` as an input so that the `ABIEncoder` can be customized.

The user should be able to set the `encoder_configuration` wih the `contract_instance` and then pass it to `methods` which uses `method_hash`. The final result could look like this:
```rust
contract_instance.with_encoder_configuration(encoder_config).methods().my_method().call()
```
The `contract_instance` should have a default `encoder_config` at initialization so that this is also possible:
```rust
contract_instance.methods().my_method().call()
```

packages/fuels-programs/src/contract.rs
```diff
@@ -754,6 +761,7 @@ pub fn method_hash<D: Tokenizable + Parameterize + Debug, T: Account>(
         datatype: PhantomData,
         log_decoder,
         decoder_config: Default::default(),
+        encoder_config: Default::default(),
```
hal3e, line 764:
```suggestion
```

packages/fuels-programs/src/script_calls.rs
```diff
@@ -83,6 +83,7 @@ pub struct ScriptCallHandler<T: Account, D> {
     // Initially `None`, gets set to the right tx id after the transaction is submitted
     cached_tx_id: Option<Bytes32>,
     decoder_config: DecoderConfig,
+    encoder_config: DecoderConfig,
```
hal3e, line 86:
```suggestion
```

packages/fuels-programs/src/script_calls.rs
```diff
@@ -117,6 +118,7 @@ where
             datatype: PhantomData,
             log_decoder,
             decoder_config: DecoderConfig::default(),
+            encoder_config: DecoderConfig::default(),
```
hal3e, line 121:
```suggestion
```

packages/fuels-programs/src/script_calls.rs
```diff
@@ -138,6 +140,11 @@ where
         self
     }
 
+    pub fn with_encoder_config(mut self, encoder_config: DecoderConfig) -> Self {
+        self.encoder_config = encoder_config;
+        self
+    }
+
```
hal3e, line 147:
```suggestion
```

hal3e (changes requested):
Co-reviewed with @MujkicA

Overall looks good but we need to enable customizing the encoding, and update the docs with examples how the new encoder is used.

packages/fuels-code-gen/src/program_bindings/abigen/bindings/contract.rs
```diff
@@ -62,8 +64,25 @@ pub(crate) fn contract_bindings(
...
+            }
+
+            pub fn with_encoder_config(&self, encoder_config: ::fuels::core::codec::EncoderConfig)
+            -> #name::<T> {
+                #name {
+                        contract_id: self.contract_id.clone(),
+                        account: self.account(),
+                        log_decoder: self.log_decoder.clone(),
+                        encoder_config,
+                }
```
hal3e, line 85: I would use the builder pattern here and take ownership. If the user wants to keep a clone he should explicitly do so. Btw this is what we do for `script_instances`

```suggestion
            pub fn with_encoder_config(mut self, encoder_config: ::fuels::core::codec::EncoderConfig)
            -> #name::<T> {
                self.encoder_config = encoder_config;

                self
            }
```

packages/fuels-code-gen/src/program_bindings/abigen/bindings/contract.rs
```diff
@@ -62,8 +64,25 @@ pub(crate) fn contract_bindings(
...
+            pub fn with_account<U: ::fuels::accounts::Account>(&self, account: U)
+            -> ::fuels::types::errors::Result<#name<U>> {
+                ::core::result::Result::Ok(
+                #name {
+                        contract_id: self.contract_id.clone(),
+                        account,
+                        log_decoder: self.log_decoder.clone(),
+                        encoder_config: self.encoder_config.clone()
+                })
+            }
```
hal3e, line 76: `script_instances` takes ownership and I suggest we do the same here.

```suggestion
            pub fn with_account<U: ::fuels::accounts::Account>(self, account: U)
            -> ::fuels::types::errors::Result<#name<U>> {
                ::core::result::Result::Ok(
                #name {
                        contract_id: self.contract_id,
                        account,
                        log_decoder: self.log_decoder,
                        encoder_config: self.encoder_config
                })
            }
```

packages/fuels-code-gen/src/program_bindings/abigen/bindings/contract.rs
```diff
@@ -40,7 +40,8 @@ pub(crate) fn contract_bindings(
         pub struct #name<T: ::fuels::accounts::Account> {
```
hal3e, line 40:
```suggestion
        #[derive(Debug, Clone)]
        pub struct #name<T: ::fuels::accounts::Account> {
```
IMO we should add `Clone` so that the user can clone the instance. See comments below.

packages/fuels-programs/src/contract.rs
```diff
@@ -724,6 +726,7 @@ pub fn method_hash<D: Tokenizable + Parameterize + Debug, T: Account>(
     args: &[Token],
     log_decoder: LogDecoder,
     is_payable: bool,
+    encoder_config: Option<EncoderConfig>,
```
hal3e, line 729: I could not see where we use `None` so I would recommend we remove it.

```suggestion
    encoder_config: EncoderConfig,
```

packages/fuels-code-gen/src/program_bindings/abigen/bindings/script.rs
```diff
@@ -40,7 +41,8 @@ pub(crate) fn script_bindings(
         pub struct #name<T: ::fuels::accounts::Account>{
```
hal3e, line 41: I would derive `Clone` also for the `script_instace`

hal3e (changes requested):
Great job! It is almost done. I just have some small requests.

packages/fuels/tests/contracts.rs
```diff
@@ -1775,3 +1775,42 @@ async fn contract_custom_call_build_without_signatures() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+#[should_panic(
+    expected = "method not found (this should never happen): InvalidType(\"Token limit (1) reached while Encoding. Try increasing it.\")"
```
segfault-magnet, line 1781: This is weird. Why are we getting the `method not found` message when reaching the token limit?

segfault-magnet (changes requested):

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
```diff
@@ -47,11 +47,6 @@ impl FunctionGenerator {
         self
     }
 
-    pub fn make_fn_associated(&mut self) -> &mut Self {
-        self.is_method = false;
-        self
-    }
-
```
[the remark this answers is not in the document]
hal3e: `is_method` is a private field and this is the corresponding setter. Even though it is currently not used, I would still leave it.

packages/fuels-core/src/codec/abi_decoder.rs
```diff
@@ -676,7 +676,7 @@ mod tests {
             max_depth: MAX_DEPTH,
             ..Default::default()
         };
-        let msg = format!("Depth limit ({MAX_DEPTH}) reached while decoding. Try increasing it.");
+        let msg = format!("Depth limit ({MAX_DEPTH}) reached while Decoding. Try increasing it.");
```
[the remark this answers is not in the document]
hal3e: I suggested we remove `strum` and this is the derived `Debug` output. Other option is to impl `Display` manually.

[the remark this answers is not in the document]
segfault-magnet: My only gripe was why the verb was capitalized. "Limit reached while decoding" vs "Limit reached while Decoding".

segfault-magnet (commented):
Feel queasy for panicking if we cannot encode the fn args due to default encoding limits. But then again adding in one more err check isn't fantastic either.

We should all (@FuelLabs/sdk-rust ) agree to either propagate the error or panic in this case. 
Propagating the error:
```rust
contract.methods().a_method(1,2,3)?.call().await?;
```
Panicking
```rust
contract.methods().a_method(1,2,3).call().await?;
```

Or maybe a third option? Encode but don't unpack the result, save it for later, unpack when you need it on `call`. We can then have our cake and eat it too:
```rust
contract.methods().a_method(1,2,3).call().await?;
```

Wdyt?

Edit: 
If we're to panic we should document it *really* well - not that apparent that you can crash your app if you add one more element to the vec you're sending.

Edit:
No need for dyn FnOnce, saving the result is enough.

digorithm (commented):
Quick update on this: This is neat, good job @iqdecay. I'm 70% through it. I should finish this last pass very soon.

packages/fuels-code-gen/src/program_bindings/abigen/bindings/function_generator.rs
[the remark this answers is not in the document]
digorithm: Why was this discussion resolved? I think I agree with @hal3e; if it's a field that the user can decide whether or not to read, its setter should exist. 

Unless the actual field isn't significant anymore?

packages/fuels-core/src/codec/abi_decoder.rs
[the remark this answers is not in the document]
digorithm: Why was this resolved? The capitalized verb is also bothering me, haha. Super minor, but still.

packages/fuels-core/src/codec/abi_encoder/bounded_encoder.rs
```diff
@@ -0,0 +1,277 @@
...
+
+        // Enums that contain only Units as variants have only their discriminant encoded.
+        if !variants.only_units_inside() {
+            let variant_param_type = variants.param_type_of_variant(*discriminant)?;
+            let enum_width_in_bytes = variants.compute_enum_width_in_bytes()?;
+
+            if enum_width_in_bytes > self.max_total_enum_width {
+                return Err(error!(
+                    InvalidData,
+                    "Cannot encode Enum with variants {variants:?} : it is {enum_width_in_bytes} bytes wide. Try increasing encoder max memory."
```
digorithm, line 195:
```suggestion
                    "Cannot encode Enum with variants {variants:?}: it is {enum_width_in_bytes} bytes wide. Try increasing encoder max memory."
```

digorithm (changes requested):
I just left some questions/nits. But other than those, LGTM.

packages/fuels-core/src/codec/abi_decoder.rs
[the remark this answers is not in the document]
hal3e: We can fix this easily with:
```rust
pub(crate) enum CodecDirection {
    Encoding,
    Decoding,
}

impl std::fmt::Display for CodecDirection {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        match self {
            CodecDirection::Encoding => write!(f, "encoding"),
            CodecDirection::Decoding => write!(f, "decoding"),
        }
    }
}
```

and later we just use the `Display` and not `Debug` anymore.

```rust
return Err(error!(
    InvalidType,
    "{} limit ({}) reached while {}. Try increasing it.",
    self.name,
    self.max,
    self.direction
```

packages/fuels-programs/src/contract.rs
```diff
@@ -710,6 +712,7 @@ pub fn method_hash<D: Tokenizable + Parameterize + Debug, T: Account>(
     args: &[Token],
     log_decoder: LogDecoder,
     is_payable: bool,
+    encoder_config: EncoderConfig,
 ) -> Result<ContractCallHandler<T, D>> {
```
hal3e, line 716: There is no reason to return `Result` here so we should remove it.

hal3e (changes requested):
Found some small nits that will improve the DevEx

hal3e (approved):
Nice work!

packages/fuels/tests/scripts.rs
```diff
@@ -387,3 +387,44 @@ async fn test_script_transaction_builder() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn test_script_encoder_config_is_applied() {
```
MujkicA, line 392: Adding some line breaks would make these tests more readable

MujkicA (dismissed):

packages/fuels-code-gen/src/program_bindings/abigen/configurables.rs
```diff
@@ -86,9 +86,11 @@ fn generate_builder_methods(resolved_configurables: &[ResolvedConfigurable]) ->
             let encoder_code = generate_encoder_code(ttype);
             quote! {
                 #[allow(non_snake_case)]
-                pub fn #name(mut self, value: #ttype) -> Self{
-                    self.offsets_with_data.push((#offset, #encoder_code));
-                    self
+                // Generate the `with_XXX` methods for setting the configurables
+                pub fn #name(mut self, value: #ttype) -> ::fuels::prelude::Result<Self> {
+                    let encoder = #encoder_code?.resolve(0);
+                    self.offsets_with_data.push((#offset, encoder));
```
segfault-magnet, line 92:
```suggestion
                    let encoded = #encoder_code?.resolve(0);
                    self.offsets_with_data.push((#offset, encoded));
```

segfault-magnet (dismissed):
Turned out great. Safe with no dex damage.

segfault-magnet (approved):

hal3e (approved):

MujkicA (approved):

digorithm (approved):
Bring it home. Great work, @iqdecay!

PR #1257, opened 2024-01-22
test: enable testing against a live node

### LINKED ISSUE #1135: Run integration tests against a live node
# Problem

Sometimes, we catch integration errors with `fuel-core` only when users try using the SDK against a live node (e.g., `beta-4`), which causes frustration and takes us more time than it should to fix these issues. These issues aren't found when testing against a local node, even if the node's version is at the latest, because the _live_ node might have been configured differently.

# Potential solution

We need to catch these errors as early as possible by running some of our integration tests (tests that interact with a provider) against the live node and not only against a local version of the node. However, running these integration tests against a live node would introduce potential flakiness and higher latency due to all the network hops inside our CI runs. Doing this inside every pull request could slow us down unnecessarily. 

An approach to this could be to run these tests against a live node only when a PR is merged to master; the CI would trigger a run against the new master branch and run these tests against a live node. This way, it wouldn't block or slow down a PR, and since the master branch hasn't been released yet, catching an issue that's happening only against a live node at this stage isn't problematic, and we can start working on it before this issue is caught after a release.

iqdecay:
This PR closes #1135 by adding a new feature `test-against-live-node` which can be turned on to run some integration tests against a live node.
For some of them, to avoid SqueezedOutTransaction errors, I added a `sleep` call.

### DIFF
--- a/packages/fuels/tests/contracts.rs
+++ b/packages/fuels/tests/contracts.rs
@@ -1730,6 +1730,11 @@ async fn test_arguments_with_gas_forwarded() -> Result<()> {
             contract = "TestContract",
             wallet = "wallet"
         ),
+        Deploy(
+            name = "contract_instance_2",
+            contract = "VectorOutputContract",
+            wallet = "wallet"
+        ),
     );
     let x = 128;
     let vec_input = vec![0, 1, 2];


packages/fuels-test-helpers/src/accounts.rs
```diff
@@ -85,6 +86,44 @@ pub async fn launch_custom_provider_and_get_wallets(
...
+) -> Result<Vec<WalletUnlocked>> {
+    if num_wallets > 3 {
+        error!(
+            InvalidData,
+            "Trying to get more than 3 wallets from beta node"
+        );
+    }
+    let provider = Provider::connect(TESTNET_NODE_URL)
+        .await
+        .expect("Should be able to connect to {TESTNET_NODE_URL}");
```
Br1ght0ne, line 100: I suspect you wanted the value to be interpolated here:

```suggestion
        .unwrap_or_else(|_| panic!("Should be able to connect to {TESTNET_NODE_URL}"));
```

> [!NOTE]
> why not `expect(&format!(...))`? See [this clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/expect_fun_call)

Br1ght0ne (commented):
LGTM, left a minor suggestion.

packages/fuels-macros/src/setup_program_test/code_gen.rs
```diff
@@ -66,21 +66,22 @@ fn generate_abigen_targets(project_lookup: &HashMap<String, Project>) -> Vec<Abi
         .collect()
 }
 
-fn wallet_initialization_code(maybe_command: Option<InitializeWalletCommand>) -> TokenStream {
+fn wallet_initialization_code(
```
segfault-magnet, line 69: The fn will always return `Ok`, no need to have the `Result` in the ret type.

packages/fuels-test-helpers/Cargo.toml
```diff
@@ -35,3 +35,4 @@ which = { workspace = true, default-features = false }
 default = ["fuels-accounts", "std"]
 std = ["fuels-accounts?/std", "fuels-core/std"]
 fuel-core-lib = ["fuel-core"]
+test-against-live-node = []
```
segfault-magnet, line 38: I'd rather we don't introduce a feature if we don't have to. Since the test code compiles always (unlike the case with full sway type paths tests) I'd rather we make the decision to target the test net based on an env variable:
```rust
if env!(TARGET_TEST_NET) ... 
```

This way our internal testing approach doesn't leak into user features.

packages/fuels-test-helpers/src/accounts.rs
```diff
@@ -85,6 +86,44 @@ pub async fn launch_custom_provider_and_get_wallets(
     Ok(wallets)
 }
 
+pub async fn connect_to_testnet_node_and_get_wallets(
```
segfault-magnet, line 89: I wouldn't put these in the test helpers crate. The test helpers contain stuff we deem to be useful to all users. This is more of our internal desire to test the testnet. Why limit our users to stuff like:
* no more than 3 wallets
* the test net must be the official one from fuel (URL hardcoded)
* the env variables must be strictly the ones we set here (TEST_WALLET_SECRET_KEY_*) 
* etc

I'd rather we move this helper to our e2e tests and have it be private there since they are the only reason this exists.

segfault-magnet (changes requested):

Issue #1247, opened 2024-01-11
Support the Sway's new `__log` encoding scheme


digorithm:
Part of https://github.com/FuelLabs/fuels-rs/issues/1246. 

Once https://github.com/FuelLabs/sway/pull/5306 is done, we should be able to start adding support to this new encoding scheme (which can be enabled with `forc`'s `--experimental-new-encoding` feature flag).

Issue #1151, opened 2023-10-04
Decoding string slices from logs


hal3e:
String slices can not be decoded by the SDK when they are logged. Issue to track on the sway side: https://github.com/FuelLabs/sway/issues/5110

If fixed, remove the error from:
```rust
    fn can_decode_log_with_type<T: Parameterize>() -> Result<()> {
        match T::param_type()
            ParamType::StringSlice => Err(error!(
                InvalidData,
                "String slices can not be decoded from logs. Convert the slice to `str[N]` with `__to_str_array`"
            )),
            _ => Ok(()),
        }
    }
```

hal3e:
This issue is resolved in https://github.com/FuelLabs/fuels-rs/pull/1259

Issue #901, opened 2023-03-22
feat: enable `Bytes` logging in contracts


iqdecay:

digorithm:
Blocked by https://github.com/FuelLabs/sway/pull/4497.

hal3e:
This issue is resolved in https://github.com/FuelLabs/fuels-rs/pull/1259

Issue #898, opened 2023-03-22
feat: enable `Vec` logging in contracts


iqdecay:

hal3e:
This issue is resolved in https://github.com/FuelLabs/fuels-rs/pull/1259

PR #1259, opened 2024-01-23
feat!: add support for experimental encoding in logs

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/1247, https://github.com/FuelLabs/fuels-rs/issues/1151, https://github.com/FuelLabs/fuels-rs/issues/901, https://github.com/FuelLabs/fuels-rs/issues/898

This PR adds support for sway's experimental encoding in logs. The new encoder is enabled with the `experimental` rustflag.

To run the tests, first build the forc projects with: 
 `forc build --path packages/fuels --experimental-new-encoding` then run the tests with:
 `RUSTFLAGS='--cfg experimental' cargo test --test logs`
 
 What was done:
 - added `ExperimentalBoundedDecoder`
 - updated how `RawSlice` was  encoded and decoded
 - added tests for the new encoding
 - add new CI step that runs tests with the new `experimental` flag
 
 BREAKING CHANGE:
 - change `RawSlice` encoding and decoding

### Checklist


packages/fuels/tests/logs/script_logs/src/main.sw
```diff
@@ -2,6 +2,15 @@ script;
 
 use std::logging::log;
 
+impl AbiEncode for (TestStruct, TestEnum) {
```
MujkicA, line 5: I assume this will be auto implemented by the compiler at some point. Can we we have a PR to track the removal of these?

hal3e: https://github.com/FuelLabs/fuels-rs/issues/1260

MujkicA (approved):
LGTM

packages/fuels-core/src/codec/abi_decoder/bounded_decoder.rs
```diff
@@ -36,7 +35,7 @@ impl BoundedDecoder {
         }
     }
 
-    pub fn decode(&mut self, param_type: &ParamType, bytes: &[u8]) -> Result<Token> {
+    pub(crate) fn decode(&mut self, param_type: &ParamType, bytes: &[u8]) -> Result<Token> {
```
digorithm, line 38: Why are we making it public only to this crate? I'm not sure I recall this correctly, but I thought some people were using this function directly, e.g.: indexer. If that's not the case then it's all good

digorithm (approved):
Excellent work -- excited to see this new ABI decoding in action 😍 

Left a minor question but it shouldn't be a blocker.

packages/fuels-core/src/codec/abi_decoder/bounded_decoder.rs
hal3e: I think this was just a lapsus. `BoundedDecoder` is `pub(crate)` and not exposed so this function would not be reachable.

Br1ght0ne (approved):

Issue #1195, opened 2023-11-08
Predicate validation using the VM within the SDK


digorithm:
Similar to dry runs.

PR #1286, opened 2024-02-29
feat: validate predicates using the VM before sending out transaction

iqdecay:
This PR normally closes (correctly this time I hope) #1195 by adding predicates
validation within the SDK.
- This means that now the SDK can detect if an invalid predicate transaction is
  about to be sent to the node and prevents it.


segfault-magnet (dismissed):
A lot smaller than expected. 

Nicely done :)

MujkicA (approved):

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -445,6 +463,21 @@ impl EstimablePredicates for CreateTransaction {
     }
 }
 
+impl ValidablePredicates for CreateTransaction {
```
hal3e, line 466: Because the implementations for `CreateTransaction` and `ScriptTransaction` are the same, I would suggest we put it in the macro `impl_tx_wrapper` like so:
```rust
        impl ValidablePredicates for $wrapper {
            fn validate_predicates(
                self,
                consensus_parameters: &ConsensusParameters,
                block_height: u32,
            ) -> Result<()> {
                let checked = self
                    .tx
                    .into_checked(block_height.into(), consensus_parameters)?;
                let check_predicates_parameters: CheckPredicateParams = consensus_parameters.into();
                checked.check_predicates(&check_predicates_parameters)?;

                Ok(())
            }
        }
```

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -488,6 +521,21 @@ impl EstimablePredicates for ScriptTransaction {
     }
 }
 
+impl ValidablePredicates for ScriptTransaction {
```
hal3e, line 524: see comment above

hal3e (changes requested):
Nice! Left a nit regarding code duplication.

segfault-magnet (dismissed):

MujkicA (approved):

Salka1988 (approved):

hal3e (approved):

PR #1291, opened 2024-03-04
feat!: add `decode_as_debug_str` to `ABIDecoder`

hal3e:
PR: https://github.com/FuelLabs/fuels-rs/pull/690 updated `ParamType` and added names for `struct` and `enum` types and their fields/variants. This was later removed in PR: https://github.com/FuelLabs/fuels-rs/pull/885 as we completely changed how log decoding is done. 

This PR returns struct and enum names (also fields/variants) so that users can use `ParamTypes` at runtime to debug log or return receipts. In addition, users are able to go directly from `ProgramABI` and some data to decoded debug. Here is an example how this is used:

```rust
// Read the abi json
let json_abi_file =
    "../../packages/fuels/tests/types/contracts/generics/out/debug/generics-abi.json";
let abi_file_contents = std::fs::read_to_string(json_abi_file)?;

// Parse into `ProgramABI`
let parsed_abi: ProgramABI = serde_json::from_str(&abi_file_contents)?;

// Create type lookup (id, TypeDeclaration)
let type_lookup = parsed_abi
    .types
    .into_iter()
    .map(|decl| (decl.type_id, decl))
    .collect::<HashMap<_, _>>();

// Helper function to extract first argument from a contract function
let get_first_fn_argument = |fn_name: &str| {
    parsed_abi
        .functions
        .iter()
        .find(|abi_fun| abi_fun.name == fn_name)
        .expect("should be there")
        .inputs
        .first()
        .expect("should be there")
};

// Get TypeApplication
let type_application = get_first_fn_argument("struct_w_generic");
// Get corresponding ParamType
let param_type = ParamType::try_from_type_application(type_application, &type_lookup)?;
// Get debug output (String)
let debug_output =  ABIDecoder::default().decode_as_debug_str(&param_type, &[0, 0, 0, 0, 0, 0, 0, 123])?;
```
There are actually not so many changes but I refactored the `param_types.rs` file into a new module with several files to separate the logical units and their tests.

BREAKING CHANGE: `EnumVariants` are now imported through `param_types::EnumVariants`

### 


packages/fuels-core/src/types/param_types/decodable_debug.rs
```diff
@@ -0,0 +1,241 @@
...
+    codec::ABIDecoder,
+    types::{
+        errors::{error, Result},
+        param_types::ParamType,
+        Token,
+    },
+};
+
+impl ParamType {
+    pub fn decode_debug(&self, bytes: &[u8]) -> Result<String> {
```
segfault-magnet, line 13: I'd move this to the decoder for two reasons:
1. It is not expected of the ParamType to decode but rather to describe a type
2. If moved to the decoder then you wouldn't have to make the decision to use the default decoder -- i.e. the user chooses the decoder. 

So something like
`decoder.decode_as_debug_str(param_type, bytes)?`

segfault-magnet (commented):
The pair of `(String, ParamType)` comes up a lot. 

Wdyt about giving it a name? Maybe `Member` so that it can refer both to struct fields and enum variants.

kayagokalp:
This is great! I think it would be very useful for my tooling side decoding needs. Would it possible to also encode similarly given a param type and value to encode?

segfault-magnet (approved):

digorithm (approved):
Great work! This is useful enough, in my opinion, to have a small section in our docs. Thoughts? It can be done in a follow-up, not necessarily in this PR.

MujkicA (approved):

hal3e:
> Great work! This is useful enough, in my opinion, to have a small section in our docs. Thoughts? It can be done in a follow-up, not necessarily in this PR.

Here is the issue for that: https://github.com/FuelLabs/fuels-rs/issues/1296

Issue #1311, opened 2024-03-26
Dynamic base asset ID


digorithm:
Soon, we'll move to a more dynamic way of loading/interacting with the base `AssetID`. This change will affect the Fuel VM, the client, and SDKs.

These are some of the _known_ the changes that need to be done in the SDK: 

## Remove `BASE_ASSET_ID`

```rust
// Bytes representation of the asset ID of the "base" asset used for gas fees.
- pub const BASE_ASSET_ID: AssetId = AssetId::BASE;
```

## Add `base_asset_id()` to `Provider` impl

The `Provider` struct will already pre-fetch the base asset ID when `connect()` is called. A simple `base_asset_id()` function should be added to make this value available.

## Replace `BASE_ASSET_ID` in codebase

```rust
    async fn adjust_for_fee<Tb: TransactionBuilder + Sync>(
        &self,
        tb: &mut Tb,
        used_base_amount: u64,
    ) -> Result<()> {
        let missing_base_amount =
            calculate_missing_base_amount(tb, used_base_amount, self.try_provider()?).await?;

        if missing_base_amount > 0 {
            let new_base_inputs = self
+               .get_asset_inputs_for_amount(self.try_provider()?.base_asset_id(), missing_base_amount)
-               .get_asset_inputs_for_amount(BASE_ASSET_ID, missing_base_amount)
                .await?;

            adjust_inputs_outputs(tb, new_base_inputs, self.address());
        };

        Ok(())
    }
```

Note that there might be other places where we're pulling the base asset ID. Part of the work here is scanning the codebase for it, and making sure we're dynamically pulling the base asset ID.

PR #1313, opened 2024-03-28
feat!: get base `AssetId` from provider

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/1311

This PR removes the `BASE_ASSET_ID` constant and adds a new `provider` endpoint `base_asset_id` which should be used instead. 

Tests that used `BASE_ASSET_ID` now use `AssetId::zeroed()`.
`ResourceFilter` now uses `Option<AssetId>`. If the `AssetId` is not set the provider will use the `base_asset_id`
`CallParameters` now uses `Option<AssetId>`. If the `AssetId` is not set the provider will use the `base_asset_id`

BREAKING CHANGE:
- removed `BASE_ASSET_ID` constant
- `prepare_message_to_output` now requires a base asset id.
- `CoinType` `asset_id` method is renamed to `coin_asset_id` and now returns `Option<AssetId>` and is `None` for `CoinType::Message`.
- `Transaction` `used_coins` now requires a base asset id.
- removed `asset_id` from `Input`
- `ResourceFilter` now uses `Option<AssetId>`. If the `AssetId` is not set the provider will use the `base_asset_id`
- `CallParameters` now uses `Option<AssetId>`. If the `AssetId` is not set the provider will use the `base_asset_id`

### Checklist


segfault-magnet (dismissed):

examples/contracts/src/lib.rs
```diff
@@ -668,7 +668,11 @@ mod tests {
         let _ = contract_instance
             .methods()
             .initialize_counter(42)
-            .add_custom_asset(BASE_ASSET_ID, amount, Some(other_wallet.address().clone()))
+            .add_custom_asset(
+                AssetId::default(),
```
MujkicA, line 672: Nit: Since it's in the docs. It's a bit more descriptive and it might be useful to familiarize users with it.

```suggestion
                AssetId::BASE,
```

examples/wallets/src/lib.rs
```diff
@@ -138,7 +138,7 @@ mod tests {
             .transfer(wallets[1].address(), 1, asset_id, TxPolicies::default())
             .await?;
 
-        let wallet_2_final_coins = wallets[1].get_coins(BASE_ASSET_ID).await?;
+        let wallet_2_final_coins = wallets[1].get_coins(AssetId::default()).await?;
```
MujkicA, line 141:
```suggestion
        let wallet_2_final_coins = wallets[1].get_coins(AssetId::BASE).await?;
```

examples/wallets/src/lib.rs
```diff
@@ -271,7 +271,7 @@ mod tests {
         let mut rng = rand::thread_rng();
 
         let asset_base = AssetConfig {
-            id: BASE_ASSET_ID,
+            id: AssetId::default(),
```
MujkicA, line 274:
```suggestion
            id: AssetId::BASE,
```

examples/wallets/src/lib.rs
```diff
@@ -317,16 +317,13 @@ mod tests {
...
-                DEFAULT_NUM_COINS,
-            },
+            prelude::{launch_provider_and_get_wallet, DEFAULT_COIN_AMOUNT, DEFAULT_NUM_COINS},
             types::AssetId,
         };
 
         let wallet = launch_provider_and_get_wallet().await?;
         // ANCHOR: get_asset_balance
-        let asset_id: AssetId = BASE_ASSET_ID;
+        let asset_id = AssetId::default();
```
MujkicA, line 326:
```suggestion
        let asset_id = AssetId::BASE;
```

packages/fuels-accounts/src/accounts_utils.rs
```diff
@@ -45,11 +44,19 @@ pub async fn calculate_missing_base_amount(
     Ok(missing_amount)
 }
 
-fn available_base_amount(tb: &impl TransactionBuilder) -> u64 {
+fn available_base_amount(tb: &impl TransactionBuilder, base_asset_id: &AssetId) -> u64 {
     tb.inputs()
         .iter()
-        .filter_map(|input| match (input.amount(), input.asset_id()) {
```
MujkicA, line 51: Looks like we were previously skipping over messages, nice catch

packages/fuels-accounts/src/provider.rs
```diff
@@ -98,6 +96,7 @@ impl ResourceQueries {
     }
 }
 
+#[derive(Default)]
```
MujkicA, line 99: Maybe this is not the behaviour we want. Using default for AssetId might be different than the asset id configured in consensus parameters.
My expectation would be that the default always filters for the base asset.

packages/fuels-core/src/types/wrappers/coin_type.rs
```diff
@@ -47,10 +44,10 @@ impl CoinType {
         }
     }
 
-    pub fn asset_id(&self) -> AssetId {
+    pub fn asset_id(&self) -> Option<AssetId> {
```
MujkicA, line 47: In this case, can we remove `asset_id()` from `wrappers/input.rs`? The assumption here is that a `CoinType` is either a coin with some asset or a message with the base asset so we can unwrap and default to base downstream.
This doesn't hold up if we call `asset_id()` on `Input` since it could also be a `Contract`. Seems like a potential footgun.

packages/fuels-programs/src/contract.rs
```diff
@@ -86,7 +86,7 @@ impl Default for CallParameters {
     fn default() -> Self {
         Self {
             amount: DEFAULT_CALL_PARAMS_AMOUNT,
-            asset_id: BASE_ASSET_ID,
+            asset_id: Default::default(),
```
MujkicA, line 89: Same concern as for ResourceFilter

packages/fuels-test-helpers/src/lib.rs
```diff
@@ -55,7 +52,7 @@ pub fn setup_multiple_assets_coins(
                 .expect("failed to fill with random data");
             random_asset_id
         })
-        .chain([BASE_ASSET_ID])
+        .chain([AssetId::default()])
```
MujkicA, line 55: Same concern here too

MujkicA (commented):
Looks good but I have one major concern.
`AssetId::BASE` and `default()` exist while there is no guarantee that they will overlap with the configured base asset.

examples/contracts/src/lib.rs
hal3e: I wanted to avoid using the name `BASE` because it could be misleading. I tried to stick to `default` in our tests as the focus is not on the asset ID.

packages/fuels-core/src/types/wrappers/coin_type.rs
hal3e: I wanted to do this and force users to match and then extract the asset_id. I needed some support :D I will do it now

packages/fuels-accounts/src/provider.rs
segfault-magnet: Also the default for asset id is going to go behind a "test-only" feature over in fuel-* crates. 
Maybe have it be an `Option` and we'll resolve it to the base asset later when querying? If we care that much about having a default.  We probably do, it seems that the ux would take a hit without being able to say `..Default::default()`

packages/fuels-core/src/types/wrappers/coin_type.rs
segfault-magnet, line 47: maybe call it `coin_asset_id`. it being the base asset isn't clear from returning `None`.

Or remove the helper altogether.

packages/fuels-programs/src/contract.rs
hal3e: This is now updated to use `Option<AssetId>`. If the `AssetId` is not set the provider will use `base_asset_id`

segfault-magnet (approved):

MujkicA (approved):

PR #1319, opened 2024-03-31
feat!: upgrade fuel-core to 24.2

MujkicA:
- With the state configuration now separate in StateConfig instead of being part of ChainConfig, I had to make some breaking changes. We agreed to keep these changes minimal until we revisit things for V1, so I went with breaking `FuelService::new` to keep `setup_test_provider` stable.

BREAKING CHANGE
- Renamed `Config` to `NodeConfig`
- `NodeConfig` no longer holds ChainConfig`
- fuel-core now separates the initial state config from `ChainConfig` in `StateConfig`
- as a consequence of the above changes, `FuelService::start()` now accepts `NodeConfig`, `ChainConfig` and `StateConfig` as arguments to startup a node.
- fuel-core has switched to `ConsensusParameters` being an enum to support versioning, when instantiating it, we need to make use of setters if we want to avoid committing to a specific version:

```rust
let mut consensus_parameters = ConsensusParameters::default();
consensus_parameters.set_tx_params(tx_params);
consensus_parameters.set_fee_params(fee_params);
```

Fields need to be accessed via methods: 

```rust
self.consensus_parameters.chain_id
// becomes
self.consensus_parameters.chain_id()
```

- The same applies to other parameter structs used when setting up the node, like `TxParameters`, `ContractParameters`, `PredicateParameters` etc.

- The `witness_index` parameters in `CreateTransactionBuilder::with_bytecode_witness_index` is now a `u16`
- `NodeInfo` no longer has `min_gas_price`
- `CreateTransaction` no longer has`bytecode_length()`

- `Header` no longer has message_receipt_root but gains:

```rust
pub message_outbox_root: Bytes32,
pub event_inbox_root: Bytes32,
pub consensus_parameters_version: u32,
pub state_transition_bytecode_version: u32
```

### Checklist


packages/fuels-test-helpers/src/node_types.rs
```diff
@@ -70,115 +64,6 @@ impl Default for Config {
...
-            utxo_validation: value.utxo_validation,
-            debug: value.debug,
-            block_production: value.block_production.into(),
-            chain_conf: value.chain_conf,
-            ..fuel_core::service::Config::local_node()
-        }
-    }
-}
-
-pub(crate) mod serde_hex {
```
MujkicA, line 100: I don't remember why this was added, but looks like it's not being used anywhere

Cargo.toml
```diff
@@ -72,21 +72,21 @@ which = { version = "6.0.0", default-features = false }
 zeroize = "1.7.0"
 
 # Dependencies from the `fuel-core` repository:
-fuel-core = { version = "0.23.0", default-features = false }
-fuel-core-chain-config = { version = "0.23.0", default-features = false }
-fuel-core-client = { version = "0.23.0", default-features = false }
-fuel-core-poa = { version = "0.23.0", default-features = false }
-fuel-core-services = { version = "0.23.0", default-features = false }
-fuel-core-types = { version = "0.23.0", default-features = false }
+fuel-core = { version = "0.24.1", default-features = false, features = ["test-helpers"] }
```
segfault-magnet, line 75: I'd have these features activated on a need-to-have basis. I.e. have the cargo workspace dep here only to make sure every crate uses (requires) the same fuel-core version but have each crate dictate its own needs as to the features of `fuel-core`. 

Cargo knows how to unify features when compiling a workspace.

packages/fuels-test-helpers/Cargo.toml
```diff
@@ -21,17 +21,13 @@ fuel-types = { workspace = true, features = ["random"] }
...
-serde_with = { workspace = true }
 tempfile = { workspace = true, default-features = false }
 tokio = { workspace = true, default-features = false }
 which = { workspace = true, default-features = false }
 
 [features]
 default = ["fuels-accounts", "std"]
 std = ["fuels-accounts?/std", "fuels-core/std"]
-fuel-core-lib = ["fuel-core"]
+fuel-core-lib = []
```
segfault-magnet, line 33: The original idea behind the feature is to speed up compilation by not having to compile the whole core.

Why is it necessary to always have the core dep present? Maybe you can use `fuel_core_chain_config` when configuring a snapshot for a binary?

packages/fuels-test-helpers/src/node_types.rs
segfault-magnet: This was probably from the time we encoded `StateConfig` entries manually. In any case, `SnapshotWriter` will take care of it all now.

packages/fuels/Cargo.toml
```diff
@@ -56,5 +56,4 @@ std = [
 ]
 # TODO: To be removed once https://github.com/FuelLabs/fuels-rs/issues/881 is unblocked.
 test-type-paths = []
-fuel-core-lib = ["fuels-test-helpers?/fuel-core-lib", "dep:fuel-core"]
-rocksdb = ["fuel-core?/rocksdb"]
+fuel-core-lib = ["fuels-test-helpers?/fuel-core-lib", "fuel-core"]
```
segfault-magnet, line 59: If you don't say `dep:` then you're implicitly making another feature flag called "fuel-core". 

Having both `fuel-core-lib` and `fuel-core` is confusing.

packages/fuels/tests/contracts.rs
```diff
@@ -1332,120 +1331,6 @@ async fn low_level_call() -> Result<()> {
     Ok(())
 }
 
-#[cfg(any(not(feature = "fuel-core-lib"), feature = "rocksdb"))]
```
segfault-magnet, line 1335: This test shouldn't be completely removed since we still support rocksdb as a database type but only when we're using the `fuel-core` as a binary.

packages/fuels/tests/providers.rs
```diff
@@ -308,6 +311,7 @@ async fn contract_deployment_respects_maturity() -> Result<()> {
     Ok(())
 }
 
+#[ignore]
```
segfault-magnet, line 314: ?

packages/fuels-test-helpers/Cargo.toml
MujkicA: yup, was gonna try that next

packages/fuels/tests/providers.rs
MujkicA: Forgot to reenable it.

MujkicA: Here we test all the remaining gas is forwarded to the call instruction if nothing else was specified by call params. 
The test is failing right now because we check :
`gas_forwarded` (gas that was automatically forwarded to CALL) + `gas_used_by_script` = `gas_limit` (set with TxParams)

I'm wondering, was this test was malformed? `gas_used_by_script` also includes gas used by CALL. The correct was to calculate it would be:
`gas_forwarded`+ (`gas_used_by_script` - `gas_used_by_call`) = `gas_limit`
where `gas_used_by_call` is impossible to retrieve afaik.

MujkicA:
Our rocksdb test is failing due to a bug on the core side. Disabling it for now

https://github.com/FuelLabs/fuels-rs/issues/1324

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -288,7 +287,7 @@ macro_rules! impl_tx_trait {
                     .any(|input| matches!(input, Input::ResourcePredicate { .. }))
             }
 
-            fn num_witnesses(&self) -> Result<u8> {
+            fn num_witnesses(&self) -> Result<u16> {
                 let num_witnesses = self.witnesses().len();
 
                 if num_witnesses + self.unresolved_signers.len() > 256 {
```
hal3e, line 293: This should be adjusted as the limit is not `256` anymore.

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -495,17 +495,13 @@ impl CreateTransaction {
...
-    pub fn bytecode_witness_index(&self) -> u8 {
+    pub fn bytecode_witness_index(&self) -> u16 {
         *self.tx.bytecode_witness_index()
     }
 
     pub fn storage_slots(&self) -> &Vec<StorageSlot> {
         self.tx.storage_slots()
     }
-
-    pub fn bytecode_length(&self) -> u64 {
```
hal3e, line 506: We should add a `BREAKING CHANGE` section and list all of the breaking changes there.

packages/fuels/tests/providers.rs
hal3e: I had a look. It seems that `gas_used_by_script` is a `static` cost of the script (maybe cost for bytecode size, witnesses, etc.). Here are the numbers for the non-experimental case:

```
    gas_used_by_script = 364
    gas_forwarded_to_call = 225_519
    total_gas_used = 796
```
`total_gas_used` should be the `static` part plus the call cost. The calculation 
`gas_used_by_script` + `gas_forwarded` = `gas_limit` makes sense. 

We should change the name of `gas_used_by_script` to reflect the fact that it is not the total cost.

packages/fuels/tests/contracts.rs
```diff
@@ -1331,6 +1336,119 @@ async fn low_level_call() -> Result<()> {
...
+    use fuels::{
+        accounts::wallet::WalletUnlocked,
+        client::{PageDirection, PaginationRequest},
+        crypto::SecretKey,
+        prelude::{setup_test_provider, DbType, Error, ViewOnlyAccount, DEFAULT_COIN_AMOUNT},
+    };
+
+    let temp_dir = tempfile::tempdir()
+        .expect("failed to make tempdir")
+        .into_path();
```
segfault-magnet, line 1353: This will stop the temp dir being deleted upon its handle being dropped. In cases of the test failing we might not execute the manual cleanup at the end of the test.

I suggest the handle is kept and its ownership moved into the future so that it may be dropped no matter how the test finishes.

segfault-magnet (approved):

Salka1988 (approved):

PR #1337, opened 2024-04-16
fix: receipt parsing when call is repeated through internal call

### LINKED ISSUE #1333: Erroneous result value on reentrant calls
`ReceiptParser` makes the wrong assumption that there is only a single Return receipt per contract id. This is not the case for reentrant calls.

For instance, executing the following codebase tests (with `forc build && cargo test`: https://github.com/IGI-111/reentrancy-repro

produces the following receipts:
```
[tests/harness.rs:43:5] res.receipts = [
    Call {
        id: 0x0000000000000000000000000000000000000000000000000000000000000000,
        to: 0xb7cac0a0521248f8baac940a0902bfd7fdc5b6bc1936b59a801df3b4a75483c8,
        amount: 0,
        asset_id: 0x0000000000000000000000000000000000000000000000000000000000000000,
        gas: 456,
        param1: 1619450810,
        param2: 10480,
        pc: 11960,
        is: 11960,
    },
    Call {
        id: 0xb7cac0a0521248f8baac940a0902bfd7fdc5b6bc1936b59a801df3b4a75483c8,
        to: 0x07822be5596a1039c73a41ffc1b8aa0b3cc0c9ffb98ebc563961b730f2e79795,
        amount: 0,
        asset_id: 0x0000000000000000000000000000000000000000000000000000000000000000,
        gas: 238,
        param1: 3752339489,
        param2: 10480,
        pc: 13168,
        is: 13168,
    },
    Call {
        id: 0x07822be5596a1039c73a41ffc1b8aa0b3cc0c9ffb98ebc563961b730f2e79795,
        to: 0xb7cac0a0521248f8baac940a0902bfd7fdc5b6bc1936b59a801df3b4a75483c8,
        amount: 0,
        asset_id: 0x0000000000000000000000000000000000000000000000000000000000000000,
        gas: 64,
        param1: 1003047684,
        param2: 0,
        pc: 14320,
        is: 14320,
    },
    Return {
        id: 0xb7cac0a0521248f8baac940a0902bfd7fdc5b6bc1936b59a801df3b4a75483c8,
        val: 1337,
        pc: 14388,
        is: 14320,
    },
    Return {
        id: 0x07822be5596a1039c73a41ffc1b8aa0b3cc0c9ffb98ebc563961b730f2e79795,
        val: 1337,
        pc: 13304,
        is: 13168,
    },
    Return {
        id: 0xb7cac0a0521248f8baac940a0902bfd7fdc5b6bc1936b59a801df3b4a75483c8,
        val: 42,
        pc: 12128,
        is: 11960,
    },
    Return {
        id: 0x0000000000000000000000000000000000000000000000000000000000000000,
        val: 1,
        pc: 10388,
        is: 10368,
    },
    ScriptResult {
        result: Success,
        gas_used: 646,
    },
]
```

The return value of this call with these receipts should be 42, as this is the original call done at `is: 11960`, however since there are two return receipts for contract `0xb7cac0a0521248f8baac940a0902bfd7fdc5b6bc1936b59a801df3b4a75483c8`, `ReceiptParser` picks the wrong one and the return value is erroneously considered to be 1337.

MujkicA:
Closes #1333 

When retrieving return data from receipts, our `ReceiptParser` was matching on the first return receipt that had a matching `ContractId`. This fails in the following case:
Some script calls contract A which internally calls contract B, then we call B from the script. The receipts then include the returns for B,A,B in that order, where the first B is the internal call.
When parsing receipts for B, we will match on the returns from the internal call instead of the one from our script.

The solution implemented here relies on a stack that will push an item whenever we encounter `Receipt::Call` and pop on `Receipt::ReturnData`. When the stack is empty, the last popped receipt is always the return for a top level call.

Co-developed with @Salka1988 

### Checklist


Salka1988 (dismissed):

packages/fuels-programs/src/receipt_parser.rs
```diff
@@ -120,6 +134,26 @@ impl ReceiptParser {
         None
     }
 
+    #[cfg(not(feature = "legacy_encoding"))]
+    fn extract_return_data(&mut self) -> Option<Vec<u8>> {
```
digorithm, line 138: This is a bit confusing for anyone without proper context. Can you write some comments here explaining why we're doing this?

digorithm (changes requested):

packages/fuels-programs/src/receipt_parser.rs
MujkicA: @hal3e is working on PR to remove the legacy encoding. I duplicated some methods here and hid them behind the legacy/not legacy flag so that @hal3e can have an easier time refactoring.

digorithm (dismissed):

packages/fuels-programs/src/receipt_parser.rs
MujkicA: I've updated to hold two separate implementations of the `ReceiptParser`

segfault-magnet (dismissed):

segfault-magnet (approved):

hal3e (approved):

digorithm (approved):

PR #1344, opened 2024-04-24
chore: bump `rust` to `1.77.2`

hal3e:


Cargo.toml
```diff
@@ -35,7 +35,7 @@ homepage = "https://fuel.network/"
 readme = "README.md"
 license = "Apache-2.0"
 repository = "https://github.com/FuelLabs/fuels-rs"
-rust-version = "1.76.0"
+rust-version = "1.77.2"
```
xgreenx, line 38: Is that possible to not update to the latest rust if you need new features?=D It makes development sometimes hard=)

hal3e: yes no problem. We need to agree how we are going to approach updating `rust` in the future :)

PR #1348, opened 2024-04-29
feat: version incompatibility is now a soft warning

segfault-magnet:
Version incompatibility is now a soft error appended to every provider error.
### Checklist

### DIFF
--- a/packages/fuels-accounts/src/provider/retryable_client.rs
+++ b/packages/fuels-accounts/src/provider/retryable_client.rs
@@ -17,10 +17,12 @@ use fuels_core::types::errors::{error, Error, Result};
 
 use crate::provider::{retry_util, RetryConfig};
 
+use super::supported_versions::{self, VersionCompatibility};
+
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum RequestError {
-    #[error(transparent)]
-    IO(#[from] io::Error),
+    #[error("io error: {0}")]
+    IO(String),
 }
 
 type RequestResult<T> = std::result::Result<T, RequestError>;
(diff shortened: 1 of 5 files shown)


digorithm (dismissed):
This is a breaking change IIUC, right?

segfault-magnet:
> This is a breaking change IIUC, right?

Not sure. No from an API perspective (we still return results in affected places). We've changed only crate internal signatures. 

From a behavior perspective, yes. If somebody relied on a hard "no" when connecting to an incompatible node this will no longer happen.

digorithm (approved):

Salka1988 (approved):

xgreenx (approved):

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -17,10 +17,12 @@ use fuels_core::types::errors::{error, Error, Result};
 
 use crate::provider::{retry_util, RetryConfig};
 
+use super::supported_versions::{self, VersionCompatibility};
+
 #[derive(Debug, thiserror::Error)]
 pub(crate) enum RequestError {
-    #[error(transparent)]
-    IO(#[from] io::Error),
+    #[error("io error: {0}")]
```
hal3e, line 24: Small nit. I would not add `io error:` here. All errors coming from the provider will now have `io::error` prepended to the actual error message like this: `provider: io error: <some error>`

hal3e (approved):

packages/fuels-accounts/src/provider/retryable_client.rs
segfault-magnet: Makes sense. Can we make this a follow up? I dislike this being part of the error at all. I'd rather make it a `log::warn` but we don't log anything else and we should probably start and document that we do and that users should enable logging in their apps in order to see the info. Maybe even expose the version compat checking fn so that users can still make this into a hard error if they want.

hal3e: Sounds good to me.

PR #1389, opened 2024-05-23
feat: make `setup_program_test!` macro hygienic

### LINKED ISSUE #1376: `setup_program_test` is unhygenic
Issues ranging from the use of `::rand` (i.e. cannot call the macro if you don't have the dep) to assuming `fuels` types are imported in the macro scope.

`::rand` (and other deps if applicable) should either be reexported (research risks to semver breaking changes) or any call to them should be delegated to a fn exported in `fuels` so that the `rand` dep is hidden behind the fn call.

Br1ght0ne:
Close #1376.

- I opted for the "hide dependency behind function" approach

### Checklist

### DIFF
--- a/packages/fuels-code-gen/src/program_bindings/abigen/bindings/script.rs
+++ b/packages/fuels-code-gen/src/program_bindings/abigen/bindings/script.rs
@@ -110,13 +110,10 @@ fn expand_fn(fn_abi: &FullABIFunction) -> Result<TokenStream> {
     let arg_tokens = generator.tokenized_args();
     let body = quote! {
             let encoded_args = ::fuels::core::codec::ABIEncoder::new(self.encoder_config).encode(&#arg_tokens);
-            let provider = ::fuels::accounts::ViewOnlyAccount::try_provider(&self.account).expect("Provider not set up")
-                .clone();
             ::fuels::programs::script_calls::ScriptCallHandler::new(
                 self.binary.clone(),
                 encoded_args,
                 self.account.clone(),
-                provider,
                 self.log_decoder.clone()
             )
     };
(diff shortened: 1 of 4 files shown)


segfault-magnet (commented):
Luv it. 

Just one worry. 

If we rely on the availability of `generate_random_salt` in `fuels` then we might have an issue with feature gating:

`setup_program_test` is available even when `generate_random_salt` is not (due to fuels-test-helpers feature being off). 

So in such a case it wouldn't be able to get the salt helper.

Maybe we should feature-gate this macro as well? It probably uses a bunch more stuff from the test helpers anyway.

packages/fuels/src/lib.rs
```diff
@@ -72,6 +72,7 @@ pub mod prelude {
             codec::{LogDecoder, LogId, LogResult},
             traits::Signer,
         },
+        macros::setup_program_test,
```
segfault-magnet, line 75: `std` feature doesn't imply `fuels-test-helpers` so having it here isn't a guarantee that it will have all it needs to work (i.e. what is gated behind `fuels-test-helpers`).

I suggest a `cargo new` with `fuels` as a dep with `default-dependencies=false` then try and see if a setup program test macro will fail to compile under various feature combos.

segfault-magnet: Ah, I'm sorry it actually does imply the test helpers :S 

saw
```
  "fuels-test-helpers?/std",
```
and went.. ok that's a problem

but above it we have
```
  "dep:fuels-test-helpers",
```

which guarantees the dep. 

Weird way to write it but :D

segfault-magnet (approved):

hal3e (approved):
Nice! I have only one question. Does it make sense that the `std` flag activates `test_helpers` ? For me it seems a bit counterintuitive.

hal3e:
> > Does it make sense that the `std` flag activates `test_helpers` ?
> 
> @hal3e Currently the `dep:fuels-test-helpers` is being enabled solely by the `std` flag. I decided to plug into the existing setup instead of introducing a new flag. I can see how breaking a `test_helpers` flag can be too

We can leave it as is in this PR  but I would add a new feature `test_helpers` and make it default. `std` would not activate `test_helpers` but `test_helpers` would activate `std`.

MujkicA (approved):

Issue #1380, opened 2024-05-16
Improve tx dependency estimation


MujkicA:
Currently, `estimate_tx_dependencies` initiates a simulation for every dependency it needs to resolve.

@segfault-magnet suggested:
To reduce the number of simulations performed, we can add a high number of variable outputs to the tx before simulation and then inspect it afterwards to see how many were used.

Additionally, we'll have to watch the tx size limit. We can only add so many variable outputs until the limit is reached, then if the tx still reverts with missing outputs we have to communicate that there is no way of constructing a valid transaction.

PR #1393, opened 2024-05-30
feat!: var output estimation optimization

segfault-magnet:
closes: #1380 
The number of variable outputs is now determined by filling the dry-run tx with as much variable outputs as the node will allow (`consensus_params.tx_params.max_ouputs` - `whatever other outputs are already present`) and counting the number of `TransferOut` receipts made by the dry run. 

I've checked with the core devs, ideally we'd be looking at the `tx` outputs themselves but since the dry run endpoint doesn't return a `tx` this is the next best thing since only `TRO` instructions require a variable output attow. 

### Breaking changes

- `.append_variable_outputs(n)` is replaced by `.with_variable_output_policy(VariableOutputPolicy::Exactly(n))`.

- `.estimate_tx_dependencies(max_attempts)` has been replaced with `.determine_missing_contracts(max_attempts)` and will only add the missing contract inputs and no longer estimate variable outputs. To estimate variable outputs you should now add `.with_variable_output_policy(VariableOutputPolicy::EstimateMinimum)`

### Checklist


digorithm (changes requested):
I spent quite some time debating whether the new API names are good or intuitive. After a while, they started growing on me, hah. I'm curious to hear everyone's opinions on these two new names. 

Great job on the actual solution; it's a very clever trick! 

I just left a small, doc-related suggestion. Other than that, lgtm.

packages/fuels-programs/src/script_calls.rs
```diff
@@ -124,6 +119,14 @@ where
         self
     }
 
+    pub fn with_variable_output_policy(
```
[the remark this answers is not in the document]
segfault-magnet: [added ](https://github.com/FuelLabs/fuels-rs/pull/1393/commits/77a04a730a256a03b3292b60b5926f14ad487efa)

Br1ght0ne (dismissed):

hal3e (dismissed):
Awesome work! I really love the refactoring especially `ScriptDryRunner`! Thanks!

hal3e:
Regarding the naming: `with_variable_output_policy` is good. However, `determine_missing_contracts` could be misleading because we will `determine` them and add them to the outputs. So maybe something like `fix_missing_contracts` or `determine_and_update_missing_contracts` ...

digorithm:
> Regarding the naming: `with_variable_output_policy` is good. However, `determine_missing_contracts` could be misleading because we will `determine` them and add them to the outputs. So maybe something like `fix_missing_contracts` or `determine_and_update_missing_contracts` ...

Yeah, I was (and maybe still am) on the fence about `determine_missing_contracts`. Maybe just `update_missing_contracts`? `update` indicates it will mutate the inner stuff and also does some magic underneath.

segfault-magnet:
Thing is previously it was called "estimate_tx_dependencies" that also didn't have anything claiming mutability. 

I share the concern but have no satisfactory name that isn't a kilometer long.

maybe `discover_missing_contract_inputs` ?

packages/fuels-core/src/types/transaction_builders/script_dry_runner.rs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            tx.inputs_mut().push(fake_input);
+
+            // Add an empty `Witness` for the `coin_signed` we just added
+            tx.witnesses_mut().push(Witness::default());
+            tx.set_witness_limit(tx.witness_limit() + WITNESS_STATIC_SIZE as u64);
+        }
+    }
+
+    fn needs_fake_base_input(
```
MujkicA, line 83: There exists a tx validity condition that requires spendable inputs to be present:
https://github.com/FuelLabs/fuel-vm/blob/898789cc58ce549a92550d680df96fa0d0480002/fuel-tx/src/transaction/validity/error.rs#L15

I didn't come across anything that indicates that there needs to be base asset inputs. Running the tests with the base asset id check removed, seems to work fine.

segfault-magnet: This is inherited. @hal3e do you remember the original reason why we've added the base asset req?

segfault-magnet: @MujkicA I've tested it and it seems you're right, dry runs are passing without any base assets in inputs.

Something must have changed in the meantime @hal3e . 

In any case I've removed the requirement of the asset being of the base id.

Br1ght0ne (approved):

MujkicA (approved):

Salka1988 (approved):

hal3e (approved):

PR #1403, opened 2024-05-31
bug: fix cargo publish

segfault-magnet:
using `include!` with paths that lead out of the current crate will lead to `cargo publish` failures since, when publishing, you only have the current crate code available.

Workaround is to have `version.rs` now live near the `Provider` and the fuel version checker will have a cargo dep on `fuels-accounts` to get it, as will the `e2e` tests. 

Even though `e2e` tests aren't published, I've updated them as well to use `fuels-accounts` even though it may add a few seconds to the build stage.

### Checklist
- [X] I have added necessary labels.
- [X] I have done my best to ensure that my PR adheres to [the Fuel Labs Code Review Standards](https://github.com/FuelLabs/rfcs/blob/master/text/code-standards/external-contributors.md).
- [X] I have requested a review from the relevant team or maintainers.


digorithm (dismissed):
Thanks for handling this so quickly!

digorithm (approved):

hal3e (approved):

PR #1404, opened 2024-06-03
feat(macros): add `SetOptions { profile }` command

### LINKED ISSUE #1360: Add build profile option to`setup_program_test!`
> change the macro to have something like `setup_program_test!(Options(profile="release"), Deploy(contract="MyContract, wallet...`

_Originally posted by @segfault-magnet in https://github.com/FuelLabs/fuels-rs/pull/1275#discussion_r1588384880_

Br1ght0ne:
Close #1360.

The command is optional, and defaults to `release` profile.

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,6 +68,7 @@ serde = { version = "1.0.193", default-features = false }
 serde_json = "1.0.108"
 serde_with = { version = "3.4.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
+strum = "0.26.2"
 syn = "2.0.39"
 tai64 = { version = "4.0.0", default-features = false }
 tar = { version = "0.4", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -68,7 +68,6 @@ serde = { version = "1.0.193", default-features = false }
 serde_json = "1.0.108"
 serde_with = { version = "3.4.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
-strum = "0.26.2"
 syn = "2.0.39"
 tai64 = { version = "4.0.0", default-features = false }
 tar = { version = "0.4", default-features = false }
--- a/packages/fuels-macros/src/setup_program_test/parsing/commands/set_options.rs
+++ b/packages/fuels-macros/src/setup_program_test/parsing/commands/set_options.rs
@@ -1,16 +1,39 @@
-use std::convert::TryFrom;
+use std::{convert::TryFrom, fmt, str::FromStr};
 
-use strum::{Display, EnumString};
 use syn::Error;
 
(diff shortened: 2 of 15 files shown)


Cargo.toml
```diff
@@ -68,6 +68,7 @@ serde = { version = "1.0.193", default-features = false }
 serde_json = "1.0.108"
 serde_with = { version = "3.4.0", default-features = false }
 sha2 = { version = "0.10.8", default-features = false }
+strum = "0.26.2"
```
segfault-magnet, line 71:
```suggestion
strum = { version="0.26.2", default-features = false }
```
The only place where default features can be disabled in a workspace dep in a workspace cargo project is here. Anywhere else is ignored by cargo.

So the only way to be of the feature-minimal mindset is to disable default features here and only enable them where needed. (even though it might only have this one usage rn).

segfault-magnet (commented):
Short and sweet, plugged in nicely at all the right places. 

I'd remove `strum` and implement the conversion to string manually.

If it were a binary I wouldn't nitpick, but wouldn't like the dep to propagate to all users of the sdk because of "debug" and "release". Besides an extra dep, our version choice (whatever it is) might also cause headaches to any present `strum` dep.

Cargo.toml
Br1ght0ne: Thank you for the comment, I will remember it for the future. I removed the dep completely as per above though. :)

segfault-magnet (approved):
I wish there was an easy way to test this (this - i.e. that debug is going to use the debug build). An e2e would require compiling both debug and release builds, and that might be overkill.

MujkicA (approved):

packages/fuels-macros/src/setup_program_test/parsing/commands/set_options.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl FromStr for BuildProfile {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "debug" => Ok(Self::Debug),
+            "release" => Ok(Self::Release),
+            _ => Err(r#"invalid build profile option: must be "default" or "release""#),
```
hal3e, line 21:
```suggestion
            _ => Err(r#"invalid build profile option: must be "debug" or "release""#),
```

hal3e (changes requested):
Nicely done! I would add 2 ui tests. One where we use `Option` with something other than `profile` and another one when `profile` is something other than `debug` or `release`

MujkicA (approved):

digorithm (approved):
Great job -- nice UI tests, too!

Salka1988 (approved):

hal3e (approved):
Great work!

Issue #1405, opened 2024-06-03
feat: Support read only contract calls without owning coins


MujkicA:
Currently, calling `simulate()` on a contract instance requires a wallet that holds some coins. Those are needed for the transaction to be valid but are not spent otherwise. We can support read-only calls by attaching fake coins and simulating without utxo-validation.

PR #1412, opened 2024-06-06
feat!: unfunded read only calls

segfault-magnet:
closes: #1405 

A transaction builder can be configured with a `BuildStrategy` to tweak how `build()` is going to work. 
```rust
pub enum ScriptBuildStrategy {
    /// Transaction is estimated and signatures are automatically added.
    #[default]
    Complete,
    /// Transaction is estimated but no signatures are added.
    /// Building without signatures will set the witness indexes of signed coins in the
    /// order as they appear in the inputs. Multiple coins with the same owner will have
    /// the same witness index. Make sure you sign the built transaction in the expected order.
    NoSignatures,
    /// No estimation is done and no signatures are added. Fake coins are added if no spendable inputs
    /// are present. Meant only for transactions that are to be dry-run with validations off.
    /// Useful for reading state with unfunded accounts.
    StateReadOnly,
}
```

Simulating a call now requires the user to tweak how the simulated call will be executed:
```rust
/// Used to control simulations/dry-runs
pub enum Execution {
    /// The transaction will be subject to all validations -- the tx fee must be covered, witnesses
    /// and UTXOs must be valid, etc.
    #[default]
    Realistic,
    /// Most validation is disabled. Witnesses are replaced with fake ones, fake base assets are
    /// added if necessary. Unused outputs are filled with variable outputs. Useful for fetching state without needing an account
    /// with base assets.
    StateReadOnly,
}
```
And that allows for actual read only calls
```rust
    let no_funds_wallet = WalletUnlocked::new_random(provider);

    let response = MyContract::new(contract_id, no_funds_wallet.clone())
        .methods()
        .get(5, 6)
        .simulate(Execution::StateReadOnly)
        .await?;

```

# Breaking changes
* `SizedAsciiString` no longer implements `AsRef<[u8]>`. To get the underlying bytes you can turn it into a `&str` via the new `AsRef<str>` and call `as_bytes()` on the `&str`: `sized_string.as_ref().as_bytes()`. 
* `build_without_signatures` is now achieved by setting the build strategy to `BuildStrategy::NoSignatures` on the transaction builder before calling `build`.
* `.simulate()` now accepts an `Execution` argument allowing for realistic or state-read-only simulations.  
### Checklist


Br1ght0ne (dismissed):
ship it (to dev) :shipit:

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -53,30 +53,48 @@ struct UnresolvedWitnessIndexes {
...
+    async fn build(self, provider: impl DryRunner, context: Self::Context) -> Result<Self::TxType>;
+}
+
+impl sealed::Sealed for ScriptTransactionBuilder {}
 
+#[derive(Debug, Clone, Default)]
+pub enum ScriptContext {
+    /// Transaction is estimated and signatures are automatically added.
+    #[default]
+    Normal,
```
hal3e, line 67: I just have this tiny nit regarding the naming `Normal`.  I think id does not reflect well what we are doing. What about something like: `CompleteTx`, `FinalizedTx` ... something that would mean that we did all the work in the background.

hal3e (approved):
Excellent work! I love the idea of building with context and also providing info what kind of a simulation we would like to do!

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -521,6 +538,37 @@ impl ScriptTransactionBuilder {
...
+            resolve_fuel_inputs(
+                self.inputs.clone(),
+                num_resolved_witnesses,
+                &self.unresolved_witness_indexes,
+            )?,
+            self.outputs.clone(),
+            vec![],
+        );
+
+        let should_saturate_variable_outputs =
```
MujkicA, line 559: Why do we need to care about variable outputs if it's read only operation?

segfault-magnet: Theoretically if it only were a simple read only we wouldn't. 

Was thinking more about some more complex logic being runnable in this way. I.e. a function mints but also returns something from storage after doing so. 

If we don't propagate the behavior of the variable output policy we're bound to get complaints about not being able to "read only" the tx. Also we might get complaints that we don't respect wishes, 2 lines above the policy was set to exactly 100, but we don't take that into account when simulating in this way.

If the policy isn't explicitly set it defaults to `Exactly(0)` so we won't add any variable outputs in the default case. 

If the user sets a policy it's going to be reflected in this simulation mode the same way it would have been reflected in the other -- because without it you might not get to the end of the contract method.

digorithm (approved):
Great work and great feature. Will make devs' lives a lot better.

Salka1988 (dismissed):

Br1ght0ne (approved):

packages/fuels-core/src/types/transaction_builders.rs
MujkicA: I vote for `Complete` or `Full`, the tx is implied already

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -50,30 +50,48 @@ struct UnresolvedWitnessIndexes {
...
+    NoSignatures,
+    /// No estimation is done and no signatures are added. Fake coins are added if no spendable inputs
+    /// are present. Meant only for transactions that are to be dry-run with validations off.
+    /// Useful for reading state with unfunded accounts.
+    StateReadOnly,
 }
 
-impl sealed::Sealed for ScriptTransactionBuilder {}
+#[derive(Debug, Clone, Default)]
+pub enum Context {
```
MujkicA, line 77: Sounds a bit too general, but I don't have an idea for a term that says "all other tx types but not script - context"

packages/fuels-core/src/types/transaction_builders.rs
```diff
@@ -50,30 +50,48 @@ struct UnresolvedWitnessIndexes {
...
+pub enum ScriptContext {
+    /// Transaction is estimated and signatures are automatically added.
+    #[default]
+    Normal,
+    /// Transaction is estimated but no signatures are added.
     /// Building without signatures will set the witness indexes of signed coins in the
     /// order as they appear in the inputs. Multiple coins with the same owner will have
     /// the same witness index. Make sure you sign the built transaction in the expected order.
-    async fn build_without_signatures(self, provider: impl DryRunner) -> Result<Self::TxType>;
+    NoSignatures,
```
MujkicA, line 69: This might be problematic when witness data is used for something other than signatures. The assumed indexes are most likely going to be wrong. 
It's not a bug per se but an easy footgun.

packages/fuels-core/src/types/transaction_builders/script_tx_estimator.rs
```diff
@@ -87,9 +95,7 @@ impl<R: DryRunner> ScriptDryRunner<R> {
             .take(self.num_unresolved_witnesses)
             .collect();
 
-        let predefined_witnesses = std::mem::take(&mut self.predefined_witnesses);
-
-        *tx.witnesses_mut() = [predefined_witnesses, dry_run_witnesses].concat();
+        *tx.witnesses_mut() = [self.predefined_witnesses.clone(), dry_run_witnesses].concat();
```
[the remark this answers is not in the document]
hal3e: I had a look. This requires a refactor of the `ScriptTxEstimator` and I would not like to do it in this PR. Also we are currently reusing the dry_run if it was done for variable outputs. If we change it to one dry_run we need to refactor that part also.

packages/fuels-core/src/types/transaction_builders.rs
hal3e: Good catch. Currently we ignore already added witnesses. I will fix this in this PR an then the comment will be correct.

digorithm (approved):
Excellent work @hal3e @segfault-magnet. The new API looks super crispy. :shipit:

digorithm (approved):

packages/fuels-core/src/types/transaction_builders.rs
hal3e: My previous comment was wrong. Manually added witnesses were considered.

digorithm (approved):

MujkicA:
I have some reservations about this that I've been discussing with @hal3e.

I don't fully understand the use case for `NoSignatures`. We are essentially putting another abstraction over how witness indexes are handled but from my perspective it seems leaky in this case. It might absolve the user from handling the witness indexes directly but it still requires them to be somewhat aware of what is happening behind the scenes. 
Since we are in the process of re-examining our interfaces for v1.0, I would suggest we hold off on this change for now until we have better picture of how we want to support transaction building and what we want the final interface to look like.

segfault-magnet:
> I have some reservations about this that I've been discussing with @hal3e.
> 
> I don't fully understand the use case for `NoSignatures`. We are essentially putting another abstraction over how witness indexes are handled but from my perspective it seems leaky in this case. It might absolve the user from handling the witness indexes directly but it still requires them to be somewhat aware of what is happening behind the scenes. 
> Since we are in the process of re-examining our interfaces for v1.0, I would suggest we hold off on this change for now until we have better picture of how we want to support transaction building and what we want the final interface to look like.
> At the very least, I would hold off until @segfault-magnet can chime in on the discussion. 

Can only offer a quick input now,  out for the next week or so.

>  I don't fully understand the use case for `NoSignatures`.

This behavior is already present, only the api changed. Iirc the motivation for it was that somebody complained about not wanting to leak real signatures to the network while simulating.

So not merging this pr doesn't stop the leakyness if any exists. This just unifies how the tx is built, you no longer call build without signatures but pass in NoSignatures as the context.

> I would suggest we hold off on this change for now until we have better picture of how we want to support transaction building

If we can afford that I vote for that as well. The issue is the same argument can be used for any public api.

Whatever the final look of tx building is, it is probably not going to resemble too much what we have now, so breaking changes are inevitable.

Not merging this means no support for read only calls until we figure out the total redesign for 1.0, because any implementation is going to change the api and why bother if we're gonna redesign.

I don't have anything much more to add I trust whatever decision you make, just wanted to underline the above.

segfault-magnet:
> I have some reservations about this that I've been discussing with @hal3e.
> 
> I don't fully understand the use case for `NoSignatures`. We are essentially putting another abstraction over how witness indexes are handled but from my perspective it seems leaky in this case. It might absolve the user from handling the witness indexes directly but it still requires them to be somewhat aware of what is happening behind the scenes. 
> Since we are in the process of re-examining our interfaces for v1.0, I would suggest we hold off on this change for now until we have better picture of how we want to support transaction building and what we want the final interface to look like.
> At the very least, I would hold off until @segfault-magnet can chime in on the discussion. 

NoSignatures is already there in master, called build_without_signatures. Whatever it's reason for existence, the leakyness is there. This is a cosmetic change wrt.

If we know we will allocate time for 1.0 redesign and if this feature (and all others that break the public api) can wait for the redesign then I vote for waiting.

Then again 1.0 will break everything so why not break this as well? We could get some feedback in the meantime if there are edge cases to the approach.

digorithm:
> > I have some reservations about this that I've been discussing with @hal3e.
> > I don't fully understand the use case for `NoSignatures`. We are essentially putting another abstraction over how witness indexes are handled but from my perspective it seems leaky in this case. It might absolve the user from handling the witness indexes directly but it still requires them to be somewhat aware of what is happening behind the scenes.
> > Since we are in the process of re-examining our interfaces for v1.0, I would suggest we hold off on this change for now until we have better picture of how we want to support transaction building and what we want the final interface to look like.
> > At the very least, I would hold off until @segfault-magnet can chime in on the discussion.
> 
> NoSignatures is already there in master, called build_without_signatures. Whatever its reason for existence, the leakyness is there. This is a cosmetic change wrt.
> 
> If we know we will allocate time for 1.0 redesign and if this feature (and all others that break the public api) can wait for the redesign then I vote for waiting.
> 
> Then again 1.0 will break everything so why not break this as well? We could get some feedback in the meantime if there are edge cases to the approach.

Yeah. Let's release this feature now; our users visibly need it (and often request it from me). We can always keep making breaking changes until 1.0. I'll continue handling the communications for these painful, breaking changes.

digorithm:
@hal3e are we ready to merge this?

Br1ght0ne (dismissed):

segfault-magnet:
After consulting with @hal3e and @MujkicA we agreed on moving the `context` out of the builder so that it is less of a `in-your-face` change. 

Before building, if you want to change how the build works, you can set a different `BuildStrategy` via `with_build_strategy` before calling `build`. 

Users that don't set it are going to get the default behavior (estimate, auto sign with added signers).

That is the best we could do without doing a cubic tonne of breaking changes.

hal3e (approved):
Looks good to me. Thanks for the refactor! :shipit:

digorithm (approved):
:shipit:

Salka1988 (approved):

packages/fuels-core/src/types/transaction_builders.rs
segfault-magnet: It's now called `BuildStrategy`. Still general. works for n-1/n tx types. 

Hopefully we'll come back to this soon and redesign the builders.

PR #1413, opened 2024-06-07
Predicate estimation fallback to API

### LINKED ISSUE #1409: Predicate estimation fallback to API
fuel core `0.28.0` is going to be releasing with upgrades to the fuel vm. Those updates will impact predicate gas usage estimation. 

Besides that `0.28.0` is backwards compatible API wise -- all endpoints should work with the current version of the SDK.

We need to release a patch version with as little breaking changes as possible that will include a fallback mechanism -- if the sdk detects that the node that we're talking to isn't the exact same version we've been developed against we should use the GQL endpoint for predicate estimation.

Releasing this as soon as possible will give users enough time to update their sdk deps so that they may continue normally estimating predicates when the `0.28.0` core rollout happens.

Br1ght0ne:
Close #1409.

This is very janky, I'm open to any and all ideas of improvement.
I'm going to collect any ideas in a short period, open issue(s) for them, and remove the TODO comments afterward.

Points of special jank:
- `estimate_predicates_with_node_fallback` (name suggestions welcome) calls node for `chain_info` to get the executor version every time a transaction is estimated. A much better approach (suggested by @segfault-magnet) would be to cache the on-chain executor version (or the whole `chain_info`) and only update it periodically.
- had to `impl TryFrom<FuelTransaction> for Transaction`, where logic depends on the enum variant being named exactly the same as the transaction type being wrapped, e.g. `FuelTransaction::Script(tx: Script)`
- in the `TryFrom` impl itself, `Err(error!(Other, "...")` feels wrong and too vague
- `RetryableClient::wrap` only supports immutable references in `self.client` calls, which is impossible with `self.client.estimate_predicates(&mut tx)`. `RetryableClient` could be refactored, but for now I opted to just forward the call to the client without applying retry logic

### Checklist

### DIFF
--- a/packages/fuels-accounts/src/provider.rs
+++ b/packages/fuels-accounts/src/provider.rs
@@ -256,6 +256,10 @@ impl Provider {
         Ok(self.client.estimate_gas_price(block_horizon).await?)
     }
 
+    pub async fn estimate_predicates(&self, tx: impl Transaction) -> Result<()> {
+        Ok(self.client.estimate_predicates(&mut tx.into()).await?)
+    }
+
     pub async fn dry_run(&self, tx: impl Transaction) -> Result<TxStatus> {
         let [tx_status] = self
             .client
--- a/packages/fuels-accounts/src/provider.rs
+++ b/packages/fuels-accounts/src/provider.rs
@@ -17,6 +17,7 @@ use fuel_core_client::client::{
         gas_price::{EstimateGasPrice, LatestGasPrice},
     },
 };
+use fuel_core_types::blockchain::header::LATEST_STATE_TRANSITION_VERSION;
 use fuel_tx::{
     AssetId, ConsensusParameters, Receipt, Transaction as FuelTransaction, TxId, UtxoId,
 };
@@ -180,7 +181,7 @@ impl Provider {
         tx.check(latest_block_height, self.consensus_parameters())?;
 
         if tx.is_using_predicates() {
-            tx.estimate_predicates(self.consensus_parameters())?;
+            self.estimate_predicates_with_node_fallback(&mut tx).await?;
             tx.clone()
                 .validate_predicates(self.consensus_parameters(), latest_block_height)?;
         }
@@ -256,8 +257,29 @@ impl Provider {
         Ok(self.client.estimate_gas_price(block_horizon).await?)
     }
 
-    pub async fn estimate_predicates(&self, tx: impl Transaction) -> Result<()> {
-        Ok(self.client.estimate_predicates(&mut tx.into()).await?)
+    async fn estimate_predicates_with_node_fallback<T: Transaction>(
+        &self,
+        tx: &mut T,
+    ) -> Result<()> {
+        if self.latest_chain_executor_version().await? > LATEST_STATE_TRANSITION_VERSION {
+            *tx = self
+                .client
+                .estimate_predicates(tx.clone().into())
+                .await?
+                .try_into()?
+        } else {
+            tx.estimate_predicates(self.consensus_parameters())?;
+        }
+        Ok(())
+    }
+
+    async fn latest_chain_executor_version(&self) -> Result<u32> {
+        self.chain_info().await.map(|chain_info| {
+            chain_info
+                .latest_block
+                .header
+                .state_transition_bytecode_version
+        })
     }
 
     pub async fn dry_run(&self, tx: impl Transaction) -> Result<TxStatus> {
--- a/packages/fuels-accounts/src/provider/retryable_client.rs
+++ b/packages/fuels-accounts/src/provider/retryable_client.rs
@@ -150,12 +150,13 @@ impl RetryableClient {
             .map(Into::into)
     }
 
-    pub async fn estimate_predicates(&self, tx: &mut Transaction) -> RequestResult<()> {
+    pub async fn estimate_predicates(&self, mut tx: Transaction) -> RequestResult<Transaction> {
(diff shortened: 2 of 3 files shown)


packages/fuels-accounts/src/provider.rs
```diff
@@ -256,6 +257,31 @@ impl Provider {
...
     }
 
+    async fn estimate_predicates_with_node_fallback<T: Transaction>(
+        &self,
+        tx: &mut T,
+    ) -> Result<()> {
+        if self.latest_chain_executor_version().await? > LATEST_STATE_TRANSITION_VERSION {
+            *tx = self
+                .client
+                .estimate_predicates(tx.clone().into())
```
Br1ght0ne, line 267: clone bad, but I don't know how much bad

packages/fuels-accounts/src/provider.rs
```diff
@@ -256,6 +257,31 @@ impl Provider {
         Ok(self.client.estimate_gas_price(block_horizon).await?)
     }
 
+    async fn estimate_predicates_with_node_fallback<T: Transaction>(
+        &self,
+        tx: &mut T,
+    ) -> Result<()> {
+        if self.latest_chain_executor_version().await? > LATEST_STATE_TRANSITION_VERSION {
```
Br1ght0ne, line 264: This calls the node for `chain_info` to get the executor version every time a transaction is estimated. A much better approach (suggested by @segfault-magnet) would be to cache the on-chain executor version (or the whole `chain_info`) and only update it periodically.

packages/fuels-accounts/src/provider.rs
```diff
@@ -180,7 +181,7 @@ impl Provider {
         tx.check(latest_block_height, self.consensus_parameters())?;
 
         if tx.is_using_predicates() {
-            tx.estimate_predicates(self.consensus_parameters())?;
+            self.estimate_predicates_with_node_fallback(&mut tx).await?;
```
Br1ght0ne, line 184: `with_node_fallback` could use a better name

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -150,6 +150,14 @@ impl RetryableClient {
             .map(Into::into)
     }
 
+    pub async fn estimate_predicates(&self, mut tx: Transaction) -> RequestResult<Transaction> {
+        self.client
+            .estimate_predicates(&mut tx)
```
Br1ght0ne, line 155: This doesn't use `RetryableClient::wrap`, because `RetryableClient::wrap` only supports immutable references in `self.client` calls, which is impossible with `self.client.estimate_predicates(&mut tx)`. Changing `RetryableClient` to accept an `FnMut` instead of `Fn` gives errors about `tx` escaping its lifetime via the closure, so for now I opted to just forward the call to the client without applying retry logic.

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -325,6 +326,20 @@ macro_rules! impl_tx_wrapper {
             }
         }
 
+        impl TryFrom<FuelTransaction> for $wrapper {
+            type Error = Error;
+
+            fn try_from(tx: FuelTransaction) -> Result<Self> {
+                match tx {
+                    FuelTransaction::$wrapped(tx) => Ok(tx.into()),
+                    _ => Err(error!(
```
Br1ght0ne, line 335: Had to `impl TryFrom<FuelTransaction> for Transaction`, where logic depends on the enum variant being named exactly the same as the transaction type being wrapped, e.g., `FuelTransaction::Script(tx: Script)`. Any other way to do a round-trip `T -> fuel_tx::Transaction -> T where T: Transaction` would be very appreciated.

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -325,6 +326,20 @@ macro_rules! impl_tx_wrapper {
...
+                        Other,
+                        "couldn't convert Transaction into a wrapper of type $wrapper"
```
Br1ght0ne, line 337: `Err(error!(Other, "...")` feels wrong and too vague

Br1ght0ne (commented):
Self-review pointing out the questionable spots.

Br1ght0ne:
Moved (3eb1a14) commit d536be4 into a separate PR (https://github.com/FuelLabs/fuels-rs/pull/1414).

packages/fuels-core/src/types/wrappers/transaction.rs
hal3e: We will probably have to re-think the whole logic and check if we can simplify this. For this PR the `try_into` is ok

packages/fuels-core/src/types/wrappers/transaction.rs
hal3e: there is a `error_transaction!` macro with `Other`.. not great but still better then only `Other`

hal3e (changes requested):
Left some ideas.

packages/fuels-accounts/src/provider.rs
Br1ght0ne: Taking `chain_info` from `prepare_transaction_for_sending` - no extra network calls, no need for caching (yet).

digorithm (approved):
Solid work :shipit:

MujkicA (approved):

Salka1988 (approved):

PR #1422, opened 2024-06-13
fix!: `dry_run` tx with non-zero base asset and tip

hal3e:
BREAKING CHANGE:
- Removed `dry_run_no_validation` in favor of `dry_run_opt`
- Removed `dry_run_no_validation_multiple` in favor of `dry_run_opt_multiple`

### Checklist

### DIFF
--- a/packages/fuels-accounts/src/provider.rs
+++ b/packages/fuels-accounts/src/provider.rs
@@ -309,13 +309,18 @@ impl Provider {
             .collect())
     }
 
-    pub async fn dry_run_no_validation(&self, tx: impl Transaction) -> Result<TxStatus> {
+    pub async fn dry_run_opt(
+        &self,
+        tx: impl Transaction,
+        utxo_validation: bool,
+        gas_price: Option<u64>,
+    ) -> Result<TxStatus> {
         let [tx_status] = self
             .client
             .dry_run_opt(
                 Transactions::new().insert(tx).as_slice(),
-                Some(false),
-                Some(0),
+                Some(utxo_validation),
+                gas_price,
             )
             .await?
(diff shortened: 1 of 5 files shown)


packages/fuels-accounts/src/provider.rs
```diff
@@ -707,7 +711,7 @@ impl DryRunner for Provider {
     async fn dry_run(&self, tx: FuelTransaction) -> Result<DryRun> {
         let [tx_execution_status] = self
             .client
-            .dry_run_opt(&vec![tx], Some(false))
+            .dry_run_opt(&vec![tx], Some(false), Some(0))
```
MujkicA, line 714: We should either use `dry_run_no_validation` or remove it in favor of dry_run_opt

MujkicA (approved):

packages/fuels-core/src/types/transaction_builders/script_dry_runner.rs
```diff
@@ -128,7 +136,7 @@ impl<R: DryRunner> ScriptDryRunner<R> {
             Default::default(),
             fake_owner,
             1_000_000_000,
-            Default::default(),
+            *base_asset_id,
```
MujkicA, line 139: This PR addresses the issue where simulations were failing because there was no base asset that would cover the set tip.
It was discovered because devnet wasn't using the `AssetId::default()` as base asset.

digorithm: Shoot, nice find.

digorithm (approved):
Great find and fix @hal3e @MujkicA

PR #1428, opened 2024-06-14
fix: use fallback predicate estimation within transaction builder

hal3e:
I have moved the `provider` fallback into the `Transaction` trait via the `DryRunner`.

I had to move `DryRunner` and `DryRun` into `fuels_core::types` because of the wasm target.

BREAKING CHNGE:
- Moved `DryRunner` and `DryRun` to `fuels_core::types`

### Checklist


packages/fuels-accounts/src/provider.rs
```diff
@@ -754,4 +749,26 @@ impl DryRunner for Provider {
...
+                self.chain_info()
+                    .await?
+                    .latest_block
+                    .header
+                    .state_transition_bytecode_version
+            }
+        };
+
+        Ok(
+            (latest_chain_executor_version > LATEST_STATE_TRANSITION_VERSION)
```
Voxelot, line 770: @xgreenx do we also need to check the consensus values version in case the gas prices change?

digorithm: Do we want to hold off the merge until this is answered?

xgreenx: No, we don't need to check consensus parameters for estimation. We may want to check remote version with local version to be sure that we use the latest consensus parameters. If local consensus parameters outdated, then we want to download them one more time.

We can do that because without new state transition function `fuel-core` can't support new gas costs variant -> If the state transition function is the same, but consensus parameters are different, we know how to download them since it is supported version.

digorithm: > We may want to check remote version with local version to be sure that we use the latest consensus parameters. If local consensus parameters outdated, then we want to download them one more time.

I believe we already do this. I need to double-check, though. cc @hal3e

hal3e: Currently we cache the `consensus_parameters` once when we connect to the node. I do not see any update mechanism. We will have to add this.

digorithm: Yeah, we're not updating it, we just check the supported version during the build.

MujkicA (dismissed):

digorithm (dismissed):

digorithm (approved):

MujkicA (approved):

packages/fuels-accounts/src/provider.rs
digorithm: Update: We'll implement this update mechanism soon as part of a follow-up work, so we don't block the release here.

PR #1462, opened 2024-07-18
refactor: canonicalize load paths in `load_from` functions

Br1ght0ne:
This helps users by showing an accurate, absolute path in error messages.

### Checklist


e2e/tests/contracts.rs
```diff
@@ -1084,16 +1086,21 @@ async fn test_add_custom_assets() -> Result<()> {
 async fn contract_load_error_messages() {
     {
         let binary_path = "sway/contracts/contract_test/out/release/no_file_on_path.bin";
```
segfault-magnet, line 1088: Can you add some no-op here so that we see the 1.79 improved err in action? 

E.g. `sway/contracts/../../sway/contracts/contract_test...`

Cargo.toml
```diff
@@ -54,6 +54,7 @@ fuel-abi-types = "0.7.0"
 futures = "0.3.29"
 hex = { version = "0.4.3", default-features = false }
 itertools = "0.12.0"
+path-clean = "1.0.1"
```
Br1ght0ne, line 57: I audited the source of `path-clean`: it's a safe, `std`-augmenting function with a set of tests and no extra dependencies.
P.S. at some point we should have a centralized repo of our crate audits, with something like `cargo-audit`

packages/fuels-accounts/src/predicate.rs
```diff
@@ -40,10 +40,17 @@ impl Predicate {
     }
 
     pub fn load_from(file_path: &str) -> Result<Self> {
-        let code = fs::read(file_path).map_err(|e| {
+        let clean_file_path = path_clean::clean(file_path);
+        let absolute_file_path = path::absolute(&clean_file_path).map_err(|e| {
```
hal3e, line 44: Not that we use `path::clean` would it make sense to use `canolicalize` instead of `absolute`? If we stick with `absolute` I would change the error messages because we currently say: `failed to canonicalize path`

digorithm:
Do you mind resolving these conflicts so we can wrap up this PR?

digorithm (approved):

packages/fuels-accounts/src/predicate.rs
```diff
@@ -40,10 +40,17 @@ impl Predicate {
...
-        let code = fs::read(file_path).map_err(|e| {
+        let absolute_file_path = path::absolute(file_path).map_err(|e| {
             error!(
                 IO,
-                "could not read predicate binary {file_path:?}. Reason: {e}"
+                "failed to make path absolute: {file_path:?}. Reason: {e}"
+            )
+        })?;
+        let clean_file_path = path_clean::clean(absolute_file_path);
+        let code = fs::read(&clean_file_path).map_err(|e| {
```
segfault-magnet, line 50: I'd change this just slightly: 
1. read the code with the original path
2. if that fails, clean up the path and report that we failed to read the clean path
3. if cleaning up fails fallback to the original path and report that

Because we don't need a "clean" path to read the data only for diagnostics. So in the eventuality of any schenanigans with symlinks we'll still function, only our reporting might or might not be nice.

Br1ght0ne: `path_clean::clean` returns a `PathBuf`, not a `Result` (can't fail). I agree, though: changed to read with the original path, and report error with cleaned up path in cb4a5cb.

packages/fuels-accounts/src/predicate.rs
```diff
@@ -40,10 +40,17 @@ impl Predicate {
...
                 IO,
-                "could not read predicate binary {file_path:?}. Reason: {e}"
+                "failed to make path absolute: {file_path:?}. Reason: {e}"
+            )
+        })?;
+        let code = fs::read(&absolute_file_path).map_err(|e| {
+            let clean_file_path = path_clean::clean(absolute_file_path);
+            error!(
+                IO,
+                "could not read predicate binary {clean_file_path:?}. Reason: {e}",
```
segfault-magnet, line 53: Can we read only the provided `file_path` and leave any prettifying (such as getting the abs path) in the Err case?

packages/fuels-programs/src/contract/regular.rs
```diff
@@ -99,17 +104,26 @@ impl Contract<Regular> {
         binary_filepath: impl AsRef<Path>,
         config: LoadConfiguration,
     ) -> Result<Contract<Regular>> {
-        let binary_filepath = binary_filepath.as_ref();
-        validate_path_and_extension(binary_filepath, "bin")?;
+        let absolute_binary_filepath = path::absolute(binary_filepath.as_ref()).map_err(|e| {
```
segfault-magnet, line 107: here as well?

packages/fuels-programs/src/contract/storage.rs
```diff
@@ -94,13 +95,23 @@ impl StorageSlots {
     }
 
     pub(crate) fn load_from_file(storage_path: impl AsRef<Path>) -> Result<Self> {
-        let storage_path = storage_path.as_ref();
-        validate_path_and_extension(storage_path, "json")?;
+        let absolute_storage_path = path::absolute(storage_path.as_ref()).map_err(|e| {
```
segfault-magnet, line 98: and here

packages/fuels-accounts/src/predicate.rs
```diff
@@ -41,10 +41,16 @@ impl Predicate {
...
-            error!(
-                IO,
-                "could not read predicate binary {file_path:?}. Reason: {e}"
-            )
+            match path::absolute(file_path).map(path_clean::clean) {
+                Ok(clean_file_path) => error!(
+                    IO,
+                    "could not read predicate binary {clean_file_path:?}. Reason: {e}",
+                ),
+                Err(e) => error!(
```
segfault-magnet, line 49: What would be ideal IMO is the following:
 read the original path - success - > all good
 read the original path -> fails -> clean it up - cleanup successful -> report nice err message
 read the original path -> fails -> clean it up -> cleanup fails -> report original err message without mention of the cleanup failing

Issue #1470, opened 2024-07-21
Blob tx / large contract support


segfault-magnet:
Adapt the SDK to support blob transactions and deploying contracts using them.

Core side: https://github.com/FuelLabs/fuel-specs/issues/589

[A nice diagram](https://github.com/FuelLabs/fuels-ts/issues/2741) by @arboleya

arboleya:
I believe the diagram comes from @JoshuaBatty. 😅

PR #1472, opened 2024-07-24
feat!: deploying large contracts (loader + blob support)

segfault-magnet:
closes: #1470 
A thank you to @hal3e for brainstorming with me.

The `Contract` is now a type-state pattern. Please refer to the below diagram to help you understand the changes:
image

The change was needed so that the following might be achieved:
* Loaders must not have configurables applied to them
* We must be able to get the contract_id at every point
* Flexibility around blob formation and upload

You basically have three states:

### A regular contract
What you're used to seeing. It is either initialized from raw code or loaded from a file:
```rust
        let contract = Contract::regular(contract_binary, Salt::zeroed(), vec![]);
```
or
```rust
    let contract = Contract::load_from(
        "sway/contracts/storage/out/release/storage.bin",
        LoadConfiguration::default(),
    )?;
```
With the notable addition of being able to set `configurables` (previously possible only when using `load_from`):
```rust
    let contract =
        Contract::regular(binary, Salt::zeroed(), vec![]).with_configurables(configurables);
```
a regular contract can be deployed via `deploy`, which hasn't changed, or via `smart_deploy` that will use blobs/loader if the contract is above what can be deployed in a create tx:
```rust
        let contract_id = Contract::load_from(
            contract_binary,
            LoadConfiguration::default().with_salt(random_salt()),
        )?
        .smart_deploy(&wallet, TxPolicies::default(), max_words_per_blob)
        .await?;
```

### Loader contract, blobs pending upload
You can turn a regular contract into a loader contract:
```rust
        let contract = Contract::load_from(
            contract_binary,
            LoadConfiguration::default(),
        )?
        .convert_to_loader(max_words_per_blob)?
```
or, if you have the blobs, create it directly:
```rust
        let contract = Contract::loader_for_blobs(blobs, random_salt(), vec![])?;
```
You can also revert back to the regular contract via `revert_to_regular`.

If you now call `deploy` the contract will first deploy the blobs and then the loader itself.

You can also split this into two parts by first calling `upload_blobs` and then `deploy`:
```rust
    let contract_id = Contract::load_from(contract_binary, LoadConfiguration::default())?
        .convert_to_loader(1024)?
        .upload_blobs(&wallet, TxPolicies::default())
        .await?
        .deploy(&wallet, TxPolicies::default())
        .await?;
```
doing so will have `deploy` only submit the create tx while the uploading will be done in `upload_blobs`.

### Loader, with blobs deployed
You arrive at this contract type by eithers having the blob ids and creating it manually:
```rust
        let contract = Contract::loader_for_blob_ids(all_blob_ids, random_salt(), vec![])?;
```
or by calling `upload_blobs` as in the previous case:
```rust
        let contract = Contract::load_from(
            contract_binary,
            LoadConfiguration::default().with_salt(random_salt()),
        )?
        .convert_to_loader(max_words_per_blob)?
        .upload_blobs(&wallet, TxPolicies::default())
        .await?;
```
Calling deploy on this contract only deploys the loader.

### Estimating max blob size
This is a bit tricky. Blob sizes are limited by two things: max transaction size and max transaction gas (both part of consensus parameters). So a blob can be as big as you want as long as the overall tx respects these two global tx limits.

So that means that properly estimating the max blob the user can send becomes a bit tricky. If you fund the tx with 2 coins instead of 1 the answer changes. If you use a predicate instead of a coin also changes the result. Basically whatever impacts the size or gas usage will impact how big the blob can be. 

Say you put a blob of 20kb, and fund the tx. Oh look you have extra size and gas to spare. You increase that a bit but you no longer have 1 coin that can cover that but need to use 2. Ok the new blob size is no longer acceptable since you took up a bit of space adding that coin. Remove it. Ok now you can increase the blob size again. But now you again need 2 coins...

As a start, and until we figure something better and robust, there is a crude estimation available in the form of:
```rust
        let max_blob_size = BlobTransactionBuilder::default()
            .estimate_max_blob_size(&provider)
            .await?;
```
The docs explain the caveats of this estimation.

### Breaking changes @digorithm 
* `Contract::new` is removed, replaced with `Contract::regular`.
* `Contract` now accepts a generic argument denoting the type of contract (regular, loader, etc)

### Checklist


e2e/tests/contracts.rs
```diff
@@ -2105,3 +2106,139 @@ async fn max_fee_estimation_respects_tolerance() -> Result<()> {
...
+        .map(|tx| match tx.transaction {
+            TransactionType::Blob(_) => "blob",
+            TransactionType::Create(_) => "create",
+            TransactionType::Script(_) => "script",
+            _ => "other",
+        })
+        .collect()
+}
+
+async fn assert_wallet_made_no_trancsactions(wallet: &WalletUnlocked) {
```
Br1ght0ne, line 2242:
```suggestion
async fn assert_wallet_made_no_transactions(wallet: &WalletUnlocked) {
```

Cargo.toml
```diff
@@ -79,6 +79,7 @@ trybuild = "1.0.85"
 uint = { version = "0.9.5", default-features = false }
 which = { version = "6.0.0", default-features = false }
 zeroize = "1.7.0"
+pretty_assertions = { version = "1.4", default-features = false }
```
Br1ght0ne, line 82: alpha-sort?

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+            // prepared fn selector and jumps to the selected contract method.
+            [
+                // 0x12 is going to hold the total size of the contract
+                op::move_(0x12, RegId::ZERO),
+                // find the start of the hardcoded blob ids, which are located after the code ends
+                op::move_(0x10, RegId::IS),
+                // 0x10 to hold the address of the current blob id
+                op::addi(0x10, 0x10, num_of_instructions * Instruction::SIZE as u16),
+                // loop counter
+                op::addi(0x13, RegId::ZERO, num_of_blobs),
```
Br1ght0ne, line 169: Excuse me if I'm wrong. Does this work?

```suggestion
                op::movi(0x13, num_of_blobs),
```

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+                // 0x11 to hold the size of the current blob
+                op::bsiz(0x11, 0x10),
+                // update the total size of the contract
+                op::add(0x12, 0x12, 0x11),
+                // move on to the next blob
+                op::addi(0x10, 0x10, BLOB_ID_SIZE),
+                // decrement the loop counter
+                op::subi(0x13, 0x13, 1),
+                // Jump backwards 3 instructions if the counter has not reached 0
+                op::jneb(0x13, RegId::ZERO, RegId::ZERO, 3),
```
Br1ght0ne, line 180:
```suggestion
                op::jnzb(0x13, RegId::ZERO, 3),
```

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+                // move the stack pointer by the contract size since we need to write the contract on the stack since only that memory can be executed
+                op::cfe(0x12),
+                // find the start of the hardcoded blob ids, which are located after the code ends
+                op::move_(0x10, RegId::IS),
+                // 0x10 to hold the address of the current blob id
+                op::addi(0x10, 0x10, num_of_instructions * Instruction::SIZE as u16),
+                // 0x12 is going to hold the total bytes loaded of the contract
+                op::move_(0x12, RegId::ZERO),
+                // loop counter
+                op::addi(0x13, RegId::ZERO, num_of_blobs),
```
Br1ght0ne, line 190: Ditto:

```suggestion
                op::movi(0x13, num_of_blobs),
```

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+                // load the current blob
+                op::bldd(0x14, 0x10, RegId::ZERO, 0x11),
+                // update the total bytes loaded
+                op::add(0x12, 0x12, 0x11),
+                // move on to the next blob
+                op::addi(0x10, 0x10, BLOB_ID_SIZE),
+                // decrement the loop counter
+                op::subi(0x13, 0x13, 1),
+                // Jump backwards 6 instructions if the counter has not reached 0
+                op::jneb(0x13, RegId::ZERO, RegId::ZERO, 6),
```
Br1ght0ne, line 207:
```suggestion
                op::jnzb(0x13, RegId::ZERO, 6),
```

Br1ght0ne (changes requested):
Nothing but nits - exceptional work. Still woozy after that asm, but I've never seen an asm block so well commented. LFG!

Cargo.toml
segfault-magnet: cargo sort you fail me.

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+            .collect();
+
+        Ok(blobs)
+    }
+
+    // This function creates a contract that loads the specified blobs into memory and delegates the call to the code contained in the blobs.
+    fn loader_contract(blob_ids: &[[u8; 32]]) -> Result<Vec<u8>> {
+        const BLOB_ID_SIZE: u16 = 32;
+        let get_instructions = |num_of_instructions, num_of_blobs| {
+            // There are main 3 steps:
```
hal3e, line 155:
```suggestion
            // There are 3 main steps:
```

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+            // 1. Calculate the total size of the contract
+            // 2. Allocate and load the contract into memory
+            // 3. Jump into the memory where the contract is loaded
+            // After that the execution continues normally with the loaded contract reading our
+            // prepared fn selector and jumps to the selected contract method.
+            [
+                // 0x12 is going to hold the total size of the contract
```
hal3e, line 162: To make this ASM block maintainable I would add where each of the 3 steps is starting.
```suggestion
                // 1. Calculate the total size of the contract
                // 0x12 is going to hold the total size of the contract
```

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+                op::bsiz(0x11, 0x10),
+                // update the total size of the contract
+                op::add(0x12, 0x12, 0x11),
+                // move on to the next blob
+                op::addi(0x10, 0x10, BLOB_ID_SIZE),
+                // decrement the loop counter
+                op::subi(0x13, 0x13, 1),
+                // Jump backwards 3 instructions if the counter has not reached 0
+                op::jnzb(0x13, RegId::ZERO, 3),
+                // move the stack pointer by the contract size since we need to write the contract on the stack since only that memory can be executed
```
hal3e, line 181:
```suggestion
                // 2. Allocate and load the contract into memory
                // move the stack pointer by the contract size since we need to write the contract on the stack since only that memory can be executed
```

packages/fuels-programs/src/contract.rs
```diff
@@ -45,6 +84,160 @@ impl Contract {
...
+                op::bldd(0x14, 0x10, RegId::ZERO, 0x11),
+                // update the total bytes loaded
+                op::add(0x12, 0x12, 0x11),
+                // move on to the next blob
+                op::addi(0x10, 0x10, BLOB_ID_SIZE),
+                // decrement the loop counter
+                op::subi(0x13, 0x13, 1),
+                // Jump backwards 6 instructions if the counter has not reached 0
+                op::jnzb(0x13, RegId::ZERO, 6),
+                // what follows is called _jmp_mem by the sway compiler
```
hal3e, line 208:
```suggestion
                // 3. Jump into the memory where the contract is loaded
                // what follows is called _jmp_mem by the sway compiler
```

hal3e (commented):
Excelent work!. Love the API and how it fits to our current workflow. Left some minor comment nits...

Br1ght0ne (approved):

Salka1988 (approved):

MujkicA (approved):

MujkicA (approved):

Salka1988 (approved):

hal3e (dismissed):

Br1ght0ne (dismissed):

e2e/tests/contracts.rs
```diff
@@ -2105,3 +2105,177 @@ async fn max_fee_estimation_respects_tolerance() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn blob_contract_deployment() -> Result<()> {
+    // TODO: segfault record transactions again
```
hal3e, line 2111:
```suggestion
```

e2e/tests/contracts.rs
```diff
@@ -2105,3 +2105,177 @@ async fn max_fee_estimation_respects_tolerance() -> Result<()> {
...
+    let contract_binary = "sway/contracts/huge_contract/out/release/huge_contract.bin";
+
+    let contract = Contract::load_from(contract_binary, LoadConfiguration::default())?
+        .convert_to_loader(1024)?
+        .upload_blobs(&wallet, TxPolicies::default())
+        .await?;
+
+    let blob_ids = contract.blob_ids();
+
+    // if this were an example for the user we'd just call `deploy` on the contract above
```
hal3e, line 2235:
```suggestion
    // if this was an example for the user we'd just call `deploy` on the contract above
```

packages/fuels-core/src/types/transaction_builders/blob.rs
```diff
@@ -0,0 +1,205 @@
...
+        input::Input,
+        transaction::{BlobTransaction, EstimablePredicates, Transaction, TxPolicies},
+        DryRunner,
+    },
+    utils::{calculate_witnesses_size, sealed},
+};
+
+#[derive(Default, Clone, Debug, PartialEq)]
+pub struct Blob {
+    pub data: Vec<u8>,
```
hal3e, line 31:
```suggestion
    data: Vec<u8>,
```

packages/fuels-programs/src/contract.rs
```diff
@@ -219,4 +115,236 @@ mod tests {
...
+        assert_eq!(
+            err.to_string(),
+            "blob 1/2 has a size of 9 bytes, which is not a multiple of 8".to_string()
+        );
+    }
+
+    #[test]
+    fn last_blob_in_loader_can_be_unaligned() {
+        // given
+        let blobs = [vec![0; 8], vec![0; 9]].map(Blob::new).to_vec();
```
hal3e, line 188: would it make sense to error out here. As this can be represented as two blobs with size 8 and then one blob of size 1. Which means that the last blob can have any size smaller than 8.

packages/fuels-programs/src/contract/loader.rs
```diff
@@ -0,0 +1,314 @@
...
+        compute_contract_id_and_state_root(&self.code(), &self.salt, &self.storage_slots)
+    }
+
+    /// Creates a loader contract for the code found in `blobs`. Calling `deploy` on this contract
+    /// does two things:
+    /// 1. Uploads the code blobs.
+    /// 2. Deploys the loader contract.
+    ///
+    /// The loader contract, when executed, will load all the given blobs into memory and delegate the call to the original contract code contained in the blobs.
+    pub fn loader_for_blobs(
```
hal3e, line 125: For me it is more natural to call the function `loader_from_blobs`. 

```suggestion
    pub fn loader_from_blobs(
```

packages/fuels-programs/src/contract/loader.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    pub fn compute_roots(&self) -> (ContractId, Bytes32, Bytes32) {
+        compute_contract_id_and_state_root(&self.code(), &self.salt, &self.storage_slots)
+    }
+
+    /// Creates a loader contract using previously uploaded blobs.
+    ///
+    /// The contract code has been uploaded in blobs with [`BlobId`]s specified in `blob_ids`.
+    /// This will create a loader contract that, when deployed and executed, will load all the specified blobs into memory and delegate the call to the code contained in the blobs.
+    pub fn loader_for_blob_ids(
```
hal3e, line 279: same here
```suggestion
    pub fn loader_from_blob_ids(
```

docs/src/deploying/large_contract.md
```diff
@@ -0,0 +1,82 @@
...
+```
+
+2. The maximum gas usage for a single transaction:
+
+```rust,ignore
+{{#include ../../../examples/contracts/src/lib.rs:show_max_tx_gas}}
+```
+
+3. The maximum HTTP body size accepted by the Fuel node.
+To estimate an appropriate size for your blobs, you can run:
```
hal3e, line 70: we need a new line here. Currently, the `estimate` line is under point `3.`
```suggestion

To estimate an appropriate size for your blobs, you can run:
```

hal3e (commented):
Great work! I love the new workflow and the typesate pattern makes sense here. Left some nits.

packages/fuels-programs/src/contract.rs
segfault-magnet: Um. Not sure I follow? 

The only thing we need to enforce is that all blobs, except the last one, need to be word-aligned. 

This test here shows that we tolerate the last blob to be unaligned. 

There is another test that shows that this cannot happen anywhere else.

packages/fuels-core/src/types/transaction_builders/blob.rs
segfault-magnet: Initially wanted to make it public but you're right. We're better of hiding the field. Could add caching later on for the id if needed.

e2e/tests/contracts.rs
segfault-magnet: wanted to say "had this been an example", so were would be closer here.

TIL it's called [subjunctive mood](https://www.grammar-monster.com/glossary/subjunctive_mood.htm)

packages/fuels-programs/src/contract.rs
hal3e: Thanks for the explanation. This makes sense. Disregard my comment because I understood the test wrongly.

hal3e (approved):
:shipit:

Br1ght0ne (approved):

packages/fuels-programs/src/contract.rs
```diff
@@ -1,169 +1,65 @@
...
-pub use load::*;
 pub use storage::*;
 
-/// [`Contract`] is a struct to interface with a contract. That includes things such as
-/// compiling, deploying, and running transactions against a contract.
-#[derive(Debug, Clone)]
-pub struct Contract {
-    binary: Vec<u8>,
+#[derive(Debug, Clone, PartialEq)]
+pub struct Contract<Code> {
```
MujkicA, line 9: Nit: having a docstring here would be nice

MujkicA (dismissed):
Got some nits. Great work 🚀

e2e/tests/configurables.rs
```diff
@@ -158,6 +158,73 @@ async fn contract_configurables() -> Result<()> {
...
+        .with_TUPLE((7, false))?
+        .with_ARRAY([252, 253, 254])?
+        .with_STRUCT(new_struct.clone())?
+        .with_ENUM(new_enum.clone())?;
+
+    let contract_id = Contract::load_from(
+        "sway/contracts/configurables/out/release/configurables.bin",
+        LoadConfiguration::default(),
+    )?
+    .with_configurables(configurables)
```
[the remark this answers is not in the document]
segfault-magnet: So that you may configure a contract you created via `Contract::regular` ie via code you had in memory. 

Previously we only applied configurables to contracts loaded via "load_from". If you wanted them applied to a contract you created from in memory bytes you would need to turn `MyConfigurables` into `Configurables` and call the method `update_constants_in` and give the binary. 

This test here tests that the "setter" exists on the contract itself and that it works.

packages/fuels-programs/src/contract.rs
segfault-magnet: The previous one was mostly false :D 

I'll add one now.

examples/providers/src/lib.rs
```diff
@@ -4,6 +4,7 @@ mod tests {
 
     use fuels::prelude::Result;
 
+    #[ignore = "testnet currently not compatible with the sdk"]
```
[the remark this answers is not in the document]
segfault-magnet: Testnet doesn't have the required gas values for blob instructions.

The gql client was written to deem that a panic-worthy offense (by mistake). 

Until the consensus parameters are updated we cannot connect to testnet.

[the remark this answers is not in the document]
segfault-magnet: Also we probably don't handle [version mismatches](https://github.com/FuelLabs/fuels-rs/issues/1477) in the consensus parameters gracefully.

packages/fuels-programs/src/contract/loader.rs
segfault-magnet: I'm also leaning towards 'for' (that's why that was the first version) 

But the leader does incorporate blobs (at least blob ids) in its code, so it is also made from the blob (ids). 

I decided to accept @hal3e suggestion because somebody might discover it more easily since there are a lot of `from_xyz` in the wild.

MujkicA (approved):

Br1ght0ne (approved):

e2e/Forc.toml
```diff
@@ -110,3 +111,5 @@ members = [
   'sway/types/scripts/script_u256',
   'sway/types/scripts/script_vectors',
 ]
+[patch.'https://github.com/fuellabs/sway']
+std = { git = "https://github.com/fuellabs/sway", branch = "esdrubal/abi_changes" }
```
digorithm, line 115: We gotta keep an eye on this. Like it happened before, _when_ the sway branches gets merged and deleted, the user will see some weird bugs.

digorithm (approved):
Great work! The API is very clean. Still, I'm nervous to see how people will use it; the documentation is very on point and hopefully it'll help people. In any case, we'll have to keep our eyes here and quickly adjust things as needed.

e2e/Forc.toml
segfault-magnet: Hopefully sway gets a release soon after we do so we can remove this.

PR #1473, opened 2024-07-25
feat!: account impersonation

MujkicA:
Enables account impersonation through  the use of`ImpersonatedAccount`. An `ImpersonatedAccount` is created from a address which can correspond to a predicate or wallet address.  It implements `Account` which means it can be used will the existing interfaces for contract calls, transfer etc.

Usage: 
```rust
// create impersonator Account
let address = 
  Address::from_str("6b63804cfbf9856e68e5b6e7aef238dc8311ec55bec04df774003a2c96e0418e")
  .unwrap();
let address = Bech32Address::from(address);
let impersonator = ImpersonatedAccount::new(address, Some(provider));

// transferring assets
impersonator
  .transfer(
   receiver_address,
   amount,
   Default::default(),
   TxPolicies::default(),
)
.await
.unwrap();

// calling contracts
let contract_instance = MyContract::new(contract_id, impersonator);
contract_instance
  .methods()
  .some_method()
  .call()
  .await?;
```

Note: impersonation requires the network to be setup with `utxo_validation: false`

### Breaking changes
The SDK previously performed transaction validity checks, including signature verification, before sending a transaction to the network. This was problematic since the checks also included signature verification even when utxo validation was turned off. To enable this feature and prevent future issues like failed validation checks due to version mismatches between the network and the SDK's upstream dependencies, we decided to remove the check. Since the SDK already abstracts building transactions for common cases (contract calls, transfers, etc.), validity issues are unlikely. If needed, we can still expose the validity checks as part of the transaction builder or our transaction structs.

### Checklist


packages/fuels-accounts/src/impersonated_account.rs
```diff
@@ -0,0 +1,77 @@
...
+use fuel_crypto::{Message, Signature};
+use fuels_core::traits::Signer;
+use fuels_core::types::transaction_builders::TransactionBuilder;
+use fuels_core::types::{bech32::Bech32Address, errors::Result};
+use fuels_core::types::{coin_type_id::CoinTypeId, input::Input, AssetId};
+
+use crate::accounts_utils::try_provider_error;
+use crate::{provider::Provider, Account, ViewOnlyAccount};
+
+/// A `WalletUnlocked` is equivalent to a [`Wallet`] whose private key is known and stored
```
Br1ght0ne, line 13: Doc comment should be changed

packages/fuels-accounts/src/impersonated_account.rs
```diff
@@ -0,0 +1,77 @@
...
+
+    fn try_provider(&self) -> Result<&Provider> {
+        self.provider.as_ref().ok_or_else(try_provider_error)
+    }
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+impl Account for ImpersonatedAccount {
+    /// Returns a vector consisting of `Input::Coin`s and `Input::Message`s for the given
+    /// asset ID and amount. The `witness_index` is the position of the witness (signature)
```
segfault-magnet, line 43: I don't see no `witness_index`.

packages/fuels-accounts/src/impersonated_account.rs
```diff
@@ -0,0 +1,77 @@
...
+        Ok(self
+            .get_spendable_resources(asset_id, amount, excluded_coins)
+            .await?
+            .into_iter()
+            .map(Input::resource_signed)
+            .collect::<Vec<Input>>())
+    }
+
+    fn add_witnesses<Tb: TransactionBuilder>(&self, tb: &mut Tb) -> Result<()> {
+        tb.add_signer(self.clone())?;
```
segfault-magnet, line 61: In the PR description it is written:

>  An ImpersonatedAccount is created from a address which can correspond to a predicate or wallet address.

Maybe that should be revised? Signatures in witnesses doesn't happen with predicates. Also only `utxo` validation is off, not sure if that means that every predicate is always going to validate to true.

packages/fuels-accounts/src/provider.rs
```diff
@@ -185,7 +185,7 @@ impl Provider {
             state_transition_bytecode_version: latest_chain_executor_version,
             ..
         } = chain_info.latest_block.header;
-        tx.check(latest_block_height, self.consensus_parameters())?;
+        //tx.check(latest_block_height, self.consensus_parameters())?;
```
segfault-magnet, line 188: ?

e2e/tests/providers.rs
```diff
@@ -302,10 +301,10 @@ async fn contract_deployment_respects_maturity() -> Result<()> {
         "should not deploy contract since block height `0` is less than the requested maturity `1`",
     );
 
-    let Error::Transaction(Reason::Validation(s)) = err else {
+    let Error::Provider(s) = err else {
         panic!("expected `Validation`, got: `{err}`");
```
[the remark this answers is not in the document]
MujkicA: Validation fails on the node side and any node side errors are wrapped as a Provider error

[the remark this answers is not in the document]
segfault-magnet: agh

e2e/tests/providers.rs
```diff
@@ -1163,3 +1162,44 @@ async fn tx_with_witness_data() -> Result<()> {
...
+    let contract_id = Contract::load_from(
+        "sway/contracts/contract_test/out/release/contract_test.bin",
+        LoadConfiguration::default(),
+    )?
+    .deploy(&wallet, TxPolicies::default())
+    .await?;
+
+    let contract_instance = MyContract::new(contract_id, impersonator.clone());
+
+    // The gas used by the script to call a contract and forward remaining gas limit.
```
hal3e, line 1197:
```suggestion
```

hal3e (commented):
Would it make sense to have the `ImpersonatedAccount` behind a flag? Maybe `test-helpers` so that regular users do not see an additional account type?

Br1ght0ne (approved):

MujkicA:
> Would it make sense to have the `ImpersonatedAccount` behind a flag? Maybe `test-helpers` so that regular users do not see an additional account type?

IMO, the name, docs and docstring makes it sufficiently clear and having a feature flag just for this seems overkill.

Salka1988 (approved):

segfault-magnet (approved):
We should discuss at some point whether we plan on mixing test-only features with the rest of the codebase. 

We already have test helpers and we do turn on test only code based on the presence of the test helpers feature.

In any case, since we're probably not consistent now, I'm fine with having this PR introduce always-on test-only code (especially since it doesn't bring in any extra crate deps).

Br1ght0ne (approved):

PR #1494, opened 2024-08-22
fix: state read dry run sets gas price to 0

MujkicA:
### Checklist

### DIFF
--- a/packages/fuels-programs/src/calls/call_handler.rs
+++ b/packages/fuels-programs/src/calls/call_handler.rs
@@ -198,7 +198,7 @@ where
                 .build(provider)
                 .await?;
 
-            provider.dry_run_opt(tx, false, None).await?
+            provider.dry_run_opt(tx, false, Some(0)).await?
         } else {
             let tx = self.build_tx().await?;
             provider.dry_run(tx).await?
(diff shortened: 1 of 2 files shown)


Br1ght0ne (approved):
Great regression test, LGTM.

packages/fuels-programs/src/calls/call_handler.rs
```diff
@@ -198,7 +198,7 @@ where
                 .build(provider)
                 .await?;
 
-            provider.dry_run_opt(tx, false, None).await?
+            provider.dry_run_opt(tx, false, Some(0)).await?
```
digorithm, line 201: Can you clarify the difference between none and some(0)?

digorithm (approved):
Thanks for tackling this so quickly!

segfault-magnet (approved):

hal3e (approved):

packages/fuels-programs/src/calls/call_handler.rs
MujkicA: We're telling the node to assume that the gas price is 0 for this dry run. `None` would mean it uses whatever the current price is.

PR #1528, opened 2024-10-11
feat: don't deploy contracts if they were already deployed

### LINKED ISSUE #1518: e2e: connect to already deployed contracts instead of redeploying with salt
This will require adding this logic to `setup_program_test!`, or moving away from it (undesirable).

Might be handled in #1471 - I'll link the issue in the PR in that case.

Br1ght0ne:
- Closes #1518

# Summary

Please see the separate commits to ease reviewing:
- [add contract_exists method to Provider and RetryableClient](https://github.com/FuelLabs/fuels-rs/pull/1528/commits/085737b30ca677eb199489db521e50ddaf9d5e35)
  - after [Add and use a custom query for contract_exists](https://github.com/FuelLabs/fuels-rs/pull/1528/commits/196d30f0e5c029a0949b1688b0cab6052dac83ff), uses a query which minimizes fetched data (doesn't download `bytecode: Hex` from the node)
- [add deploy_if_not_exists method to Contract](https://github.com/FuelLabs/fuels-rs/pull/1528/commits/11fa9b03f5002c837fdd2d21b23974332f56b7bf) - coexists with current `deploy` method, wraps it with a `contract_exists` condition
- [add random_salt opt. attribute to Deploy macro](https://github.com/FuelLabs/fuels-rs/pull/1528/commits/5ea45bbd1b5409be99f8fcfae168d09840090c56) - changes all our tests using `setup_program_test!` to use a `[0; 32]` salt instead of a random one. This doesn't break existing users who don't set this option

# Checklist


packages/fuels-programs/src/contract/loader.rs
```diff
@@ -236,6 +236,18 @@ impl Contract<Loader<BlobsNotUploaded>> {
             .await
     }
 
+    /// Deploys the loader contract after uploading the code blobs,
+    /// skipping deploy if it was already deployed with this ID.
```
segfault-magnet, line 240: Not sure if I'd say deployed with I'd.

It's deployed with this load configuration and that gave it the I'd.

Maybe avoid that altogether and say I'd there is no contract with this I'd or something like that.

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -304,5 +304,11 @@ impl RetryableClient {
         })
         .await
     }
+
+    pub async fn contract_exists(&self, contract_id: &ContractId) -> RequestResult<bool> {
+        self.wrap(|| self.client.contract(contract_id))
+            .await
+            .map(|contract| contract.is_some())
```
segfault-magnet, line 311: On my phone, but can you check that this doesn't download the whole code just to check if a contract exists.
If so, maybe let ahmedm s pr go in first and then add a custom query that will minimize the fetched data?

Br1ght0ne: I was thinking about this exactly: this query definitely downloads the whole `bytecode: HexString`. Thanks for the tip, I'll wait on #1529.

Br1ght0ne:
Converting to draft to work on top of #1529 once it's merged.

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -306,6 +307,22 @@ impl RetryableClient {
...
+            });
+            self.client.query(query)
+        })
+        .await
+        .map(|query| {
+            query
+                .contract
+                .map(|contract| ContractId::from(contract.id) == *contract_id)
+                .unwrap_or(false)
+        })
```
Br1ght0ne, line 324: This assertion looks awkward, but IMO is better than just `query.contract.is_some()`, and `cynic` (or GraphQL in general, I don't know) requires at least 1 field in the query nonetheless (the simplest for `Contract` being its `id`)

packages/fuels-accounts/src/provider/retryable_client.rs
Br1ght0ne: Added a custom query in 196d30f.

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -357,4 +376,24 @@ mod custom_queries {
         #[arguments(id: $transaction_id)]
         pub transaction: Option<TransactionIdFragment>,
     }
+
+    #[derive(cynic::QueryFragment, Clone, Debug)]
+    #[cynic(
+        schema_path = "./target/fuel-core-client-schema.sdl",
+        graphql_type = "Contract"
+    )]
+    pub struct ContractIdOnly {
```
segfault-magnet, line 385: fuel core client already has a `ContractIdFragment`, you can use that. We use it in the blob query.

segfault-magnet (approved):

MujkicA (approved):

hal3e (approved):

PR #1529, opened 2024-10-11
feat: add checksum support

MujkicA:
# Release notes

Added EIP-55 checksum address support by introducing `checksum_encode` and `is_checksum_valid` functions. Added the custom query `is_user_account` to `Provider` and `RetryableClient` to verify an address doesn't correspond to a different type of resource.

# Summary

The `checksum_encode` function produces checksummed addresses as strings, and `is_checksum_valid` verifies the correctness of a given address. 
The custom query `is_user_account` will check whether the given address corresponds to a blob, contract or transaction id.

# Checklist

- [x] All **changes** are **covered** by **tests**
- [ ] All **changes** are **documented** 
- [x] I **reviewed** the **entire PR** myself


packages/fuels-core/src/types/checksum_address.rs
```diff
@@ -0,0 +1,138 @@
+use crate::types::errors::{Error, Result};
+use sha3::{Digest, Keccak256};
+
+pub fn checksum_encode(address: &str) -> Result<String> {
+    let trimmed = address.trim_start_matches("0x");
+    pre_validate(trimmed)?;
+
+    let lowercase = trimmed.to_ascii_lowercase();
+
+    let hash = Keccak256::digest(lowercase.as_bytes());
```
mchristopher, line 10: We're not using Keccak on the TS side - can you confirm the test cases we have in https://github.com/FuelLabs/fuels-ts/pull/3308/files#diff-2c4d8f2d738f195a78c5ee15cb8b4271401b169fefbdccb0e1924b33e74b1cabR381-R400 all pass the same way here?

MujkicA: I'll switch it over to sha256 if that's what we want to do

Cargo.toml
```diff
@@ -48,6 +48,8 @@ async-trait = { version = "0.1.74", default-features = false }
 bech32 = "0.9.1"
 bytes = { version = "1.5.0", default-features = false }
 chrono = "0.4.31"
+cynic = { version = "2.2", default-features = false }
+cynic-codegen = { version = "3.8.0", default-features = false }
```
segfault-magnet, line 52: Not used

packages/fuels-accounts/Cargo.toml
```diff
@@ -33,6 +34,10 @@ fuel-tx = { workspace = true, features = ["test-helpers", "random"] }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["test-util"] }
 
+[build-dependencies]
+cynic = { workspace = true, features = ["default"], optional = true } # TODO reduce features
```
segfault-magnet, line 38: Let's maybe leave the default features. Buys us more fwd compat

packages/fuels-accounts/build.rs
```diff
@@ -0,0 +1,15 @@
+fn main() {
+    #[cfg(not(feature = "std"))]
```
segfault-magnet, line 2: Should run when std, so the opposite of this.

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -304,5 +306,55 @@ impl RetryableClient {
...
+        });
+
+        let response = self.client.query(query).await?;
+
+        let is_resource = response.blob.is_some()
+            || response.contract.is_some()
+            || response.transaction.is_some();
+
+        Ok(!is_resource)
+    }
```
segfault-magnet, line 328: Move before delegation start

segfault-magnet (approved):

hal3e (approved):

digorithm (approved):

PR #1574, opened 2025-01-15
feat!: use `total_gas` and `total_fee` from tx status

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/1394

# Release notes

In this PR we expose the tx status from succeeded translations in our higher level APIs. The user can use the (if available) `tx_status` and `tx_id` from the following transaction APIs
  - deployment
  - transfers
  - force transfers to contract
  - withdrawals to base layer
  - blob upload
  - contract, script and predicate calls/simulations
  
>***Note*** that the `TransactionCost`'s field `gas_used` was renamed to `script_gas` and a new field `total_gas` is added.

# Summary

The `GasValidation` trait is removed. This trait was used to make sure that the `script_gas_limit` is set correctly. The `script_gas_limit` is either:
-  `0` if there is no script code, 
- set by the user,
- estimated by our builders
so an additional check before sending the tx is redundant - By removing it we have one less dry run before sending the tx.

# Breaking Changes

- Removed `get_response_from` method from `CallHandlers`
- `CallResponse` refactored and added `tx_status: Success` field 
- Method `get_response` accepts `TxStatus` instead of `Vec<Receipts>`
- Method `new` is removed form `CallResponse`
- `GasValidation` trait is removed from transaction builders
- `Account`s  `transfer` method returns `Result<TxResponse>`
- `Account`s  `force_transfer_to_contract` method returns `Result<TxResponse>`
- `Account`s  `withdraw_to_base_layer` method returns `Result<WithdrawToBaseResponse>`
- `Executable<Loader>`'s `upload_blob` returns `Result<Option<TxResponse>>`
- Contract's `deploy` and `deploy_if_not_exists` return `Result<DeployResponse>` and `Response<Option<DeployResponse>>` respectively
-`TransactionCost`'s field `gas_used` renamed to `script_gas`

# Checklist


examples/contracts/src/lib.rs
```diff
@@ -636,7 +643,7 @@ mod tests {
             .await?;
         // ANCHOR_END: multi_call_cost_estimation
 
-        let expected_gas = 4402;
+        let expected_gas = 11231;
```
segfault-magnet, line 646: We should probably mention in the release notes somewhere, and maybe the pub docs of `TransactionCost` that the `gas_used` is now the total gas, including both script gas and regular gas.

Otherwise, someone might get surprised by the bump.

Also, since we're reporting details such as `metered_bytes_size`, we might include the script gas used as well, since we're loosing that from `TransactionCost` with this pr.

packages/fuels-core/src/types/tx_response.rs
```diff
@@ -0,0 +1,19 @@
+use fuel_tx::{Receipt, TxId};
+
+use crate::sealed::Sealed;
+
+#[derive(Clone, Debug)]
+pub struct TxResponse<T: TxResponseType = TxId> {
```
segfault-magnet, line 6: Make it non exhaustive?

packages/fuels-programs/src/responses/call.rs
```diff
@@ -1,57 +1,31 @@
...
 /// field holds the decoded typed value returned by the contract's method. The other field holds all
 /// the receipts returned by the call.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 // ANCHOR: call_response
 pub struct CallResponse<D> {
     pub value: D,
-    pub receipts: Vec<Receipt>,
-    pub gas_used: u64,
+    pub tx: TxResponse<Option<TxId>>,
```
segfault-magnet, line 17: Some thoughts on the new struct:

I'd go for a `Option<TxResponse>`, since there is no case where you can have a tx response without having the id of the submitted transaction.

Maybe even a `Option<TxStatus>` and a `Option<TxId>` instead of the tx response. 

Or, perhaps have the `TxStatus::Success` be a newtype variant, turn this
```rust
#[derive(Debug, Clone)]
pub enum TxStatus {
    Success {
        receipts: Vec<Receipt>,
        total_gas: u64,
        total_fee: u64,
    }, ...
```
into
```rust
#[derive(Debug, Clone)]
pub struct Success {
    pub receipts: Vec<Receipt>,
    pub total_gas: u64,
    pub total_fee: u64,
}

#[derive(Debug, Clone)]
pub enum TxStatus {
    Success(Success),
```

Then you can reuse this `Success` in the `CallResponse` since it has all the fields you added in the `TxResponse` save for the id, which can be a separate field. 

Or you can keep the `TxResponse` but have it be something like
```rust
#[derive(Clone, Debug)]
pub struct TxResponse {
    pub status: Success,
    pub id: TxId,
}
```

Since we only ever save the response if we had a successful tx inclusion, otherwise you wouldn't get a call response.

e2e/tests/predicates.rs
```diff
@@ -173,12 +166,11 @@ async fn spend_predicate_coins_messages_basic() -> Result<()> {
 
     predicate.set_provider(provider.clone());
 
-    // TODO: https://github.com/FuelLabs/fuels-rs/issues/1394
-    let expected_fee = 1;
+    let amount_to_send = predicate_balance - 1;
```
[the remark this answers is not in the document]
hal3e: Good catch I will use the `total_fee` here... This one somehow slipped through

examples/contracts/src/lib.rs
hal3e: Thanks makes sense.

packages/fuels-programs/src/responses/call.rs
hal3e: As we have the `simulate` api on the call handlers you can have a tx_status without a tx_id. In both cases, simulate and call, we return the same type.

Thank you for the feedback. I have gone through several iterations and could not decide how to structure the change. I wanted all of our higher level apis to have similar responses. I like the idea with the `Success` struct. I will try that out and see how it feels.

docs/src/calling-contracts/cost-estimation.md
```diff
@@ -6,6 +6,8 @@ With the function `estimate_transaction_cost(tolerance: Option<f64>, block_horiz
 {{#include ../../../packages/fuels-accounts/src/provider.rs:transaction_cost}}
 ```
 
+> **Note** `script_gas` reefers to the part of the gas spent on the script execution.
```
segfault-magnet, line 9:
```suggestion
> **Note** `script_gas` refers to the part of the gas spent on the script execution.
```

packages/fuels-core/src/types/tx_status.rs
```diff
@@ -15,11 +15,16 @@ use crate::{
...
+}
+
 #[derive(Debug, Clone)]
 pub enum TxStatus {
-    Success {
-        receipts: Vec<Receipt>,
-    },
+    Success(Success),
     Submitted,
     SqueezedOut {
```
segfault-magnet, line 29: Maybe, for consistency's sake, extract this into a struct as well?

segfault-magnet (dismissed):

packages/fuels-core/src/types/tx_status.rs
```diff
@@ -15,43 +15,91 @@ use crate::{
...
+            Self::SqueezedOut(SqueezedOut { reason }) => {
+                Err(Error::Transaction(Reason::SqueezedOut(reason.clone())))
+            }
+            Self::Revert(Revert {
+                receipts,
+                reason,
+                revert_id: id,
+                ..
+            }) => Err(Self::map_revert_error(&receipts, &reason, id, log_decoder)),
+            Self::Submitted => Err(Error::Transaction(Reason::SqueezedOut(
```
MujkicA, line 74: `Reason::SqueezedOut` is misplaced. 

IMO the error message doesn't need to convey details about the structures are being processed internally ("cannot take x from y").
Something like "transactions was not yet included" contains just enough info.

e2e/tests/predicates.rs
```diff
@@ -115,7 +115,7 @@ async fn transfer_coins_and_messages_to_predicate() -> Result<()> {
     let num_coins = 16;
     let num_messages = 32;
     let amount = 64;
-    let total_balance = (num_coins + num_messages) * amount;
+    let balance_to_send = ((num_coins + num_messages) * amount) - 1;
```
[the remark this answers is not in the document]
hal3e: Thanks. Will fix

packages/fuels-core/src/types/tx_status.rs
hal3e: Good catch thanks!

MujkicA (approved):

segfault-magnet (approved):

Issue #1577, opened 2025-01-16
Support new policies: Expiration and Owner


SwayStar123:
Support the new expiration policy, and transaction owner policy

https://github.com/FuelLabs/fuel-vm/pull/871
https://github.com/FuelLabs/fuel-vm/pull/872

dmihal:
I believe only expiration policy is being implemented at this time, ownership is still being discussed.

PR #1583, opened 2025-01-21
feat: add `expiration` transaction policy

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/1577

# Release notes

Added the `expiration` transaction policy. Now the user can limit until which block height the transaction is valid.

# Breaking Changes

- added `expiration: Option<u64>` field to `TxPolicies` struct
- `Transaction` trait method `maturity` now returns `Option<u64>`
- `Transaction` trait method `with_maturity` is removed.

# Checklist


digorithm (approved):
Neat -- great tests!

packages/fuels-core/src/types/wrappers/transaction.rs
```diff
@@ -248,6 +260,8 @@ pub trait Transaction:
 
     fn with_maturity(self, maturity: u32) -> Self;
 
+    fn expiration(&self) -> Option<u64>;
```
MujkicA, line 263: should we a setter `with_expiration` here, like we have for maturity

e2e/tests/predicates.rs
```diff
@@ -1245,3 +1245,80 @@ async fn predicate_configurables_in_blobs() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio::test]
+async fn predicate_transfer_respects_maturity_and_expiration() -> Result<()> {
```
MujkicA, line 1250: At some point, we should discuss how we want to approach testing for features like this, where we cover various transaction and input types. It feels like the level of granularity here (and in other cases) might be higher than what we actually benefit from.

MujkicA (dismissed):
Left a remark about our general approach to testing. 
PR looks good :shipit:

packages/fuels-core/src/types/wrappers/transaction.rs
hal3e: Good question. Well other policies are also missing the `with_` methods. We can either remove it or add them for all? What do you think?

hal3e: `with_maturity` changes the `tx_id` so I would like to remove it.

e2e/tests/contracts.rs
```diff
@@ -385,37 +385,57 @@ async fn mult_call_has_same_estimated_and_used_gas() -> Result<()> {
...
+        abi = "e2e/sway/contracts/transaction_block_height/out/release/transaction_block_height-abi.json"
+    ));
 
-    let call_w_maturity = |maturity| {
-        contract_instance
-            .methods()
-            .calling_this_will_produce_a_block()
-            .with_tx_policies(TxPolicies::default().with_maturity(maturity))
-    };
+    let wallet = launch_provider_and_get_wallet().await?;
```
segfault-magnet, line 394: Why move away from `setup_program_test`?

e2e/tests/predicates.rs
```diff
@@ -1000,6 +1000,7 @@ async fn tx_id_not_changed_after_adding_witnesses() -> Result<()> {
     tx.append_witness(witness2.into())?;
     let tx_id_after_witnesses = tx.id(chain_id);
 
+    provider.produce_blocks(20, None).await?;
```
segfault-magnet, line 1003: Why the manual block generation?

hal3e: leftover from some debugging.

MujkicA (approved):

Salka1988 (approved):

segfault-magnet (approved):

PR #1593, opened 2025-01-31
feat: switch blob ID calculation from data offset to configurable offset

Salka1988:
# Release notes

In this release, we:

~~- Previously deployed blobs are no longer accessible due to changes in blob ID calculation. Redeployment is required.~~

# Summary
Updates blob ID calculation to use the configurable offset instead of the data offset, aligning with the updated design for scripts and predicates.

`has_configurable_section_offset` function checks bytes 4 and 7 to distinguish binary versions. If byte 4 is 0x74 (JMPF), it considers byte 7: 0x02 means legacy, 0x04 means new, and any other value results in an error.

# Checklist


hal3e (approved):
LGTM

packages/fuels-programs/src/assembly/script_and_predicate_loader.rs
```diff
@@ -296,6 +321,38 @@ pub fn split_at_data_offset(binary: &[u8]) -> Result<(&[u8], &[u8])> {
...
+    if binary.len() < 8 {
+        return Err(fuels_core::error!(
+            Other,
+            "binary too short to check JMPF instruction, need at least 8 bytes but got: {}",
+            binary.len()
+        ));
+    }
+
+    match (binary[4], binary[7]) {
+        (0x74, 0x02) => Ok(true),
```
segfault-magnet, line 345: it would also be nice that the idea of the presence of 74 and 02 is not present in two places since, this is duplicated in the `is_legacy_check`.

packages/fuels-programs/src/executable.rs
```diff
@@ -313,7 +367,10 @@ mod tests {
         // that there is no data section
         let data_section_offset = 16u64;
 
-        let code = [vec![0; 8], data_section_offset.to_be_bytes().to_vec()].concat();
+        let mut initial_bytes = vec![0; 16];
+        initial_bytes[4..8].copy_from_slice(&[0x74, 0x00, 0x00, 0x02]); // Insert JMPF at position 4
+
+        let code = [initial_bytes, data_section_offset.to_be_bytes().to_vec()].concat();
```
segfault-magnet, line 373: to not manually encode the instruction
```suggestion
        let jmpf = fuel_asm::op::jmpf(0x0, 0x02).to_bytes();
        let mut initial_bytes = vec![0; 16];
        initial_bytes[4..8].copy_from_slice(&jmpf);

        let code = [initial_bytes, data_section_offset.to_be_bytes().to_vec()].concat();
```

segfault-magnet (commented):
We should also probably deprecate the `data_section_offset` on `Executable<Loader>` since that will no longer point to the data section offset but rather to the configurable offset. 

so maybe `add a #deprecated` on it.

digorithm (approved):
LGTM. Great job, @Salka1988, and great suggestions, @segfault-magnet!

packages/fuels-programs/src/assembly/script_and_predicate_loader.rs
```diff
@@ -42,16 +43,17 @@ impl LoaderCode {
...
         &self.code
     }
 
+    #[deprecated]
     pub fn data_section_offset(&self) -> usize {
-        self.data_offset
+        self.section_offset
     }
 
     fn generate_loader_code(blob_id: [u8; 32], data_section: &[u8]) -> (Vec<u8>, usize) {
```
hal3e, line 59: could be either data or configurable section at this point. Would call it `split_section` or something else
```suggestion
    fn generate_loader_code(blob_id: [u8; 32], spli_section: &[u8]) -> (Vec<u8>, usize) {
```

hal3e: in this function there are helper functions that say `data_section` I assume we would have to change those also.

hal3e: I do not know the difference between `data` and `configurable` offfset, but now that we use the `configurable`and not `data` offset do we still have to have two different loader codes ?
`loader_instructions_w_data_section` and `loader_instructions_no_data_section`

packages/fuels-programs/src/debug.rs
```diff
@@ -156,6 +156,7 @@ fn parse_loader_script(script: &[u8], data: &[u8]) -> Result<Option<(ScriptCallD
         ScriptCallData {
             code: script.to_vec(),
             data: data.to_vec(),
+            #[allow(deprecated)]
```
hal3e, line 159: I would not deprecate this rather call it `section_offset` or `split_section_offset` or something else. Tooling still needs this information to shift the the old configurables.

hal3e (changes requested):
Great work. Left some comments

packages/fuels-programs/src/executable.rs
```diff
@@ -265,33 +280,87 @@ mod tests {
...
+            jmpf,
+            data_offset,
+            configurable_offset,
+            data_section,
+        ]
+        .concat();
+        assert_eq!(blob.as_ref(), configurable_stripped_code);
+
+        // And: Loader code should match expected binary
         let loader_code = loader.code();
```
segfault-magnet, line 359: We should fix the expectation here as well.

packages/fuels-programs/src/assembly/script_and_predicate_loader.rs
```diff
@@ -296,6 +326,55 @@ pub fn split_at_data_offset(binary: &[u8]) -> Result<(&[u8], &[u8])> {
...
+        let result = match offset_val {
+            0x04 => Some(true),
+            0x02 => Some(false),
+            _ => None,
+        };
+        Ok(result)
+    } else {
+        Ok(None)
+    }
+}
```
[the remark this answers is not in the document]
Salka1988: We did this because of `handwritten` code. Previously it was `Result<bool>`. I'll revert it.

segfault-magnet (approved):

hal3e (approved):
Nice work!

digorithm (approved):

PR #1608, opened 2025-02-26
chore: pin `base64ct` version

hal3e:
our CI is failing because `base64ct` version `1.7.0` is requiring cargo feature called `edition2024`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,7 +42,7 @@ rust-version = "1.81.0"
 version = "0.70.0"
 
 [workspace.dependencies]
-base64ct = "1.6.0"
+base64ct = "=1.6.0"
 Inflector = "0.11.4"
 anyhow = { version = "1.0", default-features = false }
 dialoguer = { version = "0.11", default-features = false }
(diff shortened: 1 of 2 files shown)


Salka1988 (dismissed):

MujkicA (approved):

segfault-magnet (commented):
I have a concern:

If we specify a dependency's version in the workspace Cargo.toml, will all the workspace members automatically pick up that version? When packaging the crate for crates.io, an "effective" Cargo.toml might be generated and uploaded. I'm not sure if this effective version will include the version constraint we imposed on the dependency just by listing it in the workspace Cargo.toml—especially if our crate (e.g., fuels-accounts) isn't directly marked as dependent on it with something like base64ct = { workspace = true }.

If it doesn't, then we might need to add the dependency to each workspace crate so that the effective Cargo.toml takes it into account. But there's another issue: doing that might mean adding the dependency to crates that didn't need it at all, since they hadn't listed it in their own [dependencies] section nor did they have it as a transient dependency.

We need a way to enforce that any of our crates—whether they currently depend on this dependency (directly or indirectly) or might start to in the future via a patch bump—use a specific version.

hal3e:
The maintainer yanked version `1.7.0`

PR #1628, opened 2025-03-20
feat: bump `fuel-core` to `0.42.0`

hal3e:
# Release notes

In this release, we:

- Bumped `fuel-core` to `0.42.0`
- Added support for simulating calls at specific block height

# Breaking Changes

- **Error Variant Renaming:**
  - The transaction `Reason` error variant previously named `Reverted` in transaction errors is now renamed to `Failure`.
  - The associated `revert_id` field is now an `Option<u64>` rather than a plain `u64`.
  - **Usage:** Update pattern matches and error handling to check for `Error::Transaction(Reason::Failure { .. })` and handle the optional `revert_id` accordingly.

- **Execution Simulation API Changes:**
  - The simulation execution mode has changed from enum variants to a struct with helper functions:
    - Replace **`Execution::Realistic`** with **`Execution::realistic()`**.
    - Replace **`Execution::StateReadOnly`** with **`Execution::state_read_only()`**.
  - To simulate at a specific block height, chain the call with `.at_height(height)`. Note that the node needs to support historical execution.
- The `dry_run_opt` method's `utxo_validation` argument changed to `Option<bool>` and a new argument `at_height: Option<BlockHeight>` is added.
- The method **`determine_missing_contracts`** no longer accepts an optional parameter; it is now called without arguments and will determine all missing contracts in one simulation.

- **`tx_status` changes**
    - The `Revert` variant has been replaced by the `Failure` variant.
    - New variants have been added: `PreconfirmationSuccess` and `PreconfirmationFailure`.
    - The `revert_id` type changed to `Option<u64>`
  - **Usage:** Update your pattern matching and error handling logic.

- **Amount Type Updates:**
  - Functions across accounts, wallets, predicates, and providers have switched from using `u64` to `u128` for asset amounts.
  - **Usage:** Modify function calls (e.g., `get_asset_inputs_for_amount`, `adjust_for_fee`, etc.) to pass amounts as `u128` (or convert them using `.into()` if needed).

- **Node Configuration Adjustments:**
  - The `NodeConfig` struct now includes a new boolean field: `historical_execution`.
  - **Usage:** To use `historical_execution` launch a node (with `rocksdb` enabled) and the `historical_execution` flag set.

# Checklist


e2e/tests/contracts.rs
```diff
@@ -1270,77 +1268,75 @@ async fn low_level_call() -> Result<()> {
             random_salt = false,
         ),
     );
+    {
```
hal3e, line 1271: Nothing changed just added scope to separate the two tests

e2e/tests/contracts.rs
```diff
@@ -1993,27 +1999,77 @@ async fn simulations_can_be_made_without_coins() -> Result<()> {
...
+    let wallets = setup_node_with_high_price(true).await?;
+    let wallet = wallets.first().expect("has wallet");
+
+    let contract_id = Contract::load_from(
+        "sway/contracts/contract_test/out/release/contract_test.bin",
+        LoadConfiguration::default(),
+    )?
+    .deploy_if_not_exists(wallet, TxPolicies::default())
+    .await?
+    .contract_id;
```
segfault-magnet, line 2031: You can use setup program test if you wish, just make the wallets before the macro.

packages/fuels-accounts/src/provider.rs
```diff
@@ -366,15 +366,17 @@ impl Provider {
     pub async fn dry_run_opt(
         &self,
         tx: impl Transaction,
-        utxo_validation: bool,
+        utxo_validation: Option<bool>,
```
segfault-magnet, line 369: Do we know the difference between `None` and `Some(false)`?

packages/fuels-core/src/types/tx_status.rs
```diff
@@ -177,20 +205,32 @@ impl From<ClientTransactionStatus> for TxStatus {
...
                     total_fee,
                 })
             }
+            ClientTransactionStatus::PreconfirmationFailure {
+                reason,
+                receipts,
+                total_gas,
+                total_fee,
+                ..
+            } => TxStatus::Failure(Failure {
```
segfault-magnet, line 227: Why not have a preconfirmation failure as well on our side?

I'm wary of doing `unwrap_or_default` instead of keeping the receipts optional if that is the right way of modeling them.

packages/fuels-programs/src/calls.rs
```diff
@@ -7,16 +7,45 @@ pub mod utils;
 
 pub use call_handler::*;
 pub use contract_call::*;
+use fuel_types::BlockHeight;
 pub use script_call::*;
 
 /// Used to control simulations/dry-runs
 #[derive(Debug, Clone, Default)]
```
segfault-magnet, line 14: Perhaps it would be easier to reason about what is the default if we separate out the default impl to call `Execution::realistic`. 

This way we have to know the default behavior of all of the fields of the struct.

packages/fuels-test-helpers/src/service.rs
```diff
@@ -102,10 +103,14 @@ impl FuelService {
                 max_queries_resolver_recursive_depth: 1,
                 max_queries_directives: 10,
                 max_concurrent_queries: 1024,
+                required_fuel_block_height_tolerance: 10,
```
segfault-magnet, line 106: Do we support historical execution when running the fuel core as a lib?

hal3e: https://github.com/FuelLabs/fuels-rs/pull/1628/commits/3cc45855c4dd38743053fd7e6a9b2c7b67e3a46e

packages/fuels-programs/src/calls.rs
hal3e: https://github.com/FuelLabs/fuels-rs/pull/1628/commits/7c97c96bc50f607b12dfaa81c21a98e3bb237d5d

packages/fuels-accounts/src/provider.rs
hal3e: I could not find one so i reverted to the old code.

packages/fuels-core/src/types/tx_status.rs
hal3e: Talked with the core guys. Currently, there is no case where receipts can be `None` in pre-confirmations so I would suggest we do not add a new struct for now. We can always add it later if need be.

e2e/tests/contracts.rs
```diff
@@ -1961,15 +1961,27 @@ async fn setup_node_with_high_price(historical_execution: bool) -> Result<Vec<Wa
...
+            .file_name()
+            .expect("failed to get file name")
+            .to_string_lossy()
+            .to_string();
+        let temp_database_path = temp_dir.path().join("db");
+
+        node_config.database_type = DbType::RocksDb(Some(temp_database_path));
         node_config.historical_execution = true;
+
+        chain_config.chain_name = temp_dir_name;
```
segfault-magnet, line 1982: Does a non-unique chain_name cause problems when running concurrent nodes?

I.e. maybe we should add a comment explaining why chain name needs to be set, how it is used, why does it need to be the name of a folder and whether it should be unique.

Also the tempdir will be deleted at the end of that scope. Maybe a comment explaining that we're aware of that and that we're using the tempdir to give us a guaranteed unique path for the db.

segfault-magnet, line 1982: Does a non-unique chain_name cause problems when running concurrent nodes?

I.e. maybe we should add a comment explaining why chain name needs to be set, how it is used, why does it need to be the name of a folder and whether it should be unique.

Also the tempdir will be deleted at the end of that scope. Maybe a comment explaining that we're aware of that and that we're using the tempdir to give us a guaranteed unique path for the db.

hal3e: Good catch. The unique name is not needed. https://github.com/FuelLabs/fuels-rs/pull/1628/commits/18767a678d2863d8b88bcfb1b28cbdb5ffb6ca74

segfault-magnet (approved):

e2e/tests/contracts.rs
```diff
@@ -1959,17 +1948,33 @@ async fn setup_node_with_high_price() -> Result<Vec<Wallet>> {
...
-    let node_config = Some(NodeConfig {
+
+    let mut node_config = NodeConfig {
         starting_gas_price: 1100,
         ..NodeConfig::default()
-    });
+    };
+
+    if historical_execution {
+        // The temp dir will be deleted at the end of this scope
```
rymnc, line 1958: perhaps you want to return the `temp_dir` as part of this function then? i'm curious why the tests are passing if the directory is dropped :?

examples/cookbook/src/lib.rs
```diff
@@ -181,9 +180,9 @@ mod tests {
 
             // we don't transfer the full base asset so we can cover fees
             let output = if id == *consensus_parameters.base_asset_id() {
-                wallet_1.get_asset_outputs_for_amount(wallet_2.address(), id, amount / 2)
+                wallet_1.get_asset_outputs_for_amount(wallet_2.address(), id, (amount / 2) as u64)
```
rymnc, line 183: curious, does get_balances return an amount which is not u64?

rymnc (commented):
looks good, i also tested it with zkvm-primitives against fuel-core 0.42.0 and it works. i'm curious about all the new `u128`'s in the code though, afaik we made those changes only for aggregate values that may go over the `u64::MAX_LIMIT`, not for single values

examples/contracts/src/lib.rs
```diff
@@ -382,11 +409,25 @@ mod tests {
...
                 .value;
             // ANCHOR_END: simulate_read_state
         }
+        {
+            let provider = wallet.provider();
+            provider.produce_blocks(2, None).await?;
+            let block_height = provider.latest_block_height().await?;
+            let contract_id = contract_id.clone();
+
+            // ANCHOR: simulate_read_state_at_height
```
netrome, line 423: Nice, didn't know about these until know: https://rust-lang.github.io/mdBook/format/mdbook.html#including-portions-of-a-file

packages/fuels-core/src/types/tx_status.rs
```diff
@@ -65,50 +72,61 @@ impl TxStatus {
...
             TxStatus::Success(Success { total_fee, .. })
-            | TxStatus::Revert(Revert { total_fee, .. }) => *total_fee,
+            | TxStatus::Failure(Failure { total_fee, .. }) => *total_fee,
             _ => 0,
         }
     }
 
     fn map_revert_error(
-        receipts: &[Receipt],
+        receipts: Vec<Receipt>,
```
netrome, line 116: Why do we take ownership of the receipts here? We only use references, and I see that this is causing redundant clone operations in the code.

netrome (approved):
Relatively shallow review as I'm new to this codebase. Mostly looks good to me. Taking ownership over receipts in `map_revert_error` seems unnecessary to me so I'd prefer to see that fixed, but it's not a hard blocker.

examples/cookbook/src/lib.rs
hal3e: yes the type is `u128` as we can have multiple coins (`u64`) so that we do not overflow.

packages/fuels-core/src/types/tx_status.rs
hal3e: At the end of the function we create a `Error::Transaction(Reason::Failure` that clones the receipts. My idea was if we need ownership of the receipts better request it in the function signature.

e2e/tests/contracts.rs
hal3e: https://github.com/FuelLabs/fuels-rs/pull/1628/commits/15809ae8fb0954af18c2170eb9ed4bbc38d89739

netrome (dismissed):

Issue #1627, opened 2025-03-19
`amount` in `transfer` and other higher level API's should be `u128`


hal3e:
The `amount` in `transfer` should be `u128` as we can transfer several coins with amount `u64`. 
The outputs are coins with max amount `u64` so we should split the outputs into several coins until the whole `amount` is covered.

AurelienFT:
Closing because we don't allow transfers of more than u64::MAX in fuel-core.

PR #1643, opened 2025-04-11
Change amount to be u128 in transfer

AurelienFT:
https://github.com/FuelLabs/fuels-rs/issues/1627

# Release notes

In this release, we:

- Modified the `amount` field  in `transfer` to be type `u128`.

# Summary

 Modified the `amount` field  in `transfer` to be type `u128`.

# Breaking Changes

Before : 
```
async fn transfer(
        &self,
        to: &Bech32Address,
        amount: u64,
        asset_id: AssetId,
        tx_policies: TxPolicies,
    ) -> Result<TxResponse>
```

Now: 
```
async fn transfer(
        &self,
        to: &Bech32Address,
        amount: u128,
        asset_id: AssetId,
        tx_policies: TxPolicies,
    ) -> Result<TxResponse>
```
# Checklist


rymnc (approved):
hmm, aren't asset balances restricted to u64 though? or are just coin values restricted to u64 which is why the amount can be u128 here?

AurelienFT:
@hal3e on the issue you mentioned "other high level API" but I can't find other place where is makes sense. If oyu can point me out

AurelienFT:
@rymnc I don't fully understand your comment, the goal of this change is to take advantage of the fact that we can output 255 * u64::MAX amount and so allow transfers that fit in this

hal3e:
> hmm, aren't asset balances restricted to u64 though? or are just coin values restricted to u64 which is why the amount can be u128 here?

As far as I have understood it the asset_balance is `u128` and coins are restricted to u64. I have gone through the fuel-core-client api and the `balances` endpoint returns `u128` thgourgh the `Balance` struct. 

However, we should update the `balance` endpoint as it converts the `u128` amount to `u64`
```
    pub async fn balance(
        &self,
        owner: &Address,
        asset_id: Option<&AssetId>,
    ) -> io::Result<u64> {
        let owner: schema::Address = (*owner).into();
        let asset_id: schema::AssetId = match asset_id {
            Some(asset_id) => (*asset_id).into(),
            None => schema::AssetId::default(),
        };
        let query = schema::balance::BalanceQuery::build(BalanceArgs { owner, asset_id });
        let balance: types::Balance = self.query(query).await?.balance.into();
        Ok(balance.amount.try_into().unwrap_or(u64::MAX))
    }
```

AurelienFT:
> However, we should update the balance endpoint as it converts the u128 amount to u64

Ok will do

AurelienFT:
@hal3e it should be done !

hal3e:
> @hal3e on the issue you mentioned "other high level API" but I can't find other place where is makes sense. If oyu can point me out

I was thinking about `force_transfer_to_contract` and `withdraw_to_base_layer` but I have reviewed the functions and I think we can leave it as `u64` because of the sway code used.

packages/fuels-accounts/src/account.rs
```diff
@@ -107,14 +107,27 @@ pub trait ViewOnlyAccount: Send + Sync {
...
+        let mut outputs = Vec::new();
+        let to: Address = to.into();
+
+        // Split the amount into multiple Coin outputs, each with a maximum value of u64::MAX
+        while remaining_amount > 0 {
+            let output_amount = remaining_amount.min(u64::MAX as u128) as u64;
+            outputs.push(Output::coin(to, output_amount, asset_id));
+            remaining_amount -= output_amount as u128;
+            if outputs.len() >= 254 {
+                return Err(error!(
```
hal3e, line 122: could we do the check at the beginning and have an early exit from the fn ?

packages/fuels-accounts/src/account.rs
```diff
@@ -107,14 +107,27 @@ pub trait ViewOnlyAccount: Send + Sync {
...
+                    Other,
+                    "`amount` of transfer is too large to fit in a single transaction or 254 outputs"
+                ));
+            }
+        }
+
+        outputs.push(Output::change(self.address().into(), 0, asset_id));
+        Ok(outputs)
```
hal3e, line 130: can we please add a test for this logic?

AurelienFT:
Closing because we don't allow transfers of more than u64::MAX in fuel-core.

PR #1644, opened 2025-04-14
feat: add support for preconfirmations

### LINKED ISSUE #1642: Add support for preconfirmation statuses
`fuel-core` `v0.43.0` adds support for preconfirmation statuses. The sdk should expose the new endpoints.

AurelienFT:
Closes #1642 

# Release notes

Add support of pre confirmations.

In this release, we:

- Added two need functions in the `Provider` to access preconfirmations `send_transaction_and_await_status` and `subscribe_transaction_status`

# Checklist


e2e/tests/providers.rs
```diff
@@ -1011,6 +1012,90 @@ async fn test_build_with_provider() -> Result<()> {
...
+    // Then
+    assert_eq!(status.len(), 3);
+    assert!(status.iter().enumerate().all(|(i, tx_status)| {
+        matches!(
+            (i, tx_status.clone().unwrap()),
+            (0, TxStatus::Submitted { .. })
+                | (1, TxStatus::PreconfirmationSuccess { .. })
+                | (2, TxStatus::Success { .. })
+        )
+    }));
```
hal3e, line 1048: If we do all steps from above we can use the same asserts as with subscribe:
```suggestion
    assert!(matches!(
        status.next().await.unwrap().unwrap(),
        TxStatus::Submitted { .. }
    ));
    assert!(matches!(
        status.next().await.unwrap().unwrap(),
        TxStatus::PreconfirmationSuccess { .. }
    ));
    assert!(matches!(
        status.next().await.unwrap().unwrap(),
        TxStatus::Success { .. }
    ));
```

e2e/tests/providers.rs
```diff
@@ -1011,6 +1012,90 @@ async fn test_build_with_provider() -> Result<()> {
...
+    Ok(())
+}
+
+#[tokio::test]
+async fn send_transaction_and_subscribe_status() -> Result<()> {
+    let wallet = launch_provider_and_get_wallet().await?;
+    let provider = wallet.provider();
```
hal3e, line 1055: Maybe we can eliminate the falkyness with a custom block production interval?
```suggestion
     let block_time = 3u32;
    let config = NodeConfig {
        block_production: Trigger::Interval {
            block_time: std::time::Duration::from_secs(block_time.into()),
        },
        ..NodeConfig::default()
    };
    let wallets =
        launch_custom_provider_and_get_wallets(WalletsConfig::default(), Some(config), None)
            .await?;
    let wallet = &wallets[0];
    let provider = wallet.provider();
```

e2e/tests/providers.rs
```diff
@@ -1011,6 +1012,90 @@ async fn test_build_with_provider() -> Result<()> {
...
+            TxStatus::PreconfirmationSuccess { .. }
+        ));
+        assert!(matches!(
+            statuses.next().await.unwrap()?,
+            TxStatus::Success { .. }
+        ));
+    } else if let TxStatus::Success { .. } = status {
+    } else {
+        panic!("Unexpected status: {:?}", status);
+    }
```
hal3e, line 1094: If the custom block production interval fixes flakyness we can return to the previous asserts.
```suggestion
    assert!(matches!(
        status.next().await.unwrap().unwrap(),
        TxStatus::Submitted { .. }
    ));
    assert!(matches!(
        status.next().await.unwrap().unwrap(),
        TxStatus::PreconfirmationSuccess { .. }
    ));
    assert!(matches!(
        status.next().await.unwrap().unwrap(),
        TxStatus::Success { .. }
    ));
```

packages/fuels-accounts/src/provider.rs
```diff
@@ -200,13 +203,73 @@ impl Provider {
...
+                    }
+                    _ => {}
+                };
+                status.map_err(Into::into)
+            })
+            .collect::<Vec<Result<TxStatus>>>()
+            .await;
+
+        Ok(tx_status)
+    }
```
hal3e, line 271: If we change the fn signature in `fuel-core-client` we can simplify the fn:
```suggestion
    pub async fn send_transaction_and_await_status<T: Transaction>(
        &self,
        tx: T,
        include_preconfirmation: bool,
    ) -> Result<impl futures::Stream<Item = Result<TxStatus>>> {
        #[cfg(feature = "coin-cache")]
        let base_asset_id = *self.consensus_parameters().await?.base_asset_id();

        #[cfg(feature = "coin-cache")]
        self.check_inputs_already_in_cache(&tx.used_coins(&base_asset_id))
            .await?;

        let tx = self.prepare_transaction_for_sending(tx).await?;
        Ok(self
            .uncached_client()
            .submit_and_await_status(tx.clone().into(), include_preconfirmation)
            .await?
            .map(move |status| {
                let status: Result<TxStatus> = status.map(Into::into).map_err(Into::into);

                #[cfg(feature = "coin-cache")]
                match status {
                    Ok(TxStatus::SqueezedOut { .. }) | Ok(TxStatus::Failure { .. }) => {
                        self.coins_cache
                            .lock()
                            .map_err(|_| {
                                std::io::Error::new(
                                    std::io::ErrorKind::Other,
                                    "Failed to acquire lock on coins cache",
                                )
                            })?
                            .remove_items(tx.used_coins(&base_asset_id));
                    }
                    _ => {}
                };

                status
            }))
    }
```

packages/fuels-accounts/src/provider/retryable_client.rs
```diff
@@ -146,6 +147,30 @@ impl RetryableClient {
...
+        &'a self,
+        tx: &'a Transaction,
+        include_preconfirmation: bool,
+    ) -> RequestResult<impl Stream<Item = io::Result<TransactionStatus>> + 'a> {
+        self.wrap(|| {
+            self.client
+                .submit_and_await_status_opt(tx, None, Some(include_preconfirmation))
+        })
+        .await
+    }
```
hal3e, line 160:
```suggestion
    pub async fn submit_and_await_status(
        &self,
        tx: Transaction,
        include_preconfirmation: bool,
    ) -> RequestResult<impl Stream<Item = io::Result<TransactionStatus>>> {
        self.wrap(|| {
            self.client
                .submit_and_await_status_opt(tx.clone(), None, Some(include_preconfirmation))
        })
        .await
    }
```

packages/fuels-core/src/types/tx_status.rs
```diff
@@ -170,6 +170,13 @@ impl TxStatus {
             _ => vec![],
         }
     }
+
+    pub fn is_final(&self) -> bool {
+        matches!(
+            self,
+            TxStatus::Success(_) | TxStatus::Failure(_) | TxStatus::SqueezedOut(_)
+        )
+    }
```
hal3e, line 179: If we do all above I would not add this. Less code to maintain :)
```suggestion
```

hal3e (commented):
Thank you very much for working on this and nice work finding the workaround for the `submit_and_await_status` lifetime issues.

I had a look in `fuel-core-client` and I could not find a reason why we bind the `Stream`'s lifetime to a `Transaction` reference that we clone inside the function. This is the current code:
```
    pub async fn submit_and_await_status_opt<'a>(
        &'a self,
        tx: &'a Transaction,
        estimate_predicates: Option<bool>,
        include_preconfirmation: Option<bool>,
    ) -> io::Result<impl Stream<Item = io::Result<TransactionStatus>> + 'a> {
        use cynic::SubscriptionBuilder;
        use schema::tx::SubmitAndAwaitStatusArg;
        let tx = tx.clone().to_bytes(); // We clone here so better to take ownership
        let s = schema::tx::SubmitAndAwaitStatusSubscription::build(
            SubmitAndAwaitStatusArg {
                tx: HexString(Bytes(tx)),
                estimate_predicates,
                include_preconfirmation,
            },
        );

        let stream = self.subscribe(s).await?.map(
            |r: io::Result<schema::tx::SubmitAndAwaitStatusSubscription>| {
                let status: TransactionStatus = r?.submit_and_await_status.try_into()?;
                Result::<_, io::Error>::Ok(status)
            },
        );

        Ok(stream)
    }
```
If we change the function signature to require ownership of the `Transaction` the `Streaam's` lifetime is not bound to the reference anymore and we do not need the workaround anymore. This would be the new impl in `fuel-core-client`:
```
    pub async fn submit_and_await_status(
        &self,
        tx: Transaction,
    ) -> io::Result<impl Stream<Item = io::Result<TransactionStatus>> + '_> {
        self.submit_and_await_status_opt(tx, None, None).await
    }

    /// Similar to [`Self::submit_and_await_commit_opt`], but includes all intermediate states.
    #[cfg(feature = "subscriptions")]
    pub async fn submit_and_await_status_opt(
        &self,
        tx: Transaction,
        estimate_predicates: Option<bool>,
        include_preconfirmation: Option<bool>,
    ) -> io::Result<impl Stream<Item = io::Result<TransactionStatus>> + '_> {
        use cynic::SubscriptionBuilder;
        use schema::tx::SubmitAndAwaitStatusArg;
        let tx = tx.to_bytes();
        let s = schema::tx::SubmitAndAwaitStatusSubscription::build(
            SubmitAndAwaitStatusArg {
                tx: HexString(Bytes(tx)),
                estimate_predicates,
                include_preconfirmation,
            },
        );

        let stream = self.subscribe(s).await?.map(
            |r: io::Result<schema::tx::SubmitAndAwaitStatusSubscription>| {
                let status: TransactionStatus = r?.submit_and_await_status.try_into()?;
                Result::<_, io::Error>::Ok(status)
            },
        );

        Ok(stream)
    }
```

I am ok with merging this version as is with the workaround but would be very happy if we could change the fn signature in `fuel-core-client` so that we can return a `Stream` also in `submit_and_await_status`.

packages/fuels-accounts/src/provider.rs
```diff
@@ -200,13 +203,73 @@ impl Provider {
...
+                .map_err(|_| Error::Other("Failed to acquire lock on coins cache".to_string()))?
                 .remove_items(tx.used_coins(&base_asset_id))
         }
 
         Ok(tx_status)
     }
 
+    /// Similar to `send_transaction_and_await_commit`,
+    /// but collect all the status received until a final one and return them.
+    pub async fn send_transaction_and_await_status<T: Transaction>(
```
segfault-magnet, line 215: A slightly different approach:
```rust
        let base_asset_id = *self.consensus_parameters().await?.base_asset_id();
        let used_base_coins = tx.used_coins(&base_asset_id);

        self.check_inputs_already_in_cache(&used_base_coins).await?;

        let tx = self.prepare_transaction_for_sending(tx).await?.into();
        let mut stream = self
            .uncached_client()
            .submit_and_await_status(&tx, include_preconfirmation)
            .await?;

        let mut statuses = Vec::new();

        // Process stream items until we get a final status
        while let Some(status) = stream.next().await {
            let tx_status = status.map(TxStatus::from).map_err(Into::into);

            let is_final = tx_status.as_ref().ok().is_some_and(|s| s.is_final());

            statuses.push(tx_status);

            if is_final {
                break;
            }
        }

        // Handle cache updates for failures
        if statuses.iter().any(|status| {
            matches!(
                &status,
                Ok(TxStatus::SqueezedOut { .. }) | Ok(TxStatus::Failure { .. }),
            )
        }) {
            self.coins_cache
                .lock()
                .await
                .remove_items(used_base_coins.clone());
        }

        Ok(statuses)
```

fits better because:
* we don't have to use tokio oneshot's to simulate take_until_inclusive
* we don't have to swap `tokio::Mutex` for `std::Mutex` for the `coin_cache` because of lifetimes

or, a bit more verbose but functional, replacing the loop with:
```rust
        let statuses = self
            .uncached_client()
            .submit_and_await_status(&tx, include_preconfirmation)
            .await?
            .scan(
                false, // initial state: should we stop after this item?
                |should_stop, status| {
                    if *should_stop {
                        return futures::future::ready(None);
                    }

                    let tx_status: Result<TxStatus> = status.map(Into::into).map_err(Into::into);

                    // If this is a final status, we'll want to stop after this item
                    *should_stop = tx_status.as_ref().ok().is_some_and(|s| s.is_final());

                    futures::future::ready(Some(tx_status))
                },
            )
            .collect::<Vec<_>>()
            .await;
```

packages/fuels-accounts/src/provider.rs
```diff
@@ -301,7 +366,10 @@ impl Provider {
         self.check_inputs_already_in_cache(&used_utxos).await?;
 
         let tx_id = self.uncached_client().submit(&tx.into()).await?;
-        self.coins_cache.lock().await.insert_multiple(used_utxos);
+        self.coins_cache
+            .lock()
```
segfault-magnet, line 370: I'd rather avoid having a sync::Mutex in async code if possible. Albeit most of the operations shouldn't block for too long.

AurelienFT:
Thanks everyone for all your feedbacks I will place this PR as draft waiting for the change you proposed on fuel-core which is a great idea. @hal3e what's the policy then because it will not be part of 0.43 and so we will need to wait another version of fuel-core to have this here ?

hal3e:
> Thanks everyone for all your feedbacks I will place this PR as draft waiting for the change you proposed on fuel-core which is a great idea. @hal3e what's the policy then because it will not be part of 0.43 and so we will need to wait another version of fuel-core to have this here ?

As it might take some time to get a new release I would suggest we go with @segfault-magnet implementation idea and then update once we have a new `fuel-core` version.

e2e/tests/providers.rs
AurelienFT: Will be done after https://github.com/FuelLabs/fuels-rs/issues/1647

e2e/tests/providers.rs
AurelienFT: Done.

packages/fuels-accounts/src/provider.rs
AurelienFT: Done.

packages/fuels-accounts/src/provider.rs
AurelienFT: I think that std mutex are more performant if they are not hold while awaiting for something else which is something we don't do with `coins_cache` (a ressource about that : https://kitsu.me/posts/2020_06_01_mutex_in_async_world ) but i'm fine with keeping your standards and I updated

packages/fuels-core/src/types/tx_status.rs
AurelienFT: Still needed until https://github.com/FuelLabs/fuels-rs/issues/1647

e2e/tests/providers.rs
AurelienFT: Done.

hal3e (approved):
Thanks! LGTM

packages/fuels-accounts/src/provider.rs
segfault-magnet: I don’t feel strongly about it — just wanted to flag it for future reference.

Generally, when it comes to trade‑offs involving performance in the SDK, performance tends to be deprioritized, so other concerns usually win out.

To clarify my concern:

Right now, we don’t hold locks across `await` points, so using a `std::sync::Mutex` is possible. If we ever do introduce such a pattern, it would result in a compile‑time error, so we’d catch it early.  So I'm ok with that.

However, the more subtle risk is if we start doing more work while holding the lock — even without an `.await`. In that case, we risk starving the Tokio worker thread, since it can’t process other tasks. In fact, [one of Tokio’s authors recommends keeping](https://stackoverflow.com/a/74547875) **no more than 10–100 μs** of continuous work between `.await` points.
  
Rather than rely on everyone remembering this constraint, I’d prefer to accept the small performance overhead and use a lock type that cooperates better with the async runtime.

In practice, many optimizations are usually negligible in our context, since most operations involve a network call — which dominates overall latency. So even if a `std::Mutex` is slightly more efficient in isolation, that gain is dwarfed by the 1–200 ms it typically takes to complete an external request.

segfault-magnet (approved):

PR #1653, opened 2025-04-23
chore: bump versions

segfault-magnet:


xgreenx:
Do you know if we need to create a release?=D It looks like we don't have a lot of changes, or is it release for Sway?

kayagokalp:
> Do you know if we need to create a release?=D It looks like we don't have a lot of changes, or is it release for Sway?

This is for sway and forc-wallet

PR #1655, opened 2025-04-29
Add possibility to place a predefined gas amount for predicate

AurelienFT:
Useful to be forced to `estimate_predicates` and let the user define it if he wants. Tried to make it non-breaking by using a new constructor (I don't know if the new field in the enum variant is considered breaking)

# Release notes

In this release, we:

- Added a new constructor for `Input`:  `resource_predicate_with_gas` to let user place a predefined amount 

# Checklist

### DIFF
--- a/packages/fuels-core/src/types/wrappers/input.rs
+++ b/packages/fuels-core/src/types/wrappers/input.rs
@@ -16,6 +16,7 @@ pub enum Input {
         resource: CoinType,
         code: Vec<u8>,
         data: Vec<u8>,
+        gas_used: u64,
     },
     Contract {
         utxo_id: UtxoId,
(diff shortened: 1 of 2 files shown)


packages/fuels-core/src/types/wrappers/input.rs
```diff
@@ -16,6 +16,7 @@ pub enum Input {
         resource: CoinType,
         code: Vec<u8>,
         data: Vec<u8>,
+        gas_used: u64,
```
rymnc, line 19: this is a breaking change unless we annotated this previously with the `non_exhaustive` attribute.

AurelienFT: Added

rymnc: adding `non_exhaustive` also makes this change a breaking change, but prevents breaking changes in the future, was that intended?

rymnc: in any case, having `non_exhaustive` only on one enum variant makes it a little annoying for the consumer of this library, so perhaps we can just annotate the whole enum with it?

packages/fuels-core/src/types/wrappers/input.rs
```diff
@@ -36,6 +37,21 @@ impl Input {
             resource,
             code,
             data,
+            gas_used: 0,
+        }
+    }
+
+    pub const fn resource_predicate_with_gas(
```
hal3e, line 44: As it is already  a breaking change I would add the `gas_used` directly to `resource_predicate`

Issue #1181, opened 2023-10-31
test: show that two identical contracts can be deployed because their salt is different


iqdecay:
This need comes from #1177.

digorithm:
This should be closed now, right?

digorithm:
Ah, my bad, I didn't see the `test:` prefix! (and I was the one who suggested writing a test for it 😂)

PR #1658, opened 2025-05-13
chore: test contract deployment with salt

hal3e:
closes: https://github.com/FuelLabs/fuels-rs/issues/1181

# Checklist


e2e/tests/contracts.rs
```diff
@@ -46,6 +46,45 @@ async fn test_multiple_args() -> Result<()> {
     Ok(())
 }
 
+#[tokio::test]
+async fn contract_deployment_with_salt() {
```
MitchTurner, line 50: Can we reformat this to match our testing format standard:
https://www.notion.so/fuellabs/Engineering-Style-Guide-83ba6fd81f12443796c814ecfc746b39?pvs=4#1742f2293f3180f0931bd1355b0afa13

It looks like the `setup_program_test!` macro is obscuring some of this, so maybe just

1. rename the test to clarify what we are doing and what we are expecting: `sut__when..._then...`
2. for the `given/when/then`, maybe just put the `//given` and `//when` above the macro and the `//then` above the asserts

I don't really know what this test is doing other than deploying contract with salt. i.e. I don't know what the _alternative_ case is. Should/can we have another test with the unhappy path? This would help elucidate the full picture for a layman looking at the code.

PR #784, opened 2023-01-17
style!: use C-style getters naming for SDK methods

iqdecay:
This PR closes #783 by changing the naming of contract instance getter methods. This is a breaking change since those APIs are public-facing.
It also renames two functions whose name wasn't really adapted.
It also adds a comment about a `.gitignore` file.

Issue #783, opened 2023-01-17
style: rename getters in contracts bindings


iqdecay:
In `packages/fuels-core/src/code_gen/abigen/bindings/contract.rs`, getters do not conform to the [Rust C-getters convention](https://rust-lang.github.io/api-guidelines/naming.html#c-getter).
We should rename those getters.

PR #1714, opened 2026-06-12
Provide errors in case of `fuel_bin_service::run_node` failing

ironcev:
# Summary

This PR captures stderr and stdout in case of `run_node` not being able to get the bound address.

When the spawned `fuel-core` process exits (or times out) before logging its bound address, the resulting error was just "fuel-core process exited before reporting its bound address". The process's actual output was discarded, making the failure undiagnosable from test logs alone.

`run_node` now buffers the startup lines read from stderr and, on failure, drains stdout as well, appending everything the process printed to the error message. This matters in practice: e.g. when fuel-core resolves to the fuelup proxy, the proxy reports its failures on stdout and exits with status 0, so previously no trace of the real cause survived.

Example of the new error:

```
fuel-core process exited before reporting its bound address. captured output:
[stdout] Adding component fuel-core v0.48.0 to 'testnet-x86_64-unknown-linux-gnu'
[stdout] Could not add component fuel-core(0.48.0): No such file or directory (os error 2)
```

# Checklist

### DIFF
--- a/packages/fuels-test-helpers/src/fuel_bin_service.rs
+++ b/packages/fuels-test-helpers/src/fuel_bin_service.rs
@@ -212,7 +212,9 @@ async fn run_node(extended_config: ExtendedConfig) -> FuelResult<(SocketAddr, Jo
     let mut stdout_reader = BufReader::new(stdout).lines();
     let mut stderr_reader = BufReader::new(stderr).lines();
 
-    let bound_address = tokio::time::timeout(Duration::from_secs(30), async {
+    let mut startup_logs = Vec::new();
+
+    let scan_result = tokio::time::timeout(Duration::from_secs(30), async {
         while let Some(line) = stderr_reader
             .next_line()
             .await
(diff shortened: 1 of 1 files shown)


ironcev:
Regarding the [failing CI step](https://github.com/FuelLabs/fuels-rs/actions/runs/27418027849/job/81036036973?pr=1714#step:12:337), here is the Claude Code Fable 5 analysis:

The CI failure has nothing to do with your PR — time 0.3.48 was published today (2026-06-12) and it breaks every version of octocrab, a dependency of the change-log script.

Here's the full chain:

1. fuels-rs doesn't commit a Cargo.lock, so CI resolves dependencies fresh on every run. Your local build uses your existing lockfile, which pins time 0.3.47 — that's the entire difference between CI and local.
2. time 0.3.48 added a From impl using an associated-type projection that collides with octocrab's blanket impl<T: serde::Serialize> From<T> for Filter<T> (octocrab/src/params.rs), producing E0119: conflicting implementations. I reproduced it in a clean worktree with a fresh lockfile, and confirmed that even the newest octocrab (0.53.1) fails the same way — so bumping octocrab doesn't help.
3. The time maintainers already know: issue #783 ("New version (0.3.48) makes my project stop compiling") and a fix #784 ("avoid associated type projection in modifier! From impl") were both opened today.

Your options

Wait and re-run CI (recommended): given #784, a fixed time 0.3.49 or a yank of 0.3.48 is likely within a day or two. Re-running the job after that will pass with no changes — half the Rust ecosystem using octocrab without a lockfile is hitting this right now.

Conclusion: No rush with this PR, we can wait for a few days.

ironcev:
:+1:

xgreenx (approved):

Issue #97, opened 2022-02-11
Model `abigen!`s `Source::Path` behaviour after `include_str!`


mitchmindtree:
*Follow-up to #94.*

Currently the path passed to the right hand side argument of `abigen!` is expected to be relative to the `std::env::current_dir`:

https://github.com/FuelLabs/fuels-rs/blob/f6d4cc0399829a16766a1bbdffaf31c4476ed175/fuels-core/src/source.rs#L47

This evaluates to whatever happens to be the current directory at the time of macro expansion. This means that the necessary path given to `abigen!` can change when the project is moved to a different subdirectory of the same workspace, or if its moved from a workspace into its own project, or vice versa. Ideally, we want our `src` not to be dependent on the directory that the project happens to be executed within.

To fix this, we could model our behaviour after the std `include_str!` macro. `include_str!` accepts a path that is relative to the *file* in which it is invoked. For example if we have:

```
project/my-contract-abi.json
project/tests/harness.rs
```

the `abigen!` macro could then be invoked within the `project/tests/harness.rs` with

```rust
abigen!(MyContract, "../my-contract-abi.json");
```

JoshuaBatty:
So I took a look into this. Originally I thought we could use the `file!()` macro but then we are back to square one with not being able to have a macro evaluate as an input to another macro. Further, the path returned by `file!()` is where the macro code lives *not the file where the macro was called from*. 

I then started to look at using span to get this information out. Turns out you can use a span to get access to the calling file path, but currently it's an unstable feature. See [this open issue](https://github.com/rust-lang/rust/issues/54725)
It provided a method on span `pub fn source_file(&self) -> SourceFile;` This would allow us to get relative paths working, but the issue has been open for over 3 years, so there is a chance it might not land. Still, might be worth periodically checking in to see if there is any movement towards stabilizing this feature.

JoshuaBatty:
From memory, this needed to be solved in order for us to properly support native Windows builds. This is due to how paths are handled between windows and unix systems. See [this tracking issue](https://github.com/FuelLabs/sway/issues/1526) in the sway repo.

I just re-checked the issue I mentioned above and it seems that [this PR](https://github.com/rust-lang/rust/pull/140514) just got merged 2 weeks ago that might allow us to move forward here. Might possibly be available in the next rust release.

PR #1720, opened 2026-09-13
Resolve ABI paths relative to their source literals

pederbe:
- Ref #97

# Release notes

In this release, we:

- Resolve relative ABI paths in `abigen!` and `wasm_abigen!` from the file containing the path literal.

# Summary

Moving a package into or out of a workspace can currently change the ABI paths it needs. Resolving from the literal's source file removes that dependency on the compiler's working directory. This also migrates the repository's existing paths and documents the resolution rules.

# Breaking Changes

Existing relative ABI paths must be updated to be relative to the file containing their string literal. Review every path even if compilation succeeds: a valid ABI at the new location will be selected instead of the old one. There is no working-directory fallback.

For macro wrappers, literals written in the wrapper resolve from its definition file; forwarded literals resolve from the file where they were written. This differs from `include_str!`, which uses the outer invocation's file, so this proposal does not provide full equivalence with the behavior discussed in #97.

Absolute paths, inline JSON, `setup_program_test!` project paths, and direct `Abi::load_from` calls retain their existing behavior.

# Checklist


pederbe:
Can a maintainer please add the `breaking` label?

